PR #559, opened 2017-10-21
Coalesce anchoring status updates

### LINKED ISSUE #556: Performance issues w 200 annotations on a page
### Steps to reproduce
1. Go here https://engagements2017-18.as.virginia.edu/the-document/rockfish-gap-report/

### Expected behaviour
Page and sidebar performance should be acceptable

### Actual behaviour
Quite slow, close to unusable.

Chrome

robertknight:
Profiling the test case in #556 showed significant overhead from the
processing involved in or triggered by each `UPDATE_ANCHOR_STATUS`
action handled by the store. Previously one `UPDATE_ANCHOR_STATUS`
action was dispatched for each annotation whose anchoring status
changed.

Improve the situation by making the `UPDATE_ANCHOR_STATUS` action handle
updates for multiple annotations at once and modifying the `frameSync`
service to coalesce anchoring status updates from the host page into
a smaller number of `UPDATE_ANCHOR_STATUS` actions.

This commit also takes the opportunity to remove the logic to update
annotation tags when handling this action since it is no longer
required.

With this change the test case in #556 is not perfectly smooth, but it is significantly better and good enough to be usable on many systems.

Fixes #556

### DIFF
--- a/src/sidebar/frame-sync.js
+++ b/src/sidebar/frame-sync.js
@@ -130,7 +130,7 @@ function FrameSync($rootScope, $window, Discovery, annotationUI, bridge) {
     bridge.on('sync', function (events_) {
       events_.forEach(function (event) {
         inFrame.add(event.tag);
-        annotationUI.updateAnchorStatus(null, event.tag, event.msg.$orphan);
+        annotationUI.updateAnchorStatus(event.tag, event.msg.$orphan);
         $rootScope.$broadcast(events.ANNOTATIONS_SYNCED, [event.tag]);
       });
     });
--- a/src/sidebar/frame-sync.js
+++ b/src/sidebar/frame-sync.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var debounce = require('lodash.debounce');
+
 var events = require('./events');
 var bridgeEvents = require('../shared/bridge-events');
 var metadata = require('./annotation-metadata');
@@ -126,12 +128,24 @@ function FrameSync($rootScope, $window, Discovery, annotationUI, bridge) {
 
     bridge.on('destroyFrame', destroyFrame.bind(this));
 
+    // Map of annotation tag to anchoring status
+    // ('anchored'|'orphan'|'timeout').
+    //
+    // Updates are coalesced to reduce the overhead from processing
+    // triggered by each `UPDATE_ANCHOR_STATUS` action that is dispatched.
+    var anchoringStatusUpdates = {};
+    var scheduleAnchoringStatusUpdate = debounce(() => {
+      annotationUI.updateAnchorStatus(anchoringStatusUpdates);
+      $rootScope.$broadcast(events.ANNOTATIONS_SYNCED, Object.keys(anchoringStatusUpdates));
+      anchoringStatusUpdates = {};
+    }, 10);
+
     // Anchoring an annotation in the frame completed
     bridge.on('sync', function (events_) {
       events_.forEach(function (event) {
         inFrame.add(event.tag);
-        annotationUI.updateAnchorStatus(event.tag, event.msg.$orphan);
-        $rootScope.$broadcast(events.ANNOTATIONS_SYNCED, [event.tag]);
+        anchoringStatusUpdates[event.tag] = event.msg.$orphan ? 'orphan' : 'anchored';
+        scheduleAnchoringStatusUpdate();
       });
     });
 
--- a/src/sidebar/reducers/annotations.js
+++ b/src/sidebar/reducers/annotations.js
@@ -173,13 +173,10 @@ var update = {
 
   UPDATE_ANCHOR_STATUS: function (state, action) {
     var annotations = state.annotations.map(function (annot) {
-      var match = (annot.id && annot.id === action.id) ||
-                  (annot.$tag && annot.$tag === action.tag);
-      if (match) {
+      if (annot.$tag === action.tag) {
         return Object.assign({}, annot, {
           $anchorTimeout: action.anchorTimeout || annot.$anchorTimeout,
           $orphan: action.isOrphan,
-          $tag: action.tag,
         });
       } else {
         return annot;
@@ -274,7 +271,6 @@ function addAnnotations(annotations, now) {
             dispatch({
               type: actions.UPDATE_ANCHOR_STATUS,
               anchorTimeout: true,
-              id: orphan.id,
               tag: orphan.$tag,
             });
           });
@@ -297,19 +293,16 @@ function clearAnnotations() {
 }
 
 /**
- * Updating the local tag and anchoring status of an annotation.
+ * Update the anchoring status of an annotation.
  *
- * @param {string|null} id - Annotation ID
- * @param {string} tag - The local tag assigned to this annotation to link
- *        the object in the page and the annotation in the sidebar
+ * @param {string} tag - The annotation's local tag
  * @param {boolean} isOrphan - True if the annotation failed to anchor
  */
-function updateAnchorStatus(id, tag, isOrphan) {
+function updateAnchorStatus(tag, isOrphan) {
   return {
     type: actions.UPDATE_ANCHOR_STATUS,
-    id: id,
-    tag: tag,
-    isOrphan: isOrphan,
+    tag,
+    isOrphan,
   };
 }
 
--- a/src/sidebar/reducers/annotations.js
+++ b/src/sidebar/reducers/annotations.js
@@ -261,19 +261,21 @@ function addAnnotations(annotations, now) {
     // successfully anchor then the status will be updated.
     var ANCHORING_TIMEOUT = 500;
 
-    var anchoringAnnots = added.filter(metadata.isWaitingToAnchor);
-    if (anchoringAnnots.length) {
-      setTimeout(function () {
-        var statusUpdates = arrayUtil
-          .filterMap(anchoringAnnots, function (annot) {
-            return findByID(getState().annotations, annot.id);
-          })
-          .filter(metadata.isWaitingToAnchor)
-          .reduce((statusUpdates, orphan) => {
-            statusUpdates[orphan.$tag] = 'timeout';
-            return statusUpdates;
-          }, {});
-        dispatch(updateAnchorStatus(statusUpdates));
+    var anchoringIDs = added.filter(metadata.isWaitingToAnchor)
+                            .map(ann => ann.id);
+    if (anchoringIDs.length > 0) {
+      setTimeout(() => {
+        // Find annotations which haven't yet been anchored in the document.
+        var anns = getState().annotations;
+        var annsStillAnchoring = anchoringIDs.map(id => findByID(anns, id))
+                                             .filter(ann => ann && metadata.isWaitingToAnchor(ann));
+
+        // Mark anchoring as timed-out for these annotations.
+        var anchorStatusUpdates = annsStillAnchoring.reduce((updates, ann) => {
+          updates[ann.$tag] = 'timeout';
+          return updates;
+        }, {});
+        dispatch(updateAnchorStatus(anchorStatusUpdates));
       }, ANCHORING_TIMEOUT);
     }
   };
--- a/src/sidebar/reducers/annotations.js
+++ b/src/sidebar/reducers/annotations.js
@@ -173,13 +173,15 @@ var update = {
 
   UPDATE_ANCHOR_STATUS: function (state, action) {
     var annotations = state.annotations.map(function (annot) {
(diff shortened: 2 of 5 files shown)


src/sidebar/frame-sync.js
```diff
@@ -126,12 +128,24 @@ function FrameSync($rootScope, $window, Discovery, annotationUI, bridge) {
...
+    // ('anchored'|'orphan'|'timeout').
+    //
+    // Updates are coalesced to reduce the overhead from processing
+    // triggered by each `UPDATE_ANCHOR_STATUS` action that is dispatched.
+    var anchoringStatusUpdates = {};
+    var scheduleAnchoringStatusUpdate = debounce(() => {
+      annotationUI.updateAnchorStatus(anchoringStatusUpdates);
+      $rootScope.$broadcast(events.ANNOTATIONS_SYNCED, Object.keys(anchoringStatusUpdates));
+      anchoringStatusUpdates = {};
+    }, 10);
```
seanh: Ok, so the anonymous function passed to `debounce()` will be called at most once every 10ms, no matter how many times we try to call it. And when it does finally get called, it uses `anchoringStatusUpdates` to update `annotationUI` and broadcast `ANNOTATIONS_SYNCED`, and then it resets `anchoringStatusUpdates` to empty.

src/sidebar/frame-sync.js
```diff
@@ -126,12 +128,24 @@ function FrameSync($rootScope, $window, Discovery, annotationUI, bridge) {
...
+
     // Anchoring an annotation in the frame completed
     bridge.on('sync', function (events_) {
       events_.forEach(function (event) {
         inFrame.add(event.tag);
-        annotationUI.updateAnchorStatus(null, event.tag, event.msg.$orphan);
-        $rootScope.$broadcast(events.ANNOTATIONS_SYNCED, [event.tag]);
+        anchoringStatusUpdates[event.tag] = event.msg.$orphan ? 'orphan' : 'anchored';
+        scheduleAnchoringStatusUpdate();
```
seanh: Add this is where we add things in to `anchoringStatusUpdates` for the debounced function `scheduleAnchoringStatusUpdate()` to pick them up. And here we freely call `scheduleAnchoringStatusUpdate()` as often as we want, safe in the knowledge the it'll only actually be called at most once every 10ms.

src/sidebar/reducers/annotations.js
```diff
@@ -173,17 +173,16 @@ var update = {
...
-      var match = (annot.id && annot.id === action.id) ||
-                  (annot.$tag && annot.$tag === action.tag);
-      if (match) {
-        return Object.assign({}, annot, {
-          $anchorTimeout: action.anchorTimeout || annot.$anchorTimeout,
-          $orphan: action.isOrphan,
-          $tag: action.tag,
-        });
-      } else {
+      if (!action.statusUpdates.hasOwnProperty(annot.$tag)) {
```
seanh: Ok, so here we're looping over _all_ of the annotations in the sidebar (I guess this is why it's slow, to call this function too often) and for each annotation, _if_ the annotation is in the status updates, we proceed

src/sidebar/reducers/annotations.js
```diff
@@ -173,17 +173,16 @@ var update = {
...
         return annot;
       }
+
+      var state = action.statusUpdates[annot.$tag];
+      if (state === 'timeout') {
+        return Object.assign({}, annot, { $anchorTimeout: true });
+      } else {
+        return Object.assign({}, annot, { $orphan: state === 'orphan' });
+      }
```
seanh: And for each annotation that gets updated, we _either_ set `$anchorTimeout: true` _or_ we set `$orphan` to either `true` or `false`. Here I'm confused. Isn't "anchor timeout" and "orphan" kind of the same thing? And I'm not sure why this PR is changing this block of code, from the `if (match)` block that was here before.

src/sidebar/reducers/annotations.js
```diff
@@ -265,19 +264,16 @@ function addAnnotations(annotations, now) {
...
-              anchorTimeout: true,
-              id: orphan.id,
-              tag: orphan.$tag,
-            });
-          });
+          .reduce((statusUpdates, orphan) => {
+            statusUpdates[orphan.$tag] = 'timeout';
+            return statusUpdates;
+          }, {});
+        dispatch(updateAnchorStatus(statusUpdates));
```
seanh: I have no idea what is happening here

src/sidebar/reducers/annotations.js
robertknight: > Ok, so here we're looping over all of the annotations in the sidebar (I guess this is why it's slow, to call this function too often)

The current data structure that stores annotations in the sidebar is a list. This is indeed certainly suboptimal for certain operations. In practice however, it wasn't the (main) issue here from the profiling I looked at.

Rather the problem is that for every action that goes through the store, there are a bunch of processing steps that get triggered. These include Angular digest cycles (though these are debounced using `$applyAsync`) to trigger UI updates, plus any subscribers to state changes. Collectively this all adds up. It can be optimized, but batching updates is always going to be more efficient.

src/sidebar/frame-sync.js
robertknight: Yes, exactly.

src/sidebar/frame-sync.js
robertknight: Yes, indeed.

src/sidebar/reducers/annotations.js
robertknight: Thanks for the feedback. I rewrote this piece of code. Please let me know if it seems clearer now.

src/sidebar/reducers/annotations.js
```diff
@@ -262,22 +261,21 @@ function addAnnotations(annotations, now) {
...
+        var anns = getState().annotations;
+        var annsStillAnchoring = anchoringIDs.map(id => findByID(anns, id))
+                                             .filter(ann => ann && metadata.isWaitingToAnchor(ann));
+
+        // Mark anchoring as timed-out for these annotations.
+        var anchorStatusUpdates = annsStillAnchoring.reduce((updates, ann) => {
+          updates[ann.$tag] = 'timeout';
+          return updates;
+        }, {});
+        dispatch(updateAnchorStatus(anchorStatusUpdates));
```
robertknight: This is responsible for marking anchoring as having timed-out for newly loaded annotations if they fail to anchor within a certain time period. I modified this initially to dispatch _one_ action with N updates instead of N actions with one update each. In a second commit I rewrote this whole section for clarity in response to CR feedback.

src/sidebar/reducers/annotations.js
```diff
@@ -173,17 +173,16 @@ var update = {
 
   UPDATE_ANCHOR_STATUS: function (state, action) {
     var annotations = state.annotations.map(function (annot) {
-      var match = (annot.id && annot.id === action.id) ||
```
robertknight: It used to be the case that some annotations (existing ones that had not yet been anchored) had only IDs, some had only tags (newly created but not yet saved ones) and some had both (saved annotations that had been anchored).

Every annotation now has a `$tag` as soon as it is added to the store, so we no longer need to check for either an ID or a tag.

src/sidebar/reducers/annotations.js
robertknight: Other parts of the code make use of `$anchorTimeout`. I think we might indeed be able to have _just_ an `$orphan` flag in future, but to keep this PR smaller I preserved the existing behaviour of this action.

src/sidebar/test/frame-sync-test.js
```diff
@@ -189,16 +189,39 @@ describe('FrameSync', function () {
...
+    });
+
     it('updates the anchoring status for the annotation', function () {
       fakeBridge.emit('sync', [{tag: 't1', msg: {$orphan: false}}]);
-      assert.calledWith(fakeAnnotationUI.updateAnchorStatus, null, 't1', false);
+      clock.tick(20);
+      assert.calledWith(fakeAnnotationUI.updateAnchorStatus, { t1: 'anchored' });
+    });
+
+    it('coalesces multiple "sync" messages', () => {
```
robertknight: This is a test to check that multiple sync messages in quick succession result in a single action being dispatched.

src/sidebar/test/frame-sync-test.js
```diff
@@ -189,16 +189,39 @@ describe('FrameSync', function () {
...
+        t2: 'orphan',
+      });
     });
 
     it('emits an ANNOTATIONS_SYNCED event', function () {
       var onSync = sinon.stub();
       $rootScope.$on(events.ANNOTATIONS_SYNCED, onSync);
 
       fakeBridge.emit('sync', [{tag: 't1', msg: {$orphan: false}}]);
+      clock.tick(20);
```
robertknight: This is needed because emitting the `ANNOTATIONS_SYNCED` event no longer happens synchronously, due to debouncing, when a "sync" message is received by the sidebar.

src/sidebar/test/frame-sync-test.js
```diff
@@ -189,16 +189,39 @@ describe('FrameSync', function () {
   });
 
   context('when anchoring completes', function () {
+    var clock = sinon.stub();
```
robertknight: Due to the introduction of debouncing, we now need to play god with time in the tests.

src/sidebar/test/annotation-ui-test.js
```diff
@@ -483,28 +491,12 @@ describe('annotationUI', function () {
   });
 
   describe('#updatingAnchorStatus', function () {
-    it("updates the annotation's tag", function () {
```
robertknight: All annotations now have tags as soon as they are added to the store, so `updateAnchorStatus` no longer does what this test checks.

src/sidebar/test/annotation-ui-test.js
```diff
@@ -483,28 +491,12 @@ describe('annotationUI', function () {
...
-
     it("updates the annotation's orphan flag", function () {
       var annot = defaultAnnotation();
       annotationUI.addAnnotations([annot]);
-      annotationUI.updateAnchorStatus(annot.id, 'atag', true);
+      annotationUI.updateAnchorStatus({ [tagForID(annot.id)]: 'orphan' });
       assert.equal(annotationUI.getState().annotations[0].$orphan, true);
     });
-
-    it('updates annotations by tag', function () {
```
robertknight: This test is now irrelevant since there is only one way to reference which annotation was anchored / failed to anchor, and that is covered by the test above.

robertknight:
Thanks for the review @seanh. I updated a piece of code that you flagged up as unclear and added some comments to a few other things. Please take a look.

src/sidebar/reducers/annotations.js
seanh: Ah, I see: "This commit also takes the opportunity to remove the logic to update annotation tags when handling this action since it is no longer required." This caused me a fair bit of confusion (since the change seemed to have nothing to do with coalescing updates), if this could have been done as a separate PR or at least a separate commit it might have been easier to follow.

src/sidebar/reducers/annotations.js
seanh: The `if (state === 'timeout')` confuses me a little. This wasn't part of the logic of the previous code. It seems that it previously set _both_ `$anchorTimeout` and `$orphan`, but now it sets _either_ one or the other, depending on this new if?

src/sidebar/reducers/annotations.js
seanh: I see, so this is a function that runs after 500ms and orphans any annotations that haven't anchored yet. It'll end up calling the `UPDATE_ANCHOR_STATUS` function above, which will set `$anchorTimeout: true`. It seems odd that this is in the sidebar code rather than in the annotator code, but I appreciate that you've rewritten it in-place for clarity.

robertknight: The reason it is in the sidebar rather than the annotator code is because the "annotator" code runs in the context of the host page and there are various ways in which that could cause the timeout code to either not run or to fail to execute within the timeout period. For example, a news website bogged down by a ton of crappy JS and ads.

Putting the code in the sidebar avoids some of these failure scenarios and, depending on the browser's JS execution scheduling, avoids its execution being blocked by code in the host page.

src/sidebar/reducers/annotations.js
robertknight: Actually this was a bug in the previous code, when the `UPDATE_ANCHOR_STATUS` action was dispatched from this file, the `isOrphan` property was missing from the action, so `$orphan` was always being set to `undefined` here. That happens to be a no-op, but conceptually it is a bug.

PR #561, opened 2017-10-25
Strip unknown query params when generating Internet Archive embed URLs

robertknight:
Internet Archive details pages may include params in the query string
as well as "start" and "end" params in the path. In that case invalid
URLs were being generated, such as https://archive.org/embed/CSPAN3_20170216_193100_President_Trump_Holds_News_Conference?start=3543&end=3553?q=trump (note the _two_ query strings in this URL)

Use a more robust approach to generating embed URLs and add some missing
test cases. Any unrecognized query params are ignored when generating
the embed URLs.


seanh:
@robertknight can you add instructions on how to test this? e.g. a real IA video URL that I can embed in an annotation, that is broken on master but works on this branch.

robertknight:
@seanh - See https://hypothes-is.slack.com/archives/C1M8NH76X/p1508942654000404 for original issue reproduction.

When I tested the IA URL given there however, my experience was that the embed still _worked_ despite the invalid URLs being generated. For correctness sake, I decided to fix the URL generation anyway. See the test cases at https://github.com/hypothesis/client/pull/561/files#diff-84e04b252cb2cab90622ea7aef0c6f5bR98 for an example of what changed when generating the `/embed` links from a given `/details` link.

seanh:
Works here :+1:

PR #563, opened 2017-10-26
Add missing page index bounds check

robertknight:
When anchoring an annotation with a position and a quote selector, if
the position selector fails, then PDF anchoring searches page contents
starting with those pages nearest the position.

If the position selector's `start` offset was greater than the length of
the PDF's text, `prioritizePages` would try to fetch the text of page
indexes beyond the valid range, causing PDF.js to throw an exception and
quote anchoring to fail.

Fix this by adding a missing bounds check.

This is a partial fix for #558. It fixes anchoring of one of two test
annotations on that page. The other fails due to differences in the
extracted text between the HTML and PDF versions of the article.

### DIFF
--- a/src/annotator/anchoring/pdf.coffee
+++ b/src/annotator/anchoring/pdf.coffee
@@ -97,7 +97,8 @@ findPage = (offset) ->
   #    150    | 2
   #
   count = (textContent) ->
-    if total + textContent.length > offset
+    lastPageIndex = PDFViewerApplication.pdfViewer.pagesCount - 1
+    if total + textContent.length > offset or index == lastPageIndex
       offset = total
       return Promise.resolve({index, offset, textContent})
     else
--- a/src/annotator/anchoring/test/pdf-test.js
+++ b/src/annotator/anchoring/test/pdf-test.js
@@ -35,7 +35,7 @@ var fixtures = {
   ],
 };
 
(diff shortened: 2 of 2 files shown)


seanh:
@robertknight I'm not sure how to reproduce the problem. On the page that https://github.com/hypothesis/client/issues/558 links to I see two annotations, both of them anchor (using production client), neither orphan. Also confused because it's an HTML page not a PDF.

robertknight:
@seanh - The easiest way to reproduce the problem (IMO) is to look at this test case: https://github.com/hypothesis/client/pull/563/files#diff-15689de7b8aaba1c6fa37541583f3c01R176 and check out the version of `pdf.coffee` from master without the fix.

In the test case, we're assuming that there is _some_ annotation with a selector that has a `TextPositionSelector` type and a position value which is greater than the length of the document's actual text.

The bug report (#558) linked to an HTML version of the article but if you go to the PDF version via the "Download PDF" link on http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0183175 you should see the same annotations but orphaned. This PR should fix _one_ of the orphans.

seanh:
Tested - it does make the one production annotation anchor instead of orphaning.

seanh:
:+1:

PR #567, opened 2017-11-06
Add OAuth client registration steps to developer docs

robertknight:
Add additional steps to the documentation for using a local build of the client with a local instance of "h". These steps will be mandatory once cookie auth is no longer supported.


seanh:
I added a note that you need to be logged in to h as an admin to access the oauthclients page

seanh:
:+1: easy to follow, worked here

PR #572, opened 2017-11-09
Fix issue that could lead to user being logged into normal Hypothesis account on websites using third-party accounts

### LINKED ISSUE #571: User can end up logged into regular Hypothesis account on a page configured to use a third party account
**Steps to reproduce**

1. Run a local build of the client against a local h service
2. Login to the client using a regular account on the local h service
3. Run the publisher account test site https://github.com/hypothesis/publisher-account-test-site and do not login to the test site

**Expected result**

Client shows publisher group. User is not logged in.

**Actual result**

Client shows normal Hypothesis groups. User is logged in to their normal Hypothesis account.

**Notes**

The issue is that in this context, the host page (the publisher account test site) includes a `services` config for the client but the `grantToken` is null. If the grant token is null we [fall back to trying to load access tokens from local storage](https://github.com/hypothesis/client/blob/4e44239165c813eebdc17880bbfb57dbbc516618/src/sidebar/oauth-auth.js#L228).

Since the `apiUrl` config param is the same for this "third party service" as it is for the regular Hypothesis service, the [`storageKey()` function](https://github.com/hypothesis/client/blob/4e44239165c813eebdc17880bbfb57dbbc516618/src/sidebar/oauth-auth.js#L114) returns the same key that was used to store OAuth tokens for the regular Hypothesis service.

Consequently the client ends up using the saved access token and the user ends up logged in to their regular Hypothesis account.

robertknight:
This fixes an issue where the client could end up using OAuth tokens for a regular Hypothesis account from a previous session when visiting a website which configures the client to use third-party accounts. This happened if the user was anonymous on the third-party website and so the "grantToken" property in the service configuration was `null`.

Fixes #571

### DIFF
--- a/src/sidebar/oauth-auth.js
+++ b/src/sidebar/oauth-auth.js
@@ -264,9 +264,18 @@ function auth($http, $rootScope, $window,
       }
 
       if (Date.now() > token.expiresAt) {
+        var shouldPersist = true;
+
+        // If we are using automatic login via a grant token, do not persist the
+        // initial access token or refreshed tokens.
+        var cfg = serviceConfig(settings);
+        if (cfg && typeof cfg.grantToken !== 'undefined') {
+          shouldPersist = false;
+        }
+
         // Token expired. Attempt to refresh.
         tokenInfoPromise = refreshAccessToken(token.refreshToken, {
-          persist: true,
+          persist: shouldPersist,
         }).catch(() => {
           // If refreshing the token fails, the user is simply logged out.
           return null;
--- a/src/sidebar/oauth-auth.js
+++ b/src/sidebar/oauth-auth.js
@@ -94,14 +94,6 @@ function auth($http, $rootScope, $window,
     return $http.post(url, data, requestConfig);
   }
 
-  function grantTokenFromHostPage() {
-    var cfg = serviceConfig(settings);
-    if (!cfg) {
-      return null;
-    }
-    return cfg.grantToken;
-  }
-
   /**
    * Return the storage key used for storing access/refresh token data for a given
    * annotation service.
@@ -223,17 +215,25 @@ function auth($http, $rootScope, $window,
    */
   function tokenGetter() {
     if (!tokenInfoPromise) {
-      var grantToken = grantTokenFromHostPage();
+      var cfg = serviceConfig(settings);
 
-      if (grantToken) {
-        // Exchange host-page provided grant token for a new access token.
-        tokenInfoPromise = exchangeJWT(grantToken).then((tokenInfo) => {
-          return tokenInfo;
-        }).catch(function(err) {
-          showAccessTokenExpiredErrorMessage(
-            'You must reload the page to annotate.');
-          throw err;
-        });
+      // Check if automatic login is being used, indicated by the presence of
+      // the 'grantToken' property in the service configuration.
+      if (cfg && typeof cfg.grantToken !== 'undefined') {
+        if (cfg.grantToken) {
+          // User is logged-in on the publisher's website.
+          // Exchange the grant token for a new access token.
+          tokenInfoPromise = exchangeJWT(cfg.grantToken).then((tokenInfo) => {
+            return tokenInfo;
+          }).catch(function(err) {
+            showAccessTokenExpiredErrorMessage(
+              'You must reload the page to annotate.');
+            throw err;
+          });
+        } else {
+          // User is anonymous on the publisher's website.
+          tokenInfoPromise = Promise.resolve(null);
+        }
       } else if (authCode) {
         // Exchange authorization code retrieved from login popup for a new
         // access token.
--- a/src/sidebar/test/oauth-auth-test.js
+++ b/src/sidebar/test/oauth-auth-test.js
@@ -181,12 +181,6 @@ describe('sidebar.oauth-auth', function () {
       });
     });
 
-    it('should not persist access tokens fetched using a grant token', function () {
-      return auth.tokenGetter().then(() => {
-        assert.notCalled(fakeLocalStorage.setObject);
-      });
-    });
-
     context('when the access token request fails', function() {
       beforeEach('make access token requests fail', function () {
         fakeHttp.post.returns(Promise.resolve({status: 500}));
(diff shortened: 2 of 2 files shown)


seanh:
Works here

src/sidebar/oauth-auth.js
```diff
@@ -223,17 +215,25 @@ function auth($http, $rootScope, $window,
...
+            return tokenInfo;
+          }).catch(function(err) {
+            showAccessTokenExpiredErrorMessage(
+              'You must reload the page to annotate.');
+            throw err;
+          });
+        } else {
+          // User is anonymous on the publisher's website.
+          tokenInfoPromise = Promise.resolve(null);
+        }
```
seanh: The logic here, if I understand it correctly, is that whenever there's a truthy `serviceConfig()` (which currently I believe is equivalent to whenever the embedding page contains a `js-hypothesis-config` or a `window.hypothesisConfig()` with a `services` object) then first-party accounts are disabled and third-party accounts are used.

But in the longer term I think what we want is for the default hypothes.is service to be a service like any other, one that appears in the `services` object alongside any other service configs, it's just that it's found in a `services` object in the `app.html` page instead of in the embedding page.

Do you agree? And if so, do we need another way of deciding when to enable third-party accounts, instead of checking for the presence of a service config?

I guess this also introduces another way to disable Hypothesis on a site, by embedding the client and including a services config (e.g. one with an authority that doesn't exist server-side and a grant token that doesn't work).

src/sidebar/test/oauth-auth-test.js
```diff
@@ -359,6 +353,33 @@ describe('sidebar.oauth-auth', function () {
...
+    [{
+      // User is logged-in on the publisher's website.
+      authority: 'publisher.org',
+      grantToken: 'a.jwt.token',
+      expectedToken: 'firstAccessToken',
+    },{
+      // User is anonymous on the publisher's website.
+      authority: 'publisher.org',
+      grantToken: null,
+      expectedToken: null,
```
seanh: Do we need to add more cases to the `tokenGetter()` tests above this? In `describe('#tokenGetter'` I see `it('should request an access token if a grant token was provided',` but that's all. Looking at the code I think there are a number of cases to test, seems a likely candidate for a parametrized test like this one:

1. No service config at all
2. Service config contains no grantToken
3. Service config contains a grantToken but it's null. I believe other falsey values e.g. `false` may behave the same as null too.
4. Service config contains a truthy grantToken

Case 2 is also tested currently, in `it('should return null if no grant token was provided'`, but it's way further down in the file than `it('should request an access token if a grant token was provided',`, may be worth bringing them together in one parametrized test and adding additional cases.

src/sidebar/test/oauth-auth-test.js
```diff
@@ -359,6 +353,33 @@ describe('sidebar.oauth-auth', function () {
...
+          assert.notCalled(fakeLocalStorage.setObject);
+        });
+      });
+
+      it('should not read persisted access tokens if a grant token was set', () => {
+        fakeSettings.services = [{ authority, grantToken }];
+        return auth.tokenGetter().then(token => {
+          assert.equal(token, expectedToken);
+          assert.notCalled(fakeLocalStorage.getObject);
+        });
```
seanh: Should these tests be in `describe('#tokenGetter'`? They look like tests for `tokenGetter()` to me, although I think the existence of `describe('persistence of tokens to storage'` confuses things (all the other `describe`s in the file are for describing functions: `describe('#tokenGetter'`, `describe('#login'`, `describe('#logout',`, and then this one is different).

src/sidebar/oauth-auth.js
```diff
@@ -264,9 +264,18 @@ function auth($http, $rootScope, $window,
...
 
       if (Date.now() > token.expiresAt) {
+        var shouldPersist = true;
+
+        // If we are using automatic login via a grant token, do not persist the
+        // initial access token or refreshed tokens.
+        var cfg = serviceConfig(settings);
+        if (cfg && typeof cfg.grantToken !== 'undefined') {
+          shouldPersist = false;
+        }
```
seanh: `tokenGetter()` is getting pretty long (over 70 lines). It probably needs to be split up very soon, if not overdue. It's doing a lot of different things

1. Decides whether automatic login is in use
2. Exchanges grant token for access token
3. Shows error for expired access token
4. Exchanges auth code for access token
5. Attempts to load tokens from previous session
6. Decides whether tokens should be persisted
7. Attempts to refresh tokens
8. ...

This is probably a sign that `oauth-auth.js` itself has gotten too long and needs to be broken up soon.

src/sidebar/oauth-auth.js
robertknight: > The logic here, if I understand it correctly, is that whenever there's a truthy serviceConfig() (which currently I believe is equivalent to whenever the embedding page contains a js-hypothesis-config or a window.hypothesisConfig() with a services object) then first-party accounts are disabled and third-party accounts are used.

Here we aren't testing for first vs third party services but rather the use of automatic login via a grant token. That is what matters because this method is responsible for obtaining an access token from somewhere. In future if h becomes "just another annotation service" then it would, in theory, be eligible to use grant tokens for login as well.

> But in the longer term I think what we want is for the default hypothes.is service to be a service like any other, one that appears in the services object alongside any other service configs, it's just that it's found in a services object in the app.html page instead of in the embedding page.

Yes, that is where I expect we'll eventually end up.

src/sidebar/oauth-auth.js
robertknight: Hmm. There are three things going on here:

1. Fetch a token from one of three sources (automatic login, auth code, browser storage) if we haven't got one yet.
2. Await the fetched token
3a. Return the fetched token if valid
3b. Otherwise, refresh the token and await the result.

I could split out (1) into separate function but I'd prefer to do that as a separate PR.

robertknight:
Depends on https://github.com/hypothesis/client/pull/573, will need rebasing after that is merged.

src/sidebar/test/oauth-auth-test.js
robertknight: Yes. Reading tokens from & writing tokens to storage is part of what is triggered by by `tokenGetter()`.

robertknight:
Rebased on https://github.com/hypothesis/client/pull/573

seanh:
Rebased

src/sidebar/oauth-auth.js
```diff
@@ -223,17 +215,25 @@ function auth($http, $rootScope, $window,
...
-          showAccessTokenExpiredErrorMessage(
-            'You must reload the page to annotate.');
-          throw err;
-        });
+      var cfg = serviceConfig(settings);
+
+      // Check if automatic login is being used, indicated by the presence of
+      // the 'grantToken' property in the service configuration.
+      if (cfg && typeof cfg.grantToken !== 'undefined') {
+        if (cfg.grantToken) {
```
seanh: As [discussed in Slack](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1510596981000246?thread_ts=1510583846.000041&cid=C4K6M7P5E) there's an unhandled case here when there _is_ a service config but it _doesn't_ contain any `grantToken` (not even `null`). This should be fixed elsewhere, in a separate PR: a service config without a `grantToken` should be considered invalid and rejected. The docs for `grantToken` need to be updated to say that it's required and does _not_ default to `null`.

src/sidebar/test/oauth-auth-test.js
```diff
@@ -367,6 +388,20 @@ describe('sidebar.oauth-auth', function () {
       });
     });
 
+    function expireAndRefreshAccessToken() {
```
robertknight: Add a docstring here.

seanh:
Following a Slack chat to explain some of the details we're gonna merge this now in order to not block other ongoing work, but @robertknight has some good ideas for refactoring this file in follow up pull requests that should make the file a lot easier to follow

PR #573, opened 2017-11-13
Re-arrange OAuth service tests

robertknight:
_Original context: https://github.com/hypothesis/client/pull/572/files#r150519749_

Move the tests for fetching and persistence of tokens from/to storage
into the `describe('#tokenGetter', ...` block since this is the method
called from outside the service that initiates the fetching /
persistence.

Also reword a couple of the test cases to be clearer now that the OAuth
service now handles fetching tokens using methods other than exchanging
grant tokens.

### DIFF
--- a/src/sidebar/test/oauth-auth-test.js
+++ b/src/sidebar/test/oauth-auth-test.js
@@ -169,7 +169,7 @@ describe('sidebar.oauth-auth', function () {
   });
 
   describe('#tokenGetter', function () {
-    it('should request an access token if a grant token was provided', function () {
+    it('exchanges the grant token for an access token if provided', function () {
       return auth.tokenGetter().then(function (token) {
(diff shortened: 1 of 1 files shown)


src/sidebar/test/oauth-auth-test.js
```diff
@@ -372,6 +368,8 @@ describe('sidebar.oauth-auth', function () {
     });
 
     it('persists refreshed tokens to storage', () => {
+      fakeSettings.services = [];
```
robertknight: Previously this setup was inherited from the `beforeEach` in the parent block, but now it needs to be provided with each test case.

Setup "inheritance" is one reason why wrapper blocks like `describe`/`context` should not be deeply nested. As we've discussed before, pytest fixtures handles this in a much nicer way.

PR #574, opened 2017-11-13
Move new note btn next to selection tabs.

sheetaluk:


robertknight:
Are you intending to finish this @sheetaluk or should we close it?

PR #555, opened 2017-10-10
Change UI for Elife.

sheetaluk:
Added config options for the following:

disableToolbarCloseBtn: false/true
Disables and hides the close button on the toolbar

disableToolbarMinimizeBtn: true/false
Disables and hides the minimize button on the toolbar

disableToolbarHighlightsBtn: true/false
Disables and hides the highlights button on the toolbar

disableToolbarNewNoteBtn: true/false
Disables and hides the new note button on the toolbar

disableBucketBar: true/false
Disables and hides the bucket bar

enableSidebarDropShadow: true/false
Turns on the dropshadow for the sidebar

theme: 'clean'
Turns on the 'clean' theme for the sidebar. By default the classic view(with annotation cards) is turned on, which the rest of the Hypothesis users will see.

usernameUrl: 'someurl/'
The url to direct a user to when their username in the annotation is clicked.

PR #576, opened 2017-11-14
Add config option to enable the new note button when the notes tab is selected.

sheetaluk:


robertknight:
For future reference, we're adding this as a client side config option rather than a feature flag in the h service because we want to initially enable this for a publisher (specifically eLife) using the new look developed in https://github.com/hypothesis/client/pull/555 but the service doesn't yet support creating feature cohorts that include "all users using account namespace 'foo.org'".

PR #578, opened 2017-11-15
Remove orphans tab feature flag checks

robertknight:
This PR removes a set of obsolete code paths for supporting the case where the "Orphans" tab was disabled. After this is merged, we can remove the "orphans_tab" feature flag from the service.

There is no time pressure on this one, but I've had it sitting around for a while locally and we should always be getting rid of code that no longer serves a purpose.


seanh:
Agree, this is a lot of code removal  and simplification, it's important to do. Works here and looks good to me

PR #580, opened 2017-11-15
Add config option to display a clean onboarding look to users.

sheetaluk:


src/sidebar/templates/sidebar-tutorial.html
```diff
@@ -41,3 +41,36 @@ <h1 class="sidebar-tutorial__header">How to get started</h1>
...
+        Create page level notes
+        <button class="sidebar-tutorial__list-item-new-note-btn" h-branding="ctaBackgroundColor">
+          + New note
+        </button>
+      </p>
+    </li>
+    <li class="sidebar-tutorial__list-item sidebar-tutorial__list-item--is-theme-clean">
+      <p class="sidebar-tutorial__list-item-content">
+        View annotations through your profile
+        <i class="h-icon-account ng-scope"></i>
```
robertknight: What is the `ng-scope` for here?

src/styles/sidebar-tutorial.scss
```diff
@@ -44,10 +54,45 @@
   color: $gray-light;
 }
 
+.sidebar-tutorial__list-item--is-theme-clean {
+  font-size: 14px;
+}
+
+.sidebar-tutorial__list-item--is-theme-clean::before {
+  content: counter(sidebar-tutorial__list) ".";
+  display: table-cell; /* aha! */
```
robertknight: What is this for?

src/styles/sidebar-tutorial.scss
```diff
@@ -44,10 +54,45 @@
...
+.sidebar-tutorial__list-item-annotate {
+  background-color: #b9d9fa;
+}
+
+.sidebar-tutorial__list-item-new-note-btn {
+  background-color: #626262;
+  border: none;
+  border-radius: 3px;
+  color: #fff;
+  display: -ms-flexbox;
```
robertknight: CSS property values should be unprefixed (ie. `display: flex`). `-ms-flexbox` is a legacy value for IE 10/11. Our CSS preprocessor will add prefixes for the browsers we support. The exceptions are a few rare properties that only exist in one browser.

src/styles/sidebar-tutorial.scss
```diff
@@ -44,10 +54,45 @@
...
 .sidebar-tutorial__list-item-content {
   margin-top: 1em; /* Put vertical space in-between list items, equal to the
                       space between normal paragraphs.
                       Note: This also puts the same amount of space above the
                       first list item (i.e. between the list and whatever's
                       above it). */
 }
+
+.sidebar-tutorial__list-item-annotate {
+  background-color: #b9d9fa;
```
robertknight: Is this just for testing purposes, do we have variables that we'll eventually use here?

sheetaluk:
@robertknight I cleaned up all the placeholders and copy-paste errors. Please take a look!

robertknight:
Note for future reference: For testing purposes in a local dev environment, the only way the sidebar tutorial can be re-enabled for a user who has dismissed it currently is via a direct DB update:

```sh
docker run -it --link postgres:postgres --rm postgres sh -c 'exec psql -h "$POSTGRES_PORT_5432_TCP_ADDR" -p "$POSTGRES_PORT_5432_TCP_PORT" -U postgres '

postgres=# update "user" set sidebar_tutorial_dismissed = false;
```

robertknight:
This is what the combination of all the new UI customization options currently looks like on the publisher account test site:

screenshot 2017-11-16 10 13 39

screenshot 2017-11-16 10 13 22

If this option is enabled on its own for Hypothesis it looks like this:

screenshot 2017-11-16 10 05 41

src/sidebar/templates/sidebar-tutorial.html
```diff
@@ -41,3 +41,36 @@ <h1 class="sidebar-tutorial__header">How to get started</h1>
...
+  <h1 class="sidebar-tutorial__header sidebar-tutorial__header--is-theme-clean">
+    <i class="h-icon-annotate sidebar-tutorial__header-annotate" h-branding="accentColor"></i>
+    Start annotating
+  </h1>
+  <ol class="sidebar-tutorial__list">
+    <li class="sidebar-tutorial__list-item sidebar-tutorial__list-item--is-theme-clean">
+      <div class="sidebar-tutorial__list-item-content">
+        Select some text to
+          <span class="sidebar-tutorial__list-item-annotate">annotate</span>
+          <svg-icon width="3px" height="10px" class="sidebar-tutorial__list-item-cursor" name="'cursor'"></svg-icon>
```
robertknight: The `width` and `height` properties here are redundant and can be removed since we're now applying the styling in CSS.

src/sidebar/templates/sidebar-tutorial.html
```diff
@@ -41,3 +41,36 @@ <h1 class="sidebar-tutorial__header">How to get started</h1>
...
+        or highlight.
+      </div>
+    </li>
+    <li class="sidebar-tutorial__list-item sidebar-tutorial__list-item--is-theme-clean">
+      <div class="sidebar-tutorial__list-item-content sidebar-tutorial__list-item-content--is-theme-clean">
+        Create page level notes
+        <button class="sidebar-tutorial__list-item-new-note-btn" h-branding="ctaBackgroundColor">
```
robertknight: Hmm ... this isn't actually clickable so it might be a bit surprising that it has the affordances of a button (eg. a cursor pointer).

src/styles/sidebar-tutorial.scss
```diff
@@ -44,10 +54,56 @@
...
                       above it). */
 }
+
+.sidebar-tutorial__list-item-annotate {
+  background-color: $highlight-color-focus;
+  padding: 0px 3px;
+}
+
+.sidebar-tutorial__list-item-new-note-btn {
+  background-color: #626262;
```
robertknight: This should use the `$color-dove-gray` variable for consistency with other elements that are styled using the same palette element from the color palette.

src/annotator/config/index.js
```diff
@@ -23,6 +23,7 @@ function configFrom(window_) {
     disableToolbarHighlightsBtn: settings.hostPageSetting('disableToolbarHighlightsBtn'),
     disableToolbarNewNoteBtn: settings.hostPageSetting('disableToolbarNewNoteBtn'),
     disableBucketBar: settings.hostPageSetting('disableBucketBar'),
+    enableCleanOnboardingTheme: settings.hostPageSetting('enableCleanOnboardingTheme'),
```
robertknight: Perhaps for both consistency with how we are handling the "New note" button and also to make it clear that this isn't yet an officially supported setting, we ought to name it something like `enableExperimentalNewOnboarding`? 🤔 .

The new-look onboarding actually works fine whether or not the "clean" theme is being used, so perhaps we can consider it a separate thing.

PR #581, opened 2017-11-16
Fix default values for settings not being used in the browser extension

robertknight:
The logic for skipping host page configuration of settings that are
ignored by the browser extension did not take into account any default
values configured via the `defaultValue` property.

Also fix and add several related tests:

 - Correct the test that checks behavior when `defaultValue` is not set
   to use `undefined` rather than `null` to indicate a missing
   `defaultValue`.

 - Add test case for when a default value is overridden.

 - Add test case for default values in browser extensions.


seanh:
:+1: Works here and looks good, thanks! I agree with what you've said in an issue elsewhere that all these new settings need some tidying up

PR #597, opened 2017-11-21
Remove checks for `flag_action` feature flag.

robertknight:
This has now been enabled for all users for several months.

### DIFF
--- a/src/sidebar/components/annotation.js
+++ b/src/sidebar/components/annotation.js
@@ -471,13 +471,8 @@ function AnnotationController(
   };
 
   this.canFlag = function () {
-    if (session.state.userid === self.annotation.user) {
-      return false;
-    }
-    if (persona.isThirdPartyUser(self.annotation.user, settings.authDomain)) {
-      return true;
-    }
-    return features.flagEnabled('flag_action');
+    // Users can flag any annotations except their own.
+    return session.state.userid !== self.annotation.user;
   };
 
   this.isFlagged = function() {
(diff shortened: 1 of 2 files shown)


src/sidebar/components/annotation.js
```diff
@@ -471,13 +471,8 @@ function AnnotationController(
   };
 
   this.canFlag = function () {
-    if (session.state.userid === self.annotation.user) {
-      return false;
-    }
-    if (persona.isThirdPartyUser(self.annotation.user, settings.authDomain)) {
-      return true;
-    }
-    return features.flagEnabled('flag_action');
```
seanh: Good to see another "if is third party user do X, else do Y" branch removed, too

src/sidebar/components/test/annotation-test.js
```diff
@@ -833,40 +833,15 @@ describe('annotation', function() {
...
-
-        fakeFeatures.flagEnabled.returns(true);
-        assert.equal(controller.canFlag(), true);
-      });
-
       it('returns false if the user signed in is the same as the author of the annotation', function () {
         var ann = fixtures.defaultAnnotation();
         var controller = createDirective(ann).controller;
-
-        fakeFeatures.flagEnabled.returns(true);
```
seanh: Can `fakeFeatures` be removed entirely from `annotation-test.js` now? And the `features` dependency from `annotation.js`

robertknight: Done, although there is a decent chance that we'll need to re-add it at some point in future.

PR #601, opened 2017-11-22
Rename account ID utility functions

robertknight:
sidebar/filter/persona.js used to contain an Angular filter for use in
templates. However following several refactorings it now contains a
simple set of utility functions for parsing account IDs. This renames
the module accordingly and puts it alongside other utility function
modules.


seanh:
@sheetaluk Did you see that @robertknight requested your review on this? I'm not sure why the Jenkins build had failed, I just replayed it and it succeeded, some temporary hiccup. The Codecov fails look spurious. So I think this is ready to be reviewed?

robertknight:
> I realise that we are deriving isThirdPartyUser and username from the account id, but would it be more obvious that these methods are available in account or account-id?

The latter seems slightly more obvious to me but I think either would be reasonable. I'm going to merge this as-is for the moment but I'd be happy to rename it in future if we decide to put non-userid related functions in the same module.

PR #603, opened 2017-11-23
Extract OAuth client into a separate module

robertknight:
This PR simplifies the authorization service in oauth-auth.js by extracting the code which handles the details of interaction with the OAuth endpoints in the "h" service into a separate `OAuthClient` class in sidebar/util/oauth-client.js

This has several benefits:

- Separates the concerns of _how_ to interact with the OAuth endpoints from _when_ to do so and with _what_ values.
- Makes tests for the authorization service easier to follow because they just stub Promise-returning methods in `OAuthClient` rather than having to stub the full HTTP requests and auth window interaction.
- The oauth-client module, although not explicitly packaged for re-use yet, is something that would be much easier for third parties to use as a reference for implementing their own clients as it has minimal dependencies on other parts of the application or Angular.

### DIFF
--- a/src/sidebar/oauth-auth.js
+++ b/src/sidebar/oauth-auth.js
@@ -230,6 +230,13 @@ function auth($http, $rootScope, $window, OAuthClient,
         // Token expired. Attempt to refresh.
         tokenInfoPromise = refreshAccessToken(token.refreshToken, {
           persist: shouldPersist,
+        }).then(token => {
+          // Sanity check that prevents an infinite loop. Mostly useful in
+          // tests.
+          if (Date.now() > token.expiresAt) {
+            throw new Error('Refreshed token expired in the past');
+          }
+          return token;
         }).catch(() => {
           // If refreshing the token fails, the user is simply logged out.
           return null;
--- a/src/sidebar/oauth-auth.js
+++ b/src/sidebar/oauth-auth.js
@@ -11,10 +11,17 @@ var serviceConfig = require('./service-config');
  */
 
 /**
- * OAuth-based authorization service.
+ * Authorization service.
  *
- * A grant token embedded on the page by the publisher is exchanged for
- * an opaque access token.
+ * This service is responsible for acquiring access tokens for making API
+ * requests and making them available via the `tokenGetter()` method.
+ *
+ * Access tokens are acquired via the OAuth authorization flow, loading valid
+ * tokens from a previous session or, on some websites, by exchanging a grant
+ * token provided by the host page.
+ *
+ * Interaction with OAuth endpoints in the annotation service is delegated to
+ * the `OAuthClient` class.
  */
 // @ngInject
 function auth($http, $rootScope, $window, OAuthClient,
--- a/src/sidebar/oauth-auth.js
+++ b/src/sidebar/oauth-auth.js
@@ -1,7 +1,5 @@
 'use strict';
 
-var queryString = require('query-string');
-
 var events = require('./events');
 var resolve = require('./util/url-util').resolve;
 var serviceConfig = require('./service-config');
@@ -28,8 +26,8 @@ var serviceConfig = require('./service-config');
  * an opaque access token.
  */
 // @ngInject
-function auth($http, $rootScope, $window,
-              apiRoutes, flash, localStorage, random, settings) {
+function auth($http, $rootScope, $window, OAuthClient,
+              apiRoutes, flash, localStorage, settings) {
 
   /**
    * Authorization code from auth popup window.
@@ -45,6 +43,9 @@ function auth($http, $rootScope, $window,
    * @type {Promise<TokenInfo|null>}
    */
   var tokenInfoPromise;
+
+  /** @type {OAuthClient} */
+  var client;
 
   /**
    * Absolute URL of the `/api/token` endpoint.
@@ -64,34 +65,6 @@ function auth($http, $rootScope, $window,
         timeOut: 0,
       }
     );
-  }
-
-  /**
-   * Return a new TokenInfo object from the given tokenUrl endpoint response.
-   * @param {Object} response - The HTTP response from a POST to the tokenUrl
-   *                            endpoint (an Angular $http response object).
-   * @returns {TokenInfo}
-   */
-  function tokenInfoFrom(response) {
-    var data = response.data;
-    return {
-      accessToken:  data.access_token,
-
-      // Set the expiry date to some time slightly before that implied by
-      // `expires_in` to account for the delay in the client receiving the
-      // response.
-      expiresAt: Date.now() + ((data.expires_in - 10) * 1000),
-
-      refreshToken: data.refresh_token,
-    };
-  }
-
-  function formPost(url, data) {
-    data = queryString.stringify(data);
-    var requestConfig = {
-      headers: {'Content-Type': 'application/x-www-form-urlencoded'},
-    };
-    return $http.post(url, data, requestConfig);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/sidebar/oauth-auth.js
```diff
@@ -13,23 +11,21 @@ var serviceConfig = require('./service-config');
  */
 
 /**
- * An object holding the details of an access token from the tokenUrl endpoint.
- * @typedef {Object} TokenInfo
- * @property {string} accessToken  - The access token itself.
- * @property {number} expiresAt    - The date when the timestamp will expire.
- * @property {string} refreshToken - The refresh token that can be used to
- *                                   get a new access token.
- */
```
robertknight: This definition moved unchanged into oauth-client.js

src/sidebar/oauth-auth.js
```diff
@@ -13,23 +11,21 @@ var serviceConfig = require('./service-config');
...
- * an opaque access token.
+ * This service is responsible for acquiring access tokens for making API
+ * requests and making them available via the `tokenGetter()` method.
+ *
+ * Access tokens are acquired via the OAuth authorization flow, loading valid
+ * tokens from a previous session or, on some websites, by exchanging a grant
+ * token provided by the host page.
+ *
+ * Interaction with OAuth endpoints in the annotation service is delegated to
+ * the `OAuthClient` class.
```
robertknight: This updates the previously obsolete JSDoc comment for the auth service.

src/sidebar/oauth-auth.js
```diff
@@ -66,34 +65,6 @@ function auth($http, $rootScope, $window,
...
-  }
-
-  function formPost(url, data) {
-    data = queryString.stringify(data);
-    var requestConfig = {
-      headers: {'Content-Type': 'application/x-www-form-urlencoded'},
-    };
-    return $http.post(url, data, requestConfig);
-  }
-
```
robertknight: `tokenInfoFrom`, `formPost`, `exchangeJWT` and `exchangeAuthCode` have all been moved into the `OAuthClient` class.

src/sidebar/oauth-auth.js
```diff
@@ -296,86 +255,12 @@ function auth($http, $rootScope, $window,
    * then exchange for access and refresh tokens.
    */
   function login() {
```
robertknight: The logic below for handling interaction with the OAuth popup window has moved into the `OAuthClient` class.

src/sidebar/util/test/fake-window.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+class FakeWindow {
```
robertknight: This fake implementation of `Window` was moved from oauth-auth-test.js mostly unchanged. It is mostly used by the `OAuthClient` tests but is still used by the authorization service tests that check for handling of "storage" events that occur when the user has the client open in multiple tabs and changes their login in one of them.

src/sidebar/test/oauth-auth-test.js
```diff
@@ -139,22 +69,30 @@ describe('sidebar.oauth-auth', function () {
...
     angular.mock.module('app', {
-      $http: fakeHttp,
+      $http: {},
       $window: fakeWindow,
       apiRoutes: fakeApiRoutes,
       flash: fakeFlash,
       localStorage: fakeLocalStorage,
-      random: fakeRandom,
       settings: fakeSettings,
+      OAuthClient: () => fakeClient,
```
robertknight: An unusual feature of JavaScript constructor functions is that they can _return_ an object which becomes the _result_ of the `new FooClass` call. This provides a convenient way to mock classes (rather than instances) in tests.

We actually do this implicitly all over the place in the code (Angular "services" are essentially class constructors).

src/sidebar/test/oauth-auth-test.js
```diff
@@ -1,76 +1,23 @@
 'use strict';
```
robertknight: Most of the changes in this file are removing stubbing of HTTP requests/responses and authorization window opening and replacing them with calls to a fake `OAuthClient` instance.

seanh:
I've tested this with both first and third-party accounts and all seems to be working. The only thing I didn't test was refreshing. Did you test that @robertknight ?

src/sidebar/util/oauth-client.js
```diff
@@ -0,0 +1,200 @@
...
+    this.revokeEndpoint = config.revokeEndpoint;
+
+    // Test seam
+    this.generateState = config.generateState || generateState;
+  }
+
+  /**
+   * Exchange an authorization code for access and refresh tokens.
+   */
+  exchangeAuthCode(code) {
```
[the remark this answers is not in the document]
robertknight: The way `OAuthClient` is used, a single instance is constructed by the auth service and requests are then made each time some authorization-related action needs to happen. Therefore the only data that can be held internally in `OAuthClient` is state which is unchanged once the client is constructed.

The auth code parameter, access token, refresh token etc. are different each time the methods are used.

src/sidebar/util/oauth-client.js
```diff
@@ -0,0 +1,200 @@
...
+   * Revoke an access and refresh token pair.
+   */
+  revokeToken(accessToken) {
+    return this._formPost(this.revokeEndpoint, { token: accessToken });
+  }
+
+  /**
+   * Prompt the user for permission to access their data.
+   *
+   * Returns an authorization code which can be passed to `exchangeAuthCode`.
```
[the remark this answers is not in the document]
robertknight: It does. I've clarified that in the return type JSDoc.

src/sidebar/util/test/oauth-client-test.js
```diff
@@ -0,0 +1,251 @@
...
+describe('sidebar.util.oauth-client', () => {
+  var fakeHttp;
+  var client;
+  var clock;
+  var config = {
+    clientId: '1234-5678',
+    authorizationEndpoint: 'https://annota.te/oauth/authorize',
+    tokenEndpoint: 'https://annota.te/api/token',
+    revokeEndpoint: 'https://annota.te/oauth/revoke',
+    generateState: () => 'notrandom',
```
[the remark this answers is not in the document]
robertknight: Yes. As you can see I haven't worried too much about preventing state param forgeries in the tests ;)

robertknight:
Thanks for the detailed review 🙂

>  I've made some comments requesting some docstrings to be added to, I think once that's done I'd be happy to merge this. 

I have added improved JSDoc comments for `OAuthClient` and associated types/functions. I hope this is good to merge now.

> Second, in object oriented design I think it's good to keep the data and the methods that operate on that data together in a single class. 

I explained the rationale for what is part of `OAuthClient`'s configuration vs what is a parameter in [a comment](https://github.com/hypothesis/client/pull/603#discussion_r153471839). TL;DR - I think it makes sense given the responsibilities and usage of the class that I've chosen here.

> I wonder if this class, and the file, should be named OAuth2Client in case we ever have a reason to also implement a client for OAuth 1? (Or who knows maybe one day 3)

I personally consider that "OAuth" is now a synonym for "OAuth 2" and is relatively unlikely to substantially change for a while. I can't think of a good reason to ever implement OAuth 1 - it was way more complex because of the need to work over plain HTTP, and we really shouldn't encourage that since the rest of the client assumes a secure connection to the server.

src/sidebar/util/test/oauth-client-test.js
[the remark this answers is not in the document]
seanh: It's just that it's supposed to be a stand-in for what would be a random string, but it says "not random". Was thinking something like `"RANDOM"`

src/sidebar/util/oauth-client.js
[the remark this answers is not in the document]
seanh: I don't understand this. Why can't an `OAuth` object hold state that it changes over time? Generally it's possible for objects to do so.

This was more of a suggestion for future direction though, I'm happy for this PR to merge as is

seanh:
@robertknight You said you wanted to check again that refresh tokens were working. I'm approving this so you can merge it when you're happy

robertknight:
I've retested token refresh. Merging...

PR #609, opened 2017-11-27
WIP - settings.pageGroups configures publisher-promoted groups in sidebar groups dropdown

gobengo:
This is a work in progress. Publishing to facilitate 11/27 meetings.

We're adding the ability for publishers to configure hypothesis/clients to feature one or more groups. So the client will show groups you're a member of, but now also groups the publisher want to promote.

This PR depends on functionality added to h in [this PR](https://github.com/gobengo/h/pull/1).

Publishers can configure one or more `pageGroups` (happy to change that to whatever) like:
```
window.hypothesisConfig = function () {
    return {
        pageGroups: [
            "http://h.hypothesis:5000/groups/2JKqAjYE/scipub-com-open-group",
            "http://h.hypothesis:5000/groups/wg54vyqQ/h-hypothesis-open",
        ]
    };
};
```

The idea is here is it's "Just the group URL" that you'd get by using a web browser to find the group. The client then requests JSON from that URL to resolve up-to-date group information, before rendering the groups in the groups dropdown.

Right now this PR just gets these groups rendering in the top-right groups dropdown using the same template as groups that you're a member of. Eventually I'm told it should look like [these designs](https://docs.google.com/presentation/d/1ESiIotb91xJnUj9v6M2R2qW0o7akT1MbwsIMPjab6Ho/edit#slide=id.g2576362aea_1_2).

This PR does:
* annotator/config - new hostPageSetting `pageGroups`
* sidebar/app
  * `newGroupList` - oy. I plan to just copy this over the existing group-list, or will rename to something new to reflect component with still-required design changes.
  * `debugFilter` - an angular template filter to debug values. I found it useful but can remove if asked.
* sidebar/components/new-group-list - Like existing group-list, but now binds to groups and uses new-group-list.html template
* src/sidebar/groups - `.pageGroups` will look at the configured value of group URLs, fetch them from h API, and return those objects. It'll also cache the value, so only the first call makes an HTTP request. In the future this may have to change if we want to allow the set of configured groups to change throughout the document lifetime
* sidebar/store
  * `createAPICall` changes to accomodate for the fact that we want to dereference those group URLs directly by their value. We don't need to look up the real URL in the routes list that other methods here use. So now createAPICall can create calls that just accept a url. `apiCall({ url: true })`
  * createAPICall request method defaults to GET

Questions
* [] - cool with `pageGroups` as configuration option and method names? Or what should I change it to?
* [] - new-group-list is not a good final name. What is? Or should I just make edits to the existing groups-list

What's next
* Implement new designs for the groups dropdown in sidebar

### DIFF
--- a/scripts/gulp/live-reload-server.js
+++ b/scripts/gulp/live-reload-server.js
@@ -99,7 +99,10 @@ function LiveReloadServer(port, config) {
 
                 // Open the sidebar when the page loads
                 openSidebar: true,
-                pageGroups: ["http://h.hypothesis:5000/groups/MGkYz9j2/http-test-localhost"]
+                pageGroups: [
+                  "http://h.hypothesis:5000/groups/MGkYz9j2/http-test-localhost", // will 404
+                  "http://h.hypothesis:5000/groups/2JKqAjYE/scipub-com-open-group",
+                ]
               };
             };
 
--- a/scripts/gulp/live-reload-server.js
+++ b/scripts/gulp/live-reload-server.js
@@ -99,6 +99,7 @@ function LiveReloadServer(port, config) {
 
                 // Open the sidebar when the page loads
                 openSidebar: true,
+                pageGroups: ["http://h.hypothesis:5000/groups/MGkYz9j2/http-test-localhost"]
               };
             };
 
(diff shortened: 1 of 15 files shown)


gobengo:
superceded by rebased https://github.com/hypothesis/client/pull/640

Issue #610, opened 2017-11-29
Onboarding tutorial and empty annotations/notes minitutorial appear at the same time for new users.


sheetaluk:
### Steps to reproduce
1. Login to the partner site as a new user.

### Expected behaviour
Only the onboarding tutorial should appear on the sidebar.

### Actual behaviour
In addition to the onboarding tutorial, an empty annotations/notes minitutorial also appears.

### Browser/system information
Chrome

PR #611, opened 2017-11-30
Show the shorter version of the empty annotations/notes message when …

sheetaluk:
…the sidebar tutorial is displayed

and the new note button is enabled.

Fixes https://github.com/hypothesis/client/issues/610

### DIFF
--- a/src/sidebar/components/test/selection-tabs-test.js
+++ b/src/sidebar/components/test/selection-tabs-test.js
@@ -5,6 +5,17 @@ var angular = require('angular');
 var util = require('../../directive/test/util');
 
 describe('selectionTabs', function () {
+  var fakeSession = {
+    state: {
+      preferences: {
+        show_sidebar_tutorial: false,
+      },
+    },
+  };
+  var fakeSettings = {
+    enableExperimentalNewNoteButton: false,
+  };
+
   before(function () {
     angular.module('app', [])
       .component('selectionTabs', require('../selection-tabs'));
@@ -15,13 +26,11 @@ describe('selectionTabs', function () {
     var fakeFeatures = {
       flagEnabled: sinon.stub().returns(true),
     };
-    var fakeSettings = {
-      enableExperimentalNewNoteButton: true,
-    };
 
     angular.mock.module('app', {
       annotationUI: fakeAnnotationUI,
       features: fakeFeatures,
+      session: fakeSession,
       settings: fakeSettings,
     });
   });
@@ -33,7 +42,6 @@ describe('selectionTabs', function () {
         totalAnnotations: '123',
         totalNotes: '456',
       });
-
       var tabs = elem[0].querySelectorAll('a');
 
       assert.include(tabs[0].textContent, 'Annotations');
@@ -48,8 +56,8 @@ describe('selectionTabs', function () {
         totalAnnotations: '123',
         totalNotes: '456',
       });
-
       var tabs = elem[0].querySelectorAll('a');
+
       assert.isTrue(tabs[0].classList.contains('is-selected'));
     });
 
@@ -59,8 +67,8 @@ describe('selectionTabs', function () {
         totalAnnotations: '123',
         totalNotes: '456',
       });
-
       var tabs = elem[0].querySelectorAll('a');
+
       assert.isTrue(tabs[1].classList.contains('is-selected'));
     });
 
@@ -70,6 +78,7 @@ describe('selectionTabs', function () {
         totalAnnotations: '123',
         totalNotes: '456',
       });
+
       assert.isFalse(elem[0].querySelectorAll('.selection-tabs')[0].classList.contains('selection-tabs--theme-clean'));
     });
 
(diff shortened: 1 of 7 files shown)


sheetaluk:
@robertknight I think `/filters/sidebar-tutorial.js` is in the wrong place. Its not really a filter used in templates. Is `/util` a better place?

src/sidebar/filter/sidebar-tutorial.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+/**
+ * Returns true if the sidebar tutorial has to be shown to a user for a given session.
+ */
+function show(sessionState) {
```
robertknight: - The `preferences` property is a non-optional field in the profile type as returned by `/api/profile`, so you don't need to check for it.
- Re. your question about where this should go, the `src/sidebar/filter` directory refers to Angular filters, rather than utility functions. There are two places in the code for this kind of logic:

  1. A simple function in a module in the `util/` directory (eg. in say `util/session` or something). If you do this, please make sure the function name makes sense on its own (eg. `shouldShowSidebarTutorial`).
  2. A "selector" function in an app state module (eg. `src/sidebar/reducers/session` in this case) which takes the app state and returns something derived from it.

Also, one related note - the name "session" in various modules is a hangover from when the data came from h's `/app` endpoint, which returned data about the user's cookie session on the website. At some point we should change it to "profile" everywhere since it comes from `/api/profile` now and no longer necessarily reflects the user's state on the website (eg. if using third party accounts).

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -87,28 +96,105 @@ describe('selectionTabs', function () {
...
+      fakeSession.state.preferences.show_sidebar_tutorial = false;
+      fakeSettings.enableExperimentalNewNoteButton = false;
+
+      it('should display the longer version of the no notes message when there are no notes', function () {
+        var elem = util.createDirective(document, 'selectionTabs', {
+          selectedTab: 'note',
+          totalAnnotations: '10',
+          totalNotes: 0,
+        });
+        var unavailableMsg = elem[0].querySelectorAll('.annotation-unavailable-message__label')[0];
```
robertknight: You can use `querySelector(selector)` rather than `querySelectorAll(selector)[0]` if you are only interested in the first element.

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -87,28 +96,105 @@ describe('selectionTabs', function () {
...
+        var unavailableTutorial = elem[0].querySelectorAll('.annotation-unavailable-message__tutorial')[0];
+        var noteIcon = unavailableTutorial.querySelectorAll('i')[0];
+
+        assert.include(unavailableMsg.textContent, 'There are no annotations in this group.');
+        assert.include(unavailableTutorial.textContent, 'Create one by selecting some text and clicking the');
+        assert.isTrue(noteIcon.classList.contains('h-icon-annotate'));
+      });
+
+      it('should not display the new note button when the notes tab is active and the new note button is disabled', function () {
```
robertknight: These tests are in a `when the sidebar tutorial is not displayed` context block, but they apply whether or not the sidebar tutorial is shown, don't they? If so, then they should be outside the context block. Also, don't we already have tests for this?

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -87,28 +96,108 @@ describe('selectionTabs', function () {
...
+        var elem = util.createDirective(document, 'selectionTabs', {
+          selectedTab: 'note',
+          totalAnnotations: '10',
+          totalNotes: 0,
+        });
+        var msg = elem[0].querySelector('.annotation-unavailable-message__label');
+
+        assert.include(msg.textContent, 'There are no page notes in this group.');
+        assert.notInclude(msg.textContent, 'Create one by clicking the');
+        assert.notInclude(msg.textContent, 'Create one by selecting some text and clicking the');
```
robertknight: I think this line was included by mistake? You never need to select text to create a note.

robertknight: I think this line was included by mistake? You never need to select text to create a note.

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -87,28 +96,108 @@ describe('selectionTabs', function () {
...
+      it('should display the shorter version of the no annotations message when there are no annotations', function () {
+        var elem = util.createDirective(document, 'selectionTabs', {
+          selectedTab: 'annotation',
+          totalAnnotations: 0,
+          totalNotes: '10',
+        });
+        var msg = elem[0].querySelector('.annotation-unavailable-message__label');
+
+        assert.include(msg.textContent, 'There are no annotations in this group.');
+        assert.notInclude(msg.textContent, 'Create one by clicking the');
```
robertknight: It looks like this line is a mistake as well, since "Create one by clicking the" is only shown in the notes tab.

Just as a suggestion, it might save a few lines and make the tests more readable to extract out the logic for querying the DOM into a helper function which returns some value to indicate the displayed message that you can then test for. Alternatively if the template "logic" is very simple (eg. hiding/showing different elements depending on some controller methods which return a boolean) you can might be able to get away with just testing the controller method's output in different cases.

robertknight: Discussed this on Slack. This is a check to make sure that the annotations-related message is not showing.

PR #614, opened 2017-11-30
Open thirdPartyUsernameLink in current browser window

nlisgo:
The introduction of the `target="_blank"` is here: https://github.com/hypothesis/client/commit/396fce0e486d6d12b320a32f9026d15cf5fbdc3d

The implementation of the `usernameUrl` needs more work if it is to be an option that can serve many use cases. It should be implemented in the same way as `onLoginRequest` but have the `username` passed as a parameter, so we can determine the action of clicking on the link, the decision to open in a new window should be a configuration option.

@robertknight wrote on slack on Nov 28th 12:38pm:

"My suggestion here is that we should make the username link work the same as other service-provided handlers eg. for login & logout, namely:

- It should be part of the service configuration, rather than top-level configuration, since each service will want its own username links
- It should be a callback function (like `onLoginRequest`) that allows the host page to handle as they see fit - whether that is opening a popup window, opening a new tab or changing the current route in a web app"

Until we implement the ideal solution can we request that we drop the `target="_blank"`. It does not provide the best user experience for us.

### DIFF
--- a/src/sidebar/templates/annotation-header.html
+++ b/src/sidebar/templates/annotation-header.html
@@ -7,7 +7,6 @@
       ng-href="{{vm.serviceUrl('user',{user:vm.user()})}}"
       >{{vm.displayName()}}</a>
     <a class="annotation-header__user"
-      target="_blank"
       ng-if="vm.isThirdPartyUser() && vm.thirdPartyUsernameLink()"
       href="{{ vm.thirdPartyUsernameLink() }}"
       >{{vm.displayName()}}</a>


seanh:
Hi @nlisgo, unfortunately I don't think the Hypothesis team currently has time to implement this feature request or to test, review, merge and deploy pull requests for it.

This pull request changes the username on annotation cards to open the user's page in the same tab, but only when the display name is shown. In cases where the username rather than the display name is shown it'll still open in a new tab. Other links (tag links, annotation links, share links, etc etc) all still open in new tabs too.

I don't think we want to implement an inconsistent experience in Hypothesis where some links open in the same page but some open in new tabs, or where links open in new tabs except when Hypothesis is used on certain websites they open  in the same page.

And unfortunately, even if we decided in Hypothesis that we want to change all links to open in the same page, I don't think we have the free time right now to implement that. We have other projects that we really need to be moving on with.

Sorry, but we'll have to push this one back for now.

Issue #612, opened 2017-11-30
Broken timestamp (individual annotation page) links on publisher group annotations


seanh:
The timestamp (top right of annotation card) on publisher group annotations is linking to the individual annotation page for that annotation. For example here's an annotation as shown in the sidebar on eLife's site:

![screenshot from 2017-11-30 12-07-35](https://user-images.githubusercontent.com/22498/33430131-4c989018-d5c7-11e7-85c1-8aa44373296c.png)

But these individual annotation pages that you get to after clicking the timestamp in the top right don't work for publisher group annotations:

![screenshot from 2017-11-30 12-09-48](https://user-images.githubusercontent.com/22498/33430153-62fb0962-d5c7-11e7-9304-09beb0ad91bf.png)

Problems:

1. The page contains two bouncer links (the document title and the annotation share control) but bouncer links don't work for publisher group annotations

2. Any tags shown on the card contain links to activity pages tag searches, but activity pages don't work for publisher group annotations (lots of things are broken on activity pages, when publisher group or third-party account annotations are shown on them)

3. The annotation card has a reply button that, when clicked opens a broken annotation editor with e.g. a broken post button (can be seen in above screenshot)

4. The publisher's (e.g. eLife's) theme is not used on the annotation page. It uses the default Hypothesis look

## Suggestion

Don't hyperlink the timestamps on publisher group annotations. The individual annotation pages were always out of scope for publisher groups round 1, and the plan was always to not activate these links.

PR #617, opened 2017-11-30
Dont hyperlink annotation timestamps when the annotation has no href

seanh:
This is part of the fix for <https://github.com/hypothesis/client/issues/612>. The other part is <https://github.com/hypothesis/h/pull/4711>.

A first-party annotation from the annotation read or search API looks like:

```json
{
  ...,
  links: {
    json: "http://localhost:5000/api/annotations/QvplEtX1EeeOhK9ondpFvw",
    html: "http://localhost:5000/a/QvplEtX1EeeOhK9ondpFvw",
    incontext: "http://localhost:8000/QvplEtX1EeeOhK9ondpFvw/localhost:3000/"
  },
  ...,
}
```

Notice the `html` link. A third-party annotation has no `html` or `incontext` link:

```json
{
  ...,
  links: {
    json: "http://localhost:5000/api/annotations/QvplEtX1EeeOhK9ondpFvw",
  },
  ...,
}
```

Or at least, once I've fixed h to not include these links, they won't.

With this PR the client renders the timestamps of first-party annotations as:

```html
<div ng-switch="vm.href">
  <a ng-switch-default
     target="_blank"
     title="27 Feb"
     href="https://hypothes.is/...">
       2 months ago
  </a>
</div>
```

And third-party ones as:

```html
<div ng-switch="vm.href">
  <span ng-switch-when="">
    2 months ago
  </span>
</div>
```

Annotations that are newly-created in the client and not saved to the server yet get the `<span>` version too.

### DIFF
--- a/src/sidebar/templates/timestamp.html
+++ b/src/sidebar/templates/timestamp.html
@@ -1,0 +1,5 @@
+<a class="{{ vm.className }}"
+   target="_blank"
+   title="{{ vm.absoluteTimestamp }}"
+   href="{{ vm.href }}">{{ vm.relativeTimestamp }}
+</a>
--- a/src/sidebar/templates/timestamp.html
+++ b/src/sidebar/templates/timestamp.html
@@ -1,5 +1,14 @@
-<a class="{{ vm.className }}"
-   target="_blank"
-   title="{{ vm.absoluteTimestamp }}"
-   href="{{ vm.href }}">{{ vm.relativeTimestamp }}
-</a>
+<div ng-switch="vm.href">
+  <span ng-switch-when=""
+        class="{{ vm.className }}"
+        title="{{ vm.absoluteTimestamp }}">
+    {{ vm.relativeTimestamp }}
+  </span>
+  <a ng-switch-default
+     class="{{ vm.className }}"
+     target="_blank"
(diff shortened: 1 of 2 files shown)


src/sidebar/templates/timestamp.html
```diff
@@ -0,0 +1,12 @@
+<div ng-switch="vm.href">
+  <span ng-switch-when="">
```
robertknight: The `class` attribute is missing here so the font size and color does not match between linked vs. unlinked timestamps.

Also the title attribute is missing so the full annotation creation time won't be shown on hover.

seanh: Nice catch, cheers

seanh:
@robertknight I've fixed the class and title. There's a 3 things duplicated now between the `<a>` and the `<span>` but I don't off the top of my head know a nice way to avoid that. I think I'd be happy to merge as-is if you are

robertknight:
> There's 3 things duplicated now between the <a> and the <span> but I don't off the top of my head know a nice way to avoid that.

Switching element type's dynamically in Angular 1.x templates or sharing a set of attributes between elements is not trivial as far as I know.

Issue #613, opened 2017-11-30
Broken bouncer (share / direct link) links on publisher group annotation cards


seanh:
Share links are showing on annotation cards in publisher groups, just like on any normal annotation card:

![screenshot from 2017-11-30 12-16-58](https://user-images.githubusercontent.com/22498/33430469-82da717c-d5c8-11e7-85d8-8d630e624abd.png)

But these share links don't work for publisher group annotations. When you follow one it looks like this:

![screenshot from 2017-11-30 12-18-47](https://user-images.githubusercontent.com/22498/33430515-aa5caa08-d5c8-11e7-99c8-ae412c9b6811.png)

Problems:

1. The groups control is showing _Public_, even though the annotation is in the eLife group not Public. The eLife group is not present in the groups dropdown.

2. The publisher's theme isn't used, default Hypothesis theme appears

3. The publisher's third-party accounts don't work, you won't be logged in to your eLife account in the client.

## Suggestion

Hide the entire share control on publisher group annotations. I believe share links were always out of scope for publisher groups round 1 and the plan was always to not show them.

PR #618, opened 2017-12-01
Don't show share dialog if no incontext link

seanh:
If the annotation received from the API contains no incontext link (as is the case with third-party annotations, for example) then don't render the annotation share dialog that contains the incontext link.

This is one half of the fix for <https://github.com/hypothesis/client/issues/613>, the other half is <https://github.com/hypothesis/h/pull/4714>.


robertknight:
LGTM. Tested by removing both the "html" and "incontext" link generators in `h/links.py` in the "h" service.

Issue #615, opened 2017-11-30
Page share link in top right of sidebar shouldn't show on publisher group pages


seanh:
This Via link won't work:

![screenshot from 2017-11-30 12-44-35](https://user-images.githubusercontent.com/22498/33431481-643a4e6e-d5cc-11e7-8d8d-de4dc203ef34.png)

For the same reason as why [these bouncer links don't work](https://github.com/hypothesis/client/issues/613), publisher group annotations don't work in Via. See #613 for more details

PR #620, opened 2017-12-05
Don't show page share button on eLife pages

seanh:
Don't show the "Share this page" button in the top-right of the sidebar
on pages, such as eLife's, that have both an embedded Hypothesis client
and a services config including their own `authority` string.

The "Share this page" button is a Via link, and Via doesn't work with
pages using a third-party authority config.

Fixes https://github.com/hypothesis/client/issues/615

On first-party sites, e.g. localhost:3000, share button still shows:

screenshot from 2017-12-05 18-48-55

On third-party sites, e.g. publisher test site, no share button:

screenshot from 2017-12-05 18-49-17

### DIFF
--- a/src/sidebar/components/test/top-bar-test.js
+++ b/src/sidebar/components/test/top-bar-test.js
@@ -1,15 +1,20 @@
 'use strict';
 
 var angular = require('angular');
+var proxyquire = require('proxyquire');
 
 var util = require('../../directive/test/util');
 
 describe('topBar', function () {
   var fakeSettings = {};
+  var fakeIsThirdPartyService = sinon.stub();
 
   before(function () {
     angular.module('app', [])
-      .component('topBar', require('../top-bar'))
+      .component('topBar', proxyquire('../top-bar', {
+        '../util/is-third-party-service': fakeIsThirdPartyService,
+        '@noCallThru': true,
+      }))
       .component('loginControl', {
         bindings: require('../login-control').bindings,
       })
@@ -25,6 +30,9 @@ describe('topBar', function () {
     angular.mock.module('app', {
       settings: fakeSettings,
     });
+
+    fakeIsThirdPartyService.reset();
+    fakeIsThirdPartyService.returns(false);
   });
 
   function applyUpdateBtn(el) {
@@ -85,6 +93,39 @@ describe('topBar', function () {
 
     loginControl.onLogout();
     assert.called(onLogout);
+  });
+
+  it("checks whether we're using a third-party service", function () {
+    createTopBar();
+
+    assert.called(fakeIsThirdPartyService);
+    assert.alwaysCalledWithExactly(fakeIsThirdPartyService, fakeSettings);
+  });
+
+  context('when using a first-party service', function () {
+    it('shows the share page button', function () {
+      var el = createTopBar();
+      // I want the DOM element, not AngularJS's annoying angular.element
+      // wrapper object.
+      el = el [0];
+
+      assert.isNotNull(el.querySelector('[title="Share this page"]'));
+    });
+  });
+
+  context('when using a third-party service', function () {
+    beforeEach(function() {
+      fakeIsThirdPartyService.returns(true);
+    });
+
+    it("doesn't show the share page button", function () {
+      var el = createTopBar();
+      // I want the DOM element, not AngularJS's annoying angular.element
+      // wrapper object.
+      el = el [0];
+
+      assert.isNull(el.querySelector('[title="Share this page"]'));
+    });
   });
 
   it('displays the share page when "Share this page" is clicked', function () {
--- a/src/sidebar/util/is-third-party-service.js
+++ b/src/sidebar/util/is-third-party-service.js
@@ -1,0 +1,30 @@
+'use strict';
+
+const serviceConfig = require('../service-config');
+
+/**
+ * Return `true` if the first configured service is a "third-party" service.
+ *
+ * Return `true` if the first custom annotation service configured in the
+ * services array in the host page is a third-party service, `false` otherwise.
+ *
+ * If no custom annotation services are configured then return `false`.
+ *
+ * @param {Object} settings - the sidebar settings object
+ *
+ */
+function isThirdPartyService(settings) {
+  const service = serviceConfig(settings);
+
+  if (service === null) {
+    return false;
+  }
+
+  if (!service.hasOwnProperty('authority')) {
+    return false;
+  }
+
+  return (service.authority !== settings.authDomain);
+}
+
+module.exports = isThirdPartyService;
--- a/src/sidebar/util/test/is-third-party-service-test.js
+++ b/src/sidebar/util/test/is-third-party-service-test.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const proxyquire = require('proxyquire');
+
+describe('sidebar.util.isThirdPartyService', () => {
+  let fakeServiceConfig;
+  let fakeSettings;
+  let isThirdPartyService;
+
+  beforeEach(() => {
+    fakeServiceConfig = sinon.stub();
+    fakeSettings = {authDomain: 'hypothes.is'};
+
+    isThirdPartyService = proxyquire('../is-third-party-service', {
+      '../service-config': fakeServiceConfig,
+      '@noCallThru': true,
+    });
+  });
+
+  it('returns false for first-party services', () => {
+    fakeServiceConfig.returns({authority: 'hypothes.is'});
+
+    assert.isFalse(isThirdPartyService(fakeSettings));
+  });
+
+  it('returns true for third-party services', () => {
+    fakeServiceConfig.returns({authority: 'elifesciences.org'});
+
+    assert.isTrue(isThirdPartyService(fakeSettings));
+  });
+
+  it("returns false if there's no service config", () => {
+    fakeServiceConfig.returns(null);
+
+    assert.isFalse(isThirdPartyService(fakeSettings));
+  });
+
+  // It's not valid for a service config object to not contain an authority
+  // (authority is a required field) but at the time of writing the config
+  // isn't validated when it's read in, so make sure that isThirdPartyService()
+  // handles invalid configs.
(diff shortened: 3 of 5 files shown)


src/sidebar/components/test/top-bar-test.js
```diff
@@ -87,6 +95,33 @@ describe('topBar', function () {
...
+
+    assert.called(fakeIsThirdPartyService);
+    assert.alwaysCalledWithExactly(fakeIsThirdPartyService, fakeSettings);
+  });
+
+  context('when using a first-party service', function () {
+    it('shows the share page button', function () {
+      var el = createTopBar();
+
+      assert.equal(el.find('[title="Share this page"]').length, 1);
```
seanh: `el` doesn't appear to be a normal DOM element, it's missing lots of the methods, using `length` is the only way I could find to test whether it was found or not.

Might be an idea to add a `js-*` CSS class instead of searching by `title` here. Existing tests use `title` though.

src/sidebar/util/is-third-party-service.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const serviceConfig = require('../service-config');
```
seanh: Since this is a new file I decided to use modern JavaScript

src/sidebar/util/is-third-party-service.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
seanh: I couldn't find a better place to put this file, without first doing some reorganising that I don't feel I have time for.

The similar function `isThirdPartyUser()` is in `account-id.js` but this file is just for account stuff.

Other service-related stuff is in `service-config.js` and `service-url.js`. I think what's probably needed is an `AnnotationService` class that contains everything in `service-config.js`, `service-url.js`, `is-third-party-service.js` and probably lots of other stuff besides.

src/sidebar/components/test/top-bar-test.js
robertknight: Angular in several places wraps DOM elements using a jQuery-like interface called jqLite. See https://docs.angularjs.org/api/ng/function/angular.element.

As with jQuery, functions operate on lists of elements. Hence `length` returns the list size and indexing using `el[0]` returns the actual DOM element.

src/sidebar/util/is-third-party-service.js
robertknight: >  I think what's probably needed is an AnnotationService class that contains everything

I don't think a class would be necessary, just a module of service-related utility functions would be fine. eg. In `util/service.js`.

src/sidebar/util/is-third-party-service.js
robertknight: I'm fine with this. The unwritten policy I have been following recently is to use new JS (except for still using CommonJS) for all code except for continuing to use `var` rather than `const`/`let` in existing files since that would otherwise be a very glaring inconsistency.

FWIW there are some good tools for doing bulk modernisation of JS code in straightforward cases (eg. converting var => const/let, functions to arrows etc.) and I'll propose at some point that we use them to improve consistency.

src/sidebar/util/test/is-third-party-service-test.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const proxyquire = require('proxyquire');
+
+describe('sidebar.util.isThirdPartyService', () => {
```
robertknight: We are being a bit inconsistent in how these describe blocks are named. I've been using the file path with slashes replaced by periods. You're using a mix of path and function name.

We have already standardised on one describe block per test module, so perhaps we should just make the label a relative path from the root (ie. `sidebar/util/is-third-party-service`). SGTY?

src/sidebar/components/test/top-bar-test.js
seanh: Aha, bookmarked, good to know. I was trying to find some way in the Chrome console to find out what class or type of object this is, but all I could get was "object".

seanh: Ok, I've fixed this to use `querySelector`

src/sidebar/util/is-third-party-service.js
seanh: P.S. I'm still referring to the [AirBnB JavaScript style guide](https://github.com/airbnb/javascript) for how to write JS, at least when it doesn't obviously conflict with our existing practice

src/sidebar/components/test/top-bar-test.js
```diff
@@ -1,15 +1,20 @@
...
 
 describe('topBar', function () {
   var fakeSettings = {};
+  var fakeIsThirdPartyService = sinon.stub();
 
   before(function () {
     angular.module('app', [])
-      .component('topBar', require('../top-bar'))
+      .component('topBar', proxyquire('../top-bar', {
+        '../util/is-third-party-service': fakeIsThirdPartyService,
```
robertknight: I'm really not a huge fan of mocking simple, pure utility functions. I personally would have preferred just to create representative settings for the first and third party cases which could be as simple as:

```js
// First party
fakeSettings = {
  services: [{ authority: 'hypothes.is'} ],
  authDomain: 'hypothes.is',
};

// Third party
fakeSettings = {
  services: [{ authority: 'partner.org'} ],
  authDomain: 'hypothes.is',
};
```

However, this is subjective and we haven't hashed out a clear policy either way, so I'm OK to leave it as is.

src/sidebar/util/is-third-party-service.js
seanh: It'd be nice to only have to construct an object once, passing the settings for a particular service (from the services config) into the constructor, as well as any other dependencies it needs, and then be able to access things like `serviceUrl()`, `serviceConfig()` and `isThirdPartyUser()` on that object without having to pass in the service config settings and other dependencies every time. 

This is just nicer code, in my opinion, than the approach where each user module has to hold on to the service config settings and other dependencies and make sure to pass the right params to each service-related utility function each time it calls one.

If we end up adding support for multiple annotation services to the client (which the services array was anticipating) it'll probably be even nicer, as you can instantiate one `AnnotationService` object for each service. (Although I think we probably want to add multiple services support to h, not to the client.)

src/sidebar/components/test/top-bar-test.js
```diff
@@ -1,15 +1,20 @@
 'use strict';
 
 var angular = require('angular');
+var proxyquire = require('proxyquire');
 
 var util = require('../../directive/test/util');
 
 describe('topBar', function () {
   var fakeSettings = {};
+  var fakeIsThirdPartyService = sinon.stub();
```
robertknight: Hmm ... in most other tests we do the initialization of stubs in the `beforeEach` function via assignment. This makes it more obvious that the value is always "fresh" for each test, regardless of type.

src/sidebar/util/test/is-third-party-service-test.js
```diff
@@ -0,0 +1,47 @@
...
+  let fakeServiceConfig;
+  let fakeSettings;
+  let isThirdPartyService;
+
+  beforeEach(() => {
+    fakeServiceConfig = sinon.stub();
+    fakeSettings = {authDomain: 'hypothes.is'};
+
+    isThirdPartyService = proxyquire('../is-third-party-service', {
```
robertknight: This works, but you could avoid the mocking here by requiring the argument to be the configured service, or just passing some representative settings and not mocking `service-config`.

src/sidebar/util/is-third-party-service.js
robertknight: That guide is still a reasonable baseline. I wouldn't worry about following every detail to the letter though.

I did try using their ESLint config in a personal project but I found that a number of the rules were more prescriptive than I felt valuable (eg. mandating the use of string interpolation in places were string concatenation was just as readable IMO). More significantly, the yapf/gofmt-like tool [prettier](https://github.com/prettier/prettier) obviates the need for a lot of lint rules about formatting minutiae in JS.

src/sidebar/components/test/top-bar-test.js
seanh: Yes. Unfortunately `angular.module()` is called in a `before()` that happens _before_ the `beforeEach()` in which the rest of the fakes are initialized. So this one fake needs to be initialized sooner. I've run across this a few times. I think the solution may have to be for the `angular.module()` call or similar (construction of an angular directive, etc) to happen in a `beforeEach()` instead of a `before()`, so that params that get passed into it can be re-assigned per test.

src/sidebar/util/test/is-third-party-service-test.js
seanh: I guess the confusion is between the module name `my-thing` vs the name of the actual function thing `myThing` or class `MyThing` (that is used both in the code in the module itself and in the assignment whenever anyone imports the module and then whenever anyone uses it).

If this were a module that contained more than one thing, I think you'd want to do:

```javascript
describe('sidebar/util/my-things', () => {
  describe('firstThing', () => {
  });
  describe('secondThing', () => {
  });
});
```

I guess that kind of implies that for a module containing one thing, we should do:

```
describe('sidebar/util/my-thing', () => {
  describe('myThing', () => {
  });
});
```

src/sidebar/util/test/is-third-party-service-test.js
seanh: If `isThirdPartyService()` required the individual service object then every time anyone wanted to call it they'd have to do:

```javascript
const serviceConfig = require('../service-config');
var service = serviceConfig(settings);
if (isThirdPartyService(settings, service)) { ... };
```

Instead of just:

```javascript
if (isThirdPartyService(settings)) { ... };
```

This may mean that the tests for `isThirdPartyService()` no longer need to mock `serviceConfig()`, but the tests for everything that calls `isThirdPartyService()` would need to mock `serviceConfig()` instead.

Or to put it another way - we wouldn't be encapsulating `serviceConfig()`. If you were to draw a diagram of the architecture, instead of this:



(Here I've used `AnnotationService` as a stand-in for a new class that includes `isThirdPartyService()` and other service-related methods in one class)

You would instead have this:

PR #621, opened 2017-12-06
Remove traces of persona.js

seanh:
Fix a couple of out-of-date references to `persona.js`. which no longer exists.

### DIFF
--- a/src/sidebar/components/test/annotation-header-test.js
+++ b/src/sidebar/components/test/annotation-header-test.js
@@ -189,7 +189,7 @@ describe('sidebar.components.annotation-header', function () {
               return false;
             };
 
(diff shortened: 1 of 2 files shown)


robertknight:
LGTM, although presumably if the test was passing despite the incorrect reference then the fake implementation was not actually being used?

seanh:
Hmm, indeed. The fake itself is out of date - it fakes the `username()` function which the code does not call anymore, and it doesn't think the `isThirdPartyUser()` that the code does call. The code that calls `isThirdPartyUser()` isn't tested, that's why the tests still pass. The fake isn't actually used. I've sent a fix for this here: https://github.com/hypothesis/client/pull/622

PR #622, opened 2017-12-06
Fix out of date account-id fake

seanh:


src/sidebar/components/test/annotation-test.js
```diff
@@ -785,6 +791,19 @@ describe('annotation', function() {
...
+        var { annotation, controller } = createDirective();
+
+        var returned = controller.isThirdPartyUser();
+
+        assert.calledOnce(fakeAccountID.isThirdPartyUser);
+        assert.alwaysCalledWithExactly(
+          fakeAccountID.isThirdPartyUser, annotation.user, fakeSettings.authDomain);
+        assert.equal(returned, fakeAccountID.isThirdPartyUser());
+      });
+    });
```
seanh: This is a test that will fail if the `annotation.js` code is out of step with the `fakeAccountID` in `annotation-test.js`. For example if we changed the API of `accountID` or of `accountID.isThirdPartyUser()`, then grepped for `"fakeAccountID"` to find and update all the fakes of it, then just ran the tests, this test would be failing and we'd find the `annotation.js` code that needs updating.

This test is certainly a bit awkward and I think it shows up a design problem. It might be better, for example, if we had a `User` class that simply had a boolean `isThirdParty` property and the various modules just accessed that property instead of having to import `account-id.js` and call `accountID.isThirdPartyUser(userid, settings.authDomain)` and pass in the correct `userid` and `authDomain`. In `annotation.js` this:

```javascript
  this.isThirdPartyUser = function () {
    return isThirdPartyUser(self.annotation.user, settings.authDomain);
  };
```

would become this:

```
this.isThirdParty = self.annotation.user.isThirdParty;
```

and many places that have an `Annotation` object could then just use `myAnnotation.isThirdParty` instead of `accountID.isThirdPartyUser(myAnnotation.user, settings.authDomain)`.

seanh:
:+1: Agree it'd be nice to standardise on either AngularJS injection or `require()`. My first thought is to lean towards `require()` because it doesn't couple us to AngularJS which I think of as a legacy framework.

Issue #616, opened 2017-11-30
Tags on publisher group annotations link to broken activity pages


seanh:
Publisher group annotaitons don't work on activity pages (lots of problems). These tags on the annotation cards in the sidebar shouldn't link to activity pages for publisher group annotations:

![screenshot from 2017-11-30 12-50-16](https://user-images.githubusercontent.com/22498/33431699-40abbdc4-d5cd-11e7-8f8c-c2e0b66585cd.png)

dwhly:
> for publisher group annotations

Or perhaps "for 3rd party namespaces".  Publisher groups that are in the H namespace presumably wouldn't be a problem.

seanh:
We don't have any publisher groups that are in the h namespace. I would like to dispel any notion that such things exist!

robertknight:
> We don't have any publisher groups that are in the h namespace. 

I think we should stop calling what we've given eLife "publisher groups" 😛, because that is going to get really confusing soon!

What you mean here is "groups associated with third-party accounts", which admittedly is not as succinct.

seanh:
I _think_ what Dan (and others outside of the product team) mean by _publisher group_ is simply a group that is nominated, by a publisher's website, as sort of a "pinned group" whenever the client is launched on that website. What this means currently is that eLife's publisher group is the _only_ group that's available in the client on eLife's site, but in the future when other groups (e.g. first-party groups) are also available on the eLife site their publisher group will just be the one that's pinned to the top of the groups menu / highlighted / featured / promoted in some way as the "publisher's official group".

So this concept of a publisher group is distinct from the concept of a third-party group.

But of course right now we only have one kind of publisher group - the third-party publisher group with third-party account integration that we created for eLife.

What I'd like to dispel is any notion that publisher groups beyond this limited case is a feature that currently exists in Hypothesis. I'm sure that all sorts of things will come up, when we try to implement a publisher group that is a first-party group, or that allows people to join the group without having to be logged in to an account on the publisher's site, etc.

These are all features we can have, but will require thought and work.

PR #623, opened 2017-12-06
Don’t link tags to activity pages on eLife

seanh:
Depends on https://github.com/hypothesis/client/pull/622

Tags are still links on localhost:3000:

peek 2017-12-06 15-26

But on localhost:5050 they look the same but aren't linked:

peek 2017-12-06 15-27

You still get the "clickable" cursor when hovering over a tag but I don't think this is to do with the tags - the entire bottom area of the annotation card has a clickable cursor (and clicking does nothing).

Fixes https://github.com/hypothesis/client/issues/616

### DIFF
--- a/src/sidebar/components/annotation.js
+++ b/src/sidebar/components/annotation.js
@@ -444,6 +444,9 @@ function AnnotationController(
   };
 
   this.tagSearchURL = function(tag) {
+    if (this.isThirdPartyUser()) {
+      return null;
+    }
     return serviceUrl('search.tag', {tag: tag});
   };
 
--- a/src/sidebar/components/test/annotation-test.js
+++ b/src/sidebar/components/test/annotation-test.js
@@ -1068,22 +1068,57 @@ describe('annotation', function() {
     });
 
     describe('tag display', function () {
-      it('displays links to tags on the stream', function () {
+      beforeEach('make serviceUrl() return a URL for the tag', function() {
         fakeServiceUrl
           .withArgs('search.tag', {tag: 'atag'})
           .returns('https://test.hypothes.is/stream?q=tag:atag');
+      });
 
-        var directive = createDirective(Object.assign(fixtures.defaultAnnotation(), {
+      /*
+       * Return an annotation directive with a single tag.
+       */
+      function annotationWithOneTag() {
+        return createDirective(Object.assign(fixtures.defaultAnnotation(), {
           tags: ['atag'],
         }));
+      }
+
+      /**
+       * Return the one tag link element from the given annotation directive.
+       */
+      function tagLinkFrom(directive) {
         var links = [].slice.apply(directive.element[0].querySelectorAll('a'));
         var tagLinks = links.filter(function (link) {
           return link.textContent === 'atag';
         });
-
         assert.equal(tagLinks.length, 1);
-        assert.equal(tagLinks[0].href,
-                     'https://test.hypothes.is/stream?q=tag:atag');
+        return tagLinks[0];
+      }
+
+      context('when the annotation is first-party', function() {
+        beforeEach('configure a first-party annotation', function() {
+          fakeAccountID.isThirdPartyUser.returns(false);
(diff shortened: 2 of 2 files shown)


src/sidebar/components/annotation.js
```diff
@@ -444,6 +444,9 @@ function AnnotationController(
   };
 
   this.tagSearchURL = function(tag) {
+    if (this.isThirdPartyUser()) {
+      return null;
+    }
```
seanh: I didn't have to change anything in the template. It uses ngHref: `<a ng-href="{{vm.tagSearchURL(tag)}}" target="_blank">{{tag}}</a>` so that if `tagSearchURL()` returns null it renders an `<a>` with an `ng-href` but with no `href` attribute.

src/sidebar/components/test/annotation-test.js
```diff
@@ -1068,22 +1068,57 @@ describe('annotation', function() {
...
-      it('displays links to tags on the stream', function () {
+      beforeEach('make serviceUrl() return a URL for the tag', function() {
         fakeServiceUrl
           .withArgs('search.tag', {tag: 'atag'})
           .returns('https://test.hypothes.is/stream?q=tag:atag');
+      });
 
-        var directive = createDirective(Object.assign(fixtures.defaultAnnotation(), {
+      /*
+       * Return an annotation directive with a single tag.
```
robertknight: Technically it is a "component" now rather than a directive. I'm guessing there are quite a lot of references in this file which haven't been updated.

src/sidebar/components/test/annotation-test.js
```diff
@@ -1068,22 +1068,57 @@ describe('annotation', function() {
...
     describe('tag display', function () {
-      it('displays links to tags on the stream', function () {
+      beforeEach('make serviceUrl() return a URL for the tag', function() {
         fakeServiceUrl
           .withArgs('search.tag', {tag: 'atag'})
           .returns('https://test.hypothes.is/stream?q=tag:atag');
+      });
 
-        var directive = createDirective(Object.assign(fixtures.defaultAnnotation(), {
+      /*
```
robertknight: If this is meant to be a JSDoc comment as opposed to just an inline comment, it needs to start with `/**` rather than `/*`.

src/sidebar/components/test/annotation-test.js
```diff
@@ -1068,22 +1068,57 @@ describe('annotation', function() {
...
-                     'https://test.hypothes.is/stream?q=tag:atag');
+        return tagLinks[0];
+      }
+
+      context('when the annotation is first-party', function() {
+        beforeEach('configure a first-party annotation', function() {
+          fakeAccountID.isThirdPartyUser.returns(false);
+        });
+
+        it('displays links to tags on the stream', function () {
```
robertknight: The client links to search pages for tags rather than the stream now. See https://hypothes.is/api/links

src/sidebar/components/annotation.js
robertknight: I wondered about that. Thanks for noting this.

src/sidebar/components/test/annotation-test.js
```diff
@@ -1068,22 +1068,57 @@ describe('annotation', function() {
...
+          var tagLink = tagLinkFrom(annotationWithOneTag());
+
+          assert.equal(tagLink.href, 'https://test.hypothes.is/stream?q=tag:atag');
```
robertknight: Again, the h service actually generates `/search` rather than `/stream` tag links now.

seanh:
Rebased

src/sidebar/components/test/annotation-test.js
seanh: Oops, yeah that looks like a typo

src/sidebar/components/test/annotation-test.js
seanh: :(

PR #624, opened 2017-12-06
Bundle the clean theme related config options into the theme = 'clean…

sheetaluk:
…' setting.

Some of the config settings don't work well with the clean theme. Some others don't work
without the clean theme. It would be better to remove the following config options and
have each of these elements tied to the over all 'theme' of the sidebar for consistency:
disableToolbarCloseBtn
disableToolbarMinimizeBtn
disableToolbarHighlightsBtn
disableToolbarNewNoteBtn
disableBucketBar
enableSidebarDropShadow
enableCleanOnboardingTheme

See https://github.com/hypothesis/product-backlog/issues/380

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -18,14 +18,7 @@ function configFrom(window_) {
     // URL of the client's boot script. Used when injecting the client into
     // child iframes.
     clientUrl: settings.clientUrl,
-    disableToolbarCloseBtn:settings.hostPageSetting('disableToolbarCloseBtn', {defaultValue: true}),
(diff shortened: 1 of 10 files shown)


robertknight:
In testing this I noted a couple of issues:

1. The default "grey" background when `theme: "clean"` is set looks odd. We might want to change the default app background color as part of the theme.
2. The "New notes" option is still separate and since this is the only way to create a new note when using the "clean" theme, the minimal config is `{ theme: "clean", enableExperimentalNewNoteButton: true }`. Let's address that in a follow-up PR.

PR #627, opened 2017-12-07
Extract "create account" banner component

robertknight:
This PR is some refactoring in preparation for fixing an issue where the "create account" banner shows up on the pages of publishers using third-party accounts with a signup link that goes directly to Hypothesis [1] and text that may not be appropriate ("create a free account") depending on how the third party's accounts work.

The refactoring here is:

1. Extract the "To annotate, create a free account or log in" banner into a new `<create-account-banner>` component
2. Move the logic for deciding whether to show that banner or not out of the `hypothesis-app` template and into the controller, where it can be more easily tested.

[1] The signup link actually comes from the `/api/links` API, so if all API requests knew what the authority was, we could partially resolve this in H by returning a different link. However that doesn't fix the issue that the "create free account" text may not be appropriate.

### DIFF
--- a/src/sidebar/app.js
+++ b/src/sidebar/app.js
@@ -136,6 +136,7 @@ module.exports = angular.module('h', [
   .component('annotationShareDialog', require('./components/annotation-share-dialog'))
   .component('annotationThread', require('./components/annotation-thread'))
   .component('annotationViewerContent', require('./components/annotation-viewer-content'))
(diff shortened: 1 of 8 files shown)


robertknight:
Closing because we're planning to solve the problem by just removing the banner entirely: https://github.com/hypothesis/client/pull/628#issuecomment-349981664

PR #628, opened 2017-12-07
Hide banner for third party users

robertknight:
**Depends on https://github.com/hypothesis/client/pull/627**

This hides the "To annotate, create a free account or log in" banner on pages that use third-party accounts.


seanh:
@robertknight Can we just delete this banner from the product? It's stupid. Dan and Dawa previously agreed: https://hypothes-is.slack.com/archives/C07NXBDNW/p1510080494000386

robertknight:
OK, even better. I'll close this PR then.

PR #630, opened 2017-12-07
Remove create account banner

robertknight:
The text is potentially inappropriate in the context of publisher-managed accounts where the account might not be free and the user might not be able to create one themselves.
    
In the first party account context we agreed that the banner is a bit redundant anyway. See https://hypothes-is.slack.com/archives/C07NXBDNW/p1510080494000386


seanh:
:+1:

PR #633, opened 2017-12-11
Rename JS bundles to match corresponding source directory

robertknight:
This is part of a series of PRs to address some low hanging fruit to hopefully make it easier for newcomers to find their way around the client code.

Rename the "injector" and "app" bundles to "annotator" and "sidebar"
respectively to match the source directory which produces those bundles.


robertknight:
Rebased

PR #636, opened 2017-12-11
Update the docs for config options available to third party publishers.

sheetaluk:


docs/publishers/config.rst
```diff
@@ -81,6 +81,32 @@ loads.
...
+   ``Boolean`` - Controls whether the experimental new note button should be shown in the
+   notes tab in the sidebar. (Default: ``false``).
+
+   ``true`` - The button is shown.
+
+   ``false`` - The button is not shown by default.
+
+.. option:: usernameUrl
+
+   ``String``. This allows you to specify a url to direct a user to, in a new tab when they
```
robertknight: Can you clarify the format of the link here and how the actual username gets inserted?

docs/publishers/config.rst
```diff
@@ -81,6 +81,32 @@ loads.
...
+
+   ``String``. Controls the overall look of the sidebar.(Default: ``classic``).
+
+   ``classic`` - Enables the card view for annotations, the bucket bar, the sidebar minimize
+   button, the highlights button and the new note button in the toolbar. It also disables the
+   close button in the toolbar. The classic theme is enabled by default.
+
+   ``clean`` - Enables the clean view for annotations in the sidebar, disables the bucket bar,
+   the sidebar minimize button, the highlights button and the new note button in the toolbar and enables the
+   close button in the toolbar. It will also show a cleaner and minimal onboarding tutorial.
```
robertknight: s/a cleaner and minimal/a cleaner and _more_ minimal ?

docs/publishers/config.rst
```diff
@@ -81,6 +81,32 @@ loads.
       experimental and may change in future. ``true`` and ``false`` values
       are the stable API.
 
+.. option:: theme
+
+   ``String``. Controls the overall look of the sidebar.(Default: ``classic``).
+
+   ``classic`` - Enables the card view for annotations, the bucket bar, the sidebar minimize
```
robertknight: Since these are strings they should be quoted (see the `showHighlights` docs).

docs/publishers/config.rst
robertknight: "url" is an acronym so it should be capitalized in the documentation descriptions..

docs/publishers/config.rst
```diff
@@ -81,6 +81,32 @@ loads.
...
+   button, the highlights button and the new note button in the toolbar. It also disables the
+   close button in the toolbar. The classic theme is enabled by default.
+
+   ``clean`` - Enables the clean view for annotations in the sidebar, disables the bucket bar,
+   the sidebar minimize button, the highlights button and the new note button in the toolbar and enables the
+   close button in the toolbar. It will also show a cleaner and minimal onboarding tutorial.
+
+.. option:: enableExperimentalNewNoteButton
+
+   ``Boolean`` - Controls whether the experimental new note button should be shown in the
```
robertknight: We should probably capitalize or quote "New note" to make it clearer what button is being referred to.

docs/publishers/config.rst
```diff
@@ -81,6 +81,32 @@ loads.
...
+   notes tab in the sidebar. (Default: ``false``).
+
+   ``true`` - The button is shown.
+
+   ``false`` - The button is not shown by default.
```
robertknight: You've mentioned above that the default value for this setting is ``false``, so you can just change this to "The button is not shown".

PR #651, opened 2018-01-22
Fix login flow triggering popup blocker in Firefox and IE

### LINKED ISSUE #534: "Log in" button triggers popup blocker in Firefox
**Steps to reproduce:**

1. Go to https://hypothes.is/docs/help?__feature__[client_oauth]=1
2. Click the "Log in" button in the client

**Expected result:** A login popup window appears and then either prompts for login details or immediately disappears and logs the user in.

**Actual result:** A "Firefox prevented this site from opening a pop-up window" message appears.

**Notes:**

- Clicking the "Preferences" button in the message bar and selecting "Allow pop-ups for ..." results in the login popup appearing.
- The pop-up blocker is not triggered by the same actions in Chrome and Safari.

robertknight:
hypothesis/client#603 broke the login popup window in Firefox and IE
because the call to `window.open` no longer happens in the same turn of
the event loop as the user's click on the "Login" link. It is therefore
no longer considered in FF to have happened "in response to a user
gesture".

This PR fixes the issue by separating creation and use of the popup
window into separate functions and moving creation to happen earlier, in
the same event loop turn as the "Login" button click.

Fixes #534

The browser behaviour here is something that I think ought to be clarified in the spec, so I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1425419 and got linked to some related discussion in https://github.com/whatwg/html/issues/1903

### DIFF
--- a/src/sidebar/util/oauth-client.js
+++ b/src/sidebar/util/oauth-client.js
@@ -151,10 +151,12 @@ class OAuthClient {
    *
    * Returns an authorization code which can be passed to `exchangeAuthCode`.
    *
-   * @param {Window} $window
+   * @param {Window} $window - Window which will receive the auth response.
+   * @param {Window} authWindow - Popup window where the login prompt will be shown.
+   *   This should be created using `createLoginPopupWindow`.
    * @return {Promise<string>}
    */
-  authorize($window) {
+  authorize($window, authWindow) {
     // Random state string used to check that auth messages came from the popup
     // window that we opened.
     var state = this.generateState();
@@ -184,41 +186,6 @@ class OAuthClient {
     });
 
     // Authorize user and retrieve grant token
-
-    // In Chrome & Firefox the sizes passed to `window.open` are used for the
-    // viewport size. In Safari the size is used for the window size including
-    // title bar etc. There is enough vertical space at the bottom to allow for
-    // this.
-    //
-    // See https://bugs.webkit.org/show_bug.cgi?id=143678
-    var width  = 475;
-    var height = 430;
-    var left   = $window.screen.width / 2 - width / 2;
-    var top    = $window.screen.height /2 - height / 2;
-
-    // Generate settings for `window.open` in the required comma-separated
-    // key=value format.
-    var authWindowSettings = queryString.stringify({
-      left: left,
-      top: top,
-      width: width,
-      height: height,
-    }).replace(/&/g, ',');
-
-    // Open the auth window before fetching the `oauth.authorize` URL to ensure
-    // that the `window.open` call happens in the same turn of the event loop
-    // that was initiated by the user clicking the "Log in" link.
-    //
-    // Otherwise the `window.open` call is not deemed to be in response to a
-    // user gesture in Firefox & IE 11 and their popup blocking heuristics will
-    // prevent the window being opened. See
-    // https://github.com/hypothesis/client/issues/534 and
-    // https://github.com/hypothesis/client/issues/535.
-    //
-    // Chrome, Safari & Edge have different heuristics and are not affected by
-    // this problem.
-    var authWindow = $window.open('about:blank', 'Login to Hypothesis', authWindowSettings);
-
     var authUrl = this.authorizationEndpoint;
     authUrl += '?' + queryString.stringify({
       client_id: this.clientId,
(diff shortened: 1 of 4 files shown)


src/sidebar/util/oauth-client.js
```diff
@@ -151,10 +151,12 @@ class OAuthClient {
...
    * Returns an authorization code which can be passed to `exchangeAuthCode`.
    *
-   * @param {Window} $window
+   * @param {Window} $window - Window which will receive the auth response.
+   * @param {Window} authWindow - Popup window where the login prompt will be shown.
+   *   This should be created using `createLoginPopupWindow`.
    * @return {Promise<string>}
    */
-  authorize($window) {
+  authorize($window, authWindow) {
```
lyzadanger: The variation between `auth` (`authWindow`) and `login` (`createLoginPopupWindow`) — curious about that: is there a difference between the two (i.e. would calling this param `loginWindow` be _wrong_?). This light confusion is compounded by the fact that it's `$window` that "receive[s] the auth response", ultimately.

src/sidebar/util/oauth-client.js
```diff
@@ -245,6 +212,41 @@ class OAuthClient {
...
+
+  /**
+   * Create a pop-up window for use with `OAuthClient#authorize`.
+   *
+   * This function _must_ be called in the same turn of the event loop as the
+   * button or link which initiates login to avoid triggering the popup blocker
+   * in certain browsers. See https://github.com/hypothesis/client/issues/534
+   * and https://github.com/hypothesis/client/issues/535.
+   *
+   * @param {Window} $window - The parent of the popup window.
```
lyzadanger: Perhaps this comment would be clearer if it indicated that the popup window is what is being created (and returned) by this method (e.g. `$window - The parent context of the popup window that will be generated and returned`). I found myself initially thinking "what popup window?...oh, the one this method creates. I see"

src/sidebar/util/oauth-client.js
```diff
@@ -245,6 +212,41 @@ class OAuthClient {
...
+   * @return {Window}
+   */
+  static createLoginPopupWindow($window) {
+    // In Chrome & Firefox the sizes passed to `window.open` are used for the
+    // viewport size. In Safari the size is used for the window size including
+    // title bar etc. There is enough vertical space at the bottom to allow for
+    // this.
+    //
+    // See https://bugs.webkit.org/show_bug.cgi?id=143678
+    var width  = 475;
```
lyzadanger: Completely orthogonal to this discussion, but I'm curious as to where these hard dimensions arose from (in terms of flexibility for different viewports). You can ignore this question for now if you like!

src/sidebar/util/oauth-client.js
```diff
@@ -245,6 +212,41 @@ class OAuthClient {
...
+    // Generate settings for `window.open` in the required comma-separated
+    // key=value format.
+    var authWindowSettings = queryString.stringify({
+      left: left,
+      top: top,
+      width: width,
+      height: height,
+    }).replace(/&/g, ',');
+
+    return $window.open('about:blank', 'Login to Hypothesis', authWindowSettings);
```
lyzadanger: Pedantic: `Login` should be `Log in` (_login_ is a noun; _log in_ the verb)

src/sidebar/util/test/oauth-client-test.js
```diff
@@ -162,15 +162,35 @@ describe('sidebar.util.oauth-client', () => {
...
 
     beforeEach(() => {
       fakeWindow = new FakeWindow;
     });
 
-    it('opens a popup window at the authorization URL', () => {
-      var authorized = client.authorize(fakeWindow);
+    function authorize() {
+      var popupWindow = OAuthClient.createLoginPopupWindow(fakeWindow);
+      var authorized = client.authorize(fakeWindow, popupWindow);
```
lyzadanger: There's a coupling between `createLoginPopupWindow` and `authorize` — `authorize` now requires the second param (`authWindow`) to be successful. Is there any value in testing for `authorize`'s behavior if the second (popup window) argument is omitted or invalid in some way?

src/sidebar/util/oauth-client.js
```diff
@@ -245,6 +212,41 @@ class OAuthClient {
...
+   *
+   * This function _must_ be called in the same turn of the event loop as the
+   * button or link which initiates login to avoid triggering the popup blocker
+   * in certain browsers. See https://github.com/hypothesis/client/issues/534
+   * and https://github.com/hypothesis/client/issues/535.
+   *
+   * @param {Window} $window - The parent of the popup window.
+   * @return {Window}
+   */
+  static createLoginPopupWindow($window) {
```
lyzadanger: This method not only creates the window object for, but shows (opens) the popupWindow. Worth mentioning in the comment?

src/sidebar/util/oauth-client.js
robertknight: The URL shown in the popup is what the OAuth spec terms the "authorization endpoint". The client uses the term "Log in" in the UI for user friendliness.

Is it clearer if I rename `createLoginPopupWindow` to `openAuthPopupWindow`?

src/sidebar/util/oauth-client.js
robertknight: Agreed. Will do.

src/sidebar/util/oauth-client.js
robertknight: I didn't think it was possible to create a window without also showing it to the user using DOM APIs, short of using a hack like positioning it off-screen. I'll clarify this in any case.

src/sidebar/util/oauth-client.js
robertknight: These are preferred sizes if the user agent allows, to make the popup look like Dawa's designs, with some allowance for differences in behaviour between Safari and Chrome.

The rendered HTML page is responsive, so it will adapt to other sizes on eg. mobile browsers where only full-screen windows are permitted.

src/sidebar/util/oauth-client.js
robertknight: Good spot!

src/sidebar/util/oauth-client.js
lyzadanger: I think that would feel more consistent to me, yep!

robertknight:
> I haven't tested in browser (I can if you need another sanity check!).

A sanity check in Firefox would be worthwhile. IE users are a tiny minority of our userbase and setting it up for testing is quite a faff (I use Microsoft's free Win 10 VirtualBox VMs), so I wouldn't worry about that too much.

lyzadanger:
Sounds like I have an assignment to test this in FF; everything else is looking good.

lyzadanger:
@robertknight 

Reproduced original issue on `master` in FF:

image

On your branch, "it works!":

image

I tried a few different ways through the login flow and canceling out of it, etc.; all looking good.

👍

robertknight:
Got approval from Lyza via Slack.

PR #653, opened 2018-01-22
Add window.hypothesisConfig props to settings on client boot. Fix #652

### LINKED ISSUE #652: Boot script does not read `assetRoot` and `sidebarAppUrl` settings from `window.hypothesisConfig`
### Steps to reproduce
1. Add `window.hypothesisConfig` function that returns _custom `assetRoot` and `sidebarAppUrl`_ properties. (_Edited_)
2. Check to see if these properties are applied on boot.

### Expected behaviour
`boot/index.js` should merge properties from `window.hypothesisConfig()` in addition to those added in the `script.js-hypothesis-config` tag; allowing for `assetRoot` and `sidebarAppUrl` to be set from the `window` property.

### Actual behaviour
Properties are never added before the client is instantiated.

```
User-Agent:
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36
```

patmigliaccio:
Fixes an issue with `assetRoot` not being read if the property is set from the `window.hypothesisConfig` function.

### DIFF
--- a/src/shared/settings.js
+++ b/src/shared/settings.js
@@ -46,6 +46,33 @@ function jsonConfigsFrom(document) {
   return config;
 }
 
+/**
+ * Returns a `hypothesisConfig` object from the window, or `{}`.
+ *
+ * Find the `window.hypothesisConfig` object and merge the properties
+ * of it onto the config parameter passed in.
+ *
+ * If the `config` param doesn't exist, instantiate a new object.
+ *
+ * If the `window.hypothesisConfig` exists as a function and returns an
+ * object containing properties, add it to the return object.
+ *
+ * @param {HypothesisConfig} config - an existing config object.
+ * @returns {HypothesisConfig} - an updated config object.
+ */
+function jsonConfigsFromWindow(config) {
+  var config = config || {};
+
+  if (window && typeof window.hypothesisConfig === 'function') {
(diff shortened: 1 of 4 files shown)


src/shared/settings.js
```diff
@@ -46,6 +46,33 @@ function jsonConfigsFrom(document) {
...
+ *
+ * If the `config` param doesn't exist, instantiate a new object.
+ *
+ * If the `window.hypothesisConfig` exists as a function and returns an
+ * object containing properties, add it to the return object.
+ *
+ * @param {HypothesisConfig} config - an existing config object.
+ * @returns {HypothesisConfig} - an updated config object.
+ */
+function jsonConfigsFromWindow(config) {
```
robertknight: Hmm ... this duplicates some of the logic from `config-func-settings-from.js` but with parts missing. Now that I look into how we are using `jsonConfigsFrom` in `src/annotator/config/settings.js`. I see that there are other cases where we respect configuration provided via JSON `<script>` tags but not equivalent properties from the `window.hypothesisConfig` function. Specifically the "annotations" and "query" properties.

My inclination would be to combine those two sources into one function to avoid this hazard. However, I just noticed another issue related to `window.hypothesisConfig`'s pure-ness.

PR #658, opened 2018-01-30
Hide the direct-link CTA for third party accounts

### LINKED ISSUE #656: Logged out message lead text should be optional
https://github.com/hypothesis/client/blob/563d5a5f69e7f63377af0eb38a7f10852648a678/src/sidebar/templates/loggedout-message.html#L4

This text should be configurable.

### LINKED ISSUE #657: These link to Hypothesis, rather than draw on configuration
https://github.com/hypothesis/client/blob/563d5a5f69e7f63377af0eb38a7f10852648a678/src/sidebar/templates/loggedout-message.html#L7

A quick solution would be to make the whole text optional.

screen shot 2018-01-30 at 11 53 46

robertknight:
The CTA message shown when an anonymous user visits a direct link
contains links to the Hypothesis site and makes other assumptions, such
as the user being able to create accounts and those accounts being free,
which may not be applicable on sites that use third party accounts.

For now, just hide this CTA if the page uses third party accounts.

Fixes #656
Fixes #657

----

Example of inappropriate message with current released client:

screenshot 2018-01-30 12 20 42

### DIFF
--- a/src/sidebar/components/sidebar-content.js
+++ b/src/sidebar/components/sidebar-content.js
@@ -2,6 +2,7 @@
 
 var SearchClient = require('../search-client');
 var events = require('../events');
+var isThirdPartyService = require('../util/is-third-party-service');
 var memoize = require('../util/memoize');
 var tabs = require('../tabs');
 var uiConstants = require('../ui-constants');
@@ -289,6 +290,12 @@ function SidebarContentController(
     // If user has not landed on a direct linked annotation
     // don't show the CTA.
     if (!settings.annotations) {
+      return false;
+    }
+
+    // The CTA text and links are only applicable when using Hypothesis
+    // accounts.
+    if (isThirdPartyService(settings)) {
       return false;
     }
 
--- a/src/sidebar/components/test/sidebar-content-test.js
+++ b/src/sidebar/components/test/sidebar-content-test.js
@@ -506,6 +506,17 @@ describe('sidebar.components.sidebar-content', function () {
       $scope.$digest();
       assert.isFalse(ctrl.shouldShowLoggedOutMessage());
     });
+
+    it('does not show loggedout message if using third-party accounts', function () {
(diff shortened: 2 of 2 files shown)


nlisgo:
🙏

seanh:
As we did with another similar banner in the sidebar ([Slack discussion](https://hypothes-is.slack.com/files/U074DEU66/F7W74D04U/screenshot_from_2017-11-07_12-14-17.png), [GitHub issue 1](https://github.com/hypothesis/client/pull/628), [GitHub issue 2](https://github.com/hypothesis/client/pull/630)) I wonder if we can just remove this banner from the product entirely?

- We already allow people to try to annotate when not logged in, and when they do that we inform them that they need to log in (the UX of this you-need-to-log-in message is terrible, but that should be fixed, separate problem)

- We also have a log in button in the top right of the sidebar which, if not prominent to be noticed by users, should be made more prominent

@dawariley ?

dawariley:
I've spoken to @seanh about this card and agreed that we shouldn't surface non applicable messaging to users in the UI, such as in the eLife scenario. However before disabling this CTA to login we need to determine what scenarios this message is displayed in.

I would also add that if there is an additional way to sign in/login provided on the top bar (which according to the screenshot appears to be the case) then it becomes less critical to show this message unless the goal is: to drive H signups or we are concerned that users will miss the login/signup links at the top. Trying to understand what the original rationale was for having this additional messaging display in the first place. Perhaps @dwhly can shed some light on this. 
Thanks!

seanh:
It looks to me as if this notice appears:

1. Only when you're not logged in
2. Only when you follow a direct link (e.g. [here's one](https://hyp.is/uZQIIn3REeaxT2MRd4zuYg/hypothes.is/docs/help)) so that the sidebar is filtered to show one annotation only

seanh:
:+1: Tested that the footer still shows in first-party context but no longer shows on publisher test site

PR #662, opened 2018-02-06
Refactor export of bound selectors from Redux store

robertknight:
This is a small cleanup of the exports of the Redux modules in `src/sidebar/reducers/` in preparation for upcoming changes as part of the groups work.

Group the selector functions in each Redux module together in a `selectors` object so they can be re-exported as methods on the Redux store without having to list each of them manually.

This also just makes it easier to visually see which exported functions from a Redux module are selectors.


sheetaluk:
LGTM. Will make necessary changes in my PR(https://github.com/hypothesis/client/pull/665/files) too.

PR #663, opened 2018-02-07
Localization implementation proposal (Not a complete PR)

broccolinisoup:
Hi, 
My colleague @SiliconValet an I are working on a project where we use Hypothesis as our base solution. For our project, currently we are in the process of implementation of the localisation and we'd like to share our solution with you to get your opinion and eventually contribute this feature to Hypothesis. 

First of all, **this is not a complete solution yet**. We wanted to share it with you before going so far. 

This request includes **angular-translate module implementation, $translateProvider service configuration and the UI integration**.

This structure works for us very well. (With a few more options of supported languages and translation strings on the html. Currently there is only one translation string ({{ 'Annotation' | translate}} and one language file (english_en.json) as this PR is just a sample to show what the structure looks like. )

All these configurations so far are for sidebar as we do not have Angular on  'Guest' and 'Bucketbar'. Our solution for 'Guest' and 'Bucketbar' is that translate the strings on the sidebar using angular-translate  and transfer the translated strings over the cross communication frame. (We haven't started any development for this yet. ) What would you think about this solution? 

This PR and proposal is just a beginning of the process of internationalization. We would really want to hear what you think about this. 

Please let us know if you have any questions/concerns/comments.

Note : We haven't worked on the tests yet. 

Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "angular-route": "~1.5.6",
     "angular-sanitize": "~1.5.6",
     "angular-toastr": "^1.7.0",
+    "angular-translate": "^2.17.0",
     "angulartics": "0.17.2",
     "autofill-event": "0.0.1",
     "autoprefixer": "^6.0.3",
--- a/scripts/gulp/vendor-bundles.js
+++ b/scripts/gulp/vendor-bundles.js
@@ -14,6 +14,7 @@ module.exports = {
       'angular',
       'angular-route',
       'angular-sanitize',
+      'angular-translate',
       'ng-tags-input',
       'angular-toastr',
       'angulartics/src/angulartics',
--- a/src/sidebar/components/language-list.js
+++ b/src/sidebar/components/language-list.js
@@ -1,0 +1,20 @@
+'use strict';
+
+// @ngInject
+function LanguageListController(locals) {
+
+  this.languages = locals;
+
+  this.selectLanguage = function (languageId) {
+    locals.select(languageId);
+  };
+}
+
+module.exports = {
+  controller: LanguageListController,
+  controllerAs: 'vm',
+  bindings: {
+    auth: '<',
+  },
+  template: require('../templates/language-list.html'),
+};
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -5,6 +5,7 @@ var disableOpenerForExternalLinks = require('./util/disable-opener-for-external-
 var getApiUrl = require('./get-api-url');
 var serviceConfig = require('./service-config');
 var crossOriginRPC = require('./cross-origin-rpc.js');
+var translateProvider = require('./translate-provider.js');
 require('../shared/polyfills');
 
 var raven;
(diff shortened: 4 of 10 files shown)


xolotl:
I'm not deep enough to look at the code, but would this localization strategy include capabilities to localize metadata in annotations, like the timestamps?

SiliconValet:
Time / date localization would be a different/additional library, and this is focused only on the static user interface text (not on the annotations coming from users.) That is to say, we will change the header bar, descriptions, etc but not the user submitted content or time/date formats.

We /are/ interested in localizing time/date formats but that's a separate task for us.

robertknight:
> . Our solution for 'Guest' and 'Bucketbar' is that translate the strings on the sidebar using angular-translate and transfer the translated strings over the cross communication frame. 

From our point of view, I think it would be convenient if we could use the same system for localising JS everywhere in the client and also down the road in the "h" website's front end, although admittedly there aren't a lot of translatable strings in the latter right now.

For that reason, I would be inclined to use something non-Angular specific and just create an Angular filter for it if useful in the sidebar app. Perhaps [Polyglot.js](http://airbnb.io/polyglot.js/)? I must admit I'm not that familiar with the current i18n ecosystem in JS.

In terms of requirements, is the language selector in your app purely a debugging/development tool or is there a reason why someone would need to switch between locales at runtime? In other words, could we just rely on the browser's current locale with a fallback if no translation is available?

> Time / date localization would be a different/additional library

We use standard DOM APIs for formatting date strings which already support internationalisation, so I'm hoping we won't need additional libraries for this.

SiliconValet:
The current process relies on a filter provided by angular, if we develop a replacement down the line, we simply replace that provider. I'm hesitant to make guesses about what's going to make the most sense elsewhere in the app and this allows us to proceed.

As for the language selector, there are plenty of situations where that comes in handy. The browser language isn't a completely reliable indicator for travelers or expats using local computers. In that regard, it would be useful to persist the locale selection in the jwt/profile.  But that's an iterative enhancement.

Didn't see anything in a quick glance through the sidebar stuff, but I'd be interested in seeing what that solution looks like.

robertknight:
> As for the language selector, there are plenty of situations where that comes in handy. The browser language isn't a completely reliable indicator for travelers or expats using local computers.

How is that usually solved in other web applications?  I'm just a little wary of duplicating affordances that the browser/OS already provides.

SiliconValet:
It depends on the desired user experience, but generally there is a cascade
1. Check if user has previously expressed a preference for a locale (cookie, localstorage, etc)
2. Check browser provided locale preferences against supported locales 
3. Use IP region detection

The ux provides the user with an option to control the experience. In our use cases, we sometimes have reviewers using app chroming in native language (japanese for example) to review english content. So in this case, we've got to juggle the preference for two distinct locales. 

There's no reason the appearance of that language selector can't be a configurable option, leaving deference to browser locale as the only option.

SiliconValet:
From our point of view, I think it would be convenient if we could use the same system for localising JS everywhere in the client and also down the road in the "h" website's front end, although admittedly there aren't a lot of translatable strings in the latter right now.

I misunderstood this the first time, you were suggesting we use something generic for the bucket bar, and then port that into the sidebar. We can look into that, doing the translation on the server side with polyglot may be better overall, particularly if we can cache it. We'll do some comparison work on this end, but to meet our own obligations will likely be completing this work before pursuing that alternative.

PR #665, opened 2018-02-07
Get the list of groups from the new endpoint and update it where applicable

sheetaluk:
Get the list of groups from the new endpoint and update it wherever applicable.
    
Fetch the list of groups from the new `/api/groups` endpoint instead of using the groups list returned in the `/api/profile` result. This supports providing the current document URI as a parameter, which enables the server to change the list of open groups returned depending on the current URL.
    
In order to update the groups list correctly, taking into account the current document URI, when the user joins or leaves a group, the `streamer` service now handles "session change" notifications by triggering a re-fetch of groups from the `/api/groups` endpoint instead of updating the groups directly from the payload of the WebSocket message.

### DIFF
--- a/src/sidebar/groups.js
+++ b/src/sidebar/groups.js
@@ -14,17 +14,59 @@
 var STORAGE_KEY = 'hypothesis.groups.focus';
 
 var events = require('./events');
+var { awaitStateChange } = require('./util/state-util');
 
 // @ngInject
-function groups(localStorage, serviceUrl, session, $rootScope, store) {
+function groups(annotationUI, localStorage, serviceUrl, session, $rootScope, store) {
   // The currently focused group. This is the group that's shown as selected in
   // the groups dropdown, the annotations displayed are filtered to only ones
   // that belong to this group, and any new annotations that the user creates
   // will be created in this group.
-  var focusedGroup;
+  var focusedGroupId;
+  var groups = [];
+  var documentUri;
+
+  function getDocumentUriForGroupSearch() {
+    function mainUri() {
+      var uris = annotationUI.searchUris();
+      if (uris.length === 0) {
+        return null;
+      }
+
+      // We get the first HTTP URL here on the assumption that group scopes must
+      // be domains (+paths)? and therefore we need to look up groups based on
+      // HTTP URLs (so eg. we cannot use a "file:" URL or PDF fingerprint).
+      return uris.find(uri => uri.startsWith('http'));
+    }
+    return awaitStateChange(annotationUI, mainUri);
+  }
+
+  /**
+   * Fetch the list of applicable groups from the API.
+   *
+   * The list of applicable groups depends on the current userid and the URI of
+   * the attached frames.
+   */
+  function load() {
+    return getDocumentUriForGroupSearch().then(uri => {
+      return store.groups.list({ document_uri: uri });
+    }).then(gs => {
+      $rootScope.$apply(() => {
+        var focGroup = focused();
+        if (focGroup) {
+          var focusedGroupInFetchedList = gs.some(g => g.id === focGroup.id);
+          if (!focusedGroupInFetchedList) {
+            focus(gs[0].id);
+          }
+        }
+        groups = gs;
+      });
+      return gs;
+    });
+  }
 
   function all() {
-    return session.state.groups || [];
+    return groups;
   }
 
   // Return the full object for the group with the given id.
@@ -58,14 +100,16 @@ function groups(localStorage, serviceUrl, session, $rootScope, store) {
    * a previous session. Lastly, we fall back to the first group available.
    */
   function focused() {
-    if (focusedGroup) {
-      return focusedGroup;
+    if (focusedGroupId) {
+      return get(focusedGroupId);
     }
+
     var fromStorage = get(localStorage.getItem(STORAGE_KEY));
     if (fromStorage) {
-      focusedGroup = fromStorage;
-      return focusedGroup;
+      focusedGroupId = fromStorage.id;
+      return fromStorage;
     }
+
     return all()[0];
   }
 
@@ -74,22 +118,37 @@ function groups(localStorage, serviceUrl, session, $rootScope, store) {
     var prevFocused = focused();
     var g = get(id);
     if (g) {
-      focusedGroup = g;
+      focusedGroupId = g.id;
       localStorage.setItem(STORAGE_KEY, g.id);
       if (prevFocused.id !== g.id) {
         $rootScope.$broadcast(events.GROUP_FOCUSED, g.id);
       }
     }
   }
-
+  
   // reset the focused group if the user leaves it
   $rootScope.$on(events.GROUPS_CHANGED, function () {
-    if (focusedGroup) {
-      focusedGroup = get(focusedGroup.id);
-      if (!focusedGroup) {
-        $rootScope.$broadcast(events.GROUP_FOCUSED, focused());
+    // return for use in test
+    return load();
+  });
+
+  // refetch the list of groups when user changes
+  $rootScope.$on(events.USER_CHANGED, () => {
+    // FIXME Makes a second api call on page load. better way?
+    // return for use in test
+    return load();
+  });
+
+  // refetch the list of groups when document url changes
+  $rootScope.$on(events.FRAME_CONNECTED, () => {
+    // FIXME Makes a third api call on page load. better way?
+    // return for use in test
+    return getDocumentUriForGroupSearch().then(uri => {
+      if (documentUri !== uri) {
+        documentUri = uri;
+        load();
       }
-    }
+    });
   });
 
   return {
@@ -97,6 +156,7 @@ function groups(localStorage, serviceUrl, session, $rootScope, store) {
     get: get,
 
     leave: leave,
+    load: load,
 
     focused: focused,
     focus: focus,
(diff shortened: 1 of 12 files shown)


src/sidebar/groups.js
```diff
@@ -15,16 +15,78 @@ var STORAGE_KEY = 'hypothesis.groups.focus';
...
+ *
+ * `await` returns a Promise which resolves when a selector function,
+ * which reads values from a Redux store, returns non-null.
+ *
+ * @param {Object} store - Redux store
+ * @param {Function<T|null>} selector - Function which returns a value from the
+ *   store if the criteria is met or `null` otherwise.
+ * @return {Promise<T>}
+ */
+function awaitSelector(store, selector) {
```
robertknight: Just noting that we had a chat in Slack yesterday and agreed to change this to `awaitStateChange`.

src/sidebar/groups.js
```diff
@@ -81,22 +143,36 @@ function groups(localStorage, serviceUrl, session, $rootScope, store) {
...
-      if (!focusedGroup) {
-        $rootScope.$broadcast(events.GROUP_FOCUSED, focused());
+    // FIXME A better way of doing this?
+    load();
+  });
+
+  // refetch groups when user changes
+  $rootScope.$on(events.USER_CHANGED, () => {
+    // FIXME Makes a second api call on page load. better way?
+    load();
```
robertknight: It may be helpful to draw a flow diagram / state machine here of the desired flow.

In doing this, I think you'll find that some information is missing from this event. It is fired when the profile is (1) _initially fetched_, as well as when the (2) user logs in or (3) logs out. You only want to re-fetch groups however when (2) or (3) happen.

src/sidebar/templates/annotation-header.html
```diff
@@ -23,7 +23,11 @@
     <br>
     <span class="annotation-header__share-info">
       <a class="annotation-header__group"
-        target="_blank" ng-if="vm.group() && vm.group().url" href="{{vm.group().url}}">
+        ng-if="vm.group() && vm.group().type == 'open'">
+        <i class="h-icon-public"></i><span class="annotation-header__group-name">{{vm.group().name}}</span>
+      </a>
+      <a class="annotation-header__group"
+        target="_blank" ng-if="vm.group() && !vm.group().type == 'open' && vm.group().url" href="{{vm.group().url}}">
```
robertknight: I'll note that we don't need to change the behaviour here for open groups any more because activity pages works for them. However, I don't think this code does what you intended.

In `!vm.group().type == 'open'` the unary `!` operator has [higher precedence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence) than the binary `==` operator so if the group type is "private" this will evaluate to `!'private' == open` => `false == 'open'` => `false` and hide the link instead of showing it.

src/sidebar/util/state-util.js
```diff
@@ -0,0 +1,32 @@
...
+      if (result !== null) {
+        unsubscribe();
+        resolve(result);
+      }
+    });
+  });
+}
+
+module.exports = {
+  awaitStateChange: awaitStateChange,
```
robertknight: In object literals in new code, I've been using [object literal shorthand](https://ariya.io/2013/02/es6-and-object-literal-property-value-shorthand) to shorten `{ foo: foo }` to `{ foo }`. This isn't mandatory, but I find it a little more readable.

src/sidebar/util/test/state-util-test.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+var fakeStore = require('../../test/fake-redux-store');
+var stateUtil = require('../state-util');
+
+describe('state-util', function () {
+  var number;
```
robertknight: `number` is a Redux store, so I would suggest naming this variable `store`.

src/sidebar/util/test/state-util-test.js
```diff
@@ -0,0 +1,42 @@
...
+      var expected = 5;
+
+      number.setState({ val: 5 });
+
+      return stateUtil.awaitStateChange(number, findNum).then(function (actual) {
+        assert.equal(actual, expected);
+      });
+    });
+
+    it('should wait for state change from a null value to a non-null value', function () {
```
robertknight: I think the description here could use a slight tweak to clarify that `awaitStateChange` waits for the _selector_ function to return a non-null value.

src/sidebar/util/test/state-util-test.js
```diff
@@ -0,0 +1,42 @@
...
+
+describe('state-util', function () {
+  var number;
+
+  beforeEach(function() {
+    number = fakeStore({ val: 0 });
+  });
+
+  describe('awaitStateChange()', function () {
+    function findNum(store) {
```
robertknight: This function could use a slightly more descriptive name.

src/sidebar/groups.js
```diff
@@ -14,17 +14,59 @@
...
+   * The list of applicable groups depends on the current userid and the URI of
+   * the attached frames.
+   */
+  function load() {
+    return getDocumentUriForGroupSearch().then(uri => {
+      return store.groups.list({ document_uri: uri });
+    }).then(gs => {
+      var focGroup = focused();
+      if (focGroup) {
+        var focusedGroupInFetchedList = gs.find(g => g.id === focGroup.id);
```
robertknight: `Array.find` returns the matching item from the list, whereas the variable name `focusedGroupInFetchedList` and subsequent usage treats it as a boolean, although it happens to work because a non-null object is truthy. Use `Array.some` if you want to check whether an item in an array matches some condition.

src/sidebar/groups.js
```diff
@@ -14,17 +14,59 @@
...
   // The currently focused group. This is the group that's shown as selected in
   // the groups dropdown, the annotations displayed are filtered to only ones
   // that belong to this group, and any new annotations that the user creates
   // will be created in this group.
-  var focusedGroup;
+  var focusedGroupId;
+  var groups = [];
+  var documentUri;
+
+  function getDocumentUriForGroupSearch() {
```
robertknight: I think it would be useful to have a JSDoc comment here clarifying what this returns.

src/sidebar/test/groups-test.js
```diff
@@ -58,121 +68,156 @@ describe('groups', function() {
...
+        assert.deepEqual(groups, [
+          {name: 'Group 1', id: 'id1'},
+          {name: 'Group 2', id: 'id2'},
+          {name: 'Group 3', id: 'id3'},
+        ]);
+      });
+    });
+  });
+
+  describe('.load() method', function() {
```
robertknight: Some of our tests are not consistent about this, but the canonical naming for `describe` blocks for methods is `describe('#someMethod')`, taken from JSDoc conventions for referencing instance methods/properties.

src/sidebar/test/groups-test.js
```diff
@@ -58,121 +68,156 @@ describe('groups', function() {
...
-      assert.deepEqual(groups, [
-        {name: 'Group 1', id: 'id1'},
-        {name: 'Group 2', id: 'id2'},
-        {name: 'Group 3', id: 'id3'},
-      ]);
+    it('focuses on the first in the list of groups if user leaves the focused group', function () {
+      var svc = service();
+
+      return svc.load().then(() => {
+        return svc.focus('id2');
```
robertknight: `svc.focus()` is a synchronous method that doesn't return a value, so there is no need to wait for the result here. It happens to work because a `return` statement without a value in a Promise callback does nothing and just results in the next `then` in the chain being called.

src/sidebar/util/test/state-util-test.js
```diff
@@ -0,0 +1,42 @@
...
+
+describe('state-util', function () {
+  var store;
+
+  beforeEach(function() {
+    store = fakeStore({ val: 0 });
+  });
+
+  describe('awaitStateChange()', function () {
+    function getValWhenGreaterThanTwo(store) {
```
robertknight: 👍

src/sidebar/util/test/state-util-test.js
```diff
@@ -0,0 +1,42 @@
...
+      if (store.getState().val < 3) {
+        return null;
+      }
+      return store.getState().val;
+    }
+
+    it('should return promise that resolve to a non-null value', function () {
```
robertknight: s/resolve/resolves

src/sidebar/groups.js
```diff
@@ -14,17 +14,59 @@
...
+      var focGroup = focused();
+      if (focGroup) {
+        var focusedGroupInFetchedList = gs.find(g => g.id === focGroup.id);
+        if (!focusedGroupInFetchedList) {
+          $rootScope.$apply(() => {
+            focus(gs[0].id);
+          });
+        }
+      }
+      groups = gs;
```
robertknight: As you found in your local testing, you need to wrap the `groups` assignment in `$rootScope.$apply`. Since this variable is modified in a Promise callback, and therefore outside of whatever context triggered the `load` call, Angular won't know that it should do a digest cycle at this point.

src/sidebar/groups.js
```diff
@@ -74,29 +118,45 @@ function groups(localStorage, serviceUrl, session, $rootScope, store) {
...
-        $rootScope.$broadcast(events.GROUP_FOCUSED, focused());
+    // return for use in test
+    return load();
+  });
+
+  // refetch the list of groups when user changes
+  $rootScope.$on(events.USER_CHANGED, () => {
+    // FIXME Makes a second api call on page load. better way?
+    // return for use in test
+    return load();
```
robertknight: I only saw one call to `/api/groups` on page load. I haven't looked into why yet.

PR #668, opened 2018-02-12
Update to Angular 1.6.9

robertknight:
Update to the current stable version of Angular (1.5.6 => 1.6.9). The [full changelog](https://github.com/angular/angular.js/blob/master/CHANGELOG.md) has all the details including [breaking changes](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#breaking-changes) since 1.5.x.

The main backwards incompatible change which affects us is that by default component inputs are no longer available in controller constructors but only during/after the `$onInit` method is called [1]. There is a `preAssignBindingsEnabled` backwards compatibility helper which preserves the previous behaviour. I have enabled this in both the sidebar application and tests for components which need it.

See commit messages for details on other fixes required.

[1] Controllers in Angular are effectively classes and this change was introduced to make it possible to use ES6 classes as components, plus it has the benefit of encouraging the init code for a component to be put in one obvious place.


seanh:
It's hard to test everything and know for sure that a framework update like this didn't break anything, but on a brief look I couldn't find any problems

PR #674, opened 2018-02-16
Only wait for document URI before fetching groups in sidebar

robertknight:
Fixes a regression after 36c466084 where the stream and single
annotation page views failed to load.

36c466084 made the groups service wait for the host page's document URI
to be determined before making a call to the `/api/groups` endpoint
passing that URI as a parameter. In the stream (`/stream`) and single
annotation page routes, there is no host page and so
the promise returned by `getDocumentUriForGroupSearch` never resolved.
Avoid waiting for this URI in that case and avoid passing the
`document_uri` parameter to the `/api/groups` endpoint.

This commit introduces a new `isSidebar` test rather than using
`annotationUI.isSidebar` because that piece of app state is not yet
initialized at the point when `groups.load()` is called in the route's
`resolve` function in src/sidebar/index.js.

### DIFF
--- a/src/sidebar/groups.js
+++ b/src/sidebar/groups.js
@@ -17,7 +17,7 @@ var events = require('./events');
 var { awaitStateChange } = require('./util/state-util');
 
 // @ngInject
-function groups(annotationUI, localStorage, serviceUrl, session, $rootScope, store) {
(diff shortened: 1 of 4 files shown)


sheetaluk:
LGTM

PR #675, opened 2018-02-19
Remove activity page links for third-party open groups

robertknight:
Activity pages are currently only supported for open groups belonging to
the h service's first party authority.

This PR fixes the issue from the client side by hiding the group link in this case. An alternate solution would be to avoid generating these links and including them in the `/api/groups` response in the first place since they don't currently display anything useful.

----

**Testing**

1. Go to https://elifesciences.org/articles/35034.
2. Click “Annotations” link at top of page
3. Click “eLIFE” label at top of sidebar

**Expected**

No "View group activity" link under the eLIFE group

**Actual**

A "View group activity" link appears under the eLIFE group. Clicking it goes to a page that 404s.

### DIFF
--- a/src/sidebar/components/group-list.js
+++ b/src/sidebar/components/group-list.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var { isThirdPartyUser } = require('../util/account-id');
+var isThirdPartyService = require('../util/is-third-party-service');
 var serviceConfig = require('../service-config');
 
 // @ngInject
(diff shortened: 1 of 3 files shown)


lyzadanger:
I also like the idea of only providing a relevant URL in the API response if it's relevant. Happy to work on that change to help keep this logic out of the client.

lyzadanger:
With https://github.com/hypothesis/publisher-account-test-site using `h-client` on `master`:

screen shot 2018-02-21 at 9 35 06 am

And with `h-client` on this PR branch:

screen shot 2018-02-21 at 9 36 54 am

AKA 👍

PR #676, opened 2018-02-19
Pass "authority" query param when fetching groups list

robertknight:
Without this argument, the "h" service does not know which authority's
groups should be returned if the user is not logged in.

Consequently if you go to an eLife article page at the moment and are not logged in, you see the Hypothesis "Public" group (!) - https://elifesciences.org/articles/32077


robertknight:
I'm going to go ahead and get this merged because this is a rather visible regression on a partner's website.

PR #677, opened 2018-02-19
Specify npm package version to publish to S3

### LINKED ISSUE #602: Client release process did not upload correct version to S3
Client releases are hosted on https://cdn.hypothes.is, which is a backed by an S3 bucket. During the release build process, the client is first published to npm and then [s3-npm-publish](https://github.com/nickstenning/s3-npm-publish/blob/master/publish) is run to publish the "latest" client release to S3.

When v1.54 was tagged, the Jenkins build published the package as v1.54 to npm but apparently uploaded v1.53 to S3: https://jenkins.hypothes.is/job/client/job/master/568/console

We currently invoke `s3-npm-publish` without specifying a package name explicitly and my hypothesis is that npm is only eventually consistent with respect to updating the "latest" version of a package when a new released is published.

robertknight:
Sometimes the "Publish" step has ended up publishing version N-1 to the
CDN when building the tag for version N. I believe this happens due to
npm's package publishing API not guaranteeing immediate visiblity of the
latest version when it returns. If the publish task runs before the
update has propagated inside their network, then the publish step
re-uploads the previous version to the CDN.

This commit attempts to resolve the issue by specifying a package
version explicitly. We might find that this is not enough and that the
publish step either has to wait until the update has propagated or we
have to use a locally-built tarball.

Fixes #602

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,9 +5,14 @@ node {
 
     nodeEnv = docker.image("kkarczmarczyk/node-yarn:7.5")
     workspace = pwd()
+    pkgVersion = sh (
+      script: 'cat package.json | jq -r .version',
+      returnStdout: true
+    ).trim()
 
     stage 'Build'
     nodeEnv.inside("-e HOME=${workspace}") {
+        sh "echo Building Hypothesis client \"${pkgVersion}\""
         sh 'make clean'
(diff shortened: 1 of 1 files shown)


robertknight:
I'm going to test this out in a release.

PR #681, opened 2018-02-23
Pdf subframe annotation

birdbrained:
Allows a subframe containing PDF.js to be annotated by providing configuring hypothesis to use the PDF annotator plugin.

### DIFF
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -51,6 +51,10 @@ $.noConflict(true)(function() {
   }
 
   if (config.subFrameIdentifier) {
+    // Make sure the PDF plugin is loaded if the subframe contains the PDF.js viewer.
+    if (typeof window.PDFViewerApplication !== 'undefined') {
+      config.PDF = {};
+    }
     Klass = Guest;
 
     // Other modules use this to detect if this


birdbrained:
```
<html>
    <script src="hypothesis.js"></script>
    <!-- Out here the script gets set up for a regular html document -->
    <iframe src="/pdfjs-viewer">
        <!-- Inside the iframe, the script is also getting set up as a document annotator, but this is actually a pdf.js viewer. The config here https://github.com/hypothesis/client/blob/master/src/annotator/guest.coffee#L106 doesn't look like it knows its a pdf, because it only sets the script up as a pdf if its doing a sidebar https://github.com/hypothesis/client/blob/master/src/annotator/pdf-sidebar.coffee#L4, and in this case, it's a subframe. -->
    </iframe>
</html>
```

robertknight:
Thanks for the context @birdbrained. I'll take another look at this later today.

ajpeddakotla:
@robertknight The problem that this PR is trying to address is that when H is embedded in a frame around a PDF, there are timing issues, specifically with the time it takes to load H that results in annotations being orphaned when they shouldn't be. 

If there is an alternative strategy to fix this issue that this PR doesn't address, please let us know that.

birdbrained:
Just a quick update we realized the change we made was breaking the html annotation on the parent iframe; this change fixes that.

birdbrained:
@ajpeddakotla Hey this PR actually fixes a similar but slightly different issue. The timing issue where some annotations will show as orphaned during the time they are being loaded is separate from what this pr fixes. This pr makes a change that was needed because the pdf iframe doesn't load the correct annotation plugin- it loads the html plugin even though it should load the pdf plugin. That was leading to annotations that were permanently orphaned.

judell:
>the pdf iframe doesn't load the correct annotation plugin- it loads the html plugin even though it should load the pdf plugin.

Can you help us understand how and why that happens?

Here's a reference example:

image

The underlying HTML:

```
<html>
<head></head>

<body>

<p>This is the outer frame.</p>

<iframe width="90%" height="100%" src="https://hyp.jonudell.info/pdfjs/viewer/web/viewer.html"></iframe>

</body>
</html>
```

The iframe's src points to a hosted instance of the sample repo at https://github.com/hypothesis/pdf.js-hypothes.is.

In this case the PDF plugin is loaded into the annotation client that's injected into the PDF iframe. Can you say more about the setup that creates the problem you're addressing?

birdbrained:
(Sorry for spreading that out across 3 comments, I think GitHub has a length limit)

birdbrained:
The separate issue mentioned above I believe is regarding the fact that when annotations are loaded into a particularly large pdf, annotations will be briefly listed as 'orphaned' because the annotator hasn't (had time to?) anchor them yet. They do show up eventually; it seems the issue was is there something more constructive to do with them during that time frame? perhaps having a state called 'Anchoring...' with some sort of indicator. I have not looked at that issue in any depth.

dwhly:
> In our case, the hypothesis client is embedded into the outer frame

From our conversation this afternoon I did not realize you were embedding 2x  (inner frame and outer).  Would it be possible for you to only implement the client on the subframe w/ the PDF for the first iteration?  Just want to understand your flexibility so we can get you up and running as soon as possible.

It may be possible for us to accommodate this ... @robertknight will know that... but in the event that this is complicated, want to understand our options.

birdbrained:
@dwhly From what I understand wouldn't this break hypothesis for your users that have the extension installed?

birdbrained:
I've updated the index.js file to handle the case of a pdf subframe there. Is there anything else that you see needs attention?

robertknight:
Tested as follows:

1. Clone and build latest version of PDF.js following steps at https://github.com/mozilla/pdf.js/#getting-the-code
2. Create a temp dir and cd into it
3. Copy the built PDF.js viewer to the temp dir using `cp -R path/to/pdf.js/build/generic pdfjs-viewer`
4. Create the following `index.html` file in the temp dir:

```html
<html>
  <body>
    This is the parent page.
    <br>
    <iframe width="800" height="800" enable-annotation src="pdfjs-viewer/web/viewer.html"></iframe>
    <script src="http://localhost:5000/embed.js"></script>
  </body>
</html>
```

5. Serve the temp dir using `python -m SimpleHTTPServer 8000` and browse to http://localhost:8000
6. Navigate to http://localhost:8000 and create annotations in both the top-level HTML page and the embedded PDF.
7. Reload the page and check that both annotations anchor

robertknight:
Thank-you for the contribution @birdbrained and for your patience waiting for this to get reviewed!

robertknight:
Released as v1.76

PR #684, opened 2018-03-05
Observe and react to URL changes made via History API

robertknight:
This is a first step to towards making the Hypothesis client work better with SPAs and sites that use PJAX (eg. GitHub, Medium).

Web pages can change the current URL without loading a complete new page from the server by using `window.history` functions. This PR makes the client observe these URL changes and update the set of displayed annotations in response, as well as the URL and metadata associated with subsequently created annotations.

## Testing

1. Go to https://medium.com
2. Activate the Hypothesis client
3. Click on a link in Medium that takes you to a page for a specific article. Note that the browser doesn't actually load a new page but displays a custom progress bar at the top of the content area.
4. Annotate the document

Expected result on master: Annotation created in step (4) is associated with URL from step (1) instead of URL from step (3).

Expected result on this branch: Annotation created in step (4) is associated with URL from step (3).

## Implementation overview

- The first commit adds a utility function (`historyObservable`) which monkey-patches `window.history` methods in order to observe URL changes and returns an `Observable<string>` [1] of URL changes that consumers can subscribe to. The monkey-patching is necessary because there are no DOM events triggered the URL is changed in this way.
- The second commit uses this utility to watch for URL changes and send a message from the host frame to the sidebar with the updated URL and document metadata for the frame.
- The third commit listens for this message in the sidebar and updates the frame metadata in the app state in response. This in turn triggers the client to unload annotations for the previous URL and fetch annotations for the new URL.

[1] For those new to Observables in JS, see [this explainer](https://github.com/tc39/proposal-observable). The TL;DR is that it is a standard API for working with push-based data sources.

## Caveats

There are some limitations with this initial implementation:

1. Hypothesis assumes the new content has already been loaded by the time the URL changes, or will at least have loaded by the time it has fetched the annotations for the new URL and tries to anchor them. If the new content is still being loaded by the web app at the point when the annotations are fetched, the client may fail to anchor them.

2. There is no monitoring of URL changes in iframes.

Both of these can be resolved but I left them out to keep this first PR smaller. My current thinking for (1) is that the client will observe DOM changes using `MutationObserver` and retry anchoring of orphans when this happens. Resolving (2) just requires some additional plumbing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "katex": "^0.7.1",
     "lodash.debounce": "^4.0.3",
     "lodash.get": "^4.3.0",
+    "lodash.isequal": "^4.5.0",
     "mkdirp": "^0.5.1",
     "mocha": "^2.4.5",
     "ng-tags-input": "^3.1.1",
--- a/src/annotator/guest.coffee
+++ b/src/annotator/guest.coffee
@@ -8,6 +8,7 @@ $ = require('jquery')
 
 adder = require('./adder')
 highlighter = require('./highlighter')
+historyObservable = require('./util/history-observable')
 rangeUtil = require('./range-util')
 selections = require('./selections')
 xpathRange = require('./anchoring/range')
@@ -90,6 +91,15 @@ module.exports = class Guest extends Delegator
     @crossframe.onConnect(=> this._setupInitialState(config))
     this._connectAnnotationSync(@crossframe)
     this._connectAnnotationUISync(@crossframe)
+
+    # Observe URL changes made via
+    # `window.history.{pushState, replaceState, popState}` in SPAs and web pages
+    # using PJAX.
+    @historyChanges = historyObservable().subscribe(=>
+      this.getDocumentInfo().then((info) =>
+        @crossframe.call('updateFrame', info)
+      )
+    )
 
     # Load plugins
     for own name, opts of @options
@@ -176,6 +186,7 @@ module.exports = class Guest extends Delegator
     $('#annotator-dynamic-style').remove()
 
     this.selections.unsubscribe()
+    this.historyChanges.unsubscribe()
     @adder.remove()
 
     @element.find('.annotator-hl').each ->
--- a/src/annotator/guest.coffee
+++ b/src/annotator/guest.coffee
@@ -9,6 +9,8 @@ $ = require('jquery')
 adder = require('./adder')
 highlighter = require('./highlighter')
 historyObservable = require('./util/history-observable')
+metaObservable = require('./util/document-meta-observable')
+{ buffer, merge } = require('./util/observable')
 rangeUtil = require('./range-util')
 selections = require('./selections')
 xpathRange = require('./anchoring/range')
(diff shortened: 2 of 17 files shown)


robertknight:
I did some actual testing on Medium and it turned out that additional work was required:

1. Clearing internal caches of document metadata
2. Observing changes to `<meta>` and `<link>` tags. Medium updates the URL using the History API _before_ it updates the `<link rel="canonical">` tag. Since the canonical URL overrides the actual URL, we need to watch for changes to this as well

I've updated the PR but it is getting too large now so I'll need to split this into smaller pieces.

judell:
Wow. Thanks for tackling this!

robertknight:
Thanks @judell. It might be a little while before this gets merged, I'm just figuring out what is necessary at the moment in order to solve this well enough to work for at least some popular websites that users have encountered issues with.

judell:
>   I'm just figuring out what is necessary at the moment in order to solve
this well enough to work for at least some popular websites that users have
encountered issues with.

Yep. Appreciate that!

klemay:
It works!

I have annotated four Medium articles, and they have been associated with the correct page:

- https://hyp.is/UVTNxklqEeihsns41fm-vA/medium.com/@LanceUlanoff/facebook-is-not-a-monopoly-daa4124746be
- https://hyp.is/Qz7fzkiiEeiOE1dum119iw/medium.com/s/story/yes-you-should-delete-facebook-heres-why-bc623a3b4625
- https://hyp.is/KBcNTEciEei37BPaWrh4Pg/medium.com/basic-income/its-time-for-technology-to-serve-all-humankind-with-unconditional-basic-income-e46329764d28
- https://hyp.is/8b372kcgEeireOsPwIwqmQ/medium.com/@mijordan3/artificial-intelligence-the-revolution-hasnt-happened-yet-5e1d5812e1e7

There is one strange occurrence, but I don't think it's related to this PR, because I've also [seen & documented this on NPR's website](https://github.com/hypothesis/product-backlog/issues/535). H is grouping all of these articles together under one card on my activity page (I put the title of each article in my test annotation to make this clearer):

screen shot 2018-04-26 at 12 12 11 pm

Those `hyp.is` links on the individual annotation cards are all correct, but the `URL` to the right of the annotation cards belongs to one of the 4 articles I annotated & `Visit annotations in context` will take you to that one article.

Again, since I've seen this exact case on another website, I don't think this is the result of anything in this PR, just something I'm noting in case it's relevant.

robertknight:
> There is one strange occurrence, but I don't think it's related to this PR, because I've also seen & documented this on NPR's website. H is grouping all of these articles together under one card on my activity page (I put the title of each article in my test annotation to make this clearer):

Thanks for flagging that up. I think a dev should look into that to understand why it is happening before we merge this, just in case it is related.

robertknight:
Rebased.

klemay:
> I'd be curious to take a look at that.

There's an open issue for it here: https://github.com/hypothesis/product-backlog/issues/535

PR #686, opened 2018-03-07
Default to notes tab when only notes present

hmstepanek:
When only page notes are present default the focused tab to be page notes instead of the annotations tab. This is a fix for https://github.com/hypothesis/product-backlog/issues/375.

### DIFF
--- a/src/sidebar/reducers/selection.js
+++ b/src/sidebar/reducers/selection.js
@@ -14,6 +14,7 @@ var immutable = require('seamless-immutable');
 
 var toSet = require('../util/array-util').toSet;
 var uiConstants = require('../ui-constants');
+var tabs = require('../tabs');
 
 var util = require('./util');
 
@@ -130,6 +131,17 @@ var update = {
       sortKey: TAB_SORTKEY_DEFAULT[action.tab],
       sortKeysAvailable: TAB_SORTKEYS_AVAILABLE[action.tab],
     };
+  },
+
+  ADD_ANNOTATIONS(state, action) {
+    var counts = tabs.counts(action.annotations);
+    // If there are no annotations at all, ADD_ANNOTATIONS will not be called.
+    var haveOnlyPageNotes = counts.notes === action.annotations.length;
+    // If this is the init phase and there are only page notes, select the page notes tab.
+    if (state.annotations.length === 0 && haveOnlyPageNotes){
+      return {selectedTab: uiConstants.TAB_NOTES};
+    }
+    return {};
   },
 
   SET_FILTER_QUERY: function (state, action) {
--- a/src/sidebar/test/annotation-ui-test.js
+++ b/src/sidebar/test/annotation-ui-test.js
@@ -80,7 +80,7 @@ describe('annotationUI', function () {
         [sinon.match(annot)]);
     });
 
-    it('does not change selectedTab state if annotations are already loaded', function () {
+    it('does not change `selectedTab` state if annotations are already loaded', function () {
       var annot = defaultAnnotation();
       annotationUI.addAnnotations([annot]);
       var page = oldPageNote();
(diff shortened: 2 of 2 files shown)


judell:
Thanks @hmstepanek!

src/sidebar/reducers/selection.js
```diff
@@ -132,6 +133,17 @@ var update = {
     };
   },
 
+  ADD_ANNOTATIONS(state, action) {
+    var counts = tabs.counts(action.annotations);
+    // If there are no annotations at all, ADD_ANNOTATIONS will not be called.
```
robertknight: I'm not sure about this actually. I don't think anything prohibits this action from being dispatched with an empty list of actions. I think we probably ought to be defensive here and bail if the list of annotations is empty.

src/sidebar/reducers/selection.js
```diff
@@ -132,6 +133,17 @@ var update = {
     };
   },
 
+  ADD_ANNOTATIONS(state, action) {
+    var counts = tabs.counts(action.annotations);
```
robertknight: `tabs.counts` works but it also computes counts that we don't actually use - such as orphans and annotations. It would be slightly more efficient to just use the same code that `tabs.js` uses to compute the `notes` property: `countIf(annotations, metadata.isPageNote)`.

src/sidebar/test/annotation-ui-test.js
```diff
@@ -79,6 +80,19 @@ describe('annotationUI', function () {
         [sinon.match(annot)]);
     });
 
+    it('sets the selectedTab to page if only page notes are present', function () {
```
robertknight: I would suggest we put these tests in the test module that corresponds to the logic changed above, ie. `selections-test.js`. This will mean changing the style of the test a little bit.

I think we also need one more test for the case when annotations are already loaded.

src/sidebar/reducers/selection.js
```diff
@@ -132,6 +133,17 @@ var update = {
...
+    // If there are no annotations at all, ADD_ANNOTATIONS will not be called.
+    var haveOnlyPageNotes = counts.notes === action.annotations.length;
+    // If this is the init phase and there are only page notes, select the page notes tab.
+    if (state.annotations.length === 0 && haveOnlyPageNotes){
+      return {selectedTab: uiConstants.TAB_NOTES};
+    }
+    return {selectedTab: state.selectedTab};
```
robertknight: You can actually `return null` or `return {}` here if you don't want to make any changes. The returned changes are merged into a copy of the existing state.

src/sidebar/reducers/selection.js
hmstepanek: It is not called if there are no annotations.

src/sidebar/test/annotation-ui-test.js
hmstepanek: I made a separate PR for splitting out the tests. Seems like it might bet better to just split out all these all at once rather than doing it a little at a time. I'll add a test for where annotations are already loaded-that's definitely important.

hmstepanek:
Ok updated PTAL.  Again haven't split out the tests would like to do that in this issue here all at once: https://github.com/hypothesis/product-backlog/issues/523. Also I don't think there is a need to test with [] as ADD_ANNOTATIONS isn't called if there are no annotations.

src/sidebar/test/annotation-ui-test.js
```diff
@@ -79,6 +80,27 @@ describe('annotationUI', function () {
...
 
+    it('does not change selectedTab state if annotations are already loaded', function () {
+      var annot = defaultAnnotation();
+      annotationUI.addAnnotations([annot]);
+      var page = oldPageNote();
+      annotationUI.addAnnotations([page]);
+      assert.equal(annotationUI.getState().selectedTab, uiConstants.TAB_ANNOTATIONS);
+    });
+
+    it('sets the selectedTab to page if only page notes are present', function () {
```
robertknight: I think you meant `sets the selectedTab to "notes"` or `sets the selectedTab to "TAB_NOTES"`?

robertknight:
>  Again, haven't split out the tests would like to do that in this issue here all at once:  hypothesis/product-backlog#523. 

That makes sense.

PR #700, opened 2018-03-22
Extract `createStore` helper out of `store` service

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/695**~~

The `store` service constructor in `sidebar/store/index.js` contains some useful logic for composing a store out of a set of _modules_ which handle different parts of the application state, exposing the actions and selectors from each module as methods on the store.

To facilitate re-use of this logic in tests and to make it easier to test this logic directly, extract it out of the `store` function into a separate `createStore` helper.

Doing this enables removal of `modules/index.js` which existed to combine parts of the modules together.

Once this is landed, it will be easier to write tests for individual store modules as they can re-use the `createStore` helper.


src/sidebar/store/index.js
```diff
@@ -31,21 +31,15 @@
  *  3. Checking that the UI correctly presents a given state.
```
hmstepanek: There's a comment above here that says

> *  4. Other parts of the app can subscribe to changes in the app state.
> *     This is used to to update the UI etc.

Does this still apply?

Also does this comment block belong here-seems like it should get moved to the tests instead?
> * Tests for a given action consist of:

src/sidebar/store/test/index-test.js
```diff
@@ -289,15 +289,6 @@ describe('store', function () {
     ]);
   });
 
-  describe('#subscribe()', function () {
```
hmstepanek: I'm not seeing where subscribe got moved to in the new implementation, also not seeing any tests for this behavior?

robertknight: `subscribe` is actually part of the basic [store API](https://redux.js.org/api-reference/store#subscribe). However I added an additional test (`"notifies subscribers when state changes"`) in `create-store.js`.

Most of the tests that are currently in `index-test.js` will be moved to `store/modules/{relevant module}-test.js` but after that I think it will make sense to have some integration tests to verify that all the pieces work together as expected.

src/sidebar/store/index.js
robertknight: Yes. The API of the `store` service has not changed from the point of view of the rest of the application, so I left this initial block unchanged.

This PR is purely about making it easier to write store modules and tests.

robertknight:
Thanks for the feedback @hmstepanek. I added one additional test and responded to your comment about the docs. Please let me know if this clarifies things for you.

PR #702, opened 2018-03-27
Configurable sidebar UI placement

JCCR:
Included is an ADR that's proposing a new optional feature that would allow the client's sidebar user interface to be displayed in a chosen container element within the host page.

### DIFF
--- a/docs/developers/arch/adr-004.md
+++ b/docs/developers/arch/adr-004.md
@@ -1,0 +1,40 @@
+# Configurable Sidebar UI placement
+
+This document is about proposing a new optional feature that would allow the client's sidebar user interface to be displayed in a chosen container element within the host page.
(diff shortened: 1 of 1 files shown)


robertknight:
Hi @JCCR. The proposal generally looks good to me. ADRs are intended for code architecture decisions rather than user-facing features. Proposals such as this should go in an issue in the https://github.com/hypothesis/product-backlog repo which capture:

1. Most importantly, the user-facing need.
2. Any notes or thoughts you have about the implementation, as a clearly distinct thing from (1)

I think you've got both parts here. Did you mention somewhere that you have a demo or screenshot that you can use to illustrate what you have in mind?

JCCR:
@robertknight 
Makes sense, thanks for pointing out the product-backlog.
I've opened this as an issue there. https://github.com/hypothesis/product-backlog/issues/576

Closing this.

PR #707, opened 2018-04-04
`externalContainerSelector` config implementation

JCCR:
This is an implementation based on what's proposed in https://github.com/hypothesis/client/pull/702

Please refer to the commit messages in this PR for the details on what's been done

### DIFF
--- a/src/annotator/sidebar.coffee
+++ b/src/annotator/sidebar.coffee
@@ -162,8 +162,9 @@ module.exports = class Sidebar extends Host
     # container.
 
     if @onLayoutChange
-      rect = @frame[0].getBoundingClientRect()
-      computedStyle = window.getComputedStyle(@frame[0])
+      frame = @frame || @externalFrame
+      rect = frame[0].getBoundingClientRect()
+      computedStyle = window.getComputedStyle(frame[0])
       width = parseInt(computedStyle.width)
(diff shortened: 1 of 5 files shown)


JCCR:
I'll soon post an example project that shows the usage of this..Just got the thought of adding this in the live reload page that appears in development

JCCR:
Example project based on React with iframe'd content: https://github.com/JCCR/faux-reader

judell:
This will be extremely helpful, thanks @JCCR.

JCCR:
Rebased with latest master

MohammadHader:
Please notice this issue #744

src/annotator/sidebar.coffee
```diff
@@ -25,7 +25,13 @@ module.exports = class Sidebar extends Host
   gestureState: null
 
   constructor: (element, config) ->
+    if config.theme == 'clean' || config.externalContainerSelector
+      delete config.pluginClasses.BucketBar
```
robertknight: I think it is surprising to have the constructor mutate the `config.pluginClasses` object here. I would have preferred to instead create a modified copy. There is also potentially a problem if anyone were ever to subclass `Sidebar` in future - although I would suggest that would almost certainly be a bad idea anyway.

PR #716, opened 2018-04-12
Resolve relative URLs when getting PDF URL from PDF.js

robertknight:
This PR fixes an issue where the client would fail to show applicable open/restricted groups when Hypothesis is loaded inside a PDF.js viewer, if the PDF file path was specified using a relative URL (eg. `http://publisher.org/viewer.html?file=article1234.pdf`).

In this case, `window.PDFViewerApplication.url` returns the relative `article1234.pdf` URL. This is then returned by `PDFMetadata#getMetadata` in our code, which is then propagated through to the sidebar in the `frames` key of the store. When the client fetches groups via `/api/groups` it does not use `article1234.pdf` for the `document_uri` parameter as it only looks at absolute HTTP/HTTPS URLs.

- The first commit extracts an existing `normalizeURI` helper from `guest.coffee` and adds tests.
- The second commit modifies the `PDFMetadata` class to use this helper to normalize URLs returned from the `PDFMetadata#getUri` and `PDFMetadata#getMetadata` methods.

Fixes https://github.com/hypothesis/product-backlog/issues/579

----

**Testing**

1. Clone our PDF.js test repository: https://github.com/hypothesis/pdf.js-hypothes.is
2. Edit `viewer/web/viewer.html` to load "embed.js" from `http://localhost:5000/embed.js` instead of the production service
3. Save a PDF into the checked out repo (it doesn't matter which one)
4. Save this as test.html:
```html
<html>
  <body>
    <p>
    Test case for <a href="https://github.com/hypothesis/product-backlog/issues/579">product-backlog/#579</a>
    </p>
    <iframe src="viewer/index.html?file=yourpdf.pdf" width="800" height="800"></iframe>
  </body>
</html>
```
5. Run `http-server -c-1` in the test dir (you may need to install it first with `npm install -g http-server`)
6. Create an open group scoped to http://127.0.0.1:8080/ (note that http-server may choose a different port)
7. Navigate to `http://127.0.0.1:8080/test.html` and verify that the PDF shows up and that the client shows the open group created in step (6).

### DIFF
--- a/src/annotator/guest.coffee
+++ b/src/annotator/guest.coffee
@@ -1,4 +1,3 @@
-baseURI = require('document-base-uri')
 extend = require('extend')
 raf = require('raf')
 scrollIntoView = require('scroll-into-view')
@@ -12,6 +11,7 @@ highlighter = require('./highlighter')
 rangeUtil = require('./range-util')
 selections = require('./selections')
 xpathRange = require('./anchoring/range')
+{ normalizeURI } = require('./util/url')
 
 animationPromise = (fn) ->
   return new Promise (resolve, reject) ->
@@ -20,17 +20,6 @@ animationPromise = (fn) ->
         resolve(fn())
       catch error
         reject(error)
-
-# Normalize the URI for an annotation. This makes it absolute and strips
-# the fragment identifier.
-normalizeURI = (uri, baseURI) ->
-  # Convert to absolute URL
-  url = new URL(uri, baseURI)
-  # Remove the fragment identifier.
-  # This is done on the serialized URL rather than modifying `url.hash` due
-  # to a bug in Safari.
-  # See https://github.com/hypothesis/h/issues/3471#issuecomment-226713750
-  return url.toString().replace(/#.*/, '');
 
 module.exports = class Guest extends Delegator
   SHOW_HIGHLIGHTS_CLASS = 'annotator-highlights-always-on'
@@ -140,7 +129,7 @@ module.exports = class Guest extends Delegator
 
     return Promise.all([metadataPromise, uriPromise]).then ([metadata, href]) =>
       return {
-        uri: normalizeURI(href, baseURI),
+        uri: normalizeURI(href),
         metadata,
         frameIdentifier: this.frameIdentifier
       }
--- a/src/annotator/plugin/document.coffee
+++ b/src/annotator/plugin/document.coffee
@@ -1,7 +1,8 @@
 $ = require('jquery')
-Plugin = require('../plugin')
-
 baseURI = require('document-base-uri')
+
+Plugin = require('../plugin')
+{ normalizeURI } = require('../util/url')
 
 ###
 ** Adapted from:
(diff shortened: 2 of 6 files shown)


src/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -81,7 +95,7 @@ describe('pdf-metadata', function () {
         fakePDFViewerApplication.metadata.get = sinon.stub().returns('dcTitle');
 
         return pdfMetadata.getMetadata().then(function (actualMetadata) {
-          assert.deepEqual(expectedMetadata, actualMetadata);
+          assert.deepEqual(actualMetadata, expectedMetadata);
```
robertknight: The arguments were the wrong way round here, producing confusing error messages.

src/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -66,6 +66,20 @@ describe('pdf-metadata', function () {
           assert.equal(uri, 'urn:x-pdf:fakeFingerprint');
         });
       });
+
+      it('resolves relative URLs', () => {
+        var fakePDFViewerApplication = {
+          url: 'index.php?action=download&file_id=wibble',
```
robertknight: This is a simplified version of the scenario in the bug report that Arti filed.

src/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -14,7 +14,7 @@ describe('pdf-metadata', function () {
     window.dispatchEvent(event);
 
     return pdfMetadata.getUri().then(function (uri) {
-      assert.equal(uri, 'http://fake.com');
+      assert.equal(uri, 'http://fake.com/');
```
robertknight: The trailing `/` is being added here because `getUri()` now returns a resolved URL so if `PDFViewerApplication.url` is `http://example.com`, the result will be `http://example.com/`.

src/annotator/plugin/document.coffee
```diff
@@ -185,9 +186,7 @@ module.exports = class Document extends Plugin
 
   # Hack to get a absolute url from a possibly relative one
   _absoluteUrl: (url) ->
-    d = @document.createElement('a')
-    d.href = url
-    d.href
+    normalizeURI(url, @baseURI)
```
robertknight: The `href` property of an `HTMLAnchorElement` returns an absolute URL, resolved against the document's base URL, so it does the same thing as the `normalizeURI` function.

src/annotator/guest.coffee
```diff
@@ -140,7 +129,7 @@ module.exports = class Guest extends Delegator
 
     return Promise.all([metadataPromise, uriPromise]).then ([metadata, href]) =>
       return {
-        uri: normalizeURI(href, baseURI),
+        uri: normalizeURI(href),
```
robertknight: The `baseURI` argument is no longer needed here because `normalizeURI` uses the current document's base URI as its default value for the second argument.

seanh:
Tested following the thorough test instructions and test case (thanks!) - works here

seanh:
Briefly tested that biopub still works too. Code looks good to me - great to see coffeescript being turned in to javascript, docstrings being added, and tests being added!

PR #720, opened 2018-04-16
Use group.links\.html instead of group\.url

sheetaluk:
Because group.url is going to be soon be removed from the group response object, remove it's usage form the client.

Fixes: https://github.com/hypothesis/product-backlog/issues/542
Also see: https://github.com/hypothesis/product-backlog/issues/486

### DIFF
--- a/src/sidebar/templates/annotation-header.html
+++ b/src/sidebar/templates/annotation-header.html
@@ -23,12 +23,12 @@
     <br>
     <span class="annotation-header__share-info">
       <a class="annotation-header__group"
-        target="_blank" ng-if="vm.group() && vm.group().url" href="{{vm.group().url}}">
+        target="_blank" ng-if="vm.group() && vm.group().links.html" href="{{vm.group().links.html}}">
         <i class="h-icon-group"></i><span class="annotation-header__group-name">{{vm.group().name}}</span>
       </a>
       <span ng-show="vm.isPrivate"
         title="This annotation is visible only to you.">
-        <i class="h-icon-lock"></i><span class="annotation-header__group-name" ng-show="!vm.group().url">Only me</span>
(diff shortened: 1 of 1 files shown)


lyzadanger:
Hi @sheetaluk ! Are there any tests for this?

sheetaluk:
@lyzadanger This is only a template level change, we don't have corresponding tests that need adding/changing. I'm guessing the backend changes for adding `links.html` have tests.

src/sidebar/templates/annotation-header.html
```diff
@@ -23,12 +23,12 @@
...
     <span class="annotation-header__share-info">
       <a class="annotation-header__group"
-        target="_blank" ng-if="vm.group() && vm.group().url" href="{{vm.group().url}}">
+        target="_blank" ng-if="vm.group() && vm.group().links.html" href="{{vm.group().links.html}}">
         <i class="h-icon-group"></i><span class="annotation-header__group-name">{{vm.group().name}}</span>
       </a>
       <span ng-show="vm.isPrivate"
         title="This annotation is visible only to you.">
-        <i class="h-icon-lock"></i><span class="annotation-header__group-name" ng-show="!vm.group().url">Only me</span>
+        <i class="h-icon-lock"></i><span class="annotation-header__group-name" ng-show="!vm.group().links.html">Only me</span>
```
lyzadanger: Can you help me understand the logic here? It appears from first glance that the `span.annotation-header__group-name` is "shown" (`ng-show` will unhide the element via CSS) only if there is no link present for the group's activity page. What is `!vm.group().links.html` actually testing for, conceptually?

sheetaluk: So, if I remember right, this was a design decision made for publisher groups - to show the 'only me' text.
image
image

sheetaluk: @ajpeddakotla just confirmed that >> it only appears on 3rd party sites, but I think that this was an eLife requested language change

PR #726, opened 2018-04-20
change links in help panel

klemay:
"send a suppport ticket" now leads to `https://web.hypothes.is/get-help` and "visit our help documents" leads to `https://web.hypothes.is/help`


seanh:
Looks like some `.DS_Store` files got committed by accident, I've removed them

Issue #729, opened 2018-04-30
Problems with accented characters in the link?


SteelWagstaff:
We've been having issues with file names that have accented characters--when it comes to replace the link with the HTML5 audio or video embed element, the routine fails. Let's say we use something like `http://test.com/tèst.mp3`. What I'd expect is for the link to be replaced by an HTML5 audio player, but what actually happens is that the link is just inserted as a link--the embed replace never takes place.

I think I've traced it to this line in the code: https://github.com/hypothesis/client/blob/a5ae4aa0fbf5ede8cb52de5f032dcc892d9a37e4/src/sidebar/media-embedder.js#L304

What's happening is that the link.href has been converted into something that looks like this: `http://test.com/t%C3%A8st.mp3` while the link text remains `http://test.com/tèst.mp3`. These strings don't parse as equal and hence the rest of the function is never called. I think this could be fixed by adding this an additional condition to the if statement in the line cited earlier:
`decodeURI(link.href) !==  link.textContent`. Would you be open to a simple pull request that added this to that line?

robertknight:
Fixed by #738

PR #730, opened 2018-04-30
PR to address Issue #729

SteelWagstaff:
Addresses problem with accented characters in URLs converted to embedded elements discussed in  issue #729: https://github.com/hypothesis/client/issues/729


robertknight:
Hi @SteelWagstaff - Thanks for the PR. There are a couple of things we'll need before I can merge this:

1. When the user pastes the link into the annotation card, it may or may not be percent-encoded depending on where they got it from. We'll need to either normalize both URLs into the same form for comparison or compare the link text against both `link.href` and `decodeURI(link.href)` - I think the latter approach is probably better since neither `encodeURI` or `decodeURI` are idempotent.
2. We'll need tests in `media-embedder-test.js`. This should cover the cases when the text of the link is already percent-encoded and when it is not percent-encoded.

SteelWagstaff:
Would `if (link.href !== link.textContent || decodeURI(link.href) !== link.textContent)` match what you're describing for #1?

src/sidebar/test/media-embedder-test.js
```diff
@@ -297,6 +297,8 @@ describe('media-embedder', function () {
       'https://www.w3schools.com/html/horse.ogg',
       'https://www.w3schools.com/html/horse.ogg#fragment',
       'https://www.w3schools.com/html/horse.ogg?foo=bar&id=31',
+      'https://wisc.pb.unizin.org/frenchcscr/wp-content/uploads/sites/208/2018/03/6Léry_Conclusion.mp3,'
+      'https://wisc.pb.unizin.org%2Ffrenchcscr%2Fwp-content%2Fuploads%2Fsites%2F208%2F2018%2F03%2F6L%25C3%25A9ry_Conclusion.mp3'
```
robertknight: In an encoded URI, slashes in the path won't be encoded. This second test case should be "https://wisc.pb.unizin.org/frenchcscr/wp-content/uploads/sites/208/2018/03/6L%C3%A9ry_Conclusion.mp3"

src/sidebar/media-embedder.js
```diff
@@ -301,7 +301,7 @@ function embedForLink(link) {
  *
  */
 function replaceLinkWithEmbed(link) {
-  if (link.href !== link.textContent) {
+  if (link.href !== link.textContent || decodeURI(link.href) !== link.textContent){
```
robertknight: This `if` statement will be true if `link.href` and `decodeURI(link.href)` are different, which will be true if there are any percent-encoded characters in the URL. I think you meant to use `&&` rather than `||` here to check that the link's text is not equal to either the encoded or unencoded versions of the URL.

SteelWagstaff: I'm sorry--I feel a bit confused. Does that mean I need a third conditional statement--i.e. checking decodeURI(link.href) against decodeURI (link.textContent) as well, or ...

src/sidebar/test/media-embedder-test.js
SteelWagstaff: Thanks! Will fix shortly.

robertknight:
Hi @SteelWagstaff - Sorry to keep you waiting, I've been juggling a bunch of different things internally over the last few days. Please bear with me.

SteelWagstaff:
Not a problem. I appreciate your time and patience with me.

PR #738, opened 2018-06-08
Handle non percent-encoded URLs when replacing links with media embeds

robertknight:
When a user pastes a link to a video or audio file in an annotation,
handle the case where the link contains non-ASCII characters but the
link's text has not been percent-encoded.

In that case the link's text content will be the unencoded form of the
URL, but the `link.href` property will return the percent-encoded form,
so the two do not compare equal and therefore the link was not replaced.


robertknight:
I've been working on this alongside Steel W at the I Annotate 2018 hackday. I'm going to consider it reviewed.

Issue #679, opened 2018-02-21
Annotations are anchoring but showing under orphans


dwhly:
### Steps to reproduce
1. Go here: https://via.hypothes.is/https://www.theguardian.com/technology/2010/jun/20/internet-everything-need-to-know
2. Open sidebar

### Expected behaviour
A bunch of annotations under "Annotations"

### Actual behaviour
A bunch of *anchored* annotations under "Orphans"

### Browser/system information
About this version
Version:
1.71.0
User agent:
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.167 Safari/537.36
URL:
https://www.theguardian.com/technology/2010/jun/20/internet-everything-need-to-know
Username:
dwhly
Date:
21 Feb 2018 10:37:00 -0800

![screen shot 2018-02-21 at 10 34 00 am](https://user-images.githubusercontent.com/1413074/36498467-33e8b440-16f3-11e8-999d-0b0d3cb5523d.png)

klemay:
For what it's worth, I am not able to reproduce:
<img width="1549" alt="screen shot 2018-02-21 at 1 41 15 pm" src="https://user-images.githubusercontent.com/12087705/36498708-1f865862-170d-11e8-9638-a99acd2c611e.png">

robertknight:
I was not able to reproduce either in Via or using the Chrome extension. A possible reason that we might have had different experiences is because we might be seeing different ads or other features on the page and some of those ads might be loading JS which causes anchoring to break.

_If_ someone (perhaps @dwhly) can reproduce the problem with the dev tools open and copy/paste any errors that are reported here that would be useful.

If you cannot reproduce today @dwhly I suggest we close this for now and wait until the issue comes along in a way that can be more widely reproduced.

dwhly:
Works great today, I'll look for another opportunity to break our stuff!

PR #741, opened 2018-06-08
Make client search form accessible for screenreader users

klemay:
Part of accessibility fixes for clients as described in [Product Backlog #679](https://github.com/hypothesis/product-backlog/issues/679)
- The `form` element now contains `role="search"`
- The `text` input has `type="search"` instead of `type="text"`
- The `search` button has an `aria-label="Search"`

### DIFF
--- a/src/sidebar/templates/search-input.html
+++ b/src/sidebar/templates/search-input.html
@@ -1,8 +1,9 @@
 <form class="simple-search-form"
       name="searchForm"
+      role="search"
       ng-class="!vm.query && 'simple-search-inactive'">
   <input class="simple-search-input"
-         type="text"
+         type="search"
          name="query"
          placeholder="{{vm.loading && 'Loading' || 'Search'}}…"
          ng-disabled="vm.loading"
@@ -10,7 +11,7 @@
   <button type="button" class="simple-search-icon top-bar__btn" ng-hide="vm.loading">
     <i class="h-icon-search"></i>
   </button>
-  <button type="button" class="simple-search-icon btn btn-clean" ng-show="vm.loading" disabled>
+  <button type="button" class="simple-search-icon btn btn-clean" aria-label="Search" ng-show="vm.loading" disabled>
     <span class="btn-icon"><span class="spinner"></span></span>
   </button>
 </form>


src/sidebar/templates/search-input.html
```diff
@@ -1,16 +1,17 @@
...
+         type="search"
          name="query"
          placeholder="{{vm.loading && 'Loading' || 'Search'}}…"
          ng-disabled="vm.loading"
          ng-class="vm.inputClasses()"/>
   <button type="button" class="simple-search-icon top-bar__btn" ng-hide="vm.loading">
     <i class="h-icon-search"></i>
   </button>
-  <button type="button" class="simple-search-icon btn btn-clean" ng-show="vm.loading" disabled>
+  <button type="button" class="simple-search-icon btn btn-clean" aria-label="Search" ng-show="vm.loading" disabled>
```
robertknight: This "button" is in fact the loading spinner. It shouldn't be a `<button>` at all. I don't know how best to present a loading indicator with ARIA. Can you look into this?

src/sidebar/templates/top-bar.html
```diff
@@ -49,6 +49,7 @@
     <a class="top-bar__btn"
        ng-click="vm.onSharePage()"
        ng-if="vm.showSharePageButton()"
+       aria-label="Share this page"
```
robertknight: Do we need an `aria-label` _and_ a `title` for this button? What effects does that have for a screen reader?

klemay: `title` provides the tooltip on over, but does not reliably communicate to a screenreader. Without the `aria-label`, a screenreader will announce this simply as "link." 

We don't _need_ both, but since the other elements in this toolbar have `titles` (and therefore tooltips on hover) we probably want to keep both for consistency's sake.

src/sidebar/templates/search-input.html
klemay: Ah, I see that now - so this `aria-labe`l should actually go on the `button` in line 10. 

Though you bring up a good point - I'm not sure how to indicate loading search results to a screenreader. Will look into that as well.

PR #742, opened 2018-06-12
Fix "No results for {query}" message

### LINKED ISSUE #9: Empty string shown in _No results for “”_ when no search results
untitled screencast - edited 24

robertknight:
Usage of `vm.search.query` in the `<sidebar-content>` template was
incorrect. `vm.search` is always set and `query` is a function that
returns a string, not a string.

Fixes #9


robertknight:
Thanks Hannah 🙂

PR #748, opened 2018-07-10
Convert share dialog links to use bouncer instead of via.

sheetaluk:
Fixes https://github.com/hypothesis/product-backlog/issues/335

### DIFF
--- a/src/sidebar/components/share-dialog.js
+++ b/src/sidebar/components/share-dialog.js
@@ -10,7 +10,7 @@ function ShareDialogController($scope, $element, analytics, store) {
       return;
     }
 
-    self.sharePageLink = 'https://hyp.is/go?url='+frames[0].uri;
+    self.sharePageLink = 'https://hyp.is/go?url=' + encodeURIComponent(frames[0].uri);
   }
 
   var shareLinkInput = $element[0].querySelector('.js-share-link');
--- a/src/sidebar/components/share-dialog.js
+++ b/src/sidebar/components/share-dialog.js
@@ -1,31 +1,24 @@
 'use strict';
 
-var VIA_PREFIX = 'https://via.hypothes.is/';
-
 // @ngInject
 function ShareDialogController($scope, $element, analytics, store) {
   var self = this;
 
-  function updateViaLink(frames) {
+  function updateSharePageLink(frames) {
     if (!frames.length) {
-      self.viaPageLink = '';
+      self.sharePageLink = '';
       return;
     }
 
-    // Check to see if we are on a via page. If so, we just return the URI.
-    if (frames[0].uri.indexOf(VIA_PREFIX) === 0) {
-      self.viaPageLink = frames[0].uri;
-    } else {
-      self.viaPageLink = VIA_PREFIX + frames[0].uri;
-    }
+    self.sharePageLink = 'https://hyp.is/go?url='+frames[0].uri;
   }
 
-  var viaInput = $element[0].querySelector('.js-via');
-  viaInput.focus();
-  viaInput.select();
+  var shareLinkInput = $element[0].querySelector('.js-share-link');
+  shareLinkInput.focus();
+  shareLinkInput.select();
 
   $scope.$watch(function () { return store.frames(); },
-    updateViaLink);
+    updateSharePageLink);
 
   $scope.onShareClick = function(target){
     if(target){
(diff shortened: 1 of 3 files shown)


src/sidebar/components/share-dialog.js
```diff
@@ -8,15 +8,15 @@ function ShareDialogController($scope, $element, analytics, store) {
...
       return;
     }
 
     // Check to see if we are on a via page. If so, we just return the URI.
     if (frames[0].uri.indexOf(VIA_PREFIX) === 0) {
-      self.viaPageLink = frames[0].uri;
+      self.sharePageLink = frames[0].uri;
     } else {
-      self.viaPageLink = VIA_PREFIX + frames[0].uri;
+      self.sharePageLink = 'https://hyp.is/go?url='+frames[0].uri;
```
sheetaluk: @robertknight : I don't know if this should be hardcoded this way or if we should derive the `hyp.is` bit from somewhere.

dwhly:
Woop!

src/sidebar/components/share-dialog.js
robertknight: Hardcoding this is an OK first step but the sharing URL prefix/template ought to come from the same place that generates `hyp.is` links for annotations. This is ultimately the `BOUNCER_URL` env var in the `h` service.

I would suggest you could approach this in two steps:

1. Hard-code the production URL template as you are doing here.
2. Add the share URL template to the output of `/api/links` from the "h" service and then use the `serviceUrl` service in the client to generate a URL given the links data from the service and the frame URL provided by the client.

robertknight: Hardcoding the URL is a reasonable first step. I explained in a separate comment how I think we should get the "hyp.is" link from the service.

Since the URL is appearing in a query string param here, it ought to be URL escaped. One way to do this would be to use the existing functions for working with URL templates that get used for handling URL templates returned by `/api/links`:

```diff
diff --git a/src/sidebar/components/share-dialog.js b/src/sidebar/components/share-dialog.js
index 76f1f45e..fa2df4ac 100644
--- a/src/sidebar/components/share-dialog.js
+++ b/src/sidebar/components/share-dialog.js
@@ -1,5 +1,8 @@
 'use strict';
 
+var { replaceURLParams } = require('../util/url-util');
+
+var SHARE_URL_TEMPLATE = 'https://hyp.is/go?url=:url';
 var VIA_PREFIX = 'https://via.hypothes.is/';
 
 // @ngInject
@@ -16,7 +19,8 @@ function ShareDialogController($scope, $element, analytics, store) {
     if (frames[0].uri.indexOf(VIA_PREFIX) === 0) {
       self.sharePageLink = frames[0].uri;
     } else {
-      self.sharePageLink = 'https://hyp.is/go?url='+frames[0].uri;
+      var { url } = replaceURLParams(SHARE_URL_TEMPLATE, { url: frames[0].uri });
+      self.sharePageLink = url;
     }
   }
```

src/sidebar/components/share-dialog.js
```diff
@@ -8,15 +8,15 @@ function ShareDialogController($scope, $element, analytics, store) {
 
   function updateViaLink(frames) {
     if (!frames.length) {
-      self.viaPageLink = '';
+      self.sharePageLink = '';
       return;
     }
 
     // Check to see if we are on a via page. If so, we just return the URI.
     if (frames[0].uri.indexOf(VIA_PREFIX) === 0) {
```
sheetaluk: @dwhly This is where we were using the VIA_PREFIX var. We are checking to see if the user is already on a via page in which case we don't need to create a bouncer link for it.
The environment variable to determine the bouncer prefix i.e `hyp.is` is relevant to the `else` block of this code which will create a bouncer link instead of a via link for the url in question.

dwhly: Yeah, mostly just talking about other places we might still be using VIA_PREFIX... not this one.  I know you're changing it here to use bouncer. Sorry to interrupt. :)

src/sidebar/components/share-dialog.js
robertknight: I get your point. However it turns out that we don't actually need this branch because when on a Via page, the client believes the page's URL is the original URL, not `https://<proxy service>/<original URL>` because of the `<link rel="canonical">` tag in the page.

PR #751, opened 2018-07-25
Hide "Public" group for logged-out users under certain conditions

robertknight:
As discussed in https://github.com/hypothesis/product-backlog/issues/702, hide the "Public" group in the client if:

 - The user is logged out
 - The page has associated groups
 - The user has not visited a direct link to an annotation in the "Public" group

This is implemented by making the `store.allGroups()` selector used by
the group list exclude the "Public" group under the above conditions.

A `directLinkedAnnotationId` field has been added to the selection state
to support this. This field is initialized from the settings passed
through to the sidebar app when it starts.

----

**TODO**

Two problems with the current approach:

- [ ] If the "Public" group is initially hidden, and the user logs in and selects it, then logs out again, the "Public" group remains selected but is no longer visible in the drop-down menu. This is because the `LOAD_GROUPS` action does not reset the focused group in that case.
- [ ] If the "Public" group was focused in a previous session, and if `groups.load()` completes _before_ the profile is fetched on startup, then `groups.load()` will have to make a decision about whether to focus the "Public" group without the guarantee that the current profile has been loaded first. In other words, the user _could_ be logged in (so "Public" should be kept focused), but `profile.userid` might be `null` (because the profile is still being fetched) at the point in `groups.load()` where the focus from the previous session is restored.

Also need to consider whether there could be other issues arising from introducing a difference between the groups that are _loaded_ (exist in the store) and the ones that are _displayed_ in the drop down list.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "redux": "^3.5.2",
     "redux-thunk": "^2.1.0",
     "request": "^2.76.0",
+    "reselect": "^3.0.1",
     "retry": "^0.8.0",
     "scroll-into-view": "^1.8.2",
     "seamless-immutable": "^6.0.1",
--- a/src/sidebar/store/modules/test/annotations-test.js
+++ b/src/sidebar/store/modules/test/annotations-test.js
@@ -122,30 +122,39 @@ describe('annotations reducer', function () {
     }]);
   });
 
-  describe('#directLinkedAnnotation', () => {
+  describe('#directLinkedAnnotationGroup', () => {
     var directLinkedAnn = {
       id: 'abcdef',
+      group: 'the-group-id',
     };
 
     it('returns null if no direct-linked annotation', () => {
       var store = createStore([annotations, selection], [{}]);
-      assert.equal(store.directLinkedAnnotation(), null);
+      assert.equal(store.directLinkedAnnotationGroup(), null);
     });
 
     it('returns null if direct-linked annotation not yet loaded', () => {
       var store = createStore([annotations, selection], [{
         annotations: directLinkedAnn.id,
       }]);
-      assert.equal(store.directLinkedAnnotation(), null);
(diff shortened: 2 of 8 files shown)


dwhly:
Thanks Robert.

> A caveat with the current implementation is that if a user follows a link to a direct-linked annotation in "Public" and then switches to a different group, the "Public" group disappears after the direct-linked annotation gets unloaded. This is because at this point the client no longer knows whether the direct-linked annotation is in the "Public" group or not since it only has its ID.

How difficult is this to work around?  It's probably a pretty big deal given the number of annotations made in the Public channel.

robertknight:
We're exploring an alternate client side implementation at the moment which doesn't have that caveat. I agree it seems important enough to make sure we resolve before shipping this.

package.json
```diff
@@ -89,6 +89,7 @@
     "redux": "^3.5.2",
     "redux-thunk": "^2.1.0",
     "request": "^2.76.0",
+    "reselect": "^3.0.1",
```
robertknight: [reselect](https://github.com/reduxjs/reselect) is a small library for memoization. It is often paired with Redux to optimize selectors so that they only recompute when relevant parts of the application state have changed. Here it is used to optimize `store.allGroups()`.

src/sidebar/store/modules/test/annotations-test.js
```diff
@@ -1,28 +1,16 @@
 'use strict';
 
-var redux = require('redux');
```
robertknight: A while back a `createStore` utility was introduced to create a store from a set of _modules_. It is used by both the actual application and several of the tests. Here I'm changing the tests for the annotations module to use it for consistency.

robertknight:
Closing in favor of https://github.com/hypothesis/client/pull/757.

PR #753, opened 2018-08-01
Use a pointer cursor for highlights

dsiddy:
The `annotator-hl` class contains a `cursor: pointer;` rule already, but
it only applies to mobile devices (more precisely, devices which lack a
sufficiently accurate primary pointer, like a mouse). I simply removed
this restriction so that the rule applies across the board.

Closes: hypothesis/product-backlog#703


robertknight:
Looks good to me. Thanks @dsiddy!

PR #754, opened 2018-08-02
Fix `groups.focused()` error if localStorage is empty when client loads

### LINKED ISSUE #750: TypeError: null is not an object (evaluating 'u.focused().id')
The current version of the client is producing an enormous volume of these errors in production, exhausting our Sentry quota.

https://sentry.io/hypothesis/client/issues/604636862/

```
TypeError: null is not an object (evaluating 'u.focused().id')
  at None (src/sidebar/components/sidebar-content.js:236:12)
...
(2 additional frame(s) were not displayed)
```

robertknight:
If the client started and local storage did not contain a saved
last-viewed group ID, then `groups.load()` would attempt to call
`store.focusGroup(null)`. As a result subsequent calls to
`groups.focused()` would return null and code in various places in the
application expects that there will always be a focused group.

Fix the problem by not setting the focus group after groups are fetched
unless the previously-focused group was set and exists in the new
groups list.

Also improve the behavior of `store.focusGroup` by making it leave the
focused group untouched if the specified group has not been loaded, but
log an error so we'll know (via Sentry) if it happens again.

Fixes #750

----

**Steps to reproduce**

1. Clear local storage for http://localhost:5000
2. Visit any page with the development client active

**Expected (this branch)**: No error, appropriate default group is focused
**Actual (master)**: Error about `groups.focused().id` being null when the client starts

### DIFF
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -62,12 +62,12 @@ function groups($rootScope, store, api, isSidebar, localStorage, serviceUrl, ses
         params.document_uri = uri;
       }
       return api.groups.list(params);
-    }).then(gs => {
+    }).then(groups => {
       var isFirstLoad = store.allGroups().length === 0;
       var prevFocusedGroup = localStorage.getItem(STORAGE_KEY);
 
-      store.loadGroups(gs);
-      if (isFirstLoad) {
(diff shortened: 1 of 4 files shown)


sheetaluk:
LGTM.

PR #761, opened 2018-08-22
Update to Node 10 for Jenkins and Travis builds

robertknight:
The official Docker node builds now have yarn preinstalled, so I switched to use that instead of a community-provided image.


robertknight:
A couple of Travis tests failed with timeouts. I don't think that is related to the Node version. Retrying for now. Will need to address the flakiness at some point.

robertknight:
Travis and Jenkins builds are passing. I'm going to go ahead and merge this.

PR #550, opened 2017-09-21
Send search URIs to other frames by postMessage()

seanh:
Add a cross-origin RPC (remote procedure call) server that allows frames on other origins (e.g. the LTI app running at lti.hypothes.is) to call methods in the Hypothesis sidebar and receive the result. The only method that can be called so far is `searchUris()`, which our LTI app will call in order to get the URIs of the document that the client is annotating when the student clicks the submit assignment button in the LTI app. (A corresponding PR for the LTI app is coming soon.)

Add `cross-origin-rpc.js`, a partial implementation of a [JSON-RPC 2.0 server](http://www.jsonrpc.org/specification). The server only supports receiving JSON-RPC requests and sending back responses, it does not support notifications (JSON-RPC messages that don't require a response), does not support error responses (except "method not found"), and does not support method parameters. All of these features could be added without breaking backwards compatibility when we need them.

[`window.postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) is used as the transport layer over which to send the JSON-RPC messages. `postMessage()` is the DOM's way of allowing frames on different origins to communicate with each other.

Only frames whose origin is listed in the `rpcAllowedOrigins` config setting will be responded to. The plan is that only `lti.hypothes.is` will be listed in this config setting, so any other frames that try to send RPC messages to the client will be ignored. There will be a corresponding pull request to h coming, to add this config setting to h (h will then pass the config setting into the client).

## See also

* [h PR to enable this client feature](https://github.com/hypothesis/h/pull/4670)
* [LTI app PR to make it use this client feature](https://github.com/hypothesis/lti/pull/112)

## How to test this

1. Run a local dev instance of Canvas
1. Checkout out [this h PR](https://github.com/hypothesis/h/pull/4670) and run an h dev env on that branch.
1. Make sure that in your h dev env the `CLIENT_RPC_ALLOWED_ORIGINS` env var is set to the origin of your dev LTI app: `export CLIENT_RPC_ALLOWED_ORIGINS="http://localhost:8001"`
1. Checkout this client branch and run a client dev env on this branch.
1. Run a via dev env (yes you need this)
1. Checkout [this WIP LTI PR](https://github.com/hypothesis/lti/pull/112) and run an lti dev env on that branch.
1. Make sure that the `VIA_URL` and `CLIENT_ORIGIN` env vars are set in your LTI dev env: `export VIA_URL="http://localhost:9080"; export CLIENT_ORIGIN="http://localhost:5000"`.
[...]

PR #764, opened 2018-09-03
Log in using a grantToken received over postMessage

seanh:
If [the `grantToken` config setting](https://h.readthedocs.io/projects/client/en/latest/publishers/config/#cmdoption-arg-granttoken) has the special value `"postMessage"` then send a [`postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) message to [the LMS app](http://github.com/hypothesis/lms) asking it for a grant token for the h user corresponding to the currently logged in LMS user, and wait for a response. On receiving a response `postMessage` from the LMS app containing the grant token, exchange the grant token with h for an access token and log in.

### DIFF
--- a/src/sidebar/services/oauth-auth.js
+++ b/src/sidebar/services/oauth-auth.js
@@ -167,9 +167,27 @@ function auth($http, $rootScope, $window, OAuthClient,
     if (!tokenInfoPromise) {
       var cfg = serviceConfig(settings);
 
+      if (cfg && cfg.grantToken === 'postMessage') {
+        tokenInfoPromise = new Promise(function(resolve, reject) {
+          /** Receive postMessage() response with grant token from LMS app. */
+          function receiveMessage(event) {
+            if (event.origin !== "http://localhost:8001") {
+              return;
+            }
+
+            oauthClient().then(function(client) {
+              client.exchangeGrantToken(event.data).then(function(accessToken) {
+                resolve(accessToken)
+              });
+            });
+          }
+          window.addEventListener("message", receiveMessage, false);
+          window.parent.parent.postMessage("gimme_grant_token", "http://localhost:8001");
+        });
+      }
       // Check if automatic login is being used, indicated by the presence of
       // the 'grantToken' property in the service configuration.
-      if (cfg && typeof cfg.grantToken !== 'undefined') {
+      else if (cfg && typeof cfg.grantToken !== 'undefined') {
         if (cfg.grantToken) {
           // User is logged-in on the publisher's website.
           // Exchange the grant token for a new access token.


src/sidebar/services/oauth-auth.js
```diff
@@ -167,9 +167,27 @@ function auth($http, $rootScope, $window, OAuthClient,
     if (!tokenInfoPromise) {
       var cfg = serviceConfig(settings);
 
+      if (cfg && cfg.grantToken === 'postMessage') {
```
seanh: Here I used a special value of the `grantToken` config setting to trigger the postMessage auth behaviour. In the final version we may want to trigger it some other way :man_shrugging:

src/sidebar/services/oauth-auth.js
```diff
@@ -167,9 +167,27 @@ function auth($http, $rootScope, $window, OAuthClient,
...
+        tokenInfoPromise = new Promise(function(resolve, reject) {
+          /** Receive postMessage() response with grant token from LMS app. */
+          function receiveMessage(event) {
+            if (event.origin !== "http://localhost:8001") {
```
seanh: `"http://localhost:8001"` is the origin of my local dev LMS app. In production this should be `"https://lms.hypothes.is"`. It should be configurable.

src/sidebar/services/oauth-auth.js
```diff
@@ -167,9 +167,27 @@ function auth($http, $rootScope, $window, OAuthClient,
...
+              return;
+            }
+
+            oauthClient().then(function(client) {
+              client.exchangeGrantToken(event.data).then(function(accessToken) {
+                resolve(accessToken)
```
seanh: I've got the LMS app just sending the grant token directly as the `message` argument to `postMessage` (`event.data` on the receiving end here). In a [previous `postMessage()`-related feature](https://github.com/hypothesis/client/pull/550) we used JSON-RPC format messages with `postMessage()` as the transport layer. We should probably do that here too.

src/sidebar/services/oauth-auth.js
```diff
@@ -167,9 +167,27 @@ function auth($http, $rootScope, $window, OAuthClient,
...
+              });
+            });
+          }
+          window.addEventListener("message", receiveMessage, false);
+          window.parent.parent.postMessage("gimme_grant_token", "http://localhost:8001");
```
seanh: I haven't implemented any kind of timeout, if the response postMessage doesn't come. We should add a very generous timeout, and have the client show an error message and report an error to Sentry if it times out.

I _don't_ think we want to fall back on other methods of login or continue booting up if the postMessage method times out, we want to fail the client startup entirely:

* postMessage login should only ever be enabled in the LMS context (and any future contexts that use it)
* When in the LMS context postMessage login should _always_ be enabled and should always succeed. It would only fail if something has gone badly wrong
* When in the LMS context we would never want to use any other type of login, e.g. login to a first-party account, we only and always want to use the LMS login integration

seanh: Oh, and this relies on the fact that in the LMS context `window.parent.parent` is the LMS app's iframe, which is of course brittle and not generalised. This should be replaced by iterating over all the parents and/or all the frames reachable and messaging each one of them, until one responds. [The old LTI app's `postMessage()` feature did this](https://github.com/hypothesis/lti/blob/1f8896ee045018e2ac57945f9f1d396f8ab28325/lti/templates/includes/submission_form.html.jinja2#L23-L37) in the opposite direction (the LTI app finding the client's iframe, rather than vice-versa).

PR #766, opened 2018-09-13
Replace "var" with "const" or "let" everywhere

robertknight:
This PR is the result of using jscodeshift [1] to apply the "no-vars" transform from js-codemod [2] on all code in src/ and scripts/ with the exception of code in vendor/ directories. There were 13 places that could not be automatically converted and had to be fixed up manually. See commits for details.

It also enables the ESLint [no-var](https://eslint.org/docs/rules/no-var) rule to ensure consistency going forwards.
    
[1] https://github.com/facebook/jscodeshift
[2] https://github.com/cpojer/js-codemod


robertknight:
Got +1s for this on Slack. See https://hypothes-is.slack.com/archives/C1M8NH76X/p1536840446000100

PR #768, opened 2018-09-14
Read sidebar app config asynchronously on startup

robertknight:
In preparation for introducing an optional new postMessage-based method
for the sidebar app to fetch configuration from its embedder, make the
setting fetch flow async.

- Move the Angular app startup logic into a function which receives
  settings as an argument.
- Add a function to merge config rendered into the sidebar app HTML
  page with config fetched asynchronously from a host frame.

When reviewing, the diff-ignoring-whitespace should be easier to follow.

### DIFF
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -10,19 +10,14 @@ require('../shared/polyfills');
 let raven;
 
 // Read settings rendered into sidebar app HTML by service/extension.
-const settings = require('../shared/settings').jsonConfigsFrom(document);
+const appConfig = require('../shared/settings').jsonConfigsFrom(document);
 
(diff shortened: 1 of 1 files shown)


src/sidebar/index.js
```diff
@@ -10,20 +10,15 @@ require('../shared/polyfills');
 var raven;
 
 // Read settings rendered into sidebar app HTML by service/extension.
-var settings = require('../shared/settings').jsonConfigsFrom(document);
+var appConfig = require('../shared/settings').jsonConfigsFrom(document);
```
robertknight: I renamed this to clarify that it refers to the config rendered into the sidebar app's HTML page.

PR #771, opened 2018-09-14
Fetch config via postMessage RPC if `requestConfigFromFrame` is set

robertknight:
Implement support for fetching configuration for the sidebar app from an ancestor frame with a specified origin if the `requestConfigFromFrame` option is set.

This enables the configuration fetching described in https://github.com/hypothesis/client/pull/767.

- The first commit adds a module that handles fetching config from the embedding frame(s) and merging it with the config rendered into app.html. This makes a JSON-RPC-over-postMessage request to each ancestor frame if `requestConfigFromFrame` is set.
- The second commit uses that module to fetch config on app startup.

The error handling in this PR is very basic. If fetching configuration fails, an error is reported in the browser console and the sidebar will not appear at all. Presenting a visual error requires more work because the sidebar currently doesn't get displayed at all if the Angular app fails to start.

----

You can test this PR manually using the demo from [this gist](https://gist.github.com/robertknight/7c758a89c5b018b061dd1b040a7f3a4e).

### DIFF
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -2,7 +2,7 @@
 
 const addAnalytics = require('./ga');
 const disableOpenerForExternalLinks = require('./util/disable-opener-for-external-links');
-const getApiUrl = require('./get-api-url');
+const { fetchConfig } = require('./util/fetch-config');
 const serviceConfig = require('./service-config');
 const crossOriginRPC = require('./cross-origin-rpc.js');
 require('../shared/polyfills');
@@ -111,19 +111,6 @@ function configureCompile($compileProvider) {
 // @ngInject
 function setupHttp($http, streamer) {
   $http.defaults.headers.common['X-Client-Id'] = streamer.clientId;
-}
-
-/**
- * Merge settings rendered into sidebar app HTML page with
- * settings provided by the embedder of Hypothesis.
- */
-function fetchConfig(appConfig) {
-  const hostPageConfig = require('./host-config');
-  const mergedConfig = Object.assign({}, appConfig, hostPageConfig(window));
-
-  mergedConfig.apiUrl = getApiUrl(mergedConfig);
-
-  return Promise.resolve(mergedConfig);
 }
 
 function startAngularApp(config) {
@@ -257,5 +244,8 @@ function startAngularApp(config) {
 fetchConfig(appConfig).then(config => {
   startAngularApp(config);
 }).catch(err => {
-  console.error(err);
+  // Report error. This will be the only notice that the user gets because the
+  // sidebar does not currently appear at all if the Angular app fails to
+  // start.
+  console.error('Failed to start Hypothesis client: ', err);
 });
--- a/src/sidebar/util/fetch-config.js
+++ b/src/sidebar/util/fetch-config.js
@@ -1,0 +1,79 @@
+'use strict';
+
+const getApiUrl = require('../get-api-url');
+const hostConfig = require('../host-config');
+const postMessageJsonRpc = require('./postmessage-json-rpc');
+
+function ancestors(window_) {
+  if (window_ === window_.top) {
+    return [];
+  }
+
+  // nb. The top window's `parent` is itself!
+  const ancestors = [];
+  do {
+    window_ = window_.parent;
+    ancestors.push(window_);
+  } while (window_ !== window_.top);
+
+  return ancestors;
+}
+
+/**
+ * Fetch client configuration from an ancestor frame.
+ *
+ * @param {string} origin - The origin of the frame to fetch config from.
+ * @param {Window} window_ - Test seam.
+ * @return {Promise<any>}
+ */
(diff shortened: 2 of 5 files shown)


seanh:
I've tested this in Canvas with modified versions of my proof-of-concept Via and LMS PRs -- the login works!

src/sidebar/index.js
```diff
@@ -257,5 +244,8 @@ function startAngularApp(config) {
 fetchConfig(appConfig).then(config => {
   startAngularApp(config);
 }).catch(err => {
-  console.error(err);
+  // Report error. This will be the only notice that the user gets because the
+  // sidebar does not currently appear at all if the Angular app fails to
+  // start.
+  console.error('Failed to start Hypothesis client: ', err);
```
seanh: I think we may want a user visible error (not just in the dev console) at some point. And I think we definitely want to log a crash report somewhere when this happens -- this should never be happening in production, and we need to know if it is happening, even if it's only happening a few times. Of course Sentry is currently disabled for the client. I think this can come in follow-on PRs, but I think it's a pretty desirable thing to have, if it's not working we want to know about it.

src/sidebar/util/fetch-config.js
```diff
@@ -0,0 +1,79 @@
...
+
+  // nb. The top window's `parent` is itself!
+  const ancestors = [];
+  do {
+    window_ = window_.parent;
+    ancestors.push(window_);
+  } while (window_ !== window_.top);
+
+  return ancestors;
+}
```
seanh: So this is our way of collecting all the higher up windows / frames. Cool!

src/sidebar/util/fetch-config.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Fetch client configuration from an ancestor frame.
+ *
+ * @param {string} origin - The origin of the frame to fetch config from.
+ * @param {Window} window_ - Test seam.
+ * @return {Promise<any>}
+ */
+function fetchConfigFromAncestorFrame(origin, window_=window) {
```
seanh: Great function name

src/sidebar/util/fetch-config.js
```diff
@@ -0,0 +1,79 @@
...
+  const configResponses = [];
+
+  for (let ancestor of ancestors(window_)) {
+    const timeout = 3000;
+    const result = postMessageJsonRpc.call(
+      ancestor, origin, 'requestConfig', timeout
+    );
+    configResponses.push(result);
+  }
```
seanh: Ok, so we collect all the responses from all the ancestor frames that we messaged (and some of the responses may be rejected or timed out promises)

src/sidebar/util/fetch-config.js
```diff
@@ -0,0 +1,79 @@
...
+
+  if (configResponses.length === 0) {
+    configResponses.push(Promise.reject(new Error('Client is top frame')));
+  }
+
+  return Promise.race(configResponses);
```
seanh: So whoever resolves first wins?

src/sidebar/util/fetch-config.js
```diff
@@ -0,0 +1,79 @@
...
+function fetchConfig(appConfig, window_=window) {
+  const hostPageConfig = hostConfig(window_);
+
+  let embedderConfig;
+  if (hostPageConfig.requestConfigFromFrame) {
+    const origin = hostPageConfig.requestConfigFromFrame;
+    embedderConfig = fetchConfigFromAncestorFrame(origin, window_);
+  } else {
+    embedderConfig = Promise.resolve(hostPageConfig);
+  }
```
seanh: Ok, so we'll only read the config from the host page if getting the config over postMessage is disabled (by the config setting). Seems right

src/sidebar/util/fetch-config.js
```diff
@@ -0,0 +1,79 @@
...
+
+  return embedderConfig.then(embedderConfig => {
+    const mergedConfig = Object.assign({}, appConfig, embedderConfig);
+    mergedConfig.apiUrl = getApiUrl(mergedConfig);
+    return mergedConfig;
+  });
+}
```
seanh: Nice and clear :+1:

src/sidebar/util/fetch-config.js
robertknight: Resolves _or_ rejects first. Typically only one of the frames will respond with success and the others will _eventually_ timeout, but the `race` Promise will already have resolved by then.

robertknight:
> I'm noticing that a lot of modules seem to be getting added to the util folder.

We may want to subdivide that in future. An important note about layering is that code in `util/` has no dependencies on other top-level packages (`services/`, `components/`).

src/sidebar/index.js
robertknight: Raven should send `console.error` reports AFAIK - but first we have to fix Sentry for the client as you noted.

PR #773, opened 2018-09-14
Add aria label to share button in client menu

dwolfe:
In addition to adding the label, this changes the <a> element to a <button> as well.

Fixes https://github.com/hypothesis/product-backlog/issues/780

### DIFF
--- a/src/sidebar/templates/top-bar.html
+++ b/src/sidebar/templates/top-bar.html
@@ -46,12 +46,13 @@
       sort-key="vm.sortKey"
       on-change-sort-key="vm.onChangeSortKey({sortKey: sortKey})">
     </sort-dropdown>
-    <a class="top-bar__btn"
+    <button class="top-bar__btn"
        ng-click="vm.onSharePage()"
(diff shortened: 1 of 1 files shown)


dwolfe:
Tested in Firefox, Safari, Chrome, and Canary - works as expected in all.

dwolfe:
Actually, for a button, apparently the most accessible thing to do is use text inside the `<button>` element, so this may change again as we evolve our a11y standards. I'm still learning about best practices and which apply to us, but, to answer your second question, several tools do treat them differently, from what I've found. This article explains a tiny part of the landscape: https://www.deque.com/blog/text-links-practices-screen-readers/

robertknight:
Thanks for the info @dwolfe. On a related note we have very basic [set of Accessibility guidelines](https://github.com/hypothesis/frontend-toolkit/blob/master/docs/accessibility.md) that was started a while back as a way to share knowledge and provide a checklist for people making frontend changes. It would be good to revise and add to this as more is learned.

PR #780, opened 2018-09-29
Use custom name for global `require` function

### LINKED ISSUE #779: Major clash on websites with a global AMD module loader
Several (or possibly all) of your bundles (e.g. `polyfill.bundle.js`, `jquery.bundle.js` or `unorm.bundle.js`) define global `require()` functions for module loading. On websites which use AMD module loaders themselves, this leads to unresolved modules and exceptions, when the host code calls require after your bundles have been loaded and overwritten this function.

This can happen regularly when hypothes.is is loaded early via a browser extension. But even when it's loaded later, this may happen, e.g. when host code is executed as a result of a user action and requires a module.

You can observe this issue on Moodle, if you activate the browser extension and then navigate to the [calendar on the Moodle demo site](https://school.demo.moodle.net/calendar/view.php) and log in as guest.

There's little doubt that a browser extension should avoid defining global variables at all, or only define them within a unique namespace. Wrapping each of your bundles with an IIFE should resolve this issue, since Browserify never defines `require`  explicitly on window but just as `require = `. So wrapping it should do the trick (see [here](https://github.com/browserify/browserify/issues/403)).

robertknight:
Browserify bundles that make their modules available for other bundles to consume define a global `require` function. This is then imported by other bundles and used to resolve module lookups.

This `require` name can clash with web pages that use eg. RequireJS.
Most of the time this isn't a problem because all of the page's own code
has run by the time Hypothesis loads. If however the page loads
additional scripts after Hypothesis has been loaded, they will end up
finding Hypothesis's `require` instead of their own.

This commit resolves the issue by renaming the global `require` function
to `hypothesisRequire` in two steps:

 - The `externalRequireName` Browserify option changes the _exported_
   name.
 - A custom transform stream post-processes the Browserify bundle code
   to change the _imported_ name, since Browserify doesn't have an
   option for this itself.

A small fix was needed to the `boot.js` script build to avoid stripping
lines _after_ the sourcemap URL comment, instead of just that comment.

Fixes #779

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -287,7 +287,7 @@ function generateBootScript(manifest) {
     .pipe(replace('__SIDEBAR_APP_URL__', defaultSidebarAppUrl))
     // Strip sourcemap link. It will have been invalidated by the previous
     // replacements and the bundle is so small that it isn't really valuable.
-    .pipe(replace(/^\/\/# sourceMappingURL=[^ ]+$/m,''))
+    .pipe(replace(/^\/\/# sourceMappingURL=\S+$/m,''))
     .pipe(rename('boot.js'))
     .pipe(gulp.dest('build/'));
 }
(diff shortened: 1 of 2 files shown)


renehamburger:
Could this go live some time soon?

renehamburger:
Thanks for resolving this so quickly! I'm impressed. 😃

PR #789, opened 2018-10-17
Wrap "stage" contents in a block in Jenkinsfile

robertknight:
This fixes a warning from Jenkins that using `stage` without specifying
a block is deprecated.


robertknight:
This is a trivial build change so I'm going to go ahead and merge it.

PR #791, opened 2018-10-19
Skip builds of automated commits on Jenkins

robertknight:
As part of the deployment process Jenkins creates a new commit on master
in order to bump the npm package version.

There is no point in creating an automated build of these commits, which
can be identified by the commiter name.


robertknight:
I'm going to go ahead and merge this so we can verify it works the next time we do a release from master.

PR #794, opened 2018-10-23
Add missing polyfill for IE 11 compatibility when annotating PDFs

### LINKED ISSUE #762: JS error for annotating PDFs in IE 11
This comes from [one of our partners](https://hypothesis.zendesk.com/agent/tickets/2903), who is in the process of integrating Hypothesis.

### Steps to reproduce
1. In IE 11, attempt to annotate a PDF (examples can be found in the ticket or [here in Slack](https://hypothes-is.slack.com/archives/C03QZM0K4/p1535567961000100?thread_ts=1535567912.000100&cid=C03QZM0K4). This is a partner's staging environment so I don't want to make the urls public)
2. Highlight text and click "Annotate"

### Expected behaviour
Sidebar opens up, you're able to add an annotation

### Actual behaviour
Nothing happens. Browser throws a console error:
`SCRIPT65535: Argument not optional`

### Additional details
As [discussed in Slack](https://hypothes-is.slack.com/archives/C03QZM0K4/p1535567912000100), @robertknight noticed:

- We’re calling `document.createNodeIterator` and omitting some arguments that are supposed to be optional but are not in IE 11 (cf. https://github.com/hypothesis/client/blob/b7f4a83622550072620036677856af4744298eb6/src/annotator/range-util.js#L51)
- This line is probably buggy: https://github.com/hypothesis/client/blob/b7f4a83622550072620036677856af4744298eb6/src/annotator/anchoring/pdf.coffee#L288

robertknight:
The `seek` function from the `dom-seek` package expects its
`NodeIterator` argument to have a `referenceNode` property, which is
missing in IE 11.

Use the dom-node-iterator package recommended by the dom-seek README [1] to
polyfill optional-ness of arguments to `document.createNodeIterator` and
the missing properties on the returned `NodeIterator` object.

dom-node-iterator provides several ways to use the polyfill. This commit
uses the method that prefers the browser's native implementation if possible
and does not pollute the global scope, which could affect JS code
outside of the Hypothesis client.

There is one other place where we use `createNodeIterator` but in that
context the missing properties are not used so the polyfill is not
needed.

Fixes #762

[1] https://github.com/tilgovi/dom-seek/blob/master/README.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "dom-anchor-fragment": "^1.0.1",
     "dom-anchor-text-position": "^4.0.0",
     "dom-anchor-text-quote": "^4.0.2",
+    "dom-node-iterator": "^3.5.3",
     "dom-seek": "^4.0.3",
     "end-of-stream": "^1.1.0",
     "escape-html": "^1.0.3",
--- a/src/annotator/anchoring/pdf.coffee
+++ b/src/annotator/anchoring/pdf.coffee
@@ -1,5 +1,8 @@
 seek = require('dom-seek')
 
+# `dom-node-iterator` polyfills optional arguments of `createNodeIterator`
+# and properties of the returned `NodeIterator` for IE 11 compatibility.
+createNodeIterator = require('dom-node-iterator/polyfill')()
 xpathRange = require('./range')
 
 html = require('./html')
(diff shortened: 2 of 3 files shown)


sheetaluk:
I don't yet have IE11 running locally, but the code changes look fine. LGTM.

PR #798, opened 2018-10-25
Deploy QA release of client as part of master branch builds

robertknight:
This PR enables automatic publication of a pre-release version of the client to S3 as part of every build of the "master" branch. The current pre-release version's entry point is available at https://cdn.hypothes.is/hypothes.is@qa and has a version number of the form `X.Y.Z-{short git hash}`. This URL is in turn used by the qa deployment of h by setting the `CLIENT_URL` env var in the Elastic Beanstalk config. The result is that https://qa.hypothes.is/embed.js redirects to https://cdn.hypothes.is/hypothes.is@qa and hence anywhere that embeds the client from qa-h will load the qa client. This enables us to test the two QA parts of Hypothesis together.

1. The first commit adds a script to deploy to S3, which replaces the [s3-npm-publish](https://github.com/nickstenning/s3-npm-publish) tool that we had been using before. This script adds several features needed for the QA deploy:
   - Deploy directly to S3 from the current working directory, skipping the need for an npm deployment which is not currently needed for QA.
   - Make it easier to change the caching configuration, since for QA we'll want minimal or no caching of the entry point script.
   - Support deploying to named tags, so we can have separate deployments of the client at different stable URLs (https://cdn.hypothes.is/hypothesis is prod, https://cdn.hypothes.is/hypothesis@qa is qa)
2. The second commit uses this script in a new Jenkins build stage which creates a temporary pre-release version and deploys it to S3. After that deployment, the version change is reverted and deployment of the production build proceeds as before.
3. The third commit prevents caching of the entry point (https://cdn.hypothes.is/hypothesis@qa) for qa deployments. The entry point for prod deploys is still cached by the browser for 30 mins (1800s) as before.

In order to enable the Docker container which builds the client to deploy to S3, I created a new user (`jenkins-s3-cdn`) in AWS which has permissions to deploy to the cdn.hypothes.is bucket, but nothing else. Access keys for this user have been configured in the Jenkins credentials store.

This is part of https://github.com/hypothesis/product-backlog/issues/758


seanh:
> since for QA we'll want minimal or no caching of the entry point script

I'd suggest no caching at all? I don't think we _need_ caching for QA (it doesn't need to be fast, and isn't getting a lot of traffic) and it just seems likely to result in people accidentally testing the wrong version and not realising.

seanh:
> The entry point for prod deploys is still cached by the browser for 30 mins (1800s) as before.

So I guess when people are testing prod (after having tested on qa and just deployed to prod) we need to make sure they remember to disable their browser cache

robertknight:
> I'd suggest no caching at all? I don't think we need caching for QA (it doesn't need to be fast, and isn't getting a lot of traffic) and it just seems likely to result in people accidentally testing the wrong version and not realising.

I've actually disabled caching entirely for the QA deployment's entry point this morning (see last commit), since I found out that 30 minutes is the minimum cache duration that CloudFlare allows, when caching is enabled for a resource, on our current plan.

robertknight:
> So I guess when people are testing prod (after having tested on qa and just deployed to prod) we need to make sure they remember to disable their browser cache

Yes. We might revisit caching for the prod build, but I decided not to change that in this PR.

robertknight:
Thanks!

PR #799, opened 2018-10-26
Add service.groups setting to filter groups

hmstepanek:
service.groups config option is used to specify a subset of groups to be displayed in the group drop down list on the sidebar.

This is a fix for https://github.com/hypothesis/product-backlog/issues/852.

### DIFF
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -48,6 +48,13 @@ function groups($rootScope, store, api, isSidebar, localStorage, serviceUrl, ses
    * @return {Promise<Group[]>}
    */
   function filterGroups(groups, isLoggedIn, directLinkedAnnotationId) {
+    // If service groups are specified only return those.
+    // If a service group doesn't exist in the list of groups don't return it.
+    if (svc && svc.groups) {
+      const focusedGroups = groups.filter(g => svc.groups.includes(g.id));
+      return Promise.resolve(focusedGroups);
+    }
+
     // Logged-in users always see the "Public" group.
     if (isLoggedIn) {
       return Promise.resolve(groups);
--- a/src/sidebar/services/test/groups-test.js
+++ b/src/sidebar/services/test/groups-test.js
@@ -253,6 +253,46 @@ describe('groups', function() {
         });
       });
     });
+
+    [{
+      description: 'shows service groups',
+      services: [{ groups: ['abc123']}],
+      expected: ['abc123'],
+    },{
+      description: 'only shows service groups that exist',
+      services: [{ groups: ['abc123', 'no_exist']}],
+      expected: ['abc123'],
+    },{
+      description: 'shows no groups if no service groups exist',
+      services: [{ groups: ['no_exist']}],
+      expected: [],
+    },{
+      description: 'shows all groups if service is null',
+      services: null,
+      expected: ['__world__', 'abc123'],
+    },{
+      description: 'shows all groups if service groups does not exist',
+      services: [{}],
+      expected: ['__world__', 'abc123'],
+    }].forEach(({ description, services, expected }) => {
+      it(description, () => {
+        fakeSettings.services = services;
+        const svc = service();
+
+        // Create groups response from server.
+        const groups = [{ name: 'Public', id: '__world__' }, { name: 'ABC', id: 'abc123'}];
+
+        fakeApi.groups.list.returns(Promise.resolve({
+          token: '1234',
+          data: groups,
+        }));
+
+        return svc.load().then(groups => {
+          let displayedGroups = groups.map(g => g.id);
+          assert.deepEqual(displayedGroups, expected);
+        });
+      });
+    });
   });
 
   describe('#get', function() {
--- a/docs/publishers/config.rst
+++ b/docs/publishers/config.rst
@@ -186,6 +186,15 @@ loads.
       group. The image should be suitable for display at 16x16px and the
       recommended format is SVG.
 
+   .. option:: groups
+
+      ``String[]|null``. An array of group IDs. If provided, the list of groups
+      fetched from the API will be filtered against this list so that the user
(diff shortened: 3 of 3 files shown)


src/sidebar/services/test/groups-test.js
```diff
@@ -253,6 +253,93 @@ describe('groups', function() {
...
+      const groups = [{ name: 'Public', id: '__world__' }, { name: 'ABC', id: 'abc123'}];
+
+      fakeApi.groups.list.returns(Promise.resolve({
+        token: '1234',
+        data: groups,
+      }));
+
+      return svc.load().then(groups => {
+        const onlyGroupABCShown = groups.every(g => g.id === 'abc123');
+        assert.equal(onlyGroupABCShown, true);
```
robertknight: I would recommend using `assert.deepEqual(actualArray, expectedArray)` to compare arrays. You'll get more information about what went wrong if the test fails that way. If order is not important, you can sort the arrays beforehand.

src/sidebar/services/test/groups-test.js
```diff
@@ -253,6 +253,93 @@ describe('groups', function() {
...
+      });
+    });
+
+    it('shows no groups when none of the service groups exist', () => {
```
robertknight: This test could be implemented by parametrizing the above test. In JS with mocha that typically looks like this:

```js
[{
  paramA: valueA,
  paramB: valueB,
},{
  paramA: valueC,
  paramB: valueD,
}].forEach(({ paramA, paramB }) => {
   it('does something', () => {
     // Test which uses values of `paramA` and `paramB`
   });
});

File: src/sidebar/services/test/groups-test.js [hunk 3]
```diff
@@ -253,6 +253,93 @@ describe('groups', function() {
         });
       });
     });
+
+    it('shows service groups when logged in', () => {
```
robertknight: In the LMS context the user will always be logged in, so we don't have to worry too much about what happens if `services` is present and non-empty but the user is not logged in (ie. `services[0]` has no `grantToken` entry). However I would suggest we always respect the `groups` list if `services` is present & non-empty for consistency.

src/sidebar/services/groups.js
```diff
@@ -50,6 +50,12 @@ function groups($rootScope, store, api, isSidebar, localStorage, serviceUrl, ses
   function filterGroups(groups, isLoggedIn, directLinkedAnnotationId) {
     // Logged-in users always see the "Public" group.
     if (isLoggedIn) {
+
```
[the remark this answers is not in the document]
robertknight: It doesn't actually matter for the LMS use case but I think it would be more consistent to respect the `groups` property regardless of whether the user is logged-in or not. Since `services[index].groups` only applies when third-party accounts are being used, the "Public" group will not be in the returned list.

src/sidebar/services/test/groups-test.js
```diff
@@ -253,6 +253,46 @@ describe('groups', function() {
...
+        const groups = [{ name: 'Public', id: '__world__' }, { name: 'ABC', id: 'abc123'}];
+
+        fakeApi.groups.list.returns(Promise.resolve({
+          token: '1234',
+          data: groups,
+        }));
+
+        return svc.load().then(groups => {
+          let displayedGroups = groups.map(g => g.id);
+          assert.deepEqual(displayedGroups, expected);
```
hmstepanek: The addition of the default org being added kinda makes this comparison a bit tricky which is why I'm only comparing the ids as opposed to the entire array of objects.

src/sidebar/services/test/groups-test.js
hmstepanek: I just made them into one big test but added a description for each.

seanh:
I'll let @robertknight finish reviewing this but just FYI [the LMS app PR that this works with](https://github.com/hypothesis/lms/pull/303) has been deployed. I tested this client branch against that LMS app branch in Canvas and it seems to work great :+1: So as soon as we're happy with the code here we can deploy it, and then the show-only-one-group feature is shipped (with the don't-show-a-pointless-menu thing as a separate piece of follow-up work).

seanh:
Cool to see how small the required client code change was

src/sidebar/services/test/groups-test.js
robertknight: The approach you've taken here is exactly what I would have done, since we only really care about _which_ groups are shown, not the contents.

src/sidebar/services/test/groups-test.js
```diff
@@ -253,6 +253,46 @@ describe('groups', function() {
...
+      expected: [],
+    },{
+      description: 'shows all groups if service is null',
+      services: null,
+      expected: ['__world__', 'abc123'],
+    },{
+      description: 'shows all groups if service groups does not exist',
+      services: [{}],
+      expected: ['__world__', 'abc123'],
+    }].forEach(({ description, services, expected }) => {
```
robertknight: This is exactly what I was looking for 👍

src/sidebar/services/test/groups-test.js
```diff
@@ -253,6 +253,46 @@ describe('groups', function() {
...
+        // Create groups response from server.
+        const groups = [{ name: 'Public', id: '__world__' }, { name: 'ABC', id: 'abc123'}];
+
+        fakeApi.groups.list.returns(Promise.resolve({
+          token: '1234',
+          data: groups,
+        }));
+
+        return svc.load().then(groups => {
+          let displayedGroups = groups.map(g => g.id);
```
robertknight: nit. `displayedGroups` can be `const displayedGroups` here because it is never re-assigned.

robertknight:
I squashed the commits in this PR and also added a commit which documents the additional option in the public docs that we provide for publishers at https://h-client.readthedocs.io/en/latest/publishers/config/. CC @hmstepanek

PR #801, opened 2018-11-02
Set a useful title on the sidebar iframe

robertknight:
This makes it easier to identify the purpose of the frame when
navigating the page with assistive technology.

I noticed this when testing the behaviour of the spinner PR with VoiceOver enabled.

Fixes https://github.com/hypothesis/product-backlog/issues/784


robertknight:
I'm going to merge this as part of demoing client QA deployments today.

Issue #804, opened 2018-11-05
Finish CoffeeScript conversion and / or removal in the sidebar application


robertknight:
Pay down some technical debt by finishing the conversion / replacement of CoffeeScript code to JS, so that we benefit from being able to apply our JS-focused tooling to it and developers don't need to work in two different languages.

**Remaining modules**

```shellsession
$ find src/sidebar src/shared -name '*.coffee'
src/sidebar/services/test/view-filter-test.coffee
src/sidebar/services/stream-filter.coffee
src/sidebar/services/query-parser.coffee
src/shared/test/discovery-test.coffee
src/shared/discovery.coffee
```

https://github.com/hypothesis/client/pull/802 and https://github.com/hypothesis/client/pull/803 cover about half of this.

**Done when**

- [x] The remainder of the CoffeeScript code used by the sidebar application (in `src/shared` and `src/sidebar`) is removed or replaced by JS
- [x] CoffeeScript support is disabled when building the JS bundles for the sidebar application

PR #805, opened 2018-11-05
Convert discovery.coffee to JS

robertknight:
_This is part of https://github.com/hypothesis/client/issues/804._

This PR converts `src/shared/discovery.coffee` to JS and improves the documentation. This module is responsible for discovering other frames in the same tab that run the Hypothesis client's annotation layer or sidebar app code.

This is a mostly straight, manual conversion of the code, except that I rewrote the comments to provide more clarity and context as to how the module fits into the app. The tests are still in CoffeeScript. I will convert those in a separate PR.

I note that this module does have some problems, as explained in referenced GitHub issues, and we might want to solve the problem using a different method in future.


src/shared/discovery.js
```diff
@@ -0,0 +1,204 @@
...
+   * Send a message to other frames in the current window to inform them about
+   * the existence of this frame and tell them whether this frame is a client
+   * or server.
+   */
+  _beacon() {
+    let beaconMessage;
+    if (this.server) {
+      beaconMessage = '__cross_frame_dhcp_offer';
+    } else {
+      beaconMessage = '__cross_frame_dhcp_discovery';
```
robertknight: These messages ought to be changed to something more obviously connected to Hypothesis. I left them as-is in this PR while I have a think about how the whole discovery process ought to work.

src/shared/discovery.js
```diff
@@ -0,0 +1,204 @@
...
+    // If `origin` is 'null' the source frame is a file URL or loaded over some
+    // other scheme for which the `origin` is undefined. In this case, the only
+    // way to ensure the message arrives is to use the wildcard origin. See:
+    //
+    //   https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
+    //
+    // When sending messages to or from a Firefox WebExtension, current
+    // versions of Firefox have a bug that causes the origin check to fail even
+    // though the target and actual origins of the message match.
+    if (origin === 'null' || origin.match('moz-extension:') ||
```
robertknight: nb. That `'null'` is a string here is not a typo.

PR #807, opened 2018-11-07
Convert viewFilter tests to JS

robertknight:
Convert the last remaining CoffeeScript test module to JS.

I have mostly done a straight conversion of the existing tests, except that I removed the tests for the public `fields` property of the class and made that property a private variable, since it was an implementation detail unused by actual app code.

I also added some documentation to explain what the `viewFilter` and `searchFilter` classes are used for in the context of the sidebar application, since this is not particularly obvious.


dwhly:
repealcelebbarprohibition

PR #808, opened 2018-11-07
Add `make docs` and `make checkdocs`

seanh:
Deleted the "docs for how to build the docs" since this is now just a simple `make docs`, as for all our projects.


Makefile
```diff
@@ -21,7 +21,11 @@ lint: node_modules/.uptodate
 
 .PHONY: docs
 docs:
-	cd docs && make livehtml
+	tox -e py3-docs
```
seanh: For the client I'm using `py3` instead of `py36` since it's only ever used for building docs.

PR #809, opened 2018-11-08
Disable groups drop down

hmstepanek:
Disable groups drop down when a thirdparty and only one group.
This is a fix for https://github.com/hypothesis/product-backlog/issues/868.


src/sidebar/templates/group-list.html
```diff
@@ -1,22 +1,26 @@
 <div class="pull-right"
      dropdown
      keyboard-nav>
+  <!-- Show a drop down menu if showGroupsMenu is true. -->
```
hmstepanek: This is WIP-no sense in reviewing this too deeply right now.

src/sidebar/templates/group-list.html
```diff
@@ -1,22 +1,26 @@
...
     <img class="group-list-label__icon group-list-label__icon--organization"
          ng-src="{{ vm.focusedIcon() }}"
          alt="{{ vm.orgName(vm.groups.focused().id)}}"
          ng-if="vm.focusedIcon()">
          <i class="group-list-label__icon h-icon-{{ vm.focusedIconClass() }}"
              ng-if="!vm.focusedIcon()"></i><!-- nospace
-    !--><span class="group-list-label__label">{{vm.groups.focused().name}}</span><!-- nospace
-      !--><i class="h-icon-arrow-drop-down"></i>
+    !-->
+    <h2 class="group-list-label__label has-toggle" ng-if="vm.showGroupsMenu()"><a class="group-list-label__toggle">{{vm.groups.focused().name}}</a><!-- nospace
```
hmstepanek: This is WIP also.

src/styles/sidebar/components/group-list.scss
```diff
@@ -105,6 +105,16 @@ $group-list-spacing-below: 50px;
...
   font-weight:bold;
+  display: inline-block;
+}
+
+.group-list-label__toggle {
+  color: $gray-dark;
+  cursor: default;
+}
+
+.has-toggle group-list-label__toggle {
```
robertknight: To match both these classes on the same element, this should be `.group-list-label__toggle.has-toggle` (the lack of spacing is the important part). Otherwise this becomes a descendant selector ("any child of a `has-toggle` element, include itself, that has `group-list-label__toggle`).

src/sidebar/components/test/group-list-test.js
```diff
@@ -332,4 +332,58 @@ describe('groupList', function () {
     assert.calledWith($window.open, 'https://test.hypothes.is/groups/new',
       '_blank');
   });
+
+  it('should set showGroupMenu=false if third party service and only one group', function () {
```
robertknight: Component controllers are essentially classes. The convention for tests for class methods is to group tests for each method in a describe block:

```
describe('#showGroupMenu', () => {
  it('should return false if ...'), () => {

  });
});
```

src/sidebar/components/test/group-list-test.js
```diff
@@ -332,4 +332,58 @@ describe('groupList', function () {
...
+      links: {
+        html: PRIVATE_GROUP_LINK,
+      },
+      name: 'Hypothesis Developers',
+      organization: groupFixtures.defaultOrganization(),
+      type: 'private',
+    }];
+    fakeGroups.all = () => { return group; };
+
+    const element = createGroupList();
```
robertknight: This approach to creating the group list is fine I think since there are only a small number of tests and the extra overhead from unnecessarily rendering the template won't be too large. You asked in Slack though about testing without rendering the template. The idiomatic way to do that in Angular 1.x for components is to use `$componentController`. Have a search through the codebase for examples.

src/sidebar/templates/group-list.html
```diff
@@ -1,22 +1,26 @@
...
+      !--><i class="h-icon-arrow-drop-down" ng-if="vm.showGroupsMenu()"></i></h2>
```
robertknight: If `ng-if` on the `<i>` looks unnecessary to me since the same conditional exists on the parent element. If the parent element is hidden, the child will not be visible anyway.

src/sidebar/templates/group-list.html
```diff
@@ -1,22 +1,26 @@
...
+    <h2 class="group-list-label__label" ng-if="!vm.showGroupsMenu()"><a class="group-list-label__toggle">{{vm.groups.focused().name}}</a><!-- nospace
+      !--><i class="h-icon-arrow-drop-down" ng-if="vm.showGroupsMenu()"></i></h2>
```
robertknight: The `ng-if` on the `<i>` here will effectively never evaluate to `true` since the parent element is hidden in that case.

src/sidebar/templates/group-list.html
hmstepanek: Yeah we were just messing around. I just committed whatever i had for @dwolfe to continue playing with it.

src/sidebar/components/test/group-list-test.js
hmstepanek: As per slack discussion, I'm going to leave this as is for now (test via rendering) so I can add other assertions later to verify the ng-if's result in the rendering of expected components.

hmstepanek: I have added the rendering tests. I wanted to do a test of the <ul class=dropdown-menu> but in both the w/ and w/o drop down menu cases it had a length of 0 for the list of elements which was a little confusing so I used the <li class=dropdown-menu__row> instead which are inside that <ul>. I'm not sure if lengthof is the best assertion check for this. I was originally trying to use an isempty check but that didn't work so I changed to checking the length.

src/sidebar/components/group-list.js
```diff
@@ -75,6 +75,10 @@ function GroupListController($window, analytics, groups, settings, serviceUrl) {
   }
 
   this.isThirdPartyService = isThirdPartyService(settings);
+
+  this.showGroupsMenu = function() {
```
robertknight: For new code I would recommend that we use arrow functions rather than anonymous functions. That makes the behavior of `this` easier to reason about - it will always be whatever `this` was in the scope at the time of the declaration instead of depending on how the function is called.

At some point I plan to do a bulk update of existing code using automated tools.

src/sidebar/components/test/group-list-test.js
```diff
@@ -332,4 +332,89 @@ describe('groupList', function () {
     assert.calledWith($window.open, 'https://test.hypothes.is/groups/new',
       '_blank');
   });
+  describe('#showGroupMenu', () => {
+    it('should set showGroupMenu=false if third party service and only one group', function () {
```
robertknight: Given that this is no longer just a test for the `#showGroupMenu` controller method, the test description isn't really accurate any more.

I can see two ways we could address this:

1. Create two separate sets of tests, one that checks the result of the `showGroupMenu` method depending on the group inputs, and one that checks the rendering depending on the result of the `showGroupMenu` method. The second set of tests could simply stub `showGroupMenu` to return a fixed value instead of repeating the setup.

2. Continue to have each test checking both the `showGroupMenu` result and the rendered output, but change the name of the `describe` block and `it` description to reflect the level of the "API" which is being tested, in this case the user interface. Something like this:

    ```js
    describe('group menu visibility', () => {
       it('is hidden when using third party accounts if there is only one group', () => { ... })
   
       it('is visible when using third party accounts if there are multiple groups', () => { ... }}
    });
    ```

If the logic for `showGroupMenu` was complex and there were many permutations to test, I would definitely split up the tests, because you'd still only need two tests for the rendering but you might have many more for the controller method behaviour. In this case, it doesn't matter as much.

src/sidebar/templates/group-list.html
```diff
@@ -1,22 +1,32 @@
...
-  <ul class="dropdown-menu pull-none" role="menu">
+  <!-- Do not show a drop down menu if showGroupsMenu is false. -->
+  <div ng-if="!vm.showGroupsMenu()">
+    <img class="group-list-label__icon group-list-label__icon--organization"
+         ng-src="{{ vm.focusedIcon() }}"
+         alt="{{ vm.orgName(vm.groups.focused().id)}}"
+         ng-if="vm.focusedIcon()">
+         <i class="group-list-label__icon h-icon-{{ vm.focusedIconClass() }}"
+             ng-if="!vm.focusedIcon()"></i><!-- nospace
+    !--><h2 class="group-list-label__label">{{vm.groups.focused().name}}</h2>
```
robertknight: It is a bit unfortunate that these several lines of markup had to be duplicated. The general pattern I would like us to use when there is significant duplication is to extract the duplicated markup into a separate component, just like extracting a function in regular code.

Unfortunately in Angular there is quite a bit of boilerplate involved in creating a new component compared to eg. React, so that's probably best left for a separate PR.

src/styles/sidebar/components/group-list.scss
```diff
@@ -105,6 +105,15 @@ $group-list-spacing-below: 50px;
...
+  display: inline-block;
+}
+
+.group-list-label__toggle {
+  color: $gray-dark;
+}
+
+.dropdown-toggle {
+  cursor: pointer;
 }
```
robertknight: Could you add some notes here to clarify what the CSS changes are for?

src/sidebar/templates/group-list.html
dwolfe: I'm currently out of my depth RE: creating a new Angular component, but agree that refactoring this is probably best separated out into another PR.

dwolfe:
> The groups drop-down menu has become detached from the menu indicator on this branch, and no longer matches the style of the account / help menu.

@robertknight I'm glad you caught this, as it points out some other problems with the header. Currently the dropdown toggle is a little too small for general tap target recommendations:
screen shot 2018-11-13 at 10 02 04 am

With the introduction of the `<h2>` element we were able to bump that up to (just about) the minimum recommended 44px height:
screen shot 2018-11-13 at 9 42 37 am

And that's what's causing the discrepancy with the menu position - the menu itself is positioned at the bottom of the toggle element, so when the toggle element is bigger, the menu appears to drift downward. Looking at the other buttons on the right side of the header, they're all pretty tiny. The account menu toggle is actually less tiny because of the down caret, but still well below minimum tap target size:
screen shot 2018-11-13 at 9 42 16 am

So there's a decision to be made here. Do we:

1. Degrade the solution in this PR to go back to the old, too-small tap target size
2. Upgrade the other buttons in the header to a usable tap target size; or
3. Leave it as-is and address in a separate PR

I think my vote goes to 3, with the caveat that we should try to include it in this sprint (and I'm happy to take that on). Thoughts?

hmstepanek:
It does look different, not in a dysfunctional way though. Will users notice this is temporarily different-most likely not. Is this a bit of scope creep-yes. Would this improve UI and user experience in the long run-yes. Ideally it would be best if this was fixed and then we merged this PR but I don't think it's a good idea to block this PR that whole time. I'm not generally a fan of releasing something that's slightly broken and then fixing it in a follow up. I do think this fix prob deserves it's own PR though. What we could do is get this one approved/ready to go but not merge it until @dwolfe comes up with a fix for the styling then release them together. All this being said I think I'm generally in agreement with @dwolfe suggestion for option 3.

src/sidebar/templates/group-list.html
hmstepanek: I felt like there would be a lot of duplication between the two controllers so I chose to add it to the existing one rather than make two.

hmstepanek:
OK @robertknight I've cleaned up the commit history and addressed the changes you requested. See @dwolfe 's suggestion about what to do with the drop down menu location change.

robertknight:
> I think my vote goes to 3, with the caveat that we should try to include it in this sprint (and I'm happy to take that on). Thoughts?

For my money this is a fairly visible regression in a prominent part of the UI, which I would rather not ship as-is.

Making sure that hit targets are large enough is obviously a good thing to do and I'm glad that attention is being paid to it. Testing the current production client iOS and Android just now, I found the groups drop-down menu is actually easy to hit _on mobile_ because of the browser's hit-testing heuristics which enlarge the effective area of the drop-down trigger. The practical benefit of the change here actually comes _on desktop_ because those hit-testing heuristics don't apply when using a mouse. The small buttons on the other hand are more of a problem because those heuristics can't change the fact that the buttons are too close together and so it has to choose between them. I'd be happy to see that fixed as part of a separate PR.

On the basis of the above, I would be OK with either of the following:

- Revert the span => h2 change in this PR and do that as follow-up work (fine to do that in this sprint), together with an adjustment to the drop-down menu's positioning.
- Add additional commits to this PR to adjust the position of the dropdown menu. If you do this, please check all the places where it gets used (groups drop-down, sort button menu, account menu, "Post to" menu on annotation cards).

Does this sound reasonable?

dwolfe:
Yep, I can take on the button sizes in a separate PR. I'll revert the `h2` to a `span` for now.

src/sidebar/templates/group-list.html
```diff
@@ -1,22 +1,32 @@
...
+        data-toggle="dropdown" ng-if="vm.showGroupsMenu()">
     <img class="group-list-label__icon group-list-label__icon--organization"
          ng-src="{{ vm.focusedIcon() }}"
          alt="{{ vm.orgName(vm.groups.focused().id)}}"
          ng-if="vm.focusedIcon()">
          <i class="group-list-label__icon h-icon-{{ vm.focusedIconClass() }}"
              ng-if="!vm.focusedIcon()"></i><!-- nospace
-    !--><span class="group-list-label__label">{{vm.groups.focused().name}}</span><!-- nospace
-      !--><i class="h-icon-arrow-drop-down"></i>
+    !--><span class="group-list-label__label"><a class="group-list-label__toggle">{{vm.groups.focused().name}}</a><!-- nospace
```
hmstepanek: @dwolfe The span tags are in a different place than before-is that intentional/ok?

dwolfe: Yep, that's intentional. I literally just changed the `<h2>` to a `<span>` and looked to make sure it appeared correctly, and it did. Wouldn't hurt to have confirmation of that if you want to take a look, but my hope was that it would be just as simple to change it back with the new PR.

hmstepanek:
@robertknight this is ready for another look.

robertknight:
Apart from the above issue, this PR looks good from a functional point of view.

robertknight:
The menu is now read correctly by VoiceOver on macOS. I see that adding the `tabindex` attribute fixes an existing problem where tabbing around the app skips over this menu (👍).  Unfortunately the focus indicator is displayed rather prominently in Chrome when the menu is open, not just when the button is initially focused or being pressed.

Chrome:

tabindex-chrome

In Safari (and even more so in Firefox) the effect is more subtle:

tabindex-safari

Safari / Firefox look OK for my money, but I find the effect in Chrome a bit jarring.

I had a look around at a few sites to see how this behaves elsewhere. In [Bootstrap 4's navbar](https://getbootstrap.com/docs/4.0/components/navbar/#supported-content) the drop-down menu toggle exhibits different behaviour in Chrome and Safari:

- In Safari the toggle displays a subtle focus indicator when tabbing around with the keyboard or when using the mouse
- In Chrome the toggle displays a bold focus indicator when tabbing around with the keyboard but not when using the mouse. I couldn't immediately see what causes the behaviour difference.

The behaviours of the Bootstrap example in both of these browsers seems fine.

I'm going to merge this branch as it is. It would be good figure out though how we can:

 - Get a better understanding of the outline focus behaviour
 - Avoid triggering the bold outline in Chrome when using the mouse
 - Make the interface more consistent with regards to outline focus handling. I see we currently disable the focus ring in a whole bunch of places where we probably really shouldn't.

PR #814, opened 2018-11-13
Test the W.I.P. bit

robertknight:
This is a check to see if certain words in commit messages mark pull requests as not ready for merging.


robertknight:
This is indeed working.

PR #821, opened 2018-11-20
Update Babel and code coverage tooling

robertknight:
This PR updates our JS toolchain from Babel v6 to Babel v7 and enables usage of post-ES 2015 JS features. I have also followed the advice in the Babel docs to transpile features based on the browsers we want to support using the babel "env" preset, instead of just transpiling *everything*.

Since Babel sits at the bottom of the development tooling stack, it was necessary to update various tools which make use of it, including the tools which handle:

- Conversion of `@nginject` comments into code used by Angular's dependency injection
- Instrumentation of code to enable code coverage reports

Both of the above used to be handled by Browserify plugins. They are now handled by Babel plugins.

I also had to upgrade the reporter which processes the code coverage data and generates JSON/HTML files in the `coverage/` output dir.  A benefit of this is that we now have a simpler code instrumentation setup (just a Babel plugin to add instrumentation code) so hopefully there should be fewer headaches managing this going forwards. The use of a Babel plugin to handle instrumentation should also produce more accurate results.


robertknight:
The tests pass and the application builds and runs. The remaining issue to resolve is that the code coverage info looks incorrect for CoffeeScript files. I think some of the changes in the JS code coverage reported may be correct, but need to take a closer look.

.eslintrc
```diff
@@ -14,4 +11,10 @@
       "ObjectExpression": "first"
     }]
   },
+  "parserOptions": {
+    "ecmaVersion": 2018
```
robertknight: This means that ESLint will no longer complain if you use JS features added after ES2015.

.babelrc
```diff
@@ -1,4 +1,11 @@
 {
-  "presets": ["es2015"],
-  "ignore": "**/vendor/*"
+  "presets": [
+    ["@babel/preset-env", {
+       "targets": {
+         "browsers": ["> 0.3%"]
```
robertknight: This config means that the (new) JS we write will be compiled to old JS for compatibility with older browsers, but only those that have >= 0.3% market share. You can also explicitly specify particular browser versions that you want to support or specify date ranges (eg. "all versions of Chrome and Firefox that are less than two years old")

robertknight:
> The test files are getting included in the coverage report which shouldn't be the case. 

Ah, well-spotted. I've added configuration to avoid instrumenting those files - see last commit. 

> Also, looks like there are some conflicts that need resolving before merging.

Conflicts in `yarn.lock` (the client's "requirements.txt") happen often with this PR because it touches quite a lot of dependencies. yarn can fix these conflicts automatically by just running `yarn`. In other repos that use an npm lockfile (package-lock.json), [`npm install`](https://docs.npmjs.com/files/package-locks#resolving-lockfile-conflicts) can do the same thing. I will fix any such conflicts before merging.

> In short, it seems the new coverage is more accurate or intuitively what I would expect to see.

Yeah, that's what you should see. The instrumentation code used to be added after JavaScript was compiled down to ES5 for compatibility with IE etc. The instrumentation is now added earlier in the pipeline which should make it better match the original source.

robertknight:
After disabling coverage for tests, the coverage dropped slightly relative to master in percentage terms, but looking at a few source files I can see why. In the example below, function and variable declarations were counted as hits previously but are now not counted as either hits or misses. Hence the percentage of coverage has dropped slightly. The new output is better though.

**Before:**

screenshot 2019-01-31 06 53 39

**After:**

screenshot 2019-01-31 06 55 52

PR #822, opened 2018-11-29
Delete only invalid highlight when anchor has multiple highlights

lewdea:
When hypothesis client works with PDF.js, there is a bug here.  Description as following:

1. An anchor has multiple highlights, this happens when you select content from crosslines.
2. Click the 'search' button on the left side of pdf.js viewer.
3. Input any text but is part of the crossline highlights.
4. The BUG occurs! The part of the crossline highlights will be highlighted again!

This happens because of the following code from guest.coffee can't remove all the highlights which will be updated and re-anchor.

` else if anchor.highlights?`
`      # These highlights are no longer valid and should be removed.`
`      deadHighlights = deadHighlights.concat(anchor.highlights) `

### DIFF
--- a/src/annotator/plugin/pdf.coffee
+++ b/src/annotator/plugin/pdf.coffee
@@ -72,9 +72,9 @@ module.exports = class PDF extends Plugin
       # If the highlights are no longer in the document it means that either
       # the page was destroyed by PDF.js or the placeholder was removed above.
       # The annotations for these anchors need to be refreshed.
-      for hl in anchor.highlights
+      for hl, index in anchor.highlights
         if not document.body.contains(hl)
-          delete anchor.highlights
+          anchor.highlights.splice(index, 1)
           delete anchor.range
           refreshAnnotations.push(anchor.annotation)
           break


robertknight:
I can confirm the bug:

pdf-search-highlight-bug

The yellow highlighted lines on the right should only be highlighted once as they are part of a single annotation, but after searching get highlighted multiple times.

lewdea:
@robertknight     Thank you for confirm this bug and explain it in a simple way while my English is not that good enough to describe it clearly.

src/annotator/plugin/pdf.coffee
```diff
@@ -72,9 +72,9 @@ module.exports = class PDF extends Plugin
       # If the highlights are no longer in the document it means that either
       # the page was destroyed by PDF.js or the placeholder was removed above.
       # The annotations for these anchors need to be refreshed.
-      for hl in anchor.highlights
+      for hl, index in anchor.highlights
         if not document.body.contains(hl)
-          delete anchor.highlights
+          anchor.highlights.splice(index, 1)
```
robertknight: Note to self: I wondered how other code handled the fact that the anchor still has some highlights but no range. This is handled in [Guest#anchor](https://github.com/hypothesis/client/blob/2bc090cfa0fa471650b4e7b61622e79a8027ed1c/src/annotator/guest.coffee#L281) by adding all existing highlights to a list to remove when re-anchoring. The problem before was that the highlights that were still in the document were removed from `anchor.highlights` before `Guest#anchor` is called and therefore never got deleted.

PR #823, opened 2018-11-29
Hide non-applicable tutorial steps for third party users

robertknight:
Hide steps in the sidebar tutorial which are not available to
third-party users. It is expected that the conditions under which these
actions are available may change in future, but they will still be
unavailable in at least some contexts.

nb. This affects third party authorities _in general_ but not eLife, since they are using
a different version of the sidebar tutorial which already has the unavailable steps omitted.

Fixes https://github.com/hypothesis/product-backlog/issues/842

### DIFF
--- a/src/sidebar/components/sidebar-tutorial.js
+++ b/src/sidebar/components/sidebar-tutorial.js
@@ -1,9 +1,13 @@
 'use strict';
 
 const sessionUtil = require('../util/session-util');
+const isThirdPartyService = require('../util/is-third-party-service');
 
 // @ngInject
 function SidebarTutorialController(session, settings) {
+  // Compute once since this doesn't change after the app starts.
+  const isThirdPartyService_ = isThirdPartyService(settings);
+
   this.isThemeClean = settings.theme === 'clean';
 
   this.showSidebarTutorial = function () {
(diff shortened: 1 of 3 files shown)


lyzadanger:
@robertknight Was wondering if you might have a screen shot handy of what the tutorial looks like with those two sections disabled?

robertknight:
Sure, it looks like this:

sidebar tutorial for third party users

lyzadanger:
This is a bit of an aside, but I find codecov reports confusing. I was trying to track down why this is only hitting at 83%, and can't grok why it thinks this line is a test miss:

https://codecov.io/gh/hypothesis/client/pull/823/diff?src=pr&el=tree#D3-25

robertknight:
>  I was trying to track down why this is only hitting at 83%, and can't grok why it thinks this line is a test miss

There might be an issue with source map handling. We are not using Isparta's official solution for ES6+ support which is [babel-plugin-istanbul](https://github.com/istanbuljs/babel-plugin-istanbul). So that may be connected to this. I've been looking into upgrading Babel 6 => Babel 7 and we'll probably need to make that change as part of the work anyway.

PR #825, opened 2018-12-05
Fix selection of annotations when bucket bar indicators are clicked

robertknight:
A typo, `annotator` instead of `@annotator`, caused this functionality
to break when `window.annotator` (which was the same object as
`@annotator`) was removed in 0033f774424.

Correct the typo and add tests. The tests operate at the DOM level so
they should be easy to adapt if the BucketBar implementation is
rewritten or changed substantially in future.

Fixes https://github.com/hypothesis/product-backlog/issues/704

### DIFF
--- a/src/annotator/plugin/bucket-bar.coffee
+++ b/src/annotator/plugin/bucket-bar.coffee
@@ -76,6 +76,8 @@ module.exports = class BucketBar extends Plugin
     else
       $(element).append @element
 
+    @highlighter = options.highlighter ? highlighter  # Test seam.
+
   pluginInit: ->
     $(window).on 'resize scroll', @update
 
@@ -113,7 +115,7 @@ module.exports = class BucketBar extends Plugin
       unless anchor.highlights?.length
         return points
 
-      rect = highlighter.getBoundingClientRect(anchor.highlights)
+      rect = @highlighter.getBoundingClientRect(anchor.highlights)
       x = rect.top
       h = rect.bottom - rect.top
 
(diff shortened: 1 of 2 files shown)


dwhly:
Yay! Thank you.

src/annotator/plugin/bucket-bar.coffee
```diff
@@ -251,7 +253,7 @@ module.exports = class BucketBar extends Plugin
           scrollToClosest(@buckets[bucket], 'down')
         else
           annotations = (anchor.annotation for anchor in @buckets[bucket])
-          annotator.selectAnnotations annotations,
+          @annotator.selectAnnotations annotations,
```
robertknight: This is the actual bug fix. The value assigned to `this.annotator` / `@annotator` used to be assigned to a global variable `annotator`.

src/annotator/plugin/bucket-bar.coffee
```diff
@@ -76,6 +76,8 @@ module.exports = class BucketBar extends Plugin
     else
       $(element).append @element
 
+    @highlighter = options.highlighter ? highlighter  # Test seam.
```
robertknight: Test seam to enable `getBoundingClientRect` output below to be mocked in tests.

src/annotator/plugin/test/bucket-bar-test.coffee
```diff
@@ -1,12 +1,89 @@
...
+        assert.calledWith(bucketBar.annotator.selectAnnotations, anns, true)
+    )
+
+
   # Yes this is testing a private method. Yes this is bad practice, but I'd
   # rather test this functionality in a private method than not test it at all.
+  #
+  # Note: This could be tested using only the public APIs of the `BucketBar`
+  # class using the approach of the "when a bucket is clicked" tests above.
   describe '_buildTabs', ->
```
robertknight: There are existing tests to check that for a given set of annotations, window dimensions and scroll position, the buckets in the bucket bar are built as expected. However they rely on internals of the `BucketBar` class so can't be used if the implementation is significantly changed.

PR #828, opened 2018-12-11
Refactor Makefile

seanh:
Refactor Makefile to be the same as for our other projects:

    $ make help
    make help              Show this help message
    make dev               Run the app in the development server
    make lint              Run the code linter(s) and print any warnings
    make test              Run the unit tests
    make docs              Build docs website and serve it locally
    make checkdocs         Crash if building the docs website fails
    make clean             Delete development artefacts (cached files,
                           dependencies, etc)

Other make commands that our other projects have but the client doesn't
yet:

`make shell` -- Not relevant to the client? (Unless we can get a useful
JavaScript shell in the terminal.)

`make sql` -- Not relevant.

`make format` and `checkformatting` -- We can't use Black code
formatting for the client of course but maybe we could use Prettier or
something else. (Our Python apps could also be extended to have `make
format` format the JavaScript code as well.)

`make coverage` and `codecov` -- Print the unit test coverage report to
the terminal, and post it to codecov. These should be doable?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,19 +1,28 @@
 .PHONY: default
-default: all
+default: help
 
-.PHONY: all
-all: build/manifest.json
+.PHONY: help
+help:
+	@echo "make help              Show this help message"
+	@echo "make dev               Run the app in the development server"
+	@echo "make lint              Run the code linter(s) and print any warnings"
+	@echo "make test              Run the unit tests"
+	@echo "make docs              Build docs website and serve it locally"
+	@echo "make checkdocs         Crash if building the docs website fails"
+	@echo "make clean             Delete development artefacts (cached files, "
+	@echo "                       dependencies, etc)"
 
-## Remove build artifacts
-.PHONY: clean
-clean:
-	rm -f node_modules/.uptodate
-	rm -rf build
+.PHONY: dev
+dev: build/manifest.json
+	gulp watch
 
-## Run test suite
 .PHONY: test
 test: node_modules/.uptodate
+ifdef FILTER
+	yarn test --grep $(FILTER)
+else
 	yarn test
+endif
 
 .PHONY: lint
 lint: node_modules/.uptodate
@@ -27,7 +36,10 @@ docs:
 checkdocs:
 	tox -e py3-checkdocs
 
-################################################################################
+.PHONY: clean
+clean:
+	rm -f node_modules/.uptodate
+	rm -rf build
 
 build/manifest.json: node_modules/.uptodate
 	yarn run build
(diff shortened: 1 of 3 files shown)


robertknight:
> We can't use Black code formatting for the client of course but maybe we could use Prettier or something else. 

I would be quite keen to consider Prettier. From a little bit of testing, `prettier --single-quote --trailing-comma es5 --write <file path expressions>` should do the trick. Prettier can also reformat other file types including HTML, SCSS and JSON.

It probably would be a good idea to repeat what we've done with Python and apply it to some smaller projects before applying it to the client.

seanh:
The bouncer and lms projects both contain some js code so could be starting points. It'll be crucial to have `make checkformatting` integration as well, we need to enforce that the code stays formatted on master otherwise running `make format` on a feature branch would make unrelated formatting changes.

Makefile
```diff
@@ -1,19 +1,24 @@
...
-.PHONY: clean
-clean:
-	rm -f node_modules/.uptodate
-	rm -rf build
-
-## Run test suite
 .PHONY: test
 test: node_modules/.uptodate
-	yarn test
+	yarn run test
```
robertknight: [yarn test](https://yarnpkg.com/en/docs/cli/test) is an actual command, though just a shortcut for `yarn run test`. Was there any particular reason for the change?

docs/developers/developing.rst
```diff
@@ -135,7 +135,7 @@ To run tests and automatically re-run them whenever any source files change, run
    gulp test-watch
 
 You can filter the tests which are run by passing ``--grep <pattern>`` as an
-argument to ``gulp test``. See the documentation for Mocha's
+argument to ``make test``. See the documentation for Mocha's
```
robertknight: This section about filtering tests no longer worsk because make doesn't pass along CLI arguments to gulp. This could be revised to use `make test FILTER=<pattern>` instead using:

```diff
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,11 @@ dev: build/manifest.json

 .PHONY: test
 test: node_modules/.uptodate
-       yarn run test
+ifdef FILTER
+       yarn test --grep $(FILTER)
+else
+       yarn test
+endif
```

Makefile
seanh: I _thought_ this made it run the test just once and then finish, rather than waiting, but that seems not to be the case. For consistency with `make test` in other projects it _should_ run the tests and then exit, which I think would also mean that `make test` could be used to run the tests on Travis and Jenkins. It's nice to have a single `make test` command that's used everywhere because if you ever need to change it, you just change what `make test` does and that gets used everywhere.

seanh: For now I've just changed this back to `yarn test`

docs/developers/developing.rst
seanh: Done

PR #831, opened 2019-01-09
Upgrade to gulp 4

robertknight:
This should make future maintenance easier since Gulp v4 has better
documentation [1], more checks for incorrect usage and a simpler API for
expressing what should run in parallel vs sequentially during build
steps. This also eliminates a bunch of old transitive dependencies that
gulp v3 had, which sometimes caused problems (eg. with Node upgrades).

Changes required:

 - Make sure `gulp.task` functions return a value, so Gulp knows when
   the task is finished
 - `gulp.watch` now requires a function rather than a task name
 - Task no longer support a list of dependencies but need a task
   function. Dependencies are expressed using `gulp.series` or
   `gulp.parallel` to compose other functions
 - Give several functions names so that there are fewer `<anonymous>
   {starting, finished}` messages in log output.

[1] https://gulpjs.com/docs/en/api/concepts

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -121,17 +121,17 @@ const appBundleConfigs = appBundles.map(function (config) {
   return Object.assign({}, appBundleBaseConfig, config);
 });
 
-gulp.task('build-js', ['build-vendor-js'], function () {
+gulp.task('build-js', gulp.parallel('build-vendor-js', function () {
   return Promise.all(appBundleConfigs.map(function (config) {
     return createBundle(config);
   }));
-});
+}));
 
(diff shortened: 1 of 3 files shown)


robertknight:
> It looks like there are quite a lot of changes to the yarn.lock file-I'm assuming all of those are necessary.

Yes. That's all auto-generated by yarn. It's a much larger diff than you get to eg. requirements.txt when updating dependencies in Python because of the JavaScript ecosystem's love of small packages and liberal approach to adding dependencies. Quite a few of the upgraded dependencies literally provide a single function that gulp (or one of its dependencies) uses somewhere.

PR #832, opened 2019-01-09
Update AngularJS to 1.7.x

robertknight:
This PR updates AngularJS to the current stable version (1.7.5), which is due to be the [last major release](https://blog.angular.io/stable-angularjs-and-long-term-support-7e077635ee9c) of AngularJS.

The [migration guide](https://docs.angularjs.org/guide/migration#migrating-from-1-6-to-1-7) lists all the things that one might need to do, but the only change which affects us as far as I can see is the part about not using `$compileProvider.preAssignBindingsEnabled(true)`. This means we needed to change any places where we were accessing component inputs in the controller constructor and do that in the controller's `$onInit` method instead.


src/sidebar/components/annotation.js
```diff
@@ -569,8 +569,6 @@ function AnnotationController(
     }
     return self.group().type !== 'private';
   };
-
-  init();
```
robertknight: The `init` function has been converted to an `$onInit` method on the controller which Angular invokes automatically, so we don't need to call it ourselves.

src/sidebar/components/stream-content.js
```diff
@@ -69,6 +60,11 @@ function StreamContentController(
   store.setSortKey('Newest');
 
   this.loadMore = fetch;
+
+  this.$onInit = () => {
+    this.search.query = () => $routeParams.q || '';
+    this.search.update = q => $location.search({ q });
+  };
```
robertknight: This is the same logic as before, but I just moved accesses of `this.search` into `$onInit`.

Issue #810, opened 2018-11-08
Deploying a release that is not the latest merge to master can fail


robertknight:
Attempting to deploy a release from Jenkins which is not the latest merge to master can fail as follows:

1. Merge PR A 
2. Wait a minute or so and merge PR B
3. Wait for Jenkins build of PR A and PR B to finish QA deploy
4. Attempt to deploy first build (PR A) to prod.

Step 4 will create a Git tag and a commit on the master branch for the release and attempt to push that to GitHub. However this will fail because the release commit was not made against the latest version of master, but the commit before PR B was merged.

See also this [Slack discussion](https://hypothes-is.slack.com/archives/C1M8NH76X/p1541678684060500)

To resolve this, we'll either need to change the release process to avoid actually creating commits in the process, or ensure that Jenkins merges the release commit to master rather than attempting to force-push it.

seanh:
Jenkins is trying to do a release commit on top of the latest commit that it has (which is the commit that Jenkins has tested, and that the developer has tested on QA). Wouldn't merging with other commits from GitHub mean the release commit would not represent the same code that has been tested?

robertknight:
After some Slack discussion we concluded that avoiding having to commit to master as part of the release would be best. We can do this by:

- Removing CHANGELOG.md from the repo. The info will still be available in GitHub releases.
- Having a dummy version in package.json on master and then generating the monotonically increasing version number for a release based on the last Git tag.

robertknight:
The first part of this, removing the CHANGELOG, is happening in https://github.com/hypothesis/client/pull/911

PR #922, opened 2019-02-04
Stop updating package.json version field during deployments from Jenkins

robertknight:
This PR removes updating of the the version field in package.json during a Jenkins deployment. Instead the version field is set based on the highest-tagged previous version when building the client and publishing it to npm. This change is not actually committed however.

The upshot is that, since client deployments no longer try to push additional commits to the master branch, it is possible to deploy client releases which are no longer the latest. This means that if there are a queue of recently being merged PRs waiting to be deployed, it is possible to deploy them incrementally rather than all at once.

Fixes https://github.com/hypothesis/client/issues/810


Jenkinsfile
```diff
@@ -108,21 +103,13 @@ node {
                 """
             }
         }
-
-        // Revert back to the pre-QA commit.
-        sh "git checkout ${lastCommitHash}"
```
robertknight: No git commit is created during the QA release process, so this is unnecessary

Jenkinsfile
```diff
@@ -178,7 +162,7 @@ node {
 String bumpMinorVersion(String version) {
     def parts = version.tokenize('.')
     if (parts.size() != 3) {
-        throw new IllegalArgumentException("${version} is not a valid MAJOR.MINOR.PATCH version")
+        error "${version} is not a valid MAJOR.MINOR.PATCH version"
```
robertknight: The previous code never worked in Jenkins because you can't `new` things in this script due to security restrictions.

package.json
```diff
@@ -148,7 +148,6 @@
     "test": "gulp test",
     "report-coverage": "codecov -f coverage/coverage-final.json",
     "version": "make clean build/manifest.json",
-    "postversion": "./scripts/postversion.sh",
```
robertknight: The logic that used to be here has been moved to the Jenkinsfile.

Jenkinsfile
```diff
@@ -29,38 +29,33 @@ node {
...
+    sh "git fetch --quiet --prune origin 'refs/tags/*:refs/tags/*' "
+
+    // Determine version number for next release.
     pkgVersion = sh (
-      script: 'cat package.json | jq -r .version',
+      script: 'git tag --list | sort --version-sort --reverse | head -n1 | tail -c +2',
       returnStdout: true
     ).trim()
-
-    stage('Build') {
```
robertknight: This stage doesn't actually do anything useful. `make` on its own just prints help output. The client actually gets built during the "QA" and "prod" steps with the appropriate version number. If there is eg. a syntax error, it'll get flagged either during the Test or QA steps.

PR #924, opened 2019-02-05
Format code with Prettier

robertknight:
Following the successful introduction of automated code formatting for h, this PR introduces the same in the client using [Prettier](https://prettier.io). The first two commits add configuration for Prettier and run checks during CI builds. I've set the configuration to the minimal config which best matches the current style of our code. The last commit is a mega-commit that runs Prettier over our existing code. I suggest not trying to review all of that by hand 😛.

When working on code, you can run formatting using `make formatting` or check it using `make checkformatting`, as in our Python projects. Vim users auto-format code while working using [vim-prettier](https://github.com/prettier/vim-prettier). There are [integrations for other editors](https://prettier.io/docs/en/editors.html) as well.


robertknight:
The one decision that gives me a little bit of pause here is whether to leave the line width setting at the default of 80 or whether to make it a little larger. Black for example defaults to 88.

80 chars feels a little aggressive to me, although Prettier doesn't strictly wrap at this length in all cases. In particular `it` blocks in tests are allowed to be longer.

On the other hand, the docs make a point of explicitly recommending this default and it reduces the number of configuration options we need to have. Both of the remaining settings we configure (single quotes, ES5 trailing comma) are changes they are considering [making the default in future](https://github.com/prettier/prettier/issues/3503).

PR #930, opened 2019-02-06
Add custom client version header to api requests

hmstepanek:
Add a custom http request header named 'Hypothesis-Client-Version' to all api requests sent from the client to h. 

This will enable us to track which versions of the clients are in use and also help us distinguish between api calls issued from our client vs api calls issued from another party. This is useful so we can tell for instance which api features our customers depend on and which features we can deprecate, or what clients are still in-play so we don't cause any backwards compatibility issues between client updates and server updates.


hmstepanek:
Oooh that’s a good idea!

PR #931, opened 2019-02-06
Add servetests to makefile for interactive test debug

hmstepanek:
Add `make servetests` command to Makefile. This command will run the unit tests and serve them on http://localhost:9877 so that any `debugger;` breakpoints take effect and the browser console can be used to interact live with the code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ help:
 	@echo "make lint              Run the code linter(s) and print any warnings"
 	@echo "make checkformatting   Check code formatting"
 	@echo "make format            Automatically format code"
-	@echo "make test              Run the unit tests"
+	@echo "make test              Run the unit tests once"
+	@echo "make servetests        Start the unit test server on localhost"
 	@echo "make docs              Build docs website and serve it locally"
 	@echo "make checkdocs         Crash if building the docs website fails"
 	@echo "make clean             Delete development artefacts (cached files, "
@@ -25,6 +26,10 @@ ifdef FILTER
 else
 	yarn test
 endif
+
+.PHONY: servetests
+servetests: node_modules/.uptodate
+	gulp test-watch
 
 .PHONY: lint
 lint: node_modules/.uptodate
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -7,6 +7,7 @@ help:
 	@echo "make dev               Run the app in the development server"
 	@echo "make lint              Run the code linter(s) and print any warnings"
 	@echo "make test              Run the unit tests"
+	@echo "make debug             Run the unit tests in interactive mode on http://localhost:9877"
```
robertknight: The default port is **9876**. Karma will however automatically increment this if this is already in use. I would suggest leaving the port out here. Karma prints the chosen port number to the console when it runs.

robertknight: I don't think that `make debug` makes it super clear what the difference is from `make test`. The main difference is that `make test` runs the tests once and then exits. `gulp test-watch` runs the tests once and then starts watching for changes to source files and re-runs whenever they change, or whenever a new browser connects. The test server is still run on localhost:9876 in both cases, its just that when running the tests once, it is hard to connect a browser after they start.

Perhaps something like this:

```
make test - Run the tests once
make servetests - Start the test server
```

It might also be worthwhile adding a note in the development docs (in docs/developers/developing.rst) about how to use an interactive debugger with tests.

hmstepanek:
Ok, I've updated the name of the command and added to the docs.

docs/developers/developing.rst
```diff
@@ -128,15 +128,20 @@ Hypothesis uses Karma and mocha for testing. To run all the tests once, run:
...
 To run tests and automatically re-run them whenever any source files change, run:
 
 .. code-block:: sh
 
-   gulp test-watch
+   make servetests
+
+This command will also serve the tests on localhost (typically `http://localhost:9876`)
+so that break points can be set and the browser's console can be used for interactive
+debug.
```
robertknight: s/debug/debugging

PR #932, opened 2019-02-06
Add tests for Authorization header on API requests

robertknight:
Add missing tests to check that Authorization header is set on API requests.

### DIFF
--- a/src/sidebar/services/test/api-test.js
+++ b/src/sidebar/services/test/api-test.js
@@ -17,9 +17,11 @@ const util = require('../../../shared/test/util');
 const routes = require('./api-index.json').links;
 
 describe('sidebar.services.api', function() {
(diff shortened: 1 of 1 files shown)


src/sidebar/services/test/api-test.js
```diff
@@ -17,9 +17,11 @@ const util = require('../../../shared/test/util');
...
 
 describe('sidebar.services.api', function () {
-  let $httpBackend = null;
-  let sandbox = null;
-  let api = null;
+  let $httpBackend;
+  let $q;
+  let fakeAuth;
+  let sandbox;
+  let api;
```
robertknight: The `= null` part was left-over from a conversion from CoffeeScript, which has no way to declare variables without initializing them AFAIK.

PR #937, opened 2019-02-06
Support filtering annotations by display name

robertknight:
Allow `user:{term}` filters in the search bar to match display names as
well as usernames.

I have allowed for annotations having no `user_info` field as with other
code in the client that handles annotation objects. However "h" will always
populate that field currently.

Fixes https://github.com/hypothesis/product-backlog/issues/943

### DIFF
--- a/src/sidebar/services/test/search-filter-test.js
+++ b/src/sidebar/services/test/search-filter-test.js
@@ -57,6 +57,21 @@ describe('sidebar.search-filter', () => {
       assert.equal(result.any[2], 'hi-fi');
       assert.equal(result.any[3], 'a:bc');
     });
+
+    it('supports quoting terms', () => {
+      const parsed = searchFilter.toObject('user:"Dan Whaley"');
+      assert.deepEqual(parsed, {
+        user: ['Dan Whaley'],
+      });
+    });
+
+    it('assigns unquoted terms to "any" category', () => {
+      const parsed = searchFilter.toObject('user:Dan Whaley');
+      assert.deepEqual(parsed, {
+        any: ['Whaley'],
+        user: ['Dan'],
+      });
+    });
   });
 
   describe('#generateFacetedFilter', () => {
--- a/src/sidebar/services/test/view-filter-test.js
+++ b/src/sidebar/services/test/view-filter-test.js
@@ -153,10 +153,11 @@ describe('sidebar/services/view-filter', () => {
       const anns = [
         annotationWithUser('johnsmith'),
         annotationWithUser('jamesdean'),
+        annotationWithUser('johnjones'),
       ];
-      const result = viewFilter.filter(anns, userQuery('johnsmith'));
+      const result = viewFilter.filter(anns, userQuery('john'));
 
-      assert.deepEqual(result, [anns[0].id]);
+      assert.deepEqual(result, [anns[0].id, anns[2].id]);
     });
 
     it("matches user's display name if present", () => {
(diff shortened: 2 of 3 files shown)


src/sidebar/services/test/search-filter-test.js
```diff
@@ -57,6 +57,21 @@ describe('sidebar.search-filter', () => {
       assert.equal(result.any[2], 'hi-fi');
       assert.equal(result.any[3], 'a:bc');
     });
+
```
robertknight: nb. `services/search-filter.js` is the query parser that turns strings into objects representing the query. `services/view-filter.js` is the query executor that filters annotations against a query object.

src/sidebar/services/test/search-filter-test.js
```diff
@@ -57,6 +57,21 @@ describe('sidebar.search-filter', () => {
...
+    });
+
+    it('assigns unquoted terms to "any" category', () => {
+      const parsed = searchFilter.toObject('user:Dan Whaley');
+      assert.deepEqual(parsed, {
+        any: ['Whaley'],
+        user: ['Dan'],
+      });
+    });
   });
```
robertknight: There is intentionally no behaviour change here, I just wrote these tests to clarify some behaviour that was asked about on Slack.

hmstepanek: Ah interesting! This is good to know! It's also different than activity pages.

src/sidebar/services/test/search-filter-test.js
```diff
@@ -57,6 +57,21 @@ describe('sidebar.search-filter', () => {
       assert.equal(result.any[2], 'hi-fi');
       assert.equal(result.any[3], 'a:bc');
     });
+
+    it('supports quoting terms', () => {
+      const parsed = searchFilter.toObject('user:"Dan Whaley"');
+      assert.deepEqual(parsed, {
+        user: ['Dan Whaley'],
```
hmstepanek: Not that I think Dan cares but do we want to have his name in the code?

src/sidebar/services/test/view-filter-test.js
```diff
@@ -132,6 +132,57 @@ describe('sidebar/services/view-filter', () => {
...
+        annotationWithUser('jsmith', 'John Smith'),
+        annotationWithUser('jdean', 'James Dean'),
+
+        // User with no display name.
+        annotationWithUser('fmercury'),
+
+        // Annotation with no extended user info.
+        { id: 100, user: 'acct:jim@example.com' },
+      ];
+      const result = viewFilter.filter(anns, userQuery('james'));
```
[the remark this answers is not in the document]
robertknight: Hi Hannah,

I have extended these test cases to check:

- That all annotations are returned if several match
- That partial matches in usernames match
- That usernames are still matched if display names are present

src/sidebar/services/test/search-filter-test.js
robertknight: I used this here because this was literally the example we were using in the Slack discussion that prompted me to write these tests. I think I can get away with it ;)

robertknight:
Hi @hmstepanek - I've extended the tests to cover some additional cases. PTAL.

src/sidebar/services/test/search-filter-test.js
hmstepanek: May it be forever in the git history! lol

PR #940, opened 2019-02-07
Remove Google Plus share options from the client

robertknight:
(_This is some tidy-up of the share dialog while I was looking at https://github.com/hypothesis/product-backlog/issues/882_)

- Google Plus is shutting down, so remove share icons for it from the
  client
- Add padding around remaining icons in annotation share dialog and
  spacer after "Share" label to avoid having too much space between
  icons

Fixes https://github.com/hypothesis/product-backlog/issues/947

----

**Screenshots after:**

share-page-dialog

share-annotation-dialog


robertknight:
Ah, good catch. I'll sort that out tomorrow. Updating the icomoon font is a bit of a faff.

robertknight:
The icon font has been updated following the steps in ICOMOON.md to remove the Google Plus icons. Note that it has been a long time since Icomoon was last used and so the updated CSS and selection.json file looks quite different. I committed the exported files from icomoon without any further edits for reproducibility.

robertknight:
This is ready for a re-review @hmstepanek. Since the icon font is a binary file and the selection JSON output changed in the newer version of icomoon, the easiest way to check the changes is to look at a dev and prod client side-by-side and check that the icons look the same and have the same hover states.

PR #943, opened 2019-02-11
Use token instead of store.profile to calc isLoggedIn

hmstepanek:
This is a fix for: https://github.com/hypothesis/product-backlog/issues/957. 
In summary, since groups and profile are called concurrently on startup, we cannot depend on store.profile().userid to determine whether a user is logged in or not because that call may not have completed by the time we need it in the groups service.

### DIFF
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -27,6 +27,7 @@ function groups(
   serviceUrl,
   session,
   settings,
+  auth,
   features
 ) {
   const svc = serviceConfig(settings);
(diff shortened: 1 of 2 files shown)


src/sidebar/services/groups.js
```diff
@@ -27,6 +27,7 @@ function groups(
   serviceUrl,
   session,
   settings,
+  auth,
   features
```
robertknight: I forgot to flag this up before, but longer injected parameter lists (functions with `@ngInject` above them) are easiest to work with if kept sorted.

src/sidebar/services/groups.js
```diff
@@ -158,17 +159,23 @@ function groups(
...
           const profileGroupsApi = api.profile.groups.read(profileParams);
           const listGroupsApi = api.groups.list(params);
-          return Promise.all([profileGroupsApi, listGroupsApi]).then(
-            ([myGroups, featuredGroups]) =>
-              combineGroups(myGroups, featuredGroups)
-          );
+          return Promise.all([
+            profileGroupsApi,
+            listGroupsApi,
+            auth.tokenGetter(),
```
robertknight: This approach looks cleaner to me that trying to thread the auth token through the API response as I did originally with `{ includeMetadata: true }`. For consistency it might be a good idea to use the same approach in both branches.

PR #973, opened 2019-02-22
Add Currently Viewing, Featured, & My Groups sections to group drop down

hmstepanek:
Organize the group drop down menu into sections `Currently Viewing`, `Featured Groups`, and `My Groups`. `My Groups` is a list of groups for which the user is a member and this list never changes based on the page being annotated. `Featured Groups` is a list of groups that are scoped to the page with no concept of membership and may change depending on the page. All other groups show up in `Currently Viewing`. Cases for a group showing in `Currently Viewing` are the Public group when it is linked to or a restricted group that was linked to that would otherwise not show up on the page. Some groups in `My Groups` may not be scoped to the page so while they still show up they are not selectable to be annotated in and are greyed out. 

This is a fix for https://github.com/hypothesis/product-backlog/issues/938.
Based on @dwolfe 's [PR](https://github.com/hypothesis/client/pull/830).

If the user is logged out:

1. There is only one publicly visible group so show it without the drop down. 
image
2. There are no publicly visible groups so show the public group without the drop down. 
image
3. There are multiple publicly visible groups so show them in the drop down. 
image

4. The public group's annotations have been linked to but the group has publicly visible groups so show the public group under the `Currently Viewing` header. 
image

If the user is logged in:

1. There are no featured groups so don't show them. 
image

2. There is only one group so don't show the drop down menu. 
image

3. There are open groups and they show up under the `Featured Groups`. 
image

4. There is an out-of-scope restricted group which the user is a member of shown in the My Groups list as un-selectable group. 
image

5. There is a restricted group which is in-scope and the user is a member shown in the My Groups list. 
image

6. There is a restricted group which is in-scope and the user is not a member so it's shown in the Featured Groups list. 
image

### DIFF
--- a/src/sidebar/components/group-list.js
+++ b/src/sidebar/components/group-list.js
@@ -67,6 +67,17 @@ function GroupListController(
   };
 
   /**
+   * This is a quick and dirty hack to show styling, and needs refactoring.
+   */
+  this.toggleGroupDetails = function() {
+    const groupDetails = $window.document.getElementById(
+      'group-details-out-of-scope'
+    );
+    groupDetails.classList.toggle('expanded');
+    $window.event.stopPropagation();
+  };
+
+  /**
    * Show the share link for the group if it is not a third-party group
    * AND if the URL needed is present in the group object. We should be able
    * to simplify this once the API is adjusted only to return the link
--- a/src/sidebar/components/group-list.js
+++ b/src/sidebar/components/group-list.js
@@ -4,16 +4,25 @@ const { isThirdPartyUser } = require('../util/account-id');
 const isThirdPartyService = require('../util/is-third-party-service');
 const serviceConfig = require('../service-config');
 const memoize = require('../util/memoize');
-const groupOrganizations = memoize(require('../util/group-organizations'));
+const groupsByOrganization = require('../util/group-organizations');
+
+const groupOrganizations = memoize(groupsByOrganization);
+
+const myGroupOrgs = memoize(groupsByOrganization);
+
+const featuredGroupOrgs = memoize(groupsByOrganization);
+
+const currentlyViewingGroupOrgs = memoize(groupsByOrganization);
 
 // @ngInject
 function GroupListController(
   $window,
   analytics,
+  features,
   groups,
   settings,
   serviceUrl,
-  features
+  store
 ) {
   this.groups = groups;
 
@@ -55,6 +64,18 @@ function GroupListController(
 
   this.groupOrganizations = function() {
     return groupOrganizations(this.groups.all());
+  };
+
+  this.currentlyViewingGroupOrganizations = function() {
+    return currentlyViewingGroupOrgs(store.getCurrentlyViewingGroups());
+  };
+
+  this.featuredGroupOrganizations = function() {
+    return featuredGroupOrgs(store.getFeaturedGroups());
+  };
+
+  this.myGroupOrganizations = function() {
+    return myGroupOrgs(store.getMyGroups());
   };
 
   this.viewGroupActivity = function() {
--- a/src/sidebar/components/group-list.js
+++ b/src/sidebar/components/group-list.js
@@ -26,6 +26,9 @@ function GroupListController(
 ) {
   this.groups = groups;
 
+  // Track which non-selectable groups have their group details expanded.
+  this.groupDetailsExpanded = {};
+
   this.createNewGroup = function() {
     $window.open(serviceUrl('groups.new'), '_blank');
   };
@@ -87,15 +90,21 @@ function GroupListController(
     groups.focus(groupId);
   };
 
+  this.isGroupDetailsExpanded = function(groupId) {
+    if (!(groupId in this.groupDetailsExpanded)) {
+      this.groupDetailsExpanded[groupId] = false;
+    }
+    return this.groupDetailsExpanded[groupId];
+  };
+
   /**
-   * This is a quick and dirty hack to show styling, and needs refactoring.
+   * Toggle the expanded setting on un-selectable groups.
    */
-  this.toggleGroupDetails = function() {
-    const groupDetails = $window.document.getElementById(
-      'group-details-out-of-scope'
-    );
-    groupDetails.classList.toggle('expanded');
-    $window.event.stopPropagation();
+  this.toggleGroupDetails = function(event, groupId) {
+    event.stopPropagation();
+    // Call the isGroupDetailsExpanded method so that if the groupId doesn't exist,
+    // it gets added before toggling it.
+    this.groupDetailsExpanded[groupId] = !this.isGroupDetailsExpanded(groupId);
   };
 
   /**
--- a/src/sidebar/components/test/group-list-test.js
+++ b/src/sidebar/components/test/group-list-test.js
@@ -19,6 +19,7 @@ describe('groupList', function() {
     name: 'Private',
     organization: groupFixtures.defaultOrganization(),
     type: 'private',
+    isScopedToUri: true,
   });
 
   const restrictedGroup = immutable({
@@ -29,6 +30,18 @@ describe('groupList', function() {
     name: 'Restricted',
     organization: groupFixtures.defaultOrganization(),
     type: 'restricted',
+    isScopedToUri: true,
+  });
+
+  const restrictedOutOfScopeGroup = immutable({
+    id: 'restrictedoos',
+    links: {
+      html: 'https://hypothes.is/groups/restricto',
+    },
+    name: 'Restricted',
+    organization: groupFixtures.defaultOrganization(),
+    type: 'restricted',
+    isScopedToUri: false,
   });
 
   const publicGroup = immutable({
@@ -39,6 +52,7 @@ describe('groupList', function() {
     name: 'Public',
     organization: groupFixtures.defaultOrganization(),
     type: 'open',
+    isScopedToUri: true,
   });
 
   let groups;
@@ -47,6 +61,7 @@ describe('groupList', function() {
   let fakeServiceUrl;
   let fakeSettings;
   let fakeFeatures;
+  let fakeStore;
 
   before(function() {
     angular
@@ -60,6 +75,12 @@ describe('groupList', function() {
   beforeEach(function() {
     fakeFeatures = {
       flagEnabled: sinon.stub().returns(false),
+    };
+
+    fakeStore = {
+      getCurrentlyViewingGroups: sinon.stub().returns([]),
+      getFeaturedGroups: sinon.stub().returns([]),
+      getMyGroups: sinon.stub().returns([]),
     };
 
     fakeAnalytics = {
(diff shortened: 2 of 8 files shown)


robertknight:
I know this is WIP, but I have a UX question. What is the plan for being able to access the group activity page from client in the modified design?

ajpeddakotla:
Hey @hmstepanek @robertknight can we have a quick sync to make sure we're on the same page of what's being implemented here? The designs are misleading because we're not implementing everything that was originally scoped. I think quick chat to iron things out is needed.

src/sidebar/components/group-list.js
```diff
@@ -66,6 +95,17 @@ function GroupListController(
     groups.focus(groupId);
   };
 
+  /**
+   * This is a quick and dirty hack to show styling, and needs refactoring.
+   */
+  this.toggleGroupDetails = function() {
```
robertknight: This expands the wrong group if there are multiple out-of-scope groups. Given this setup:

Screenshot 2019-03-14 14 02 57

Clicking on the "Test restricted group" will toggle the top one.

What you need to do instead is record the expanded state (eg. in a local variable in the `GroupListController` function) and then query that in the template to decide whether to add the `expanded` class to a particular entry.

lyzadanger:
Hi, @hmstepanek !

@robertknight and I chatted about this PR together today. It's looking really good. To streamline the review process, Rob is going to be on point for the actual review.

src/sidebar/components/test/group-list-test.js
```diff
@@ -363,6 +385,89 @@ describe('groupList', function() {
...
+          fakeFeatures.flagEnabled.withArgs('community_groups').returns(true);
+          // In order to show the group drop down there must be at least two groups.
+          groups = currentlyViewingGroups
+            .concat(featuredGroups)
+            .concat(myGroups);
+          fakeStore.getCurrentlyViewingGroups = sinon
+            .stub()
+            .returns(currentlyViewingGroups);
+          fakeStore.getFeaturedGroups = sinon.stub().returns(featuredGroups);
+          fakeStore.getMyGroups = sinon.stub().returns(myGroups);
```
robertknight: This is fine, but if something is already a stub you can just call `returns(...)` on the existing instance to change the result.

src/sidebar/store/modules/groups.js
```diff
@@ -120,42 +121,42 @@ function getGroup(state, id) {
...
-  );
-}
-
 /**
  * Return groups the logged in user is a member of.
  *
  * @return {Group[]}
  */
-function getMyGroups(state) {
+const getMyGroups = memoize(function(state) {
```
robertknight: These should be arrow functions as they don't access `this`. I'll look at adding a lint rule for that.

src/sidebar/templates/group-list.html
```diff
@@ -28,8 +28,10 @@
     !--><span class="group-list-label__label">{{vm.groups.focused().name}}</span>
   </div>
   <!-- Only build the drop down menu if showGroupsMenu is true. -->
-  <div class="dropdown-menu__top-arrow" ng-if="vm.showGroupsMenu()"></div>
-  <ul class="dropdown-menu pull-none" role="menu" ng-if="vm.showGroupsMenu()">
+
+  <!-- Original groups menu. -->
+  <div class="dropdown-menu__top-arrow" ng-if="vm.showGroupsMenu()  && !vm.isFeatureFlagEnabled('community_groups')"></div>
```
robertknight:
```suggestion
  <div class="dropdown-menu__top-arrow" ng-if="vm.showGroupsMenu() && !vm.isFeatureFlagEnabled('community_groups')"></div>
```

src/sidebar/templates/group-list.html
```diff
@@ -82,4 +84,132 @@
       </div>
     </li>
   </ul>
+
+  <!-- Show new menu if community_groups feature flag is enabled. -->
+  <div class="dropdown-menu" role="menu" ng-if="vm.showGroupsMenu() && vm.isFeatureFlagEnabled('community_groups')">
+
+    <!-- Display Currently Viewing. -->
+    <h2 class="dropdown-menu__section-heading" ng-show="vm.currentlyViewingGroupOrganizations().length > 0">Currently Viewing</h2>
+    <ul class="dropdown-menu__section" ng-show="vm.currentlyViewingGroupOrganizations().length > 0">
```
robertknight: Are each of these sections the same except for the input list of groups?

src/styles/sidebar/components/group-list.scss
```diff
@@ -21,6 +21,27 @@ $group-list-spacing-below: 50px;
     }
   }
 
+  .dropdown-menu__section--no-header {
+    border-top: solid 1px rgba(0, 0, 0, 0.15);
+
+    .group-details {
+      font-weight: 400;
+    }
+  }
```
robertknight: I see this CSS is attempting to fit into the existing structure. However, the existing structure doesn't follow our conventions well as it was written some time ago.

The preferred pattern is that each `styles/components/{component-name}.scss` file defines one "top level" class and its subsections and variants (each CSS rule is of the form `.{component-name}__{subcomponent}--{variant}` where the latter two parts are optional. This makes it easy to know what file a particular component is found in.

Each UI component (whether written in Angular or React) will then have one associated style component of the same name. There will also be style components that don't have a corresponding Angular component (because they are just components that just need styling and no complex markup or controller logic).

As long as all the new styles defined here are only used in the `<group-list>` component then I think we can leave this as-is and improve the situation when we come to migrate this to React.

src/styles/sidebar/components/group-list.scss
```diff
@@ -61,10 +114,63 @@ $group-list-spacing-below: 50px;
...
+
+      .group-details__unavailable-message,
+      .group-details__actions {
+        display: block;
+      }
+    }
+  }
+
+  .group-details__toggle {
+    font-size: 0.8em;
```
robertknight: I know that Dave Woolfe had differing opinions on this, but we generally avoid em-sizing in our current CSS and prefer specifying pixel sizes directly. em-sizes are a scaling of the font-size in the parent. The problem is that if you nest elements that define em-sizes they compound (eg. if you had an element with font-size 0.8em and a child also with font-size 0.8em then the actual font size in the child will be 0.8 * 0.8 * {inherited font size in parent}).

The final font size here is 12px, which happens to match the `$body1-font-size`, so we can just replace 0.8em with that.

src/styles/variables.scss
```diff
@@ -125,7 +125,7 @@ $highlight-color: rgba(255, 255, 60, 0.3);
 $highlight-color-second: rgba(206, 206, 60, 0.4);
 $highlight-color-third: rgba(192, 192, 49, 0.4);
 $highlight-color-focus: rgba(156, 230, 255, 0.5);
-$top-bar-height: 40px;
+$top-bar-height: 48px;
```
robertknight: It doesn't look to me like this change is an essential part of this PR. Since it also isn't feature-flagged, I would be inclined to leave this change out of this PR and propose it separately.

robertknight:
This is an example of the issue I mentioned when a group name is long:

Screenshot 2019-03-14 13 34 43

src/sidebar/templates/group-list.html
hmstepanek: I had a simillar thought. Yes except for My Groups which handles out of scope groups differently than the Currently Viewing section (There shouldn't be any out of scope groups in the Featured section.) I wasn't sure how to apply a loop over these neatly with that caveat. I suppose I could keep track of the actual list name and use an ng-if in there. Perhaps that's not _too_ complicated.

src/sidebar/store/modules/groups.js
hmstepanek: Ah sorry.

src/sidebar/templates/group-list.html
robertknight: I think it will be easier to do (and review) if the sub-list is broken out into a separate component, which can then be parametrized like a function. I would suggest to do that in a separate PR after this is merged though.

src/styles/variables.scss
hmstepanek: To answer your question, yes I believe it was done to make more clickable space but I'm not 100% sure as I can't fine an exact reason.

hmstepanek: I do think it emphasizes the heading of the sidebar and it looks a lot better with it but I'm happy to move it to another PR.
w/o it
image
w/ it
image

lyzadanger:
I've noticed a small detail in the screenshots here (which are looking more and more awesome). There's a fine horizontal line above `Featured Groups`. It looks a trifle odd.

hmstepanek:
@lyzadanger was just about to post a note about that in the frontend channel. Because I'm using ng-show here the css is still getting applied because ng-show actually renders the element still in the DOM so I think I'm gonna go back to using ng-if's.

hmstepanek:
@robertknight @lyzadanger I see two options for out of scope groups with super long group names that run over.
1) Remove the out of scope text that comes after the group name.
2) Make the groups menu wider to accommodate that text.

I'm kinda aiming for toward option 1. What do you think?

lyzadanger:
> 1. Remove the out of scope text that comes after the group name.

> 2. Make the groups menu wider to accommodate that text.
> 
> I'm kinda aiming for toward option 1. What do you think?
> image

You know...I'm looking at your screenshot here and wondering if we should remove that text IN ALL CASES. There are good affordances to get information as to why the group is not available. Is it really useful to use the 'out of scope' wording for users?

hmstepanek:
Yes, that's what I intended-remove it entirely not just for super long names. Apologies for not being clear. Sounds like we're on the same page here so I'll go ahead and make that change.

src/sidebar/components/test/group-list-test.js
```diff
@@ -363,6 +385,125 @@ describe('groupList', function() {
...
+    );
+  });
+  describe('group details expanded on out of scope groups', () => {
+    it('sets the default for the given groupid to false and returns it', () => {
+      const element = createGroupList();
+
+      const expanded = element.ctrl.isGroupDetailsExpanded('groupid');
+
+      assert.isFalse(expanded);
+      assert.isFalse(element.ctrl.groupDetailsExpanded.groupid);
```
robertknight: I would avoid testing internal properties in tests, as it creates more work if you later change the implementation but not the user-facing behaviour.

src/sidebar/components/test/group-list-test.js
```diff
@@ -363,6 +385,125 @@ describe('groupList', function() {
...
+      let fakeEvent = { stopPropagation: sinon.stub() };
+
+      element.ctrl.toggleGroupDetails(fakeEvent, 'groupid');
+      assert.isTrue(element.ctrl.groupDetailsExpanded.groupid);
+
+      element.ctrl.toggleGroupDetails(fakeEvent, 'groupid');
+      assert.isFalse(element.ctrl.groupDetailsExpanded.groupid);
+    });
+
+    it('stops the event from propogating when toggling', () => {
```
robertknight:
```suggestion
    it('stops the event from propagating when toggling', () => {
```

PR #980, opened 2019-02-25
Load polyfills only in browsers that need them

robertknight:
We deliver a lot of unnecessary code to modern browsers to polyfill functionality that they have natively built in. This PR refactors our polyfill bundles and the loading of them to avoid loading this code in new browsers.

The upshot is that this reduces the amount of minified JS that modern browsers have to download, parse and compile by ~245KB (~70KB gzipped). This is roughly equivalent in size to our largest dependencies: Angular (~220KB) and KaTeX (~230KB).

The changes are as follows:

1. The polyfills are grouped into sets, each of which has a `src/shared/polyfills/{set}.js` module to load it
2. There is an index file which exports a function `requiredPolyfillSets` that returns which sets of polyfills from an input list are needed by the current browser
3. The boot script uses this function to only add `<script>` tags for the required polyfills to the page

Additionally the "unicode" service which handles Unicode normalization when filtering annotations has been modified to use [String.prototype.normalize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) instead of unorm (unorm also provides a polyfill of this function).

----

An alternative to the approach taken here would be to use something like [Polyfill.io](https://polyfill.io/v3/) which generates polyfill bundles on-demand based on the User Agent string. This would however add an external service dependency and it would also mean changing some of the polyfills used because the service doesn't use exactly the same ones that we are already shipping. I thought it was probably best to stick with the ones we are already using.


gulpfile.js
```diff
@@ -117,9 +117,24 @@ const appBundles = [{
...
+  name: `polyfills-${set}`,
+  entry: `./src/shared/polyfills/${set}`,
+  transforms: ['babel'],
+}));
+
+const appBundleConfigs = appBundles
+  .concat(polyfillBundles)
+  .map(config => {
+    return Object.assign({}, appBundleBaseConfig, config);
+  });
```
robertknight: This creates a new `polyfills-{set}.bundle.js` JS bundle per group of polyfills. The polyfill sets are grouped by JS version, with larger polyfills and DOM polyfills separated out into their own bundles.

scripts/gulp/vendor-bundles.js
```diff
@@ -21,7 +20,6 @@ module.exports = {
     ],
     katex: ['katex'],
     showdown: ['showdown'],
-    unorm: ['unorm'],
```
robertknight: We no longer create the "unorm" bundle because the code goes into the bundle that polyfills String.prototype.normalize.

src/annotator/index.js
```diff
@@ -1,15 +1,6 @@
...
 const configFrom = require('./config/index');
-require('../shared/polyfills');
-
-// Polyfills
-
-// `document.evaluate` polyfill for IE 11.
-if (!window.document.evaluate) {
-  const wgxpath = require('wicked-good-xpath');
-  wgxpath.install();
-}
```
robertknight: This polyfill is now loaded by the boot script in `src/boot/boot.js`, if needed, before this code runs.

src/shared/polyfills/document.evaluate.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
+
+const wgxpath = require('wicked-good-xpath');
+wgxpath.install();
```
robertknight: There is no feature detection in these bundles because they are only loaded if they are needed in the first place.

src/sidebar/index.js
```diff
@@ -5,7 +5,6 @@ const disableOpenerForExternalLinks = require('./util/disable-opener-for-externa
 const { fetchConfig } = require('./util/fetch-config');
 const serviceConfig = require('./service-config');
 const crossOriginRPC = require('./cross-origin-rpc.js');
-require('../shared/polyfills');
```
robertknight: The boot script now loads any required polyfills before the sidebar app code runs.

src/boot/boot.js
```diff
@@ -79,14 +95,22 @@ function bootHypothesisClient(doc, config) {
  * Bootstrap the sidebar application which displays annotations.
  */
 function bootSidebarApp(doc, config) {
+  const polyfills = polyfillBundles([
```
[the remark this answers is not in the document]
robertknight: That's correct. Unicode normalization is used only when filtering annotations in the sidebar. An annotation's text (or other metadata) may use a different Unicode representation of the same characters than the user's filter query, so we need to normalize both to the same representation before matching.

PR #996, opened 2019-03-07
Remove angulartics dependency

### LINKED ISSUE #976: Replace Angulartics with alternate Google Analytics client
The client uses Google Analytics to track:

- Which delivery methods are being used (embed, bookmarklet, browser extension etc.)
- Which pages are embedding the client
- How frequently certain interactions happen

This integration currently uses the [Angulartics](http://angulartics.github.io) library and its Google Analytics integration. As part of migrating the client [away from AngularJS](https://github.com/hypothesis/client/issues/972), we need to replace this with something else. Given our usage of GA is relatively simple, I would be inclined to just use Google's [analytics.js](https://developers.google.com/analytics/devguides/collection/analyticsjs/) directly and create our own minimal wrapper around it.

robertknight:
As part of the migration away from AngularJS, replace Angulartics with
direct usage of the Google Analytics client.

Since we only use a single analytics provider (Google Analytics) and
only basic event tracking functionality in a single-route application,
its easiest just to replace the code with direct calls to the
analytics.js API.

See https://developers.google.com/analytics/devguides/collection/analyticsjs/events

Fixes #976

----

**Testing**

If you want to do end-to-end testing with this, you can use our staging Google Analytics account:

1. Run h with the `GOOGLE_ANALYTICS_CLIENT_TRACKING_ID` env var set to `UA-26026798-6` (this value is also available from the Google Analytics console in step 2)
2. Log in to https://analytics.google.com/ and select the Hypothesis account and Client Staging/Dev property from the menu at the top left corner of the screen. If you don't have access, please ask me.
3. Go to Reports => Real Time => Overview
4. Run the client and do stuff. Check that the active user count in Google Analytics is incremented and your rough location (to about a state/region level) is shown on the map, depending on the accuracy of IP geolocation. If you go to the Real Time => Events view and do things like switching groups or creating annotations in the client, you should also see events reported.

### DIFF
--- a/src/sidebar/services/analytics.js
+++ b/src/sidebar/services/analytics.js
@@ -2,13 +2,44 @@
 
 const VIA_REFERRER = /^https:\/\/(qa-)?via.hypothes.is\//;
 
-const globalGAOptions = function(win, settings) {
-  settings = settings || {};
+const events = {
+  ANNOTATION_CREATED: 'annotationCreated',
+  ANNOTATION_DELETED: 'annotationDeleted',
+  ANNOTATION_FLAGGED: 'annotationFlagged',
+  ANNOTATION_SHARED: 'annotationShared',
+  ANNOTATION_UPDATED: 'annotationUpdated',
+  DOCUMENT_SHARED: 'documentShared',
+  GROUP_LEAVE: 'groupLeave',
+  GROUP_SWITCH: 'groupSwitch',
+  GROUP_VIEW_ACTIVITY: 'groupViewActivity',
+  HIGHLIGHT_CREATED: 'highlightCreated',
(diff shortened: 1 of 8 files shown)


src/sidebar/services/analytics.js
```diff
@@ -2,38 +2,108 @@
 
 const VIA_REFERRER = /^https:\/\/(qa-)?via.hypothes.is\//;
 
-const globalGAOptions = function(win, settings) {
-  settings = settings || {};
-
-  const globalOpts = {
-    category: '',
-  };
+const events = {
```
robertknight: The `events` mapping was hoisted up from inside a function to be a module-level constant.

PR #1018, opened 2019-03-21
Refactor new groups menu

hmstepanek:
Breakup the new groups menu (currently in the groups-list) into a group-section, a group, and a group-out-of-scope.


src/sidebar/templates/group-section.html
```diff
@@ -0,0 +1,13 @@
+<h2 class="heading" ng-if="vm.sectionGroups.length > 0">
+  {{ vm.heading }}
+</h2>
+<ul class="content" ng-if="vm.sectionGroups.length > 0">
+  <li class="dropdown-community-groups-menu__row dropdown-menu__row--unpadded"
+      ng-repeat="group in vm.sectionGroups track by group.id">
+
+    <group class="group" group="group" ng-if="vm.isSelectable(group.id)"></group>
```
robertknight: In Angular, an actual DOM element gets created for custom components like `<group>`, so the `class` property will have an effect. In React however this is not the case. The component tree is separate from the "rendered tree" of DOM elements. Therefore it will make the migration easier if classes are only applied to DOM elements (the top-level element inside `group.html` in this case) rather than to components. The exception is if for some reason you need to style different uses of the component with different classes.

src/styles/sidebar/components/group-section.scss
```diff
@@ -0,0 +1,47 @@
+/* The groups section. */
+
+.group-section {
+  .dropdown-menu {
+    width: $group-list-width;
+    max-height: 500px;  // fallback for browsers lacking support for vh/calc
```
robertknight: All of the browsers we support have support for both `calc` and viewport units. See https://caniuse.com/#search=calc and https://caniuse.com/#search=viewport%20units, so we might not need the fallback here. There are mentions of bugs in some older browsers though. I haven't looked in detail to see if that affects us.

src/styles/sidebar/sidebar.scss
```diff
@@ -15,6 +15,9 @@ $base-line-height: 20px;
 @import './components/dropdown-menu-btn';
 @import './components/excerpt';
 @import './components/group-list';
+@import './components/group-section';
+@import './components/group';
+@import './components/group-out-of-scope';
```
robertknight: A small request - could you keep this list sorted please? It makes it more obvious where to slot new values into place in future.

src/sidebar/components/group-out-of-scope.js
```diff
@@ -0,0 +1,62 @@
...
+  // Track whether the group details are expanded.
+  this.isDetailsExpanded = false;
+
+  this.leaveGroup = function() {
+    const groupName = this.group.name;
+    const message =
+      'Are you sure you want to leave the group "' + groupName + '"?';
+    if ($window.confirm(message)) {
+      analytics.track(analytics.events.GROUP_LEAVE);
+      store.groups.leave(this.group.id);
```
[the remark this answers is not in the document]
hmstepanek: I'm thinking I'm just going to remove this from here entirely for now since it's not used currently and not clear whether it will be needed.

src/sidebar/templates/group-section.html
hmstepanek: Yeah I didn't want to have to style it but it was causing a problem (shrinking the group row). It's styling is the same as the inner component so I expect when we move to React that styling will no longer be needed. If you can see a way to avoid it lmk (my css foo is somewhat limited).

src/styles/sidebar/components/group-list-out-of-scope-item.scss
```diff
@@ -0,0 +1,46 @@
+/* The group-list out of scope item. */
+@import "./group-list-item";
+@import "./group-list";
```
robertknight: `@import` in SASS causes a file to be included _each_ time it is used because it is a crude textual inclusion. This will cause the `group-list` CSS to be included twice in the final bundle.

The current rule of thumb is that things that will be used in multiple components (variables, mixins, functions) need to be put in separate files. Currently we have a single [variables.scss file](https://github.com/hypothesis/client/blob/d0ed54a5b3bbc45e6ccca67f416e5517a5fd2674/src/styles/variables.scss#L123) for this. We could also create a more narrowly scoped one in `src/styles/sidebar/variables.scss` in future.

There is some possibility that SASS might actually get a [proper module system](http://sass.logdown.com/posts/502818-request-for-comments-module-system-proposal) later this year, but that is still being worked on.

src/styles/sidebar/components/group-list-item.scss
```diff
@@ -0,0 +1,89 @@
+/* The group. */
+
+.group-list-item {
+  display: flex;
+  flex-direction: row;
+  flex-grow: 1;
+
+  .item {
```
robertknight: I forgot to point this out before, but our [CSS style guide](https://github.com/hypothesis/frontend-toolkit/blob/master/docs/css-style-guide.md) recommends avoiding descendant selectors  in most cases (created by nesting in SASS) and instead use `.component__element` names for sub-elements (so. `group-list-item__item` in this case). The advantage is that you can tell more easily what the final styles are on an element just by looking at the element itself, without needing to understand the surrounding context.

In this context, it avoids the possibility that another file could define rules for an `item` class and end up influencing the result here.

src/sidebar/templates/group-list-section.html
```diff
@@ -0,0 +1,13 @@
+<h2 class="heading">
+  {{ vm.heading }}
+</h2>
+<ul class="content">
+  <li class="dropdown-menu__row dropdown-menu__row--unpadded"
+      ng-repeat="group in vm.sectionGroups track by group.id">
+
+    <group-list-item class="group-list-item" group="group" ng-if="vm.isSelectable(group.id)"></group-list-item>
+
+    <group-list-out-of-scope-item class="group-list-out-of-scope-item" group="group" ng-if="!vm.isSelectable(group.id)"></group-list-out-of-scope-item>
```
robertknight: Small request, can you format this onto multiple lines so that it is easier to review. For new template files, you can use the same [Prettier](https://prettier.io) formatter that we use for JavaScript for HTML files. If you have the Vim plugin installed, this can be done with `:Prettier`. Unlike JS, we haven't done a project-wide formatting using Prettier for HTML/CSS files yet.

src/styles/sidebar/components/group-list-out-of-scope-item.scss
hmstepanek: This is one of the minor tweaks I wanted to make after yesterday.

src/sidebar/util/test/group-list-item-common-test.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+const proxyquire = require('proxyquire');
+
+describe('sidebar.util.groupListItemCommon', () => {
```
robertknight: I forgot to mention this before, we decided to use slash-separated paths in top-level `describe` blocks going forwards, since that is what import paths use in JS. The Python-style period separators was a weird misstep that some files still have.

src/styles/sidebar/components/group-list-item-out-of-scope.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  &.expanded {
+    .group-list-item-out-of-scope__details--toggle {
+      display: none;
+    }
+
+    .group-list-item-out-of-scope__details--unavailable-message,
+    .group-list-item-out-of-scope__details--actions {
+      display: block;
```
robertknight: Rather than hiding and showing these elements with CSS, I think it will be clearer to hide/show them in the template using `ng-if` or `ng-show` for consistency, unless there is some reason that I missed why it is difficult to do in this context.

src/styles/sidebar/components/group-list-section.scss
```diff
@@ -0,0 +1,18 @@
+/* The groups section. */
+
+.dropdown-menu__row {
```
robertknight: This should go in the file for the `dropdown-menu` component, assuming we have one.

src/styles/sidebar/components/group-list-section.scss
```diff
@@ -0,0 +1,18 @@
...
+  border: none;
+}
+
+.group-list-section__content {
+  border-bottom: solid 1px rgba(0, 0, 0, 0.15);
+}
+
+.group-list-section__heading {
+  color: $gray-light;
+  font-size: 12px;
```
robertknight: This should ideally use a standard font size from our palette.

src/styles/sidebar/components/group-list.scss
```diff
@@ -22,26 +21,13 @@ $group-list-spacing-below: 50px;
   }
 
   .dropdown-menu__section--no-header {
-    border-top: solid 1px rgba(0, 0, 0, 0.15);
+    //border-top: solid 1px rgba(0, 0, 0, 0.15);
```
robertknight: Commented-out code. Just remove it?

src/sidebar/templates/group-list-item-out-of-scope.html
```diff
@@ -0,0 +1,55 @@
...
+      <path
+        d="M11 15h2v2h-2zm0-8h2v6h-2zm.99-5C6.47 2 2 6.48 2 12s4.47 10 9.99 10C17.52 22 22 17.52 22 12S17.52 2 11.99 2zM12 20c-4.42 0-8-3.58-8-8s3.58-8 8-8 8 3.58 8 8-3.58 8-8 8z"
+      />
+    </svg>
+    <a
+      class="group-list-item-out-of-scope__name-link"
+      href=""
+      title="Group not annotatable on this domain."
+    >
+      {{vm.group.name}} </a
```
robertknight: The redness of the group name looks like a mistake to me, is it?

out-of-scope-group-link

I also think it is surprising here that the group name is a link, even though it doesn't visually look like one, whereas the "Why is this group unavailable?" message looks like a link but isn't.

src/sidebar/components/group-list-section.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+// @ngInject
+function GroupListSectionController() {
+  this.isSelectable = function(group_id) {
```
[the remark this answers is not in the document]
hmstepanek: Switching between languages is hard sometimes. lol

src/sidebar/components/group-list-section.js
```diff
@@ -0,0 +1,20 @@
...
+  };
+}
+
+module.exports = {
+  controller: GroupListSectionController,
+  controllerAs: 'vm',
+  bindings: {
+    sectionGroups: '<',
+    heading: '<',
+    disableOosGroupSelection: '<',
```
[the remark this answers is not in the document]
hmstepanek: I wasn't sure if there was an official way to document module exports? I just put a comment above each var. lmk if you had something else in mind.

src/sidebar/components/test/group-list-item-test.js
```diff
@@ -0,0 +1,111 @@
...
+    });
+  };
+
+  it('changes the focused group when group is clicked', () => {
+    const fakeGroup = { id: 'groupid' };
+
+    const element = createGroupListItem(fakeGroup);
+    const group = element.find('.group-list-item__item');
+
+    // click the second group
```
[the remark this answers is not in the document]
hmstepanek: Ah left over comment from porting the test. I fix...

src/sidebar/components/test/group-list-section-test.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return util.createDirective(document, 'groupListSection', config);
+  };
+
+  describe('isSelectable', () => {
+    [
+      {
+        description: 'always returns true if disableOosGroupSelection is false',
+        fakeDisableOosGroupSelection: false,
+        expectedIsSelectable: [true, true],
```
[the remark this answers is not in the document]
hmstepanek: The truth table is against fakeDisableOosGroupSelection and group.isScopedToUri. Note each of these cases tests two groups; [{ isScopedToUri}, { isScopedToUri: false}]. That should cover all the cases. I could break them up into cases that only test one type of group at a time if you think it's confusing as it's currently written?

src/styles/sidebar/components/group-list-section.scss
hmstepanek: I added a dropdown-menu__row--no-border in common with the other dropdown-menu__row css classes.

src/sidebar/components/group-list-section.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+// @ngInject
+function GroupListSectionController() {
+  this.isSelectable = function(group_id) {
+    const group = this.sectionGroups.find(g => g.id === group_id);
+    return !this.disableOosGroupSelection || group.isScopedToUri;
```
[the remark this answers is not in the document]
hmstepanek: The groupId is computed from the sectionGroups so unless there's some kind of race condition where the sectionGroups gets updated and drops the group but the old groupid call still happens or vise versa no it's not possible. I don't know enough about the inner workings of angularjs's update process to know whether that's plausible or not? Does that seem like a thing that we should guard against?

src/sidebar/templates/group-list-item-out-of-scope.html
hmstepanek: You can click anywhere on the group area to expand or select a group. I think the pointer does a nice job at indicating that when you hover over the space (although I realize that isn't a mobile accessible thing). I think the -Why is this group unavailable?- text is made to look more link like so the user knows the group is "clickable" even though the group is greyed out/visually not selectable. I'm not really that great at UI stuff so it's possible this could use some changes but it seems like something that should be done in a larger forum. All that being said I rather not make any changes to the styling of this PR since it's just a refactor. (The red name change was unintentional thank you for catching that!)

src/sidebar/templates/group-list-item.html
```diff
@@ -0,0 +1,25 @@
+<div
+  ng-class="{'group-list-item__item': true, 'is-selected': vm.isSelected()}"
+  ng-click="vm.focusGroup()"
+  tabindex="0"
+>
+  <!-- the group icon !-->
+  <div class="group-list-item__icon--container">
```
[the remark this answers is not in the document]
hmstepanek: I assume you are talking about making sure the parent div's styling doesn't impact this styling? There isn't any strict mode/etc set on this component. Not sure how to do that? I found [this](This feature isn't supported on all our browsers yet. https://developer.mozilla.org/en-US/docs/Web/CSS/contain) but it's not supported on all browser's yet.

src/sidebar/components/group-list-section.js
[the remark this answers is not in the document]
robertknight: > . I don't know enough about the inner workings of angularjs's update process to know whether that's plausible or not?

Provided that this logic is only called synchronously (ie. not in a timeout, `.then()` handler of a Promise or other callback which gets invoked asynchronously) then you're fine.

src/sidebar/components/group-list-section.js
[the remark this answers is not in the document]
robertknight: Just a comment as you've done here is great. When we switch to React we'll use the [prop-types](https://reactjs.org/docs/typechecking-with-proptypes.html) library to document component inputs.

src/sidebar/components/test/group-list-section-test.js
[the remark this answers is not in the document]
robertknight: Oh right, sorry. I see. I was looking at an _output_ and thinking it was an input. You're testing three states of `fakeDisableOosGroupSelection`, so that's fine.

src/sidebar/components/test/group-list-test.js
```diff
@@ -85,11 +75,7 @@ describe('groupList', function() {
 
     fakeAnalytics = {
       track: sinon.stub(),
-      events: {
-        GROUP_LEAVE: 'groupLeave',
-        GROUP_SWITCH: 'groupSwitch',
-        GROUP_VIEW_ACTIVITY: 'groupViewActivity',
-      },
+      events: events,
```
robertknight: This is fine, but idiomatically if the key and value are the same it can be shortened to just the key name (ie. `fakeAnalytics = { track: sinon.stub(), events }`

src/sidebar/components/test/group-list-test.js
```diff
@@ -585,14 +517,13 @@ describe('groupList', function() {
       const showGroupsMenu = element.ctrl.showGroupsMenu();
       const dropdownToggle = element.find('.dropdown-toggle');
       const arrowIcon = element.find('.h-icon-arrow-drop-down');
-      const dropdownOptions = element.find(
-        '.dropdown-community-groups-menu__row'
-      );
+      // const dropdownMenu = element.find('class[dropdown-menu]');
```
robertknight: Commented out code here.

src/sidebar/templates/group-list-item-out-of-scope.html
robertknight: OK. Let's just get the red-ness fixed and we can solicit feedback from others.

src/styles/sidebar/common.scss
```diff
@@ -260,6 +260,10 @@ html {
   min-width: 120px;
 }
 
+.dropdown-menu__row--no-border{
+  border: none;
+}
```
robertknight: 👍

src/styles/sidebar/components/group-list-item-out-of-scope.scss
```diff
@@ -0,0 +1,50 @@
...
+  width: auto;
+}
+
+.group-list-item-out-of-scope__details {
+  flex-grow: 1;
+  flex-shrink: 1;
+  font-weight: 500;
+}
+
+.group-list-item-out-of-scope__details--toggle {
```
robertknight: It feels to me like the `--toggle`, `--unavailable-message`, `--actions` and `--group-page-link` "modifiers" are really subcomponents rather than variants/modifiers, so they should be eg. `...__details-toggle`, `...-details-actions`.

The naming scheme we use is a variation of the classic [Block Element Modifier](http://getbem.com/introduction/) scheme. See that page for some more examples.

hmstepanek: Ok yeah, that makes sense.

hmstepanek:
Not seeing an issue with the red text anymore on my end. Can you verify it's fixed now?

robertknight:
> Not seeing an issue with the red text anymore on my end. Can you verify it's fixed now?

Yes, I've just rebuilt and I don't see this issue any more.

PR #1019, opened 2019-03-21
Enable conditional inclusion of code based on NODE_ENV checks

robertknight:
Packages in the React ecosystem often include debugging checks guarded
by:

```
if (process.env.NODE_ENV === 'development') { ... }
```

OR

```
if (process.env.NODE_ENV !== 'production') { ... }
```

This commit uses loose-envify to replace the `process.env.$VAR`
expression with a literal string value when building JS bundles. This enables
debugging checks in dev builds and causes the minifier to remove the
code entirely in production builds.

Some packages don't need this because their package.json file includes a
"browserify" key which already configures this transform, however not
all of them do. Therefore the transform is configured to run globally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "katex": "^0.10.0",
     "lodash.debounce": "^4.0.3",
     "lodash.get": "^4.3.0",
+    "loose-envify": "^1.4.0",
     "mkdirp": "^0.5.1",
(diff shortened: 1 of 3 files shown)


robertknight:
Note this doesn't affect any of our current dependencies, but it will affect some that I'm looking to add soon.

PR #1029, opened 2019-03-29
Convert `<group-list-item-out-of-scope>` to React

robertknight:
I'm doing this now before making any changes to the component relating to going to the group activity page or leaving the group. Also, it's a good simple example of a stateful component (it remembers whether it is expanded or collapsed between renders).

There should be no aesthetic or functional changes to how it works here.

The tests have been modified to work in terms of the UI instead of
internal controller methods, since there is no controller instance any
more.


src/sidebar/components/group-list-item-out-of-scope.js
```diff
@@ -1,37 +1,106 @@
 'use strict';
 
+const classnames = require('classnames');
+const { Fragment, createElement } = require('preact');
+const { useState } = require('preact/hooks');
```
robertknight: `useState` is one way of persisting local state between renders. Here it is used to track the expanded/collapsed state. [See docs](https://reactjs.org/docs/hooks-state.html).

src/sidebar/components/test/group-list-item-out-of-scope-test.js
```diff
@@ -1,106 +1,98 @@
...
-      '../group-list-item-out-of-scope',
-      {
-        '../util/group-list-item-common': fakeGroupListItemCommon,
-        '@noCallThru': true,
-      }
-    );
+    GroupListItemOutOfScope = proxyquire('../group-list-item-out-of-scope', {
+      // Use same instance of Preact module in tests and mocked module.
+      // See https://robertknight.me.uk/posts/browserify-dependency-mocking/
+      preact,
```
robertknight: The need to pass through `preact` here is due to a rather confusing aspect of proxyquire and the reason I think we should replace it with some other means of mocking. The issue is not specific to Preact/React, but it comes up with any library that relies on certain `instanceof` checks or global state internally. This included Angular in the past.

src/sidebar/components/group-list-item-out-of-scope.js
```diff
@@ -1,37 +1,106 @@
...
+            className="group-list-item__icon group-list-item__icon--organization"
+            alt={groupOrgName}
+            src={group.logo}
+          />
+        )}
+      </div>
+      {/* the group name */}
+      <div
+        className={classnames({
+          'group-list-item-out-of-scope__details': true,
```
lyzadanger: I'm glad that care was put on BEM class naming here.

src/sidebar/components/test/group-list-item-out-of-scope-test.js
lyzadanger: Hmmm, this _is_ a bit fiddly, eh? Is there any way to put this in another layer in the long term? Or does it absolutely need to be done locally in each test module?

src/sidebar/components/test/group-list-item-out-of-scope-test.js
```diff
@@ -1,106 +1,98 @@
...
-      group: fakeGroup,
-    });
+    return mount(
+      <GroupListItemOutOfScope analytics={fakeAnalytics} group={fakeGroup} />
+    );
   };
 
-  it('calls groupListItemCommon.trackViewGroupActivity when trackViewGroupActivity is called', () => {
-    const fakeGroup = { id: 'groupid' };
+  it('calls trackViewGroupActivity when "Go to group page" link is clicked', () => {
```
lyzadanger: This seems like duplicate functionality with the `group-list-item` component, yes? Perhaps on the list for a down-the-road refactor?

src/sidebar/components/group-list-item-out-of-scope.js
```diff
@@ -1,37 +1,106 @@
...
+          expanded: isExpanded,
+        })}
+      >
+        <svg
```
lyzadanger: Might be nice to have some sort commenting somewhere, if possible, about what this SVG _is_

src/sidebar/components/group-list-item-out-of-scope.js
```diff
@@ -1,37 +1,106 @@
...
+          </p>
+        )}
+        {isExpanded && (
+          <Fragment>
+            <p className="group-list-item-out-of-scope__details-unavailable-message">
+              This group has been restricted to selected URLs by its
+              administrators.
+            </p>
+            <p className="group-list-item-out-of-scope__details-actions">
+              <a
```
lyzadanger: Looks like we need to add some conditional logic here to prevent this link from showing up for third-party groups, as most restricted/open groups are third-party at this time, and there is no activity page to go to (yet).

src/sidebar/components/group-list-item-out-of-scope.js
robertknight: Yes. Something I'm not sure about at the moment is whether there will be use cases for which we'll still need to use classes or are better done with classes. This is something we'll need to figure out as we go.

src/sidebar/components/test/group-list-item-out-of-scope-test.js
robertknight: My preferred solution is to replace Proxyquire with a different method of mocking that doesn't have this problem. That's on my plate to do in the near future.

src/sidebar/components/group-list-item-out-of-scope.js
```diff
@@ -1,37 +1,106 @@
...
   orgName,
   trackViewGroupActivity,
 } = require('../util/group-list-item-common');
 
-// @ngInject
-function GroupListItemOutOfScopeController(analytics) {
-  // Track whether the group details are expanded.
-  this.isDetailsExpanded = false;
+function GroupListItemOutOfScope({ analytics, group }) {
+  const [isExpanded, setExpanded] = useState(false);
```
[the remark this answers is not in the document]
robertknight: Not that I can think of at the moment.

src/sidebar/components/group-list-item-out-of-scope.js
```diff
@@ -1,37 +1,106 @@
...
+  const trackViewActivity = event => {
+    event.stopPropagation();
     trackViewGroupActivity(analytics);
   };
+
+  return (
+    <div
+      className="group-list-item__item group-list-item-out-of-scope__item"
+      onClick={toggleGroupDetails}
+      tabIndex="0"
```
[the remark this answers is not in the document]
robertknight: I think this came from Dave Woolfe's initial redesign of the menu. The end goal is that the sidebar app should be keyboard navigable and tabindex should be used where appropriate to that end. I haven't closely examined the overall state of the current groups menu in that respect.

src/sidebar/components/group-list-item-out-of-scope.js
```diff
@@ -1,37 +1,106 @@
...
+        >
+          <path fill="none" d="M0 0h24v24H0V0z" />
+          <path d="M11 15h2v2h-2zm0-8h2v6h-2zm.99-5C6.47 2 2 6.48 2 12s4.47 10 9.99 10C17.52 22 22 17.52 22 12S17.52 2 11.99 2zM12 20c-4.42 0-8-3.58-8-8s3.58-8 8-8 8 3.58 8 8-3.58 8-8 8z" />
+        </svg>
+        <a
+          className="group-list-item__name-link"
+          href=""
+          title="Group not annotatable on this domain."
+        >
+          {group.name}{' '}
```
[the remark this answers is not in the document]
robertknight: It creates a space in the rendered HTML output - literal strings become text nodes. Unlike HTML there is no implicit whitespace created at the end of a line. I'll double check whether this is actually necessary though.

src/sidebar/components/test/group-list-item-out-of-scope-test.js
robertknight: Yes, this will most likely change quite soon, pending the ongoing design exploration around this menu in the #design channel.

src/sidebar/components/group-list-item-out-of-scope.js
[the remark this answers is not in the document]
robertknight: Update: I'm pretty sure this is unnecessary since the link is immediately followed by `<br/>`.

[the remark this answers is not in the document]
robertknight: Dave Woolfe pointed out to me that whether we should use `title` attributes at all is [something to perhaps reconsider](https://ux.stackexchange.com/a/19922), since not all user agents display tooltips. We didn't come up with a firm recommendation yet so I've left it as-is.

src/sidebar/components/group-list-item-out-of-scope.js
robertknight: Done. I've extracted this into a separate, named, constant.

src/sidebar/components/group-list-item-out-of-scope.js
robertknight: Done. This link is no longer rendered if the group has no HTML link.

PR #1036, opened 2019-04-02
Add annotations:group:GROUP_ID fragment url

hmstepanek:
Add ability to link to a group in the client. This is a partial fix for https://github.com/hypothesis/product-backlog/issues/987 Note this does not include handling for groups that the user doesn't have permission to, don't exist, or aren't viewable on the page-that will be done in a follow up PR.

This should be merged after https://github.com/hypothesis/client/pull/1060. As it is built on top of that branch.

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -23,6 +23,7 @@ function configFrom(window_) {
     enableExperimentalNewNoteButton: settings.hostPageSetting(
       'enableExperimentalNewNoteButton'
     ),
+    group: settings.group,
     theme: settings.hostPageSetting('theme'),
     usernameUrl: settings.hostPageSetting('usernameUrl'),
     onLayoutChange: settings.hostPageSetting('onLayoutChange'),
--- a/src/annotator/config/settings.js
+++ b/src/annotator/config/settings.js
@@ -102,6 +102,28 @@ function settingsFrom(window_) {
     return jsonConfigs.annotations || annotationsFromURL();
   }
 
+  /**
+   * Return the `#annotations:group:*` ID from the given URL's fragment.
+   *
+   * If the URL contains a `#annotations:group:<GROUP_ID>` fragment then return
+   * the group ID extracted from the fragment. Otherwise return `null`.
+   *
+   * @return {string|null} - The extracted ID, or null.
+   */
+  function group() {
+    function groupFromURL() {
+      const groupFragmentMatch = window_.location.href.match(
+        /#annotations:group:([A-Za-z0-9_-]+)$/
+      );
+      if (groupFragmentMatch) {
+        return groupFragmentMatch[1];
+      }
+      return null;
+    }
+
+    return jsonConfigs.group || groupFromURL();
+  }
+
   function showHighlights() {
     let showHighlights_ = hostPageSetting('showHighlights');
 
@@ -178,6 +200,9 @@ function settingsFrom(window_) {
     },
     get clientUrl() {
       return clientUrl();
+    },
+    get group() {
+      return group();
     },
     get showHighlights() {
       return showHighlights();
--- a/src/annotator/config/test/settings-test.js
+++ b/src/annotator/config/test/settings-test.js
@@ -289,6 +289,29 @@ describe('annotator.config.settingsFrom', function() {
     });
   });
 
+  [
+    {
+      description:
+        "returns an object with the group ID when there's a valid #annotations:group:<ID> fragment",
+      url: 'http://localhost:3000#annotations:group:alphanum3ric_-only',
+      returns: 'alphanum3ric_-only',
+    },
+    {
+      description: "returns null when there's a non-alphanumeric group ID",
+      url: 'http://localhost:3000#annotations:group:not%20alphanumeric',
+      returns: null,
+    },
+    {
+      description: "return null when there's an empty group ID",
+      url: 'http://localhost:3000#annotations:group:',
+      returns: null,
+    },
+  ].forEach(test => {
+    it(test.description, () => {
+      assert.deepEqual(settingsFrom(fakeWindow(test.url)).group, test.returns);
+    });
+  });
+
   describe('#query', function() {
     context(
       'when the host page has a js-hypothesis-config with a query setting',
--- a/src/sidebar/host-config.js
+++ b/src/sidebar/host-config.js
@@ -17,6 +17,9 @@ function hostPageConfig(window) {
     // Direct-linked annotation ID
     'annotations',
 
+    // Direct-linked group ID
+    'group',
+
     // Default query passed by url
     'query',
 
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -59,9 +59,15 @@ function groups(
    * @param {Group[]} groups
    * @param {boolean} isLoggedIn
    * @param {string|null} directLinkedAnnotationId
+   * @param {string|null} directLinkedGroupId
    * @return {Promise<Group[]>}
    */
-  function filterGroups(groups, isLoggedIn, directLinkedAnnotationId) {
+  function filterGroups(
+    groups,
+    isLoggedIn,
+    directLinkedAnnotationId,
+    directLinkedGroupId
+  ) {
     // If service groups are specified only return those.
     // If a service group doesn't exist in the list of groups don't return it.
     if (svc && svc.groups) {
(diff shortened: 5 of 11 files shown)


src/sidebar/services/groups.js
```diff
@@ -190,6 +219,11 @@ function groups(
         store.loadGroups(groups);
         if (isFirstLoad && groups.some(g => g.id === prevFocusedGroup)) {
           store.focusGroup(prevFocusedGroup);
+        } else if (
+          isFirstLoad &&
+          groups.some(g => g.id === directLinkedGroup)
+        ) {
+          store.focusGroup(directLinkedGroup);
```
robertknight: This condition should take precedence over the one above, otherwise any user who has used the client at least once before in the current browser, and is still a member of the last group they selected, will see that group selected instead of the direct-linked one.

src/annotator/config/index.js
```diff
@@ -23,6 +23,7 @@ function configFrom(window_) {
     enableExperimentalNewNoteButton: settings.hostPageSetting(
       'enableExperimentalNewNoteButton'
     ),
+    group: settings.group,
```
robertknight: Once we've done a bit more testing and we're happy with this config setting, we'll need to document it. Would you mind adding an item in the relevant GitHub issue so this doesn't get forgotten.

src/annotator/config/settings.js
```diff
@@ -102,6 +102,29 @@ function settingsFrom(window_) {
...
+  /**
+   * Return the `#annotations:group:*` ID from the given URL's fragment.
+   *
+   * If the URL contains a `#annotations:group:<GROUP_ID>` fragment then return
+   * the group ID extracted from the fragment. Otherwise return `null`.
+   *
+   * @return {string|null} - The extracted ID, or null.
+   */
+  function group() {
+    /** Return the group id from the URL, or null. */
```
robertknight: I think this comment here is redundant, given the JSDoc above.

src/annotator/config/test/settings-test.js
```diff
@@ -289,6 +289,29 @@ describe('annotator.config.settingsFrom', function() {
     });
   });
 
+  [
+    {
+      description:
+        "returns an object with the group ID when there's a valid #annotations:group:<ID> fragment",
+      url: 'http://localhost:3000#annotations:group:alphanum3ric_-only',
+      returns: 'alphanum3ric_-only',
```
robertknight: Strictly speaking the character set is "url-safe base 64" (see comments in `annotationsFromURL`), but I think this is fine.

src/sidebar/services/groups.js
```diff
@@ -60,7 +60,12 @@ function groups(
    * @param {string|null} directLinkedAnnotationId
    * @return {Promise<Group[]>}
    */
-  function filterGroups(groups, isLoggedIn, directLinkedAnnotationId) {
+  function filterGroups(
+    groups,
+    isLoggedIn,
+    directLinkedAnnotationId,
+    directLinkedGroupId
```
robertknight: The `directLinkedGroupId` param here is missing from the JSDoc comment above.

src/sidebar/services/groups.js
```diff
@@ -141,6 +146,7 @@ function groups(
     if (isSidebar) {
       uri = getDocumentUriForGroupSearch();
     }
+    let directLinkedGroup = settings.group;
```
robertknight: It doesn't look like this var is re-assigned later, in which case it should be declared with `const`.

src/sidebar/services/test/groups-test.js
```diff
@@ -176,6 +177,26 @@ describe('groups', function() {
...
+      const svc = service();
+
+      fakeSettings.group = 'selected-id';
+      const groups = [
+        { id: 'groupa', name: 'GroupA' },
+        { id: 'groupb', name: 'GroupB' },
+        { id: fakeSettings.group, name: 'Selected Group' },
+      ];
+
+      fakeApi.profile.groups.read.returns(Promise.resolve(groups));
```
robertknight: Since this call returns all three groups here, it looks like this test would pass even if the selected group was not set.

src/sidebar/services/test/groups-test.js
```diff
@@ -209,6 +231,10 @@ describe('groups', function() {
           fakeApi.groups.list,
           sinon.match({ expand: ['organization', 'scopes'] })
         );
+        assert.calledWith(
+          fakeApi.group.read,
+          sinon.match({ expand: ['organization', 'scopes'] })
```
robertknight: It doesn't look like there is a test in this file to check that the right group ID is passed to `api.group.read`.

src/sidebar/util/groups.js
```diff
@@ -35,10 +35,10 @@ function combineGroups(userGroups, featuredGroups, uri) {
 }
 
 function isScopedToUri(group, uri) {
-  // If the group has scope info, the scoping is enforced,
-  // and the uri patterns don't include this page's uri
-  // the group is not selectable, otherwise it is.
-  if (group.scopes && group.scopes.enforced) {
+  // If the group has scope info and the uri patterns
+  // don't include this page's uri the group is not
+  // selectable, otherwise it is.
```
robertknight: It is still the case that the group is selectable if the URI patterns don't match, if scoping is not enforced. I don't see anything in this comment which isn't obvious from the code below any more, so perhaps just remove it.

src/sidebar/util/test/groups-test.js
```diff
@@ -102,14 +102,14 @@ describe('sidebar.util.groups', () => {
...
+        isScopedToUri: true,
         uri: 'https://foo.com/bar',
       },
       {
         description:
-          'sets `isScopedToUri` to true if `scopes.enforced` is false',
-        scopes: { enforced: false },
-        shouldBeSelectable: true,
+          'sets `isScopedToUri` to false if `scopes.enforced` is false and patterns do not match',
+        scopes: { enforced: false, uri_patterns: ['http://bar.com*'] },
```
robertknight: Is this test still relevant now that `isScopedToUri`'s return value no longer depends on the `enforced` property?

src/sidebar/util/groups.js
robertknight: Can you remind me what the reason for removing the `scopes.enforced` check here is?

hmstepanek: Ah sorry I should have rephrased that last bit but point taken on the "this is a repeat of what the code is doing". isScopedToUri previously was the same as selectable, now it is simply whether the group is within scope. The enforced check was removed since we need to be able to distinguish groups that are in scope from groups that are out of scope but not enforced. Groups that are out of scope but not enforced when we link to them should show up in "Currently Viewing" and groups that are in scope should show up in "Featured Groups".

src/sidebar/util/test/groups-test.js
hmstepanek: Originally I left it to make sure the change I made was behaving correctly but yes, since it isn't dependent on enforced anymore this isn't necessary.

hmstepanek:
I'm not super thrilled about how large this PR is. If you have suggestions on or would like to break it up even more I'm happy to do so.

hmstepanek:
@robertknight I've made some tweaks so I've requested a re-review.

src/sidebar/host-config.js
```diff
@@ -17,6 +17,9 @@ function hostPageConfig(window) {
     // Direct-linked annotation ID
     'annotations',
 
+    // Direct-linked annotation ID
```
robertknight:
```suggestion
    // Direct-linked group ID
```

src/sidebar/services/groups.js
```diff
@@ -89,23 +95,28 @@ function groups(
...
 
-    if (!directLinkedAnnotationId) {
+    if (!directLinkedAnnotationId && !directLinkedGroupId) {
       return Promise.resolve(nonWorldGroups);
     }
 
-    return api.annotation
-      .get({ id: directLinkedAnnotationId })
+    // If the directLinkedGroupId or the directLinkedAnnotationId's group
+    // is the "Public" group show the "Public" group otherwise do not.
```
robertknight: If `directLinkedGroupId` is set to `__world__` then we always return `groups` rather than `nonWorldGroups` here. Therefore I think the code can be simplified by adding an early return in that case and leaving the rest of the function as it was before:

```
if (directLinkedGroupId === '__world__') {
  return Promise.resolve(groups);
}
```

src/sidebar/services/test/groups-test.js
```diff
@@ -236,6 +275,25 @@ describe('groups', function() {
       });
     });
 
+    it('sets the linked group to take precedence over the group saved in local storage', () => {
```
robertknight:
```suggestion
    it('sets the direct-linked group to take precedence over the group saved in local storage', () => {
```

A small suggestion to make the terminology a little more consistent with the comments in the implementation of the service.

src/sidebar/services/test/groups-test.js
```diff
@@ -300,6 +358,71 @@ describe('groups', function() {
       });
     });
 
+    it('both groups are in the final groups list when an annotation and a group are linked to', () => {
+      // This can happen if the linked to annotation and group are configured by
+      // the frame embedding the client.
```
robertknight: This comment is helpful.

src/sidebar/services/test/groups-test.js
```diff
@@ -300,6 +358,71 @@ describe('groups', function() {
...
+        const linkedToAnnGroupShown = groups.some(g => g.id === '__world__');
+        assert.isTrue(linkedToAnnGroupShown);
+      });
+    });
+
+    it('includes the "Public" group if the user links to it', () => {
+      /* Set up the test under conditions that would otherwise
+       * not return the Public group. Aka: the user is logged
+       * out and there are associated groups.
+       */
```
robertknight: Small nit. The convention in our code is to use `//` for inline comments, including multi-line ones and `/**` only for JSDoc comments.

src/sidebar/services/groups.js
```diff
@@ -89,18 +95,26 @@ function groups(
...
 
-    return api.annotation
-      .get({ id: directLinkedAnnotationId })
+    // If directLinkedGroup is the "Public" group, always return groups.
+    if (directLinkedGroupId === '__world__') {
+      return groups;
+    }
+
+    // If the directLinkedGroupId or the directLinkedAnnotationId's group
+    // is the "Public" group show the "Public" group otherwise do not.
```
robertknight: The first part of this comment is now redundant (`If the directLinkedGroupId`) given the part above.

PR #1040, opened 2019-04-04
Port h-tooltip directive to React

robertknight:
This "directive" is responsible for showing a customized tooltip based on an element's "aria-label" attribute when it is hovered.

I have some slight doubts about whether we want to continue using custom tooltips. They don't work on mobile at the moment, so no UI in the client should be reliant on them to be usable, although maybe they may still be helpful on desktop. If we did want to enable them on mobile in future, we could follow the example of [Material Design's tooltips](https://material.io/design/components/tooltips.html#placement) which use long-press activation.

**Current example of custom tooltips in the client:**

Screenshot 2019-04-04 16 28 25

----

**Implementation notes**

There are various ways that an Angular directive could be translated to
React. The approach taken here is to create a [hook](https://reactjs.org/docs/hooks-intro.html), `useTooltip`, which
creates a function that accepts a DOM element and shows a tooltip when
that element is hovered. This function can be passed as the
`ref` prop for an element and React will call it when the DOM node is
created, changed or removed.

This was informed by the approach in
https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

One disadvantage is that because the same hooks must be called in the
same order on every render, `useTooltip` must be called if it _might_ be
used, whether or not the element with the tooltip is actually rendered.

### DIFF
--- a/src/sidebar/components/test/render-utils.js
+++ b/src/sidebar/components/test/render-utils.js
@@ -1,0 +1,56 @@
+'use strict';
+
+/**
+ * Additional utilities for testing UI components.
+ */
+
+const { options } = require('preact');
+
+/**
+ * Run a function that does Preact component rendering and flush any async tasks
+ * that are enqueued (`setState` calls, `useEffect` hooks etc.). Repeat until
+ * everything stabilizies and no more tasks are enqueued.
+ *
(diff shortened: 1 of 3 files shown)


robertknight:
This is not a high priority at the moment. I'll come back to it later.

PR #1060, opened 2019-04-10
Fix feature flag race condition

hmstepanek:
Remove conditional loading of groups based on the community-groups feature flag. Instead, load all the groups as if the feature flag is enabled, and filter out the groups that shouldn't be shown for the old groups menu in groups.all().
- [x] Remove the feature flag from the group service load function.
- [x] Add getInScopeGroups to group store.
- [x] Call getInScopeGroups to get groups for old groups menu in group service all function.
- [x] Modify the isScopedToUri logic to not be based on scopes.enforced.
- [x] Modify the isSelectable logic inside the GroupListSectionController to be based on scopes.enforced since isScopedToUri no longer takes it into account.
This is a partial fix for https://github.com/hypothesis/product-backlog/issues/975.
This should be merged before .

#1052 with some additional changes:
- [x] Handle uri=null for cases such as viewing a direct annotation without the sidebar 
- [x] Remove disableOosGroupSelection from the group-list-section controller as it's not needed anymore.

### DIFF
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -77,7 +77,9 @@ function groups(
 
     // If the main document URL has no groups associated with it, always show
     // the "Public" group.
-    const pageHasAssociatedGroups = groups.some(g => g.id !== '__world__');
+    const pageHasAssociatedGroups = groups.some(
+      g => g.id !== '__world__' && g.isScopedToUri
+    );
     if (!pageHasAssociatedGroups) {
       return Promise.resolve(groups);
     }
@@ -144,7 +146,7 @@ function groups(
     return uri
       .then(uri => {
         const params = {
-          expand: 'organization',
+          expand: ['organization', 'scopes'],
         };
         if (authority) {
           params.authority = authority;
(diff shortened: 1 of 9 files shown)


src/sidebar/services/groups.js
```diff
@@ -196,8 +191,25 @@ function groups(
       });
   }
 
+  const sortGroups = memoize(groups => {
+    // Sort in the following order: scoped, public, private.
+    // This is for maintaining the order of the old groups menu so when
+    // the old groups menu is removed this can be removed.
+    const worldGroups = groups.filter(g => g.id === '__world__');
+    const nonWorldScopedGroups = groups.filter(
+      g => g.id !== '__world__' && ['open', 'restricted'].includes(g.type)
```
robertknight: I don't think we need to worry since we're expecting this code to go away very soon, but I wonder if it is possible that private groups could be scoped in future.

src/sidebar/services/groups.js
```diff
@@ -77,7 +77,9 @@ function groups(
 
     // If the main document URL has no groups associated with it, always show
     // the "Public" group.
-    const pageHasAssociatedGroups = groups.some(g => g.id !== '__world__');
+    const pageHasAssociatedGroups = groups.some(
```
robertknight: When you say "asynchronously" in the commit message I think you mean "concurrently".

PR #1064, opened 2019-04-12
Handle illegal group link states

hmstepanek:
Show an error message in the sidebar when a direct-linked group cannot be displayed. This may be due to the following circumstances:
- The api request for the group returned a 404 error.
- The group is out of scope and scope is enforced.

When the user is logged in:
image

When the user is logged out:
image

This is a fix for https://github.com/hypothesis/product-backlog/issues/990.
This should be merged after https://github.com/hypothesis/client/pull/1072.

### DIFF
--- a/src/sidebar/components/search-status-bar.js
+++ b/src/sidebar/components/search-status-bar.js
@@ -15,7 +15,8 @@ module.exports = {
     onClearSelection: '&',
     searchQuery: '<',
     selectedTab: '<',
-    selectionCount: '<',
+    // Boolean indicating all annotations are visible (none are hidden).
+    areAllAnnotationsVisible: '<',
     totalAnnotations: '<',
     totalNotes: '<',
   },
--- a/src/sidebar/components/sidebar-content.js
+++ b/src/sidebar/components/sidebar-content.js
@@ -285,6 +285,17 @@ function SidebarContentController(
     true
   );
 
+  this.showSelectedTabs = function() {
+    if (
+      this.selectedAnnotationUnavailable() ||
+      this.selectedGroupUnavailable() ||
+      this.search.query()
+    ) {
+      return false;
+    }
+    return true;
+  };
+
   this.setCollapsed = function(id, collapsed) {
     store.setCollapsed(id, collapsed);
   };
--- a/src/sidebar/components/sidebar-content.js
+++ b/src/sidebar/components/sidebar-content.js
@@ -299,15 +299,15 @@ function SidebarContentController(
   this.focus = focusAnnotation;
   this.scrollTo = scrollToAnnotation;
 
-  this.selectedAnnotationCount = function() {
+  this.areAllAnnotationsVisible = function() {
     if (this.directLinkedGroupFetchFailed) {
-      return 1;
+      return true;
     }
     const selection = store.getState().selectedAnnotationMap;
     if (!selection) {
-      return 0;
+      return false;
     }
-    return Object.keys(selection).length;
+    return Object.keys(selection).length > 0;
   };
 
   this.selectedGroupUnavailable = function() {
--- a/src/sidebar/components/sidebar-content.js
+++ b/src/sidebar/components/sidebar-content.js
@@ -161,7 +161,7 @@ function SidebarContentController(
     searchClient.get({ uri: uris, group: group });
   }
 
-  function isLoading() {
+  this.isLoading = function() {
     if (
       !store.frames().some(function(frame) {
         return frame.uri;
@@ -177,7 +177,7 @@ function SidebarContentController(
     }
 
     return false;
-  }
+  };
 
   /**
    * Load annotations for all URLs associated with `frames`.
@@ -272,7 +272,7 @@ function SidebarContentController(
         // of switching to the group containing a direct-linked annotation.
         //
         // In that case, we don't want to trigger reloading annotations again.
-        if (isLoading()) {
+        if (this.isLoading()) {
           return;
         }
         store.clearSelectedAnnotations();
@@ -307,7 +307,9 @@ function SidebarContentController(
 
   this.selectedAnnotationUnavailable = function() {
     const selectedID = firstKey(store.getState().selectedAnnotationMap);
-    return !isLoading() && !!selectedID && !store.annotationExists(selectedID);
+    return (
+      !this.isLoading() && !!selectedID && !store.annotationExists(selectedID)
+    );
   };
 
   this.shouldShowLoggedOutMessage = function() {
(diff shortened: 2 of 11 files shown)


src/sidebar/components/sidebar-content.js
```diff
@@ -298,13 +298,20 @@ function SidebarContentController(
   this.scrollTo = scrollToAnnotation;
 
   this.selectedAnnotationCount = function() {
```
hmstepanek: This name needs to change to something like hideAnnotations.

src/sidebar/components/sidebar-content.js
```diff
@@ -297,17 +312,26 @@ function SidebarContentController(
   this.focus = focusAnnotation;
   this.scrollTo = scrollToAnnotation;
 
-  this.selectedAnnotationCount = function() {
+  this.annotationsAreSelected = function() {
```
hmstepanek: I'm kinda thinking this should maybe even be isAnnotationSelected because right now you can only 'select' one annotation at a time. I'm guessing this was intended for sometime in the future when we can select multiple annotations but who knows when that's gonna be-might be simpler to just name it according to what we have right now.

src/sidebar/components/sidebar-content.js
```diff
@@ -365,6 +389,8 @@ function SidebarContentController(
 
     store.clearSelectedAnnotations();
     store.selectTab(selectedTab);
+    // Clear direct-linked group fetch failed state.
```
hmstepanek: I just realized there's no test for this so I'll have to add that tomorrow. In fact there is not test for this method at all in the tests currently. These tests desperately need some refactoring but I didn't want to do all that here.

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -150,7 +150,7 @@ describe('sidebar.components.sidebar-content', function() {
     });
   }
 
-  beforeEach(
+  const makeSidebarContentController = () => {
```
hmstepanek: There might be a better way to do this and I feel like what I did was kinda hacky but I didn't want to have to refactor all the tests so I abstracted this to a function so that I could call it to re-build the controller after I initialized the things needed to calculate directLinkedGroupFetchFailed as the digest wasn't doing it for me.

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -265,8 +357,8 @@ describe('sidebar.components.sidebar-content', function() {
         $scope.$digest();
       });
 
-      it('selectedAnnotationCount is > 0', function() {
-        assert.equal(ctrl.selectedAnnotationCount(), 1);
+      it('annotationsAreSelected is true', function() {
```
hmstepanek: Didn't convert this to `() =>` here cause I figured I'd save that for the refactor of these tests and all I did was update the description slightly.

robertknight:
Rebased.

src/sidebar/services/groups.js
```diff
@@ -68,6 +68,19 @@ function groups(
     directLinkedAnnotationId,
     directLinkedGroupId
   ) {
+    // Filter the directLinkedGroup out if it is out of scope and scope is enforced.
+    if (directLinkedGroupId) {
+      const directLinkedGroup = groups.find(g => g.id === directLinkedGroupId);
+      if (
+        directLinkedGroup &&
+        !directLinkedGroup.isScopedToUri &&
```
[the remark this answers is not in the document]
hmstepanek: yes

src/sidebar/components/test/sidebar-content-test.js
hmstepanek: > that only gets initialized when the controller is constructed, so it is expected that it can only change in one direction afterwards

Yes, exactly. I can add a comment about that so it's clearer-should have done that earlier.

src/sidebar/components/search-status-bar.js
```diff
@@ -15,7 +15,9 @@ module.exports = {
     onClearSelection: '&',
     searchQuery: '<',
     selectedTab: '<',
-    selectionCount: '<',
+    // Boolean indicating there are specially selected annotations present
+    // such as an error message or a direct-linked annotation.
```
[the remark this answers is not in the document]
hmstepanek: Ah ok that way of thinking about it is clearer.

src/sidebar/components/sidebar-content.js
```diff
@@ -49,6 +49,8 @@ function SidebarContentController(
   streamFilter
 ) {
   const self = this;
+  this.directLinkedGroupFetchFailed =
+    !!settings.group && settings.group !== store.focusedGroup().id;
```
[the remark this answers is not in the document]
hmstepanek: I'm looking into that issue with the focused group you pointed out earlier when direct linking to a private group and then logging in which _might_ require this state to be captured in multiple components but I haven't decided yet.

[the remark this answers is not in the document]
hmstepanek: It's looking like this is gonna have to move to the store but I'll do that in a follow up PR where I fix the private group login issue cause I don't want this one to get too large.

src/sidebar/components/sidebar-content.js
robertknight: Is this resolved or still TODO?

src/sidebar/components/sidebar-content.js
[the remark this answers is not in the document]
robertknight: Leaving that change for a follow-up PR makes sense.

src/sidebar/components/search-status-bar.js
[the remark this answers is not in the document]
robertknight: The comment above either needs to be updated, or simply removed now that the naming is clearer. I'd vote for just removing it.

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -163,13 +163,50 @@ describe('sidebar.components.sidebar-content', function() {
...
     return sandbox.restore();
   });
 
+  describe('showSelectedTabs', () => {
+    beforeEach(() => {
+      setFrames([{ uri: 'http://www.example.com' }]);
+      ctrl.search = { query: sinon.stub().returns(undefined) };
+    });
+
+    it('sets showSelectedTabs to false if there is a search query', () => {
```
robertknight: `showSelectedTabs` is a method that returns a value rather than one which sets/does something as a side effect, so I would suggest writing these descriptions as "returns {true, false} if {condition}"

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -255,7 +292,62 @@ describe('sidebar.components.sidebar-content', function() {
...
+        $scope.$digest();
+        // Re-construct the controller after the environment setup.
+        makeSidebarContentController();
+      });
+
+      it('sets directLinkedGroupFetchFailed to true', () => {
+        assert.isTrue(ctrl.directLinkedGroupFetchFailed);
+      });
+
+      it('sets areAllAnnotationsVisible to true since there is an error message', () => {
```
robertknight: Similar to above, I think this test and the one below it would be slightly clearer if written as "{methodName} returns {value}", because `areAllAnnotationsVisible` and `selectedGroupUnavailable` just return a value rather than being a mutable property which gets set (unlike `directLinkGroupFetchFailed`, it makes sense that gets "set")

src/sidebar/components/sidebar-content.js
hmstepanek: Just added it a moment ago.

PR #1072, opened 2019-04-17
Add sidebar-content-error react component

hmstepanek:
Split the sidebar error message (currently used to notify the user when they encounter an error with a direct-linked annotation) into a separate react controller. 

This should be merged before https://github.com/hypothesis/client/pull/1064.

### DIFF
--- a/src/sidebar/components/sidebar-content-error.js
+++ b/src/sidebar/components/sidebar-content-error.js
@@ -1,0 +1,52 @@
+'use strict';
+
+const { Fragment, createElement } = require('preact');
+const propTypes = require('prop-types');
+
+/**
+ * An error message to display in the sidebar.
+ */
+function SidebarContentError({
+  loggedOutErrorMessage,
+  loggedInErrorMessage,
+  onLoginRequest,
+  isLoggedIn,
+}) {
+  return (
+    <div className="annotation-unavailable-message">
+      <div className="annotation-unavailable-message__icon" />
+      <p className="annotation-unavailable-message__label">
+        {!isLoggedIn ? (
+          <Fragment>
+            {loggedOutErrorMessage}
+            <br />
+            You may need to{' '}
+            <a
+              className="loggedout-message__link"
+              href=""
+              onClick={onLoginRequest}
+            >
+              log in{' '}
+            </a>
+            to see it.
+          </Fragment>
+        ) : (
+          <span>{loggedInErrorMessage}</span>
+        )}
+      </p>
+    </div>
+  );
+}
+
+SidebarContentError.propTypes = {
+  /* A short error message to be displayed when logged out along with a login prompt. */
+  loggedOutErrorMessage: propTypes.string,
+  /* A detailed error message explaining why the error message happened when logged in. */
+  loggedInErrorMessage: propTypes.string,
+  /* A function that will launch the login flow for the user. */
+  onLoginRequest: propTypes.func.isRequired,
+  /* A boolean indicating whether the user is logged in or not. */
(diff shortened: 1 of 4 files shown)


src/sidebar/components/sidebar-content-error.js
```diff
@@ -0,0 +1,44 @@
...
+ * An error message to display in the sidebar.
+ */
+function SidebarContentError({ errorMessage, errorDetails, onLogin, auth }) {
+  return (
+    <Fragment>
+      <div className="annotation-unavailable-message">
+        <div className="annotation-unavailable-message__icon" />
+        <p className="annotation-unavailable-message__label">
+          {auth.status === 'logged-out' ? (
+            <span>
```
robertknight: The `<span>` here isn't needed as far as the DOM structure is concerned, but you do need some kind of container for the child elements. This is where `Fragment` is useful, it provides a container in JSX which does not appear in the DOM.

src/sidebar/components/sidebar-content-error.js
```diff
@@ -0,0 +1,44 @@
...
+      </div>
+    </Fragment>
+  );
+}
+
+SidebarContentError.propTypes = {
+  /* A short error message to be displayed when logged out along with a login prompt. */
+  errorMessage: propTypes.string,
+  /* A detailed error message explaining why the error message happened when logged in. */
+  errorDetails: propTypes.string,
```
robertknight: Hmm, the names here imply that the "details" message is longer and "errorMessage" is just a short summary, but that isn't really the case. The difference is just which one gets shown depending on whether the user is logged-in or not. Perhaps `loggedOutMessage` and `loggedInMessage` might be more obvious?

src/sidebar/components/sidebar-content-error.js
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <Fragment>
+      <div className="annotation-unavailable-message">
+        <div className="annotation-unavailable-message__icon" />
+        <p className="annotation-unavailable-message__label">
+          {auth.status === 'logged-out' ? (
+            <span>
+              {errorMessage}
+              <br />
+              <a> You may need to </a>
```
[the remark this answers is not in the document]
hmstepanek: That's a clever trick.

src/sidebar/components/sidebar-content-error.js
hmstepanek: I debated this for a long time. I wasn't sure what made sense. I was concerned loggedOutMessage vs loggedInMessage wasn't entirely accurate because technically loggedOutMessage is only the first part of the message. I thought it might be confusing to set loggedOutMessage and then see some additional text there that you didn't set. It's also a little odd to me that the short error message doesn't appear when the user is logged in, just the detailed description but I'm not sure I wanna open up that can of worms. If those names make sense to you though I'm happy to use them.

src/sidebar/components/sidebar-content-error.js
```diff
@@ -0,0 +1,52 @@
...
+          <Fragment>
+            {loggedOutErrorMessage}
+            <br />
+            You may need to{' '}
+            <a
+              className="loggedout-message__link"
+              href=""
+              onClick={onLoginRequest}
+            >
+              log in{' '}
```
robertknight:
```suggestion
              log in{' '}
```

The space at the end here should be moved outside of the anchor.

robertknight: Eh... I'm going to merge this anyway so I can look at the other open PR.

PR #1074, opened 2019-04-18
Remove fragment anchoring support

### LINKED ISSUE #479: Client doesn't open on LinkedIn Pulse pages
For example go to this page: <https://www.linkedin.com/pulse/feed/channel/227?trk=pulse-cat-rec-chn_card-img-5> or any of these <https://www.linkedin.com/pulse/discover>. When you click the Chrome extension button to open the client, nothing happens (no sidebar). I didn't see any failed network requests or any errors in the console (either host page or extension's background page).

robertknight:
_(For context on why this issue came up now, see this [Slack support issue](https://hypothes-is.slack.com/archives/C1EQELNMS/p1555520595003600))_

dom-anchor-fragment is shipped as a UMD module which causes problems on
pages that declare a global function called `define` (see #479).

While we could fix the module, fragment anchoring is not very useful at
the moment. The client primarily anchors to the quote and uses other
anchors only as an optimization. The Wikipedia baseline tests changed in
this PR show how little value a fragment has on real websites.
Therefore, just remove support for generating and using fragment selectors.

This change should not cause anchoring regressions, as anchoring would
only have succeeded previously if the quote selector matched. Now it will just
fall through to anchoring using the range, position or quote selectors instead.

Fixes #479

----

**Testing**

1. Go to https://www.fulcrum.org/concern/monographs/dj52w5765?locale=en and try to activate the prod boomarklet or extension. It should fail to load.
2. Go to https://www.fulcrum.org/concern/monographs/dj52w5765?locale=en and activate a bookmarklet / extension from this branch. It should load. Note that you will either need to serve the dev client over HTTPS or disable your browser's mixed-content blocking.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "diff": "^4.0.1",
     "diff-match-patch": "^1.0.0",
     "document-base-uri": "^1.0.0",
-    "dom-anchor-fragment": "^1.0.1",
     "dom-anchor-text-position": "^4.0.0",
     "dom-anchor-text-quote": "^4.0.2",
     "dom-node-iterator": "^3.5.3",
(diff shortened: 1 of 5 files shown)


dwhly:
Nice.

PR #1084, opened 2019-04-23
Fix direct linked private login flow

hmstepanek:
This fixes a problem where a direct-linked private group's annotations do not show up after the user logs in. Fixes https://github.com/hypothesis/product-backlog/issues/1000.

Dependent on https://github.com/hypothesis/client/pull/1111 and https://github.com/hypothesis/client/pull/1112.

----

(Note from @robertknight)

This PR resolves the issue where a user, who is not logged in to the client, following a direct link to a private group, would not see the group after logging in. The issue was that `services/groups.js` would only focus the direct-linked group after fetching groups for the first time on startup, not when re-fetching groups during an account switch. This PR clears the groups state when switching accounts, so that account switching behaves the same way as startup where the "best" group (direct linked, or last used if no direct link, or a fallback otherwise) is focused after groups have been fetched.


src/sidebar/templates/sidebar-content.html
```diff
@@ -36,7 +36,7 @@
 <sidebar-content-error
   class="sidebar-content-error"
   logged-out-error-message="'This group is not available.'"
-  logged-in-error-message="'You either do not have permission to view this group, the group does not exist, or the group is not visible at this URL.'"
+  logged-in-error-message="'This group is not available.'"
```
hmstepanek: I discussed this with @lyzadanger earlier today. She said just making them the same message would be fine in her opinion.

src/sidebar/components/sidebar-content.js
```diff
@@ -252,8 +250,14 @@ function SidebarContentController(
   // Re-fetch annotations when focused group, logged-in user or connected frames
   // change.
   $scope.$watch(
-    () => [groups.focused().id, store.profile().userid, ...store.searchUris()],
-    ([currentGroupId], [prevGroupId]) => {
+    () => [groups.focused(), store.profile().userid, ...store.searchUris()],
```
hmstepanek: I haven't checked to see if there are any unforeseen implications of this change. We would be comparing references instead of id's here but since the focused group is grabbed from the groups list which is static until it's reloaded, I think that should be ok. I might dig a little deeper into that tomorrow. Let me know if you notice/know of any!

src/sidebar/components/sidebar-content.js
```diff
@@ -252,8 +250,14 @@ function SidebarContentController(
...
+    ([currentGroup], [prevGroup]) => {
+      // If the currentGroup is null, clear the selected annotations and return immediatly.
+      if (!currentGroup) {
+        store.clearSelectedAnnotations();
+        return;
+      }
+
       // FIXME - There is a bug here where the set of displayed annotations can
```
hmstepanek: I'd really like to fix this bug noted in the comments here. I wonder if we might be able to do that with the clearGroups that I added. Perhaps set the focusGroup to null under that particular circumstance thus ensuring the first load of annotations when the focused group is not set doesn't happen.

robertknight:
I noticed an exception in the console after clicking the "Log in" link in the "Group not available" error message and completing the login:

Screenshot 2019-04-23 11 36 44

src/sidebar/components/hypothesis-app.js
```diff
@@ -112,6 +112,14 @@ function HypothesisAppController(
...
    */
   this.login = function() {
+    // Clear the groups so that the groups service treats this load
+    // as the "firstLoad" and overrides the focused group. This is
+    // necessary for direct-linked private groups since this group
+    // won't be focused until the user logs in (because it doesn't
+    // exist in the list when the user is logged out).
+    if (store.getState().directLinkedGroupFetchFailed) {
+      store.clearGroups();
+    }
```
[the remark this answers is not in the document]
robertknight: Here is one way that the logic might be kept within the groups service:

```diff
diff --git a/src/sidebar/components/hypothesis-app.js b/src/sidebar/components/hypothesis-app.js
index 2fe930b6..1467be32 100644
--- a/src/sidebar/components/hypothesis-app.js
+++ b/src/sidebar/components/hypothesis-app.js
@@ -112,14 +112,6 @@ function HypothesisAppController(
    *   completes. For non-OAuth logins, always resolves immediately.
    */
   this.login = function() {
-    // Clear the groups so that the groups service treats this load
-    // as the "firstLoad" and overrides the focused group. This is
-    // necessary for direct-linked private groups since this group
-    // won't be focused until the user logs in (because it doesn't
-    // exist in the list when the user is logged out).
-    if (store.getState().directLinkedGroupFetchFailed) {
-      store.clearGroups();
-    }
     if (serviceConfig(settings)) {
       // Let the host page handle the login request
       bridge.call(bridgeEvents.LOGIN_REQUESTED);
diff --git a/src/sidebar/services/groups.js b/src/sidebar/services/groups.js
index 2affc64e..d57f6741 100644
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -174,6 +174,14 @@ function groups(
       uri = getDocumentUriForGroupSearch();
     }
     const directLinkedGroup = settings.group;
+
+    // Use the direct link if this is the first load, or if a previous attempt
+    // to fetch a direct-linked group failed. The latter can happen if the
+    // direct link is to a private group and the user is logged out.
+    const useDirectLink =
+      store.allGroups().length === 0 ||
+      store.getState().directLinkedGroupFetchFailed;
+
     return uri
       .then(uri => {
         const params = {
@@ -247,7 +255,7 @@ function groups(
         const prevFocusedGroup = localStorage.getItem(STORAGE_KEY);
 
         store.loadGroups(groups);
-        if (isFirstLoad && groups.some(g => g.id === directLinkedGroup)) {
+        if (useDirectLink && groups.some(g => g.id === directLinkedGroup)) {
           store.focusGroup(directLinkedGroup);
         } else if (isFirstLoad && groups.some(g => g.id === prevFocusedGroup)) {
           store.focusGroup(prevFocusedGroup);
```

Are there any scenarios where this would result in the wrong thing happening?

src/sidebar/components/sidebar-content.js
```diff
@@ -252,8 +250,14 @@ function SidebarContentController(
   // Re-fetch annotations when focused group, logged-in user or connected frames
   // change.
   $scope.$watch(
-    () => [groups.focused().id, store.profile().userid, ...store.searchUris()],
-    ([currentGroupId], [prevGroupId]) => {
+    () => [groups.focused(), store.profile().userid, ...store.searchUris()],
+    ([currentGroup], [prevGroup]) => {
+      // If the currentGroup is null, clear the selected annotations and return immediatly.
+      if (!currentGroup) {
```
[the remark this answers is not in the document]
hmstepanek: This is necessary so that we don't load the annotations while the groups are still loading and also don't get deeper into this function when the focused group is null. By clearing the focused group we are able to prevent any load that may happen before the groups have been re-grabbed by the groups service. With the changed you proposed below, I run into the race condition described below this logic. The userid changes and the annotations are reloaded for the open group, then when the private group is focused on, the private group annotations are skipped because an annotation load is already in progress. I think clearing out the groups is a fairly simple way to handle this but I see your point about it being a bit fragile. I'll have to think about this a little more. Perhaps we can sync up and brainstorm some ideas tomorrow? I think there are a couple options here:
1) Re-think 'watching' these values and place this logic in the functions that trigger those changes. Perhaps moving the logic surrounding the loading annotations into a service or util.
2) Storing the state of when the group service has completed reloading the groups and only reload annotations when it's finished.
3) Add a more official clear state method into the groups service and call that instead of clearing the groups on the store directly.

To clarify the issue we are trying to resolve here, I think it runs deeper than just direct-linked groups, it really is any group that will change from being fetchable to not being fetchable (or vise versa) when the user logs in or out is broken due to this race condition.

src/sidebar/components/sidebar-content.js
```diff
@@ -269,7 +273,7 @@ function SidebarContentController(
       // logged-in users is currently a superset of those for logged-out users on
       // any given page.
 
-      if (currentGroupId !== prevGroupId) {
+      if (currentGroup.id !== prevGroup.id) {
```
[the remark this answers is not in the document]
hmstepanek: Ah! Not sure how I missed that. Durp.

src/sidebar/store/modules/direct-linked-group.js
```diff
@@ -0,0 +1,57 @@
...
+  return {
+    /**
+     * Indicates that an error occured in retrieving/showing the direct linked group.
+     * This could be because:
+     * - the group does not exist
+     * - the user does not have permission
+     * - the group is out of scope for the given page
+     * @type {boolean}
+     */
+    directLinkedGroupFetchFailed: false,
```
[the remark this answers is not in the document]
hmstepanek: I have mixed feeling about this one. Yes, it is groups related so it could possibly go in the groups module. I'm not sure that's enough of an argument to put it there though. The groups module is already very large, it consistently holds the state of the groups and the focused group and while the direct-linked group has influence on the focused group, I'm not sure tracking it's error state belongs there. The fetch failed state is specific to the direct-linked group, not to all groups. In keeping with the current placement in the store modules, the state of direct-linked/selected annotation(s) are not stored in the annotations module, they are stored in the selection module. So if we were to place this into an existing module I would propose to place it into the selection module as opposed to the groups module. On the other hand, selection seems to be a bit of a dumping ground for UI state and I'm not sure if it's a good idea to keep dumping things in there. I would like this to be easy to find, and give direct-linked group things a place to live (something that has come up a couple times already is reporting more specific error messages and I could see the need for that kind of information to be stored in the store but I don't want to get too ahead of myself) so putting it into it's own module paves the way for that. The groups are data whereas the error state is really a UI concern and those feel like separate things to me.

src/sidebar/components/hypothesis-app.js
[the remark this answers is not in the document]
hmstepanek: See below, we still need to handle the race condition, otherwise this doesn't actually load the groups for the private group but rather the group that was focused on when the user was logged out.

src/sidebar/services/groups.js
```diff
@@ -251,6 +256,34 @@ function groups(
...
+   * Clear the state of the groups if there is a direct-linked group
+   * fetch failure or the currently selected group is a private group.
+   *
+   * This is intended to be called during user state changes (aka on
+   * login or logout) so that any events that reference the focused
+   * group wait until the new focused group has been loaded before
+   * executing.
+   *
+   */
+  function clearState() {
```
hmstepanek: I need help with this name. I wasn't sure how to put all those comments into a short function name.

src/sidebar/components/sidebar-content.js
hmstepanek: Update: Didn't find any.

robertknight: There is a subtle issue here. Because the last argument to `$scope.$watch` is `true`, a deep equality check is done for all of the arguments. For a complex object graph that would be inefficient. The deep check is needed for `searchUris` because that can currently return a different reference but with the same content if some non-URL property of the connected documents changes. If `groups.focused()` can return `null` then the previous code was already broken though because it assumed otherwise.

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -178,6 +178,7 @@ describe('sidebar.components.sidebar-content', function() {
   describe('clearSelection', () => {
     it('sets selectedTab to Annotations tab if selectedTab is null', () => {
       store.selectTab(uiConstants.TAB_ORPHANS);
+      store.setDirectLinkedGroupFetchFailed();
```
robertknight: This change doesn't appear to have anything to do with the condition that the test is checking for.

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -201,11 +202,9 @@ describe('sidebar.components.sidebar-content', function() {
     });
 
     it('clears the directLinkedGroupFetchFailed state', () => {
-      ctrl.directLinkedGroupFetchFailed = true;
```
robertknight: The logic to set the flag has been removed, but it hasn't been replaced with anything else. Did you mean to put the `store.setDirectLinkedGroupFetchFailed()` call in this test instead of the one above?

src/sidebar/services/groups.js
```diff
@@ -199,12 +200,16 @@ function groups(
...
         if (directLinkedGroup) {
+          // Clear any previously set state.
+          store.clearDirectLinkedGroupFetchFailed();
           const selectedGroupApi = api.group
             .read({
               id: directLinkedGroup,
               expand: params.expand,
             })
             .catch(() => {
+              // Set the state of directLinkedGroupFetchFailed to true.
```
robertknight: This comment effectively just states what the code itself says, so I would just omit it.

src/sidebar/services/groups.js
robertknight: It is getting quite tricky to reason about all the possible states here. I think it would simplify things if we could just always clear all groups when logging out/logging in. Then startup, login and logout will all be the same state.

The reason for not doing that in the past was to avoid losing the focused group on login/logout. Since it is persisted to local storage however, it will be restored from local storage after the login/logout completes, as long as the persisted focused group isn't lost when it briefly becomes `null`. That will require a small change to the logic lower down the file which subscribes to focused group changes.

robertknight: For consistency with the other methods, I would have preferred this method to be named just `clear` - that would only make sense if there weren't a bunch of conditions attached to it though.

src/sidebar/components/test/sidebar-content-test.js
hmstepanek: agh yes!

src/sidebar/services/groups.js
hmstepanek: So as discussed, if we clear the groups on logout/login and there is a direct link it will display the direct linked group/annotation again even though the user may have 'cleared' that state by clicking on another group. I'm going to look into perhaps clearing the direct-linked settings in that case which is more explicit and will hopefully allow us to do as you suggested and simplify this logic.

src/sidebar/components/group-list-item.js
```diff
@@ -7,9 +7,14 @@ const { createElement } = require('preact');
 const { orgName } = require('../util/group-list-item-common');
 const { withServices } = require('../util/service-context');
 
-function GroupListItem({ analytics, group, store }) {
+function GroupListItem({ analytics, group, settings, store }) {
   const focusGroup = () => {
     analytics.track(analytics.events.GROUP_SWITCH);
+
+    // Clear the direct-linked annotations and group from settings.
+    delete settings.annotations;
```
hmstepanek: I wanted to keep the settings as they would have appeared had these settings not been set and in those cases the property is literally missing.

src/sidebar/components/sidebar-content.js
hmstepanek: Hmmm...well technically it could have been null but that never happened in reality because it was always set before this func began watching. I'm not sure what to do about what you describe above. Any ideas? There are quite a few attrs on groups. Not enough to cause a significant slow down at the moment but I see what you are saying.

hmstepanek: Perhaps this is one of those leave it for now and address this in a follow-up later? In general I'd like to see if there's a way we could get rid of this watch method all together but that requires a larger discussion and I'm not gonna tackle that in this PR.

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -219,6 +219,13 @@ describe('sidebar.components.sidebar-content', function() {
 
       assert.isFalse(store.getState().directLinkedGroupFetchFailed);
     });
+
+    it("clears the direct linked id's from the settings", () => {
```
robertknight: The test description here needs updating to reflect the fact that the state was moved to the store.

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -588,8 +595,7 @@ describe('sidebar.components.sidebar-content', function() {
     }
 
     beforeEach(function() {
-      // There is a direct-linked annotation
-      fakeSettings.annotations = 'test';
+      store.setDirectLinkedAnnotationsId('test');
```
robertknight: Hmm, I can't remember if multiple annotation IDs are actually supported. If not I'd change this to singular rather than plural.

robertknight:
Rebased

src/sidebar/store/modules/direct-linked.js
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Set the direct linked group id.
+ */
+function setDirectLinkedGroupId(groupId) {
+  return {
+    type: actions.UPDATE_DIRECT_LINKED_GROUP_ID,
+    directLinkedGroupId: groupId,
+  };
+}
```
robertknight: I'm going to leave these two actions here for now as they are used by sidebar content tests. However as a design principle, I should note that it isn't necessary to have "setters" for all the state in the store like you might have for objects in OO design. Idiomatically actions should be "things that happened" in the program (for example, data was fetched, the selection was cleared, the active document or tab changed).

robertknight:
I pushed some small PRs to clean up some typos and outdated comments from earlier versions of the PR, as well as adding a couple of explanatory comments.

PR #1085, opened 2019-04-23
Convert mocking to mockable-imports (2/4)

robertknight:
This is a follow-up to https://github.com/hypothesis/client/pull/1061 which converts remaining uses of proxyquire in .js files to use `$imports.$mock` instead - see earlier PR for a detailed explanation.

What remains to do after this is convert uses in .coffee files and remove proxyquire as a dependency.

### DIFF
--- a/src/sidebar/components/test/annotation-header-test.js
+++ b/src/sidebar/components/test/annotation-header-test.js
@@ -1,9 +1,9 @@
 'use strict';
 
 const angular = require('angular');
-const proxyquire = require('proxyquire');
 
 const fixtures = require('../../test/annotation-fixtures');
+const annotationHeader = require('../annotation-header');
 
 const fakeDocumentMeta = {
   domain: 'docs.io',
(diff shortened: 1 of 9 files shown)


src/sidebar/components/test/annotation-header-test.js
```diff
@@ -27,18 +27,20 @@ describe('sidebar.components.annotation-header', function() {
   });
 
   beforeEach('Import and register the annotationHeader component', function() {
-    const annotationHeader = proxyquire('../annotation-header', {
+    angular.module('app', []).component('annotationHeader', annotationHeader);
```
hmstepanek: Nit-Seems like we should be mocking the imports before attaching the component just in terms of flow for the reader but functionally it doesn't matter.

robertknight: You're right that it makes no functional difference, but I've switched it around anyway.

PR #1086, opened 2019-04-23
Convert mocking to mockable imports (3/4)

robertknight:
This is a follow-up to #1061 which converts remaining uses of proxyquire in .coffee files to use $imports.$mock instead - see earlier PR for a detailed explanation.

After this PR and https://github.com/hypothesis/client/pull/1085 are merged, the final PR will remove proxyquire as a dependency.

### DIFF
--- a/src/annotator/guest.coffee
+++ b/src/annotator/guest.coffee
@@ -7,6 +7,7 @@ Delegator = require('./delegator')
 $ = require('jquery')
 
 adder = require('./adder')
+htmlAnchoring = require('./anchoring/html')
 highlighter = require('./highlighter')
 rangeUtil = require('./range-util')
 selections = require('./selections')
@@ -37,7 +38,7 @@ module.exports = class Guest extends Delegator
     TextSelection: {}
 
   # Anchoring module
-  anchoring: require('./anchoring/html')
+  anchoring: htmlAnchoring
 
   # Internal state
   plugins: null
--- a/src/annotator/guest.coffee
+++ b/src/annotator/guest.coffee
@@ -38,7 +38,7 @@ module.exports = class Guest extends Delegator
     TextSelection: {}
 
   # Anchoring module
-  anchoring: htmlAnchoring
+  anchoring: null
 
   # Internal state
   plugins: null
@@ -49,7 +49,7 @@ module.exports = class Guest extends Delegator
   html:
     adder: '<hypothesis-adder></hypothesis-adder>'
 
-  constructor: (element, config) ->
+  constructor: (element, config, anchoring = htmlAnchoring) ->
     super
 
     this.adder = $(this.html.adder).appendTo(@element).hide()
@@ -77,6 +77,8 @@ module.exports = class Guest extends Delegator
     # Set the frame identifier if it's available.
     # The "top" guest instance will have this as null since it's in a top frame not a sub frame
     this.frameIdentifier = config.subFrameIdentifier || null
+
+    this.anchoring = anchoring
 
     cfOptions =
       config: config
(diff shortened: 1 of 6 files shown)


src/annotator/guest.coffee
```diff
@@ -37,7 +38,7 @@ module.exports = class Guest extends Delegator
     TextSelection: {}
 
   # Anchoring module
-  anchoring: require('./anchoring/html')
+  anchoring: htmlAnchoring
```
robertknight: The new approach to mocking only currently supports dependencies declared at the top level of the module, so this hoists the `require` to the top of the file. This is needed anyway if we later convert this module to JS with `import` and `export` statements, since `import` can only appear at the top level as well.

yarn.lock
```diff
@@ -1550,9 +1550,9 @@ babel-plugin-istanbul@^5.1.0:
     test-exclude "^5.2.2"
 
 babel-plugin-mockable-imports@^1.1.0:
-  version "1.2.0"
-  resolved "https://registry.yarnpkg.com/babel-plugin-mockable-imports/-/babel-plugin-mockable-imports-1.2.0.tgz#151db640cf6339a6a878a3b70e336eec4baa1895"
-  integrity sha512-KqKIFsPDQcXf2XZZazAxGa7xUgbGueLfCQb4mD7G16lvLYfz13SakC6IAcRaBueXwNn2O0K6VGlXZzPs0aLL3Q==
+  version "1.3.0"
+  resolved "https://registry.yarnpkg.com/babel-plugin-mockable-imports/-/babel-plugin-mockable-imports-1.3.0.tgz#29183a735fc3a27395b15b519e059fba7f61fde4"
+  integrity sha512-1kUED0ZXOLMKh1P90zHX9juD6IRWXgMB5yPtdupsaTJ9lKueAdGmRsuMnKXOlqtyaMYCW6oKwSz/gFLpiXaONA==
```
robertknight: This update resolves an issue with the way CoffeeScript compiles groups of `<name> = require(<module>)` statements.

src/annotator/test/guest-test.coffee
```diff
@@ -67,10 +67,16 @@ describe 'Guest', ->
     }
     selections = null
     guestConfig = {pluginClasses: {}}
+    highlighter = {
+      highlightRange: sinon.stub()
+      removeHighlights: sinon.stub()
+    }
+    htmlAnchoring = {
+      anchor: sinon.stub()
+    }
```
robertknight: This ensures that the mocks for the highlighter and HTML anchoring modules are initialized for each test, for consistency with how we set up mocks in other tests.

src/annotator/test/guest-test.coffee
```diff
@@ -50,12 +46,16 @@ describe 'Guest', ->
...
+  highlighter = null
   guestConfig = null
+  htmlAnchoring = null
 
   createGuest = (config={}) ->
     config = Object.assign({}, guestConfig, config)
     element = document.createElement('div')
-    return new Guest(element, config)
+    guest = new Guest(element, config)
+    guest.anchoring = htmlAnchoring
```
hmstepanek: Why is this here instead of in the mocked imports? This feels hacky to me.

robertknight: This is necessary because `guest.anchoring` is initialized when the module is first evaluated, and so isn't affected by mocking that happens later. That's one limitation of this approach to mocking. See [this node in the mocking plugin's README](https://github.com/robertknight/babel-plugin-mockable-imports#mocking-code-that-runs-when-a-module-is-imported). The same issue applies to `patch` in Python.

An alternative solution would be to use constructor injection. Something conceptually like this:

```js
class Guest(element, config, anchoring = htmlAnchoring) {
   this.anchoring = anchoring
}
```

With this approach, the default initialization expression is evaluated when the `Guest` constructor is called, so mocking would work.

robertknight: I've pushed a change which does this to handle the issue in a cleaner way ^

PR #1088, opened 2019-04-23
Convert timestamp component to React

robertknight:
_(There is no urgency on this PR, I had it sitting around on my system waiting for https://github.com/hypothesis/client/pull/1061 to be merged)_

A notable change is that the "current" time used to generate the relative
timestamp is now stored as an explicit piece of state, rather than
implicitly using the current time when calling `toFuzzyString`. This
provides a clean way to trigger a re-render of timestamp periodically:
an effect updates the "now" time, and that triggers a recalculation of
the relative timestamp string.

The tests have also been revised to better capture the user-facing
behavior of the component.


src/sidebar/util/time.js
```diff
@@ -209,14 +209,13 @@ function decayingInterval(date, callback) {
  * Formats a date as a string relative to the current date.
  *
  * @param {number} date - The absolute timestamp to format.
+ * @param {Date} [now] - The date to consider as the current time.
  * @return {string} A 'fuzzy' string describing the relative age of the date.
  */
-function toFuzzyString(date, Intl) {
+function toFuzzyString(date, now = new Date(), Intl) {
```
hmstepanek: It's weird to me that this works but I guess that's a difference between js and python. In python this wouldn't create a new Date every time but looks like that's not the case in js.

hmstepanek: What is `Intl` here-shouldn't that be documented as a param?

src/sidebar/util/test/time-test.js
```diff
@@ -109,7 +109,7 @@ describe('sidebar.util.time', function() {
       const d = new Date().toISOString();
       sandbox.clock.tick(year * 2 * 1000);
 
-      assert.equal(time.toFuzzyString(d, null), 'Thu Jan 01 1970');
+      assert.equal(time.toFuzzyString(d, undefined, null), 'Thu Jan 01 1970');
```
hmstepanek: Seems like we should be taking advantage of being able to pass the date to compare against as an argument in these tests now. It would make these tests a lot more explicit and easier to understand.

src/sidebar/util/test/time-test.js
```diff
@@ -77,15 +77,15 @@ describe('sidebar.util.time', function() {
...
+      assert.equal(time.toFuzzyString(t, undefined, mockIntl()), expect);
     });
 
     const testFixture = function(f) {
       return function() {
         const t = new Date().toISOString();
         const expect = f[1];
         sandbox.clock.tick(f[0] * 1000);
-        assert.equal(time.toFuzzyString(t, mockIntl()), expect);
+        assert.equal(time.toFuzzyString(t, undefined, mockIntl()), expect);
```
hmstepanek: I find this really hard to follow and have a couple thoughts about how to fix that. Moving the test cases to the location where they are used would be helpful here in terms of readability, also passing the params explicitly by name here instead of using f[0], f[1]. If we pass in the time to the toFuzzyString rather than undefined we won't have to tick the clock and this test will be much easier to read IMO.

src/sidebar/util/time.js
robertknight: `Intl` here is the browser's `Intl` global object which is a namespace for internationalization objects. Yes it should be documented :P

robertknight: > It's weird to me that this works but I guess that's a difference between js and python.

Yes indeed. In JS the default value is evaluated every time. I think this is much more useful than Python 😉

robertknight:
@hmstepanek volunteered to address the issues she raised during code review, so I've assigned to her.

robertknight:
Hi @hmstepanek - Can you help me get this merged? The tests that you commented that you want to refactor are only lightly touched by this PR for the purposes of being able to test the timestamp component. If you have time to look at them then great, otherwise can that be addressed separately?

hmstepanek:
Ah yes let me do this now!

src/sidebar/util/test/time-test.js
```diff
@@ -49,67 +34,132 @@ describe('sidebar.util.time', function() {
     // test expectations
     const offset = new Date().getTimezoneOffset();
     if (offset > 0) {
-      sandbox.clock.tick(offset * 60 * 1000);
+      sandbox.clock.tick(offset * 60 * msPerSecond);
     }
+    fakeIntl = sinon.stub().returns({
```
robertknight: `Intl` is a namespace rather than a function. In the browser you access it via `Intl.DateTimeFormat` rather than say `Intl().DateTimeFormat` or `new Intl().DateTimeFormat`.

src/sidebar/components/timestamp.js
```diff
@@ -1,56 +1,67 @@
...
+ *
+ * The timestamp automatically refreshes at an appropriate frequency.
+ */
+function Timestamp({ className, href, timestamp }) {
+  // "Current" time, used when calculating the relative age of `timestamp`.
+  const [now, setNow] = useState(new Date());
+
+  // Fuzzy, relative timestamp (eg. '6 days ago')
+  const prevTime = timestamp ? new Date(timestamp) : null;
+  const relativeTimestamp = useMemo(() => toFuzzyString(prevTime, now), [
```
robertknight: If you move the `prevTime` calculation inside the `useMemo` callback, that will save a little bit of work because the timestamp will only be parsed into a `Date` when it changes, rather than every time the component is rendered. Probably completely immaterial compared to all the other work that is happening, but I thought I'd flag it up as an unintended change.

hmstepanek:
Ok @robertknight. This is ready for a detailed review now.

src/sidebar/util/test/time-test.js
```diff
@@ -5,128 +5,166 @@ const time = require('../time');
...
+        options: { day: 'numeric', month: 'short', year: 'numeric' },
+      },
+      {
+        now: '1978-01-01T00:00:00.000Z',
+        text: '1 Jan 1970',
+        options: { day: 'numeric', month: 'short', year: 'numeric' },
+      },
+    ].forEach(test => {
+      it(
+        'passes correct arguments to format fuzzy string date for fixture ' +
```
robertknight:
```suggestion
        'passes correct arguments to `Intl.DateTimeFormat.format` for fixture ' +
```

src/sidebar/util/test/time-test.js
```diff
@@ -5,128 +5,166 @@ const time = require('../time');
...
+          test.now,
+        () => {
+          const timeStamp = fakeDate('1970-01-01T00:00:00.000Z');
+          const now = fakeDate(test.now);
+
+          fakeIntl.DateTimeFormat().format.returns(test.text); // eslint-disable-line new-cap
+          assert.equal(time.toFuzzyString(timeStamp, now, fakeIntl), test.text);
+          assert.calledWith(fakeIntl.DateTimeFormat, undefined, test.options);
```
robertknight: These two lines can be combined into one `calledWith` check

src/sidebar/util/test/time-test.js
```diff
@@ -5,128 +5,166 @@ const time = require('../time');
...
       // more than one day ago falls back to a simple date string.
-      const d = new Date().toISOString();
-      sandbox.clock.tick(day * 2 * 1000);
+      const timeStamp = fakeDate('1970-01-01T00:00:00.000Z');
+      timeStamp.toDateString = sinon.stub().returns('Thu Jan 01 1970');
+      const now = fakeDate('1970-01-02T00:00:00.000Z');
 
-      assert.equal(time.toFuzzyString(d, null), 'Thu Jan 01 1970');
+      assert.equal(time.toFuzzyString(timeStamp, now, null), 'Thu Jan 01 1970');
+      assert.calledOnce(timeStamp.toDateString);
```
robertknight: You probably want to reverse the order of these asserts since the first one will lead to the second one.

src/sidebar/util/test/time-test.js
```diff
@@ -5,128 +5,166 @@ const time = require('../time');
...
       // more than one year ago falls back to a simple date string.
-      const d = new Date().toISOString();
-      sandbox.clock.tick(year * 2 * 1000);
+      const timeStamp = fakeDate('1970-01-01T00:00:00.000Z');
+      timeStamp.toDateString = sinon.stub().returns('Thu Jan 01 1970');
+      const now = fakeDate('1972-01-01T00:00:00.000Z');
 
-      assert.equal(time.toFuzzyString(d, null), 'Thu Jan 01 1970');
+      assert.equal(time.toFuzzyString(timeStamp, now, null), 'Thu Jan 01 1970');
+      assert.calledOnce(timeStamp.toDateString);
```
robertknight: Same here, I think it makes sense to check what happens in the function before the return value, and then the return value

src/sidebar/util/time.js
```diff
@@ -208,19 +222,20 @@ function decayingInterval(date, callback) {
 /**
  * Formats a date as a string relative to the current date.
  *
- * @param {number} date - The absolute timestamp to format.
+ * @param {Date} date - The date to consider as the timestamp to format.
+ * @param {Date} now - The date to consider as the current time.
+ * @param {Object} Intl - The internationalization.
```
robertknight:
```suggestion
 * @param {Object} Intl - JS internationalization API implementation.
```

src/sidebar/util/test/time-test.js
hmstepanek: How are you thinking of combining these together? I'm afraid I'm not seeing it. I could see just getting rid of the first check all-together except that we might perhaps still want to check to see that the value from the format function was actually returned.

robertknight: Please ignore this. I misread the two "calledWith" asserts as having the same first argument. I've just realized that is not the case.

PR #1099, opened 2019-05-02
Fix GitHub release generation script

robertknight:
Git has two types of tags, lightweight tags and annotated tags. Only
annotated tags have a "tag date" associated with them. Recently it
appears that releases have been tagged with lightweight tags rather than
annotated tags, which the `generate-change-list.js` script did not
handle.

 - When fetching the last tag, use the commit date as a fallback for the
   tag date, and bail with an error if no date can be found

 - Link to the HTML rather than JSON representation of a PR in release
   notes

 - Fix a warning about use of deprecated GitHub API pagination methods
   by switching to the pagination APIs described at
   https://octokit.github.io/rest.js/#pagination

 - Support running the `generate-change-list.js` script as a CLI tool
   for debugging using `node generate-change-list.js <tag name>`

 - Skip Dependabot PRs in the change list, as these add a lot of noise.
   Generally readers are going to be interested in non-maintenance PRs.

### DIFF
--- a/scripts/generate-change-list.js
+++ b/scripts/generate-change-list.js
@@ -6,19 +6,30 @@ const wrapText = require('wrap-text');
 
 /**
  * Return a `Date` indicating when a Git tag was created.
+ *
+ * The tag creation date is used for annotated tags, otherwise there is a
+ * fallback to the date of the tagged commit.
  */
 function getTagDate(tag) {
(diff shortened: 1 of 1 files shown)


robertknight:
I'm going to go ahead and merge this so I can test it out with a release.

PR #1106, opened 2019-05-07
Fix bug where direct linked group is duplicated

hmstepanek:
Do not add the direct-linked group to featuredGroups if it is already
present. This happens when a group is direct-linked to but the group
is already scoped to the page so the direct-linked group shows up twice.

### DIFF
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -213,9 +213,11 @@ function groups(
         }
         return Promise.all(groupApiRequests).then(
           ([myGroups, featuredGroups, token, selectedGroup]) => [
+            // Don't add the selectedGroup if it's already in the featuredGroups.
             combineGroups(
               myGroups,
-              selectedGroup !== undefined
+              selectedGroup !== undefined &&
+                !featuredGroups.some(g => g.id === selectedGroup.id)
                 ? featuredGroups.concat([selectedGroup])
                 : featuredGroups,
               documentUri
--- a/src/sidebar/services/test/groups-test.js
+++ b/src/sidebar/services/test/groups-test.js
@@ -230,6 +230,25 @@ describe('groups', function() {
       });
     });
 
+    // TODO: Add a de-dup test for the direct-linked annotation.
+
+    it('does not duplicate groups if the direct-linked group is also a featured group', () => {
+      const svc = service();
+
+      // Set the direct-linked group to dummyGroups[0].
+      fakeSettings.group = dummyGroups[0].id;
+      fakeApi.group.read.returns(Promise.resolve(dummyGroups[0]));
+
+      // Include the dummyGroups[0] in the featured groups.
+      fakeApi.profile.groups.read.returns(Promise.resolve([]));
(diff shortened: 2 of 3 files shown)


src/sidebar/services/groups.js
```diff
@@ -213,9 +213,11 @@ function groups(
         }
         return Promise.all(groupApiRequests).then(
           ([myGroups, featuredGroups, token, selectedGroup]) => [
+            // Don't add the selectedGroup if it's already in the featuredGroups.
             combineGroups(
               myGroups,
-              selectedGroup !== undefined
+              selectedGroup !== undefined &&
```
hmstepanek: I wasn't sure if it was better to put this logic here or push it into the combineGroups function. The pro's of having this here is the combineGroups is responsible for determining scope and user membership of the group and combining the two groups lists: featuredGroups and myGroups (technically the directLinkedGroup/directLinkedAnnGroup can go into one of the already existing groups lists which simplifies the inputs to the combineGroups function). The con's this adds more logic to the load in the groups service because it has to only put the group in there under certain circumstances as opposed to passing the directLinkedGroup and directLinkedAnnGroup and letting the combineGroups sort out when to add them (if any). The third option would be to de-dup the featuredGroups list in combineGroups but I feel like that's making an assumption about the input that doesn't necessarily make sense (that the groups in the passed list could be duplicated).

src/sidebar/services/test/groups-test.js
```diff
@@ -230,6 +230,25 @@ describe('groups', function() {
       });
     });
 
+    // TODO: Add a de-dup test for the direct-linked annotation.
```
hmstepanek: I'm not sure which PR will get merged first (this one or https://github.com/hypothesis/client/pull/1097). Since this one is simpler I'm guessing this one.

src/sidebar/services/groups.js
robertknight: > The third option would be to de-dup the featuredGroups list

Third option? I'm not entirely clear on how this differs from the first two. By "push it into the combineGroups function" did you mean adding an extra argument to `combineGroups`?

Honestly I don't think it makes that much difference at present. If there were several more sources of groups being combined then I think it would become simpler to de-duplicate in `combineGroups`,  because you could just add a few lines at the end to de-duplicate the final `groups` array.

src/sidebar/services/test/groups-test.js
```diff
@@ -230,6 +230,25 @@ describe('groups', function() {
...
+
+    it('does not duplicate groups if the direct-linked group is also a featured group', () => {
+      const svc = service();
+
+      // Set the direct-linked group to dummyGroup[0].
```
robertknight:
```suggestion
      // Set the direct-linked group to dummyGroups[0].
```

And the same below.

src/sidebar/services/groups.js
hmstepanek: In the third option I mean add the group into the featuredGroups list so there are two of the same group and just add a de-dup in the combineGroups function.

hmstepanek:
I've cleaned up the docs on the combineGroups method a bit.

PR #1109, opened 2019-05-09
Use same direct-linked group error when logged-in

hmstepanek:
Use the same short error message for notifying the user that the direct-linked group in unavailable regardless of login state.

### DIFF
--- a/src/sidebar/components/test/sidebar-content-test.js
+++ b/src/sidebar/components/test/sidebar-content-test.js
@@ -6,6 +6,7 @@ const EventEmitter = require('tiny-emitter');
 const events = require('../../events');
 const sidebarContent = require('../sidebar-content');
 const uiConstants = require('../../ui-constants');
+const util = require('../../directive/test/util');
 
 let searchClients;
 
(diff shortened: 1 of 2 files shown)


src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -149,6 +150,19 @@ describe('sidebar.components.sidebar-content', function() {
     });
   }
 
+  function createSidebarContent(
```
hmstepanek: I plan on cleaning up these tests but for now I wanted to make sure I had some tests for this.

src/sidebar/components/test/sidebar-content-test.js
```diff
@@ -149,6 +150,19 @@ describe('sidebar.components.sidebar-content', function() {
...
+    { userid } = { userid: 'acct:person@example.com' }
```
robertknight: Small aside - This can also be written as `{ userid = 'acct:person@example.com ' } = {}`. If you have an object with many keys, its easier to put the defaults on the left side of the main assignment.

PR #1113, opened 2019-05-10
Fix overlapping text in headings in annotation bodies

robertknight:
Parent elements of annotation cards set line heights to absolute pixel
values. Annotation card headings then inherit this and as a result,
their line height is too small for the font size.

We ought to reconsider the way line heights are set elsewhere in the
app, but resetting the line height in the `styled-text` mixin ensures
that annotation body styles are protected from whatever the parent
elements do with it.

**Master:**

Screenshot 2019-05-10 17 12 05

**This branch:**

Screenshot 2019-05-10 17 08 49

Fixes https://github.com/hypothesis/product-backlog/issues/162

This same fix can also be applied in h.

### DIFF
--- a/src/styles/sidebar/styled-text.scss
+++ b/src/styles/sidebar/styled-text.scss
@@ -1,4 +1,7 @@
 @mixin styled-text() {
+  // Reset the line-height in case any parent elements have set it.
+  line-height: normal;
+
   h1, h2, h3, h4, h5, h6, p, ol, ul, img, pre, blockquote {
     margin: .618em 0;
   }


src/styles/sidebar/styled-text.scss
```diff
@@ -1,4 +1,7 @@
 @mixin styled-text() {
+  // Reset the line-height in case any parent elements have set it.
+  line-height: normal;
```
robertknight: `normal` here means "whatever the browser uses by default", which according to MDN, is a small multiple (eg. 1.2) of _the element's font size_.

PR #1124, opened 2019-05-14
Simplify groups service `load` method using async/await

robertknight:
This is a follow up to https://github.com/hypothesis/client/pull/1122 which removes about 20 lines of code from the `load()` method and reduces the number of scopes and conditionals.

I suggest a commit-by-commit review, but in brief I used async/await syntax to eliminate Promise chains, then removed short-lived variables that were no longer needed and finally moved some variables closer to where they were first used.

This should make it easier to amend this function in future if there are any further changes to the business rules.

### DIFF
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -222,7 +222,7 @@ function groups(
       });
     }
 
-    let [
+    const [
       myGroups,
       featuredGroups,
       token,
@@ -241,11 +241,12 @@ function groups(
     // don't already have it.
 
     // If there is a direct-linked group, add it to the featured groups list.
-    featuredGroups =
+    if (
       directLinkedGroup &&
       !featuredGroups.some(g => g.id === directLinkedGroup.id)
-        ? featuredGroups.concat([directLinkedGroup])
-        : featuredGroups;
+    ) {
+      featuredGroups.push(directLinkedGroup);
+    }
 
     // If there's a direct-linked annotation it may require an extra API call
     // to fetch its group.
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -171,7 +171,7 @@ function groups(
 
     // Step 2: Concurrently fetch the groups the user is a member of,
     // the groups associated with the current document and the annotation
-    // or group that was direct-linked (if any).
+    // and/or group that was direct-linked (if any).
     const params = {
       expand: ['organization', 'scopes'],
     };
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -290,15 +290,14 @@ function groups(
 
     store.loadGroups(groups);
 
-    if (
-      isFirstLoad &&
-      groups.some(g => g.id === directLinkedAnnotationGroupId)
-    ) {
-      store.focusGroup(directLinkedAnnotationGroupId);
-    } else if (isFirstLoad && groups.some(g => g.id === directLinkedGroupId)) {
-      store.focusGroup(directLinkedGroupId);
-    } else if (isFirstLoad && groups.some(g => g.id === prevFocusedGroup)) {
-      store.focusGroup(prevFocusedGroup);
+    if (isFirstLoad) {
+      if (groups.some(g => g.id === directLinkedAnnotationGroupId)) {
+        store.focusGroup(directLinkedAnnotationGroupId);
+      } else if (groups.some(g => g.id === directLinkedGroupId)) {
+        store.focusGroup(directLinkedGroupId);
+      } else if (groups.some(g => g.id === prevFocusedGroup)) {
+        store.focusGroup(prevFocusedGroup);
+      }
     }
 
     return groups;
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -158,8 +158,7 @@ function groups(
    * group.
    *
    * The groups that are fetched depend on the current user, the URI of
-   * the current document, and whether any direct-links were followed (either
-   * to an annotation or group).
+   * the current document, and the direct-linked group and/or annotation.
    *
    * @return {Promise<Group[]>}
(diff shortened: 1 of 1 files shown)


src/sidebar/services/groups.js
```diff
@@ -163,164 +163,144 @@ function groups(
...
+
+    // If there's a direct-linked annotation it may require an extra API call
+    // to fetch its group.
+    let directLinkedAnnotationGroupId = null;
+    if (directLinkedAnn) {
+      // Set the directLinkedAnnotationGroupId to be used later in
+      // the filterGroups method.
+      directLinkedAnnotationGroupId = directLinkedAnn.group;
+
+      const directLinkedAnnGroup = myGroups
```
hmstepanek: While reviewing I had to really think about why this was checking for the group in both lists. Maybe this was just me having a durp moment but I was thinking that might indicate that a comment might be in order here. I was really struggling with the why part of this which is actually in the comment after this line.

`// Check to see if the group has already been fetched.`

src/sidebar/services/groups.js
```diff
@@ -163,164 +163,144 @@ function groups(
...
+    const groups = filterGroups(
+      combineGroups(myGroups, featuredGroups, documentUri),
+      isLoggedIn,
+      directLinkedAnnotationGroupId,
+      directLinkedGroupId
+    );
+
+    injectOrganizations(groups);
+
+    // Step 5. Load the groups into the store and focus the appropriate
```
hmstepanek: The step comments help break up the problem. Just thinking out loud here but, they might be good markers for refactoring these into functions later.

src/sidebar/services/groups.js
```diff
@@ -163,164 +163,144 @@ function groups(
    *
    * @return {Promise<Group[]>}
    */
-  function load() {
-    let uri = Promise.resolve(null);
+  async function load() {
+    // Step 1: Get the URI of the active document, so we can fetch groups
+    // associated with that document.
+    let uri = null;
```
hmstepanek: I could be wrong but I think this code could be simplified by doing:
```
let documentUri = null;
.
.
.
    if (isSidebar) {
      documentUri = await getDocumentUriForGroupSearch();
    }
```

src/sidebar/services/groups.js
```diff
@@ -163,164 +163,144 @@ function groups(
...
-      uri = getDocumentUriForGroupSearch();
+      uri = await getDocumentUriForGroupSearch();
     }
-    const directLinkedGroupId = store.getState().directLinkedGroupId;
-    const directLinkedAnnId = store.getState().directLinkedAnnotationId;
+    documentUri = uri;
+
+    // Step 2: Concurrently fetch the groups the user is a member of,
+    // the groups associated with the current document and the annotation
+    // or group that was direct-linked (if any).
```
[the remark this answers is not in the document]
robertknight: It doesn't make sense to allow a direct link to an annotation AND a group, because the client can only follow one or the two instructions. The fact that a user can specify both in the client config, now that I reconsider it, a flaw in the API design. If we had designed it like this:

```js
{
  // Direct link to an annnotation...
  directLink: { 'type': 'annotation', id: '123' }, 

  // ...OR direct link to a group
  directLink: { 'type': 'group', id: '456' }, 
}
```

That would have eliminated that possibility.

Given that we've shipped this, I think we should simply decide that if both settings are present, one takes precedence over the other.

src/sidebar/services/groups.js
robertknight: Perhaps. The goal of this PR was to make the `load` function as "direct" as possible, ie. a set of sequential steps with minimal branching.

src/sidebar/services/groups.js
robertknight: I'll move the comment below up to make this clearer.

src/sidebar/services/groups.js
```diff
@@ -163,164 +163,144 @@ function groups(
...
-        // the groups associated with the current document and the annotation
-        // or group that was direct-linked (if any).
-        if (authority) {
-          params.authority = authority;
-        }
-        if (uri) {
-          params.document_uri = uri;
+    // If there is a direct-linked group, add an API request to get that
+    // particular group since it may not be in the results returned by
+    // group.list or profile.groups.
```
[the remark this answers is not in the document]
robertknight: I like this suggestion, but I also see a risk here that the comment gets out of date, so I've simplified it to just refer to groups that have been fetched by "other requests".

src/sidebar/services/groups.js
robertknight: You're right. I'll do this.

robertknight:
Please take another look when convenient @hmstepanek

src/sidebar/services/groups.js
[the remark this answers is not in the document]
hmstepanek: Here's the thread on slack where I considered what to do here: https://hypothes-is.slack.com/archives/C1M8NH76X/p1554397519016100. In summary, the direct-linked annotation does take precedents but it still may grab the direct-linked group that wasn't in the list of groups and add it to the list.

src/sidebar/services/groups.js
hmstepanek: Just a thought certainly not something I intended to be done for this particular PR. :)

src/sidebar/services/groups.js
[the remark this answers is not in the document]
hmstepanek: While I do like that suggested approach/change, I do think this comment still needs to be adjusted to accurately describe what's _currently_ happening.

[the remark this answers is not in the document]
hmstepanek: @robertknight This was the comment I was referring to that needs addressing.

robertknight:
Hi @hmstepanek - I have updated the comments and made some notes in the commits to reflect the fact that both an annotation and a group can currently be direct-linked. PTAL.

PR #1127, opened 2019-05-17
Fix/simplify manifest rebuild when running "gulp watch"

robertknight:
On my system "gulp watch" was failing to generate the client's entry
point script `build/boot.js` correctly. This may have happened after
recent upgrades to gulp or Node.

The issue was that `generateManifest` was run after "gulp
watch" started, before build/scripts/boot.bundle.js was generated,
producing an empty output file in build/boot.js. It was not re-run later
after build/scripts/boot.bundle.js was generated.

After some experimentation, removing use of `gulp-batch` in the
`watch-manifest` task resolved the issue. I haven't verified whether the
node upgrade or a gulp upgrade was the root cause. Since gulp-batch was
non-essential, I've just removed it to simplify things and reduce the
risk of similar issues happening in future.


gulpfile.js
```diff
@@ -347,11 +345,7 @@ function generateManifest() {
 }
 
 gulp.task('watch-manifest', function () {
-  gulp.watch(MANIFEST_SOURCE_FILES, batch(function (events, done) {
-    endOfStream(generateManifest(), function () {
-      done();
-    });
-  }));
+  gulp.watch(MANIFEST_SOURCE_FILES, { delay: 500 }, generateManifest);
```
robertknight: The use of `endOfStream` was redundant here because `generateManifest` returns a `Stream` and `gulp.watch` already knows how to watch a stream and mark a task as done when the stream finishes.

robertknight:
I've gone ahead and merged this as it was blocking my local development work.

PR #1135, opened 2019-05-24
Format gulpfile.js and files in scripts/ using Prettier

robertknight:
Eliminate an inconsistency in the client codebase that most files are
formatted with Prettier but files in scripts/ and gulpfile.js were not.


robertknight:
We already agreed to format JS files in the codebase using Prettier some time ago, so I'm just going to merge this once CI passes.

PR #1137, opened 2019-05-24
Make it easier to create custom builds of the client

robertknight:
Add a script `./scripts/build-app <OAuth client ID> [<h API URL>]` which will create a production build of the client using a given OAuth client ID which can then be served from the pre-registered URL associated with that OAuth client. This is useful for early-stage experiments or third-parties wanting to customize the client in ways that are not suitable for inclusion in the main codebase.

More generally, I think it makes sense that the client repo should document what a minimal version of the sidebar app's HTML should look like, even if we may need to generate a custom version in h or the browser extension in order to set the OAuth client ID.

 - Add a template for a minimal version of "app.html", the sidebar app's
   HTML file which is normally hosted by h. At build time, this template
   is processed and emitted into the "build" directory by the
   "build-html" task

 - Make it possible to serve the client from a URL without having to
   bake the URL into the client at build time by using
   `document.currentScript` to resolve the location of the client's
   assets. This avoids the need to rebuild the client if it is moved to
   a different URL.

I wrote this tool originally for creating an experimental fork of the client that supports embedded H5P activities (https://hypothesis-h5p.s3.us-east-2.amazonaws.com/index.html), but I think it is more generally useful.

**Testing**

1. Check that the embedded client at http://localhost:3000 still works
2. Check that a dev browser extension which bundles the client from this branch still works
3. Create and host a custom build of the client as follows:
   1. Register an OAuth client at http://localhost:5000/admin/oauthclients with the Redirect URL set to http://localhost:4040
   2. In the client repo on this branch run `scripts/build-app <OAuth client ID> http://localhost:5000/api`
   3. In the client repo run ``python3 -m http.server 4040``
   4. Create an HTML file ``test.html`` with the following content in the root of the client repo:
        ```html
        <h1>Hello World</h1>
        <script src="http://localhost:4040/build/boot.js"></script>
        ```
   5. Navigate to http://localhost:4040/test.html and check that the custom build of the client loads and that you can log in

With Step 3, you can also register a custom client that works with the production Hypothesis service at https://hypothes.is/admin/oauthclients. The custom client build can then be hosted in any static hosting site -  for example GitHub Pages, S3 etc.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -4,6 +4,7 @@
 
 const { mkdirSync, readdirSync } = require('fs');
 const path = require('path');
+const url = require('url');
 
 const changed = require('gulp-changed');
 const commander = require('commander');
@@ -209,6 +210,24 @@ gulp.task(
   })
 );
 
+gulp.task('build-html', () => {
+  const clientIdRequiredMsg =
+    'You must configure an OAuth client ID with the "OAUTH_CLIENT_ID" env var';
+  const apiUrl = process.env.API_URL || 'https://hypothes.is/api/';
+  const config = {
+    apiUrl,
+    authDomain: url.parse(apiUrl).hostname,
+
+    oauthClientId: process.env.OAUTH_CLIENT_ID || clientIdRequiredMsg,
+  };
+  const htmlSafeJsonConfig = JSON.stringify(config).replace(/</g, '\\u003c');
+  return gulp
+    .src('src/sidebar/app.html')
+    .pipe(replace('__CONFIG__', htmlSafeJsonConfig))
+    .pipe(rename('app.html'))
+    .pipe(gulp.dest('./build'));
+});
+
 const MANIFEST_SOURCE_FILES =
   'build/@(fonts|images|scripts|styles)/*.@(js|css|woff|jpg|png|svg)';
 
@@ -235,7 +254,7 @@ function generateBootScript(manifest, { usingDevServer = false } = {}) {
   } else {
     defaultAssetRoot = '{current_scheme}://{current_host}:3001/hypothesis';
   }
-  defaultAssetRoot = `${defaultAssetRoot}/${version}/`;
+  defaultAssetRoot = `${defaultAssetRoot}/${version}/build/`;
 
   if (isFirstBuild) {
     log(`Sidebar app URL: ${defaultSidebarAppUrl}`);
@@ -295,6 +314,7 @@ const buildAssets = gulp.parallel(
   'build-js',
   'build-css',
   'build-fonts',
+  'build-html',
   'build-images'
 );
 gulp.task('build', gulp.series(buildAssets, generateManifest));
--- a/src/boot/index.js
+++ b/src/boot/index.js
@@ -17,13 +17,44 @@ import { isBrowserSupported } from './browser-check';
 
 if (isBrowserSupported()) {
   const settings = parseJsonConfig(document);
+  // Use the asset root and sidebar app locations specified in the host page,
+  // if present. This is used by the Hypothesis browser extensions to make the
+  // boot script load assets bundled with the extension.
+  let assetRoot;
+  if (settings.assetRoot) {
+    // The `assetRoot` setting is assumed to point at the root of the contents of
+    // the npm package.
+    assetRoot = settings.assetRoot + 'build/';
+  }
+  let sidebarAppUrl = settings.sidebarAppUrl;
+
+  // Otherwise, try to determine the default root URL for assets and the sidebar
+  // application from the location where the boot script is hosted.
+  try {
+    const script = /** @type {HTMLScriptElement} */ (document.currentScript);
+    let scriptUrl = new URL(script.src);
+
+    // We only use the bundled sidebar HTML and assets if the boot script has
+    // its original name. If the `<script>` tag references a custom name
+    // (eg. as in "https://hypothes.is/embed.js") then we skip this and fall
+    // back to the URLs embedded in the boot script.
+    if (scriptUrl.pathname.endsWith('/boot.js')) {
+      assetRoot = assetRoot || new URL('./', scriptUrl).href;
+      sidebarAppUrl = sidebarAppUrl || new URL('app.html', scriptUrl).href;
+    }
+  } catch (e) {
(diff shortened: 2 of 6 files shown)


judell:
>  useful for early-stage experiments

Very useful. 

@LMS007, this is one of the methods for decoupling client development from the service. I've cherry-picked an early version of it a couple of times. it will be great to get it merged. Would've saved you a couple of days of headbanging.

robertknight:
I've updated this branch against recent changes in the gulpfile and to simplify it a little. This has been sitting around a while but is still relevant so I'm going to get it finished. I still need to:

- [ ] Test in IE
- [ ] Test with the browser extension
- [ ] Review the current code against the PR description and check that the description is correct

src/boot/index.js
```diff
@@ -16,10 +16,40 @@ import processUrlTemplate from './url-template';
...
+try {
+  let scriptUrl = new URL(document.currentScript.src);
+
+  // We only use the bundled sidebar HTML and assets if the boot script has
+  // its original name. If the `<script>` tag references a custom name
+  // (eg. as in "https://hypothes.is/embed.js") then we skip this and fall
+  // back to the URLs embedded in the boot script.
+  if (scriptUrl.pathname.endsWith('/boot.js')) {
+    assetRoot = assetRoot || new URL('./', scriptUrl).href;
+    sidebarAppUrl = sidebarAppUrl || new URL('app.html', scriptUrl);
```
bibliotechie:
```suggestion
    sidebarAppUrl = sidebarAppUrl || new URL('app.html', scriptUrl).href;
```

cgorringe:
I'm currently building on a fork of this branch since I use the `build-app` script, and was wondering if/when this could be merged to master?

src/sidebar/app.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
```
LMS007, line 1: My first instinct is to not include `build-html` with `build`. I guess that means this file gets produced for every build now and it's a little unnecessary for anything other than a custom build correct?

 Its confusing because our own app.html is owned by h./ and this file is really only used for ad-hoc hosting of the a forked client.

I'm not really giving direction here, but just speaking to the larger picture.  Here is kind of what I was thinking....

```
const buildCustomAssets = gulp.parallel(
  'build-js',
  'build-css',
  'build-fonts',
  'build-html',
  'build-images'
);
const buildAssets = gulp.parallel(
  'build-js',
  'build-css',
  'build-fonts',
  'build-images'
);
gulp.task('build', gulp.series(buildAssets, generateManifest));
gulp.task('build-custom', gulp.series(buildAssets, generateManifest));
```

And then the script calls `build-custom` for example... in my mind, then the /build dir would more accurately reflect what was run rather than being a swiss army know output.

LMS007 (commented):
I think this is really good start, just had some general feedback after giving it some testing and examination.

I also had a hell of a time trying to get it to work running my own server because from what I understand, cloudflare adds headers that I think may be missing and necessary for CORS to work correctly.

Screen Shot 2020-10-22 at 10 06 14 AM


I noticed that my local instance of h./ was not returning `access-control-allow-origin: *` from the /api call so Chome was blowing up. This of course in most cases won't be much of an issue because I take it most devs are going to want to connect to our own h, but if a dev wanted to use a hosted h/, then it seems like ours is not set up to be CORS friendly with app.html hosted on another port as near as I can tell.

The final point is we just need to think/have a plan for getting oauth ids to devs that need them to connect to our h. I just just used the one from https://hypothes.is/docs/help in my DOM and I was able to login. But we may want something a bit more polished :)

src/boot/index.js
```diff
@@ -9,18 +9,59 @@
...
+}
+let sidebarAppUrl = settings.sidebarAppUrl;
+
+// Otherwise, try to determine the default root URL for assets and the sidebar
+// application from the location where the boot script is hosted.
+try {
+  const script = /** @type {HTMLScriptElement} */ (document.currentScript);
+  let scriptUrl = new URL(script.src);
+
+  // We only use the bundled sidebar HTML and assets if the boot script has
```
LMS007, line 46: I don't propose a better solution, but this does feel fragile and hard to understand.  I wonder if there is something that can be driven from the config itself to trigger this condition to make it feel a bit more explicit.

src/boot/index.js
```diff
@@ -17,13 +17,44 @@ import { isBrowserSupported } from './browser-check';
...
+  // application from the location where the boot script is hosted.
+  try {
+    const script = /** @type {HTMLScriptElement} */ (document.currentScript);
+    let scriptUrl = new URL(script.src);
+
+    // We only use the bundled sidebar HTML and assets if the boot script has
+    // its original name. If the `<script>` tag references a custom name
+    // (eg. as in "https://hypothes.is/embed.js") then we skip this and fall
+    // back to the URLs embedded in the boot script.
+    if (scriptUrl.pathname.endsWith('/boot.js')) {
```
LMS007, line 41: I don't propose a better solution, but this does feel fragile and hard to understand.  I wonder if there is something that can be driven from the config itself to trigger this condition to make it feel a bit more explicit.

gulpfile.js
```diff
@@ -216,6 +217,24 @@ gulp.task(
...
+  const clientIdRequiredMsg =
+    'You must configure an OAuth client ID with the "OAUTH_CLIENT_ID" env var';
+  const apiUrl = process.env.API_URL || 'https://hypothes.is/api/';
+  const config = {
+    apiUrl,
+    authDomain: url.parse(apiUrl).hostname,
+
+    oauthClientId: process.env.OAUTH_CLIENT_ID || clientIdRequiredMsg,
+  };
+  const htmlSafeJsonConfig = JSON.stringify(config).replace(/</g, '\\u003c');
```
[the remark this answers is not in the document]
robertknight: Even though it is unlikely that the JSON configuration will be unsafe for naive insertion into HTML, I think we should always do the proper safe serialization just in case.

tgiardina:
There's a reference to `/build/styles/annotator.css` at the top of `src/annotator/util/shadow-root.js`. When creating a custom build, this must be `/build/styles/annotator.css`, otherwise the adder css won't load correctly.

PR #1145, opened 2019-05-29
Move firstKey function from sidebar-content to store

hmstepanek:
- add getFirstSelectedAnnotationId to selection store
- move tests from index-tests to a new selection-tests module specific to the selection module

This is part of https://github.com/hypothesis/product-backlog/issues/1016.


src/sidebar/store/modules/selection.js
```diff
@@ -314,6 +314,22 @@ function clearSelectedAnnotations() {
...
 
+/**
+ * Returns the annotation ID of the first annotation in `selectedAnnotationMap`.
+ *
+ * @return {string|null}
+ */
+function getFirstSelectedAnnotationId(state) {
+  const selection = state.selectedAnnotationMap || {};
+  for (const annId in selection) {
+    if (!selection.hasOwnProperty(annId)) {
```
hmstepanek: Wondering if this ever happens?

hmstepanek: Seems the short answer is no, it is always true or not present but maybe leave it for now until the data structure is potentially adjusted at sometime in the future.

PR #1148, opened 2019-05-29
Implement new groups menu design built on groups menu

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/1146**~~

This is the initial implementation of a new design for the groups menu based on the recently added `Menu`, `MenuItem` and `MenuSection` components. It re-introduces actions to go to the activity page for a group and leave a group directly from the groups menu, but in a way that fixes the previous UX problems with small icons and an activity page link that was too easy to click accidentally.

The "Copy to clipboard" function from the mocks is missing in this PR. It will be added later. The menus will also need some further updates to match the latest designs once this lands.

Screenshot 2019-05-29 19 41 27

The main commit removes quite a lot of code and CSS overall as much of the functionality and styling has been moved from group list-related components to the shared menu.


src/sidebar/components/group-list-item.js
```diff
@@ -22,53 +39,82 @@ function GroupListItem({ analytics, group }) {
...
-            group.type === 'private'
-              ? `Show and create annotations in ${group.name}`
-              : 'Show public annotations'
-          }
-        >
-          {group.name}
-        </a>
-      </div>
-    </div>
+    <Fragment>
```
lyzadanger: Mentally parsing what kind of markup/DOM structure results from this rendering was a little challenging. It may help to leave a comment with the rough structure of what results here. I believe it may be:

```
<MenuItem />
{isExpanded && (<ul>...submenu</ul>
  {!isSelectable && (<p>URL restriction disclaimer</p>)}
}
```

Meaning that you end up with 1-3 sibling elements which consist of: `<MenuItem />` (always), `<ul>` (when expanded), `<p>` (if URL-restricted and expanded). Is that correct?

It can be derived from the JSX, of course, but for me the `<Fragment>`s make it harder to parse easily. Took me a few minutes and I had to run the branch and inspect the DOM to grok.

src/sidebar/components/group-list-item.js
```diff
@@ -22,53 +39,82 @@ function GroupListItem({ analytics, group }) {
...
+        isExpanded={isExpanded}
+        isSelected={isSelected}
+        isSubmenuVisible={isExpanded}
+        label={group.name}
+        onClick={isSelectable ? focusGroup : toggleSubmenu}
+        onToggleSubmenu={toggleSubmenu}
+      />
+      {isExpanded && (
+        <Fragment>
+          <ul onClick={collapseSubmenu}>
```
lyzadanger: I think I was also thrown here by the use of a plain ol' `ul` instead of a `<Menu>`. It didn't occur to me that you could use `<MenuItem>`s outside of a  (direct) `<Menu>` context.

src/sidebar/components/group-list-v2.js
```diff
@@ -0,0 +1,119 @@
...
+  );
+
+  const { authDomain } = settings;
+  const canCreateNewGroup = userid && !isThirdPartyUser(userid, authDomain);
+  const newGroupLink = serviceUrl('groups.new');
+
+  let label;
+  if (focusedGroup) {
+    const icon = focusedGroup.organization.logo;
+    label = (
```
lyzadanger: I wonder if there is any way to tap into the `SVGIcon` component here in the future. I know that right now `SVGIcon` depends on a hard-coded mapping of `name => svg source` but I wonder if it could be expanded upon to make rendering these plausible...

src/sidebar/components/group-list-v2.js
```diff
@@ -0,0 +1,119 @@
...
+  } else {
+    label = <span>…</span>;
+  }
+
+  // If there is only one group and no actions available for that group,
+  // just show the group name as a label.
+  const actionsAvailable = !isThirdPartyService(settings);
+  if (
+    !actionsAvailable &&
+    currentGroups.length + featuredGroups.length + myGroups.length < 2
```
lyzadanger: `<= 1` might be easier to read here? Also, I know that order of operations means that this will evaluate in the anticipated order, but perhaps wrapping the "left side" of the equation here in parens would be helpful for human parsing?

`(currentGroups.length + featuredGroups.length + myGroups.length) < 2`

src/sidebar/components/group-list-item.js
```diff
@@ -22,53 +39,82 @@ function GroupListItem({ analytics, group }) {
...
+      )}
+    </Fragment>
   );
 }
 
 GroupListItem.propTypes = {
   group: propTypes.object.isRequired,
 
   analytics: propTypes.object.isRequired,
+  groups: propTypes.object.isRequired,
```
lyzadanger: What is `groups`? Aha! I got confused between a list of groups and the groups service. This is the latter.

src/sidebar/components/test/group-list-item-test.js
```diff
@@ -29,6 +41,10 @@ describe('GroupListItem', () => {
       orgName: sinon.stub(),
     };
 
+    fakeGroups = {
```
lyzadanger: For me this highlighted the confusion in naming overload for `groups`: it can mean either the groups service or a list of groups. In this case, I'm assuming it's the service.

src/sidebar/components/test/group-list-item-test.js
```diff
@@ -43,55 +59,58 @@ describe('GroupListItem', () => {
...
       />
     );
   };
 
   it('changes the focused group when group is clicked', () => {
-    const fakeGroup = { id: 'groupid' };
-
     const wrapper = createGroupListItem(fakeGroup);
-    wrapper.find('.group-list-item__item').simulate('click');
+    wrapper
```
lyzadanger: You could do this with `shallow` rendering, too, no?

src/sidebar/components/test/group-list-item-test.js
```diff
@@ -43,55 +59,58 @@ describe('GroupListItem', () => {
...
+      .find('MenuItem')
+      .props()
+      .onClick();
```
lyzadanger: Didn't know you could do this directly like this in one statement!

src/styles/variables.scss
```diff
@@ -126,7 +126,7 @@ $highlight-color-second: rgba(206, 206, 60, 0.4);
 $highlight-color-third: rgba(192, 192, 49, 0.4);
 $highlight-color-focus: rgba(156, 230, 255, 0.5);
 $top-bar-height: 40px;
-$group-list-width: 300px;
+$group-list-width: 280px;
```
lyzadanger: Hmmmm....locally, the menu is rendering at <200px wide for me (i.e. too narrow):

image

src/sidebar/components/group-list-item.js
```diff
@@ -22,53 +39,82 @@ function GroupListItem({ analytics, group }) {
     actions.focusGroup(group.id);
   };
 
-  const focusedGroupId = useStore(store => store.focusedGroupId());
-  const isSelected = group.id === focusedGroupId;
-  const groupOrgName = orgName(group);
+  const leaveGroup = () => {
+    const message = `Are you sure you want to leave the group "${group.name}"?`;
+    if (window.confirm(message)) {
```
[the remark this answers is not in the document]
robertknight: This is just the `window` global that is available to any JS running in a browser context. Our module/build system doesn't do anything to hide those.

src/sidebar/components/group-list-item.js
robertknight: Yeah. It is a bit confusing in this context ... but we can't rename it here because dependency injection looks up the service by name. It can be aliased to another name in the destructuring that happens in the `function GroupListItem(...)` line though, eg. `{ groups: groupsService, ... }`

src/sidebar/components/group-list-v2.js
robertknight: `<SvgIcon>` currently renders an `<svg>` inline for CSS style-ability. Here we use an `` a) because this needs to be loaded from a remote source and b) because this is untrusted publisher-provided markup.

src/sidebar/components/test/group-list-item-test.js
robertknight: I'll rename it to `fakeGroupsService` where possible.

src/styles/variables.scss
robertknight: This is resolved. I've added the ability for a `Menu` consumer to specify a custom class for the dropdown content area and used that to set a minimum width for the groups menu.

src/sidebar/components/group-list-item.js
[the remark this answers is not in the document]
lyzadanger: Ah. I was way over-thinking this.

src/sidebar/components/test/group-list-item-test.js
robertknight: I ran into an issue with that. IIRC it used to be the case that `GroupListItem` was a leaf component, which is why it used `mount` originally. Trying to switch it to `shallow` just now I ran into a limitation of Enzyme's shallow rendering to do with calling `wrapper.update()` [1] with a component that renders multiple elements at the top level.

[1] `wrapper.update()` synchronizes Enzyme's view of the rendered output from a component with the actual rendered output. This is called automatically when using eg. `wrapper.simulate(...)` but needs to be called manually if you do something which causes a re-render without Enzyme being aware of it.

robertknight: I've added a note about that issue

src/sidebar/components/test/group-list-v2-test.js
```diff
@@ -0,0 +1,166 @@
...
+    );
+  });
+
+  it('displays "My Groups" section if user is a member of any groups', () => {
+    fakeStore.getMyGroups.returns([testGroup]);
+    const wrapper = createGroupList();
+    assert.isTrue(wrapper.exists('GroupListSection[heading="My Groups"]'));
+  });
+
+  it('sorts groups within each section by organization', () => {
```
lyzadanger: I'm having a little trouble discerning exactly what this is testing...I'm wondering if the term "sort" in the test description is misleading. Is this test checking to make sure that groups are _slotted_ into the appropriate section?

robertknight: This is about making sure that the order of groups within a section matches what `groupOrganizations(groupsForSection)` returns rather than say, being random or whatever came back from the server.

PR #1155, opened 2019-06-05
Move onForceVisible to annotation-thread

hmstepanek:
Move the implementation of onForceVisible into where it's used in the
annotation-thread component instead of passing it through several layers
of components. This simplifies the implementation and makes the code
easier to follow.

### DIFF
--- a/src/sidebar/components/stream-content.js
+++ b/src/sidebar/components/stream-content.js
@@ -59,9 +59,6 @@ function StreamContentController(
   fetch(20);
 
   this.setCollapsed = store.setCollapsed;
-  this.forceVisible = function(id) {
-    store.setForceVisible(id, true);
-  };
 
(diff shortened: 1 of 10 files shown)


src/sidebar/components/stream-content.js
```diff
@@ -59,9 +59,6 @@ function StreamContentController(
   fetch(20);
 
   this.setCollapsed = store.setCollapsed;
-  this.forceVisible = function(id) {
```
hmstepanek: I find this implementation confusing and I'm not sure how it was working as expected before?

robertknight: It looks like this was indeed broken, given that the corresponding method in `<sidebar-content>` expects a thread rather than a thread ID.

PR #1158, opened 2019-06-07
Remove wrapping `<span>` around SVG icons

robertknight:
This makes it easier to get icons positioned correctly because there is one
less element in the hierarchy to deal with and any class passed to
`SvgIcon` is applied to the root element rendered by that component.

Unfortunately React/Preact don't provide a way to trivially return an HTML
or SVG string as the output from a component without a wrapping element.

The implementation therefore works by first using the browser's HTML parser (via
`innerHTML`) to parse the SVG markup and then converting the returned
DOM tree to a Preact element, which the `SvgIcon` component returns as
its render output.

### DIFF
--- a/src/sidebar/components/svg-icon.js
+++ b/src/sidebar/components/svg-icon.js
@@ -1,7 +1,7 @@
 'use strict';
 
-const { createElement } = require('preact');
-const { useLayoutEffect, useRef } = require('preact/hooks');
+const { cloneElement, createElement } = require('preact');
+const { useMemo } = require('preact/hooks');
 const propTypes = require('prop-types');
 
 // The list of supported icons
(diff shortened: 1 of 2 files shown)


robertknight:
To expand on the PR description, the layout problem this PR fixes is that the wrapping span previously had inline layout, which meant that its height depended on the height of line boxes in the element, which depends on the computed `line-height`. In the context of the top bar for example, this worked out at 18.4px, rounded down to 18px, which is taller than the 15px height of the icons themselves.

Removing the wrapping span is one way to make the height of the elements rendered by `SvgIcon` match the `<svg>` content. The other, which Lyza and me discussed, is to apply `display: flex` or `display: block; line-height: 0` to the `<span>`. The latter approach is a bit simpler so it is what we've gone with for the moment.

I think this PR may still have a slight performance advantage in that it avoids changing the CSS class of the content after it is rendered, and thereby potentially triggering some style recalculation.

robertknight:
I'm closing this for now as we've found an alternative solution that was simpler to implement. If we find reasons to pursue this again, we can reopen it.

PR #1159, opened 2019-06-07
Open sidebar when highlighting while logged-out

hmstepanek:
When the user creates a highlight while logged out, they appear to
work however, they silently do not get saved. An error message
does pop up in the sidebar informing them they cannot create annotations
when not logged in but they don't see it because the sidebar does not
open when creating a highlight. This makes the error message more
apparent to the user by popping open the sidebar.

Previously the error message only mentioned annotations now it says,
"You must be logged in to create annotations **and highlights**.".

This is a slight adjustment from https://github.com/hypothesis/client/pull/967 but the same concept.

### DIFF
--- a/src/sidebar/components/annotation.js
+++ b/src/sidebar/components/annotation.js
@@ -53,6 +53,7 @@ function AnnotationController(
   store,
   annotationMapper,
   api,
+  bridge,
   drafts,
   flash,
   groups,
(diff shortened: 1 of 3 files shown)


dwhly:
+1

hmstepanek:
Just following up with notes based on our discussion. Yes! 100% agree with the annotation card after login issue. Also I noticed there's a slight annoyance here where the "You must be logged in to create annotations and highlights" message doesn't always show up at the top of the annotation list depending on where the user attempts to highlight. I think this should behave as the other error messages do where the annotation message card is "selected" and thus the other annotations are hidden and it appears at the top.

PR #1169, opened 2019-06-11
Make sidebar buttons transparent

diegodlh:
The three buttons protruding from the sidebar sometimes cover content from the underlying webpages (e.g. https://en.wikipedia.org/wiki/Hypothes.is, covering part of the "Log in" link at the top-right corner). Making these buttons transparent (though still solid on hover) may help seeing the underlying content.
I wasn't sure what the transparency should be (I set it to 0.7), or whether I had to file an issue/feature request first.

Previously:
image

Now:
image

### DIFF
--- a/src/styles/annotator/annotator.scss
+++ b/src/styles/annotator/annotator.scss
@@ -108,7 +108,7 @@ $base-font-size: 14px;
   .annotator-frame-button {
     transition: background-color .25s .25s;
     @include smallshadow;
-    background: $white;
+    background: rgba(255, 255, 255, 0.7);
     border: solid 1px $gray-lighter;
     border-radius: 4px;
     color: $gray-light;
--- a/src/styles/annotator/annotator.scss
+++ b/src/styles/annotator/annotator.scss
@@ -119,15 +119,15 @@ $base-font-size: 14px;
     font-size: 16px;
     margin-bottom: 5px;
 
-    &:active {
-      transition: background-color .25s;
-      background-color: $gray-light;
-    }
-
     &:focus, &:hover {
       background: $white;
       outline: 0;
       color: $text-color;
+    }
+
+    &:active {
+      transition: background-color .25s;
+      background-color: $gray-light;
     }
   }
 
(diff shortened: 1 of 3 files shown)


hmstepanek:
Hi @diegodlh, 
Thanks for the contribution!

Just some preliminary thoughts: I think if we are going to make the buttons transparent we should also make the annotation counts transparent as well. Notice 4 in the screenshot is not transparent which makes it really the only thing left on that bar that isn't.
image

I think we'll have to discuss this change as a team and get back to you so it might be a couple weeks before you hear anything from us again.

diegodlh:
Hi Hannah! Thank you for taking the time to consider my suggestion.
It is just an idea, based on my regular use of Hypothesis. However, I perfectly understand this kind of suggestions have to be discussed with the team before deciding whether to include them or not. I hope you find it useful!
I made some further changes:
- af6ae5e: Used CSS variables instead of hard-coding the background color, as suggested in your [CSS guidelines](https://github.com/hypothesis/frontend-toolkit/blob/master/docs/css-style-guide.md).
- d778cf8: There was a transition effect on the background-color property. Added the color property as well, as it looked weird otherwise.
Before:
before
After:
after
- 44e0335: When I committed 6a9bbf9 to make toolbar buttons solid again on hover, they would no longer change to gray on active (click). Changed CSS pseudo-class order from active > focus-hover to focus-hover > active, as suggested [here](https://bitsofco.de/when-do-the-hover-focus-and-active-pseudo-classes-apply/).
- d5e123d: As per your suggestion, I changed the bucket-bar indicators to be transparent as well. However, the triangle to the left, top or bottom remains solid white:
triangle
I tried to fix it, playing with border-right, border-bottom or border-top properties in :before and :after pseudo-elements, but if I managed to make the triangle transparent, it would lose its gray border. I could consider this more deeply if you think this PR makes sense.

Btw, let me ask you: in future cases, is it OK to propose small changes like this (i.e. ideas, not bug-fixes) with a pull request? Or would you prefer that I send a draft pull request instead? Or file an issue before the PR?

Best,
Diego

PR #1172, opened 2019-06-13
Remove passing of search

hmstepanek:
Removing passing of search from hypothesis-app component to search-status-bar component and just use store directly.

This is a partial fix for https://github.com/hypothesis/product-backlog/issues/1016.

This is dependent on https://github.com/hypothesis/client/pull/1161.

### DIFF
--- a/src/sidebar/components/search-status-bar.js
+++ b/src/sidebar/components/search-status-bar.js
@@ -37,15 +37,21 @@ function SearchStatusBarController(store, rootThread) {
     }
     return Object.keys(selection).length > 0;
   };
+
+  this.filterQuery = function() {
+    return store.getState().filterQuery;
+  };
+
+  this.filterActive = function() {
+    return !!store.getState().filterQuery;
+  };
 }
 
 module.exports = {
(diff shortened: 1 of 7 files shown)


src/sidebar/components/search-status-bar.js
```diff
@@ -37,15 +37,21 @@ function SearchStatusBarController(store, rootThread) {
     }
     return Object.keys(selection).length > 0;
   };
+
+  this.searchQuery = function() {
```
robertknight: I think we should take the opportunity to rename this `filterQuery` for consistency with the state key name and the `filterActive` function. There is the wider issue of perhaps renaming the entire component to reflect the fact that it really is a filter rather than a search (the difference IMO being that it always shows a subset of what you see when no filter is active, and the ordering of annotations is not affected).

hmstepanek: That makes sense to me. I'm going to leave the renaming of the component for another PR though as there are several components with search in the name so it'd be a pretty big change from what's already here.

PR #1186, opened 2019-06-20
Refactor clearSelection out of sidebar-content and into the store

hmstepanek:
Move clearSelection out of the sidebar-content and into the store so that it's not getting passed around from the sidebar-content to child components. 

This is a partial fix for https://github.com/hypothesis/product-backlog/issues/1028.

### DIFF
--- a/src/sidebar/store/modules/selection.js
+++ b/src/sidebar/store/modules/selection.js
@@ -64,6 +64,29 @@ function freeze(selection) {
   }
 }
 
+const setTab = (selectedTab, newTab) => {
+  // Do nothing if the "new tab" is not a valid tab.
+  if (
+    [
+      uiConstants.TAB_ANNOTATIONS,
+      uiConstants.TAB_NOTES,
+      uiConstants.TAB_ORPHANS,
+    ].indexOf(newTab) === -1
+  ) {
+    return {};
+  }
+  // Shortcut if the tab is already correct, to avoid resetting the sortKey
+  // unnecessarily.
+  if (selectedTab === newTab) {
+    return {};
+  }
+  return {
+    selectedTab: newTab,
+    sortKey: TAB_SORTKEY_DEFAULT[newTab],
+    sortKeysAvailable: TAB_SORTKEYS_AVAILABLE[newTab],
+  };
+};
+
 function init(settings) {
   return {
     // Contains a map of annotation tag:true pairs.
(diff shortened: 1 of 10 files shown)


src/sidebar/store/modules/selection.js
```diff
@@ -97,7 +120,28 @@ function init(settings) {
 }
 
 const update = {
-  CLEAR_SELECTION: function() {
+  CLEAR_SELECTION: function(state) {
+    let selectedTab = state.selectedTab;
+    if (selectedTab === uiConstants.TAB_ORPHANS) {
```
hmstepanek: There was a !selectedTab || selectedTab === uiConstants.TAB_ORPHANS in here before but I don't see how selectedTab could be falsey.

src/sidebar/store/modules/selection.js
```diff
@@ -97,7 +120,22 @@ function init(settings) {
...
+      selectedTab = uiConstants.TAB_ANNOTATIONS;
+    }
+    const tabSettings = setTab(state.selectedTab, selectedTab);
+    return Object.assign(
```
robertknight: Pro-tip: If you want to merge two objects together you can do that a bit more simply in new code using "object spread" syntax:

```js
return { filterQuery: null, selectedAnnotationMap: null, ...tabSettings };
```

hmstepanek:
Ack! Sorry, it must have dropped the removal of the onClearSelection binding change during my final rebase cleanup. Should be fixed now!

src/sidebar/store/modules/selection.js
```diff
@@ -64,6 +64,29 @@ function freeze(selection) {
...
+  }
+  // Shortcut if the tab is already correct, to avoid resetting the sortKey
+  // unnecessarily.
+  if (selectedTab === newTab) {
+    return {};
+  }
+  return {
+    selectedTab: newTab,
+    sortKey: TAB_SORTKEY_DEFAULT[newTab],
+    sortKeysAvailable: TAB_SORTKEYS_AVAILABLE[newTab],
```
robertknight: `sortKeysAvailable` could be refactored into a selector in future, since it can always be determined from the selected tab and doesn't need to be maintained as a separate piece of state.

src/sidebar/store/test/index-test.js
```diff
@@ -61,6 +61,63 @@ describe('store', function() {
     });
   });
 
+  describe('clearSelection', () => {
+    // Test clearSelection here over the entire store as it triggers the
+    // CLEAR_SELECTION action in multiple store modules.
```
robertknight: Hmm ... I've mixed thoughts about this. On the one hand, it is useful that all the tests for this function are in one place. On the other, someone looking at just the tests for a particular module might think that this action is untested.

hmstepanek: Yeah I did too but I felt like the integrationy store tests should go in the index and while some of these tests could have gone in the corresponding individual modules I thought it was simpler to keep them all in one place. We could put a comment into the individual modules saying look here for these tests or something.

PR #1193, opened 2019-06-24
Separate hover areas for main action and submenu toggle

robertknight:
For menu items that have associated submenus, make it clearer that there
are two separate click zones by highlighting only the clickable area for
the main action when hovering that part of the item.

(See commit message for implementation notes)

_I'm not really sure whether I prefer this change vs the version on master. I'm putting up this PR to solicit feedback._

----

**Before (group name/icon hovered):**

before-main-action

**Before (submenu toggle hovered):**

before-submenu-toggle

**After (group name/icon hovered):**

after-main-action

**After (submenu toggle hovered):**

after-submenu-toggle

### DIFF
--- a/src/sidebar/components/menu-item.js
+++ b/src/sidebar/components/menu-item.js
@@ -69,23 +69,25 @@ function MenuItem({
       role="menuitem"
       {...(onClick && onActivate('menuitem', onClick))}
     >
-      {hasLeftIcon && (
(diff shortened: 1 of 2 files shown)


dwhly:
Nice

robertknight:
Okay, I'm going to deploy this. We can always change it later.

PR #1195, opened 2019-06-25
Remove old groups menu and "community_groups" feature flag checks

robertknight:
This PR removes the remaining feature flag checks for the "community_groups" menu from `src/sidebar/services/groups.js`, as well as the old implementation of the `<groups-list>` component.

The new implementation is still in a module called `groups-list-v2.{js, scss}`. To make the diff easier to follow I didn't rename those files in this commit but will do that in a follow-up PR.


robertknight:
Thanks. I think I'll defer merging this until tomorrow morning just to give a little more time to let the dust settle.

PR #1211, opened 2019-06-27
Move load annotations to service

hmstepanek:
Extract the loading of annotations from the sidebar-content component into a service.
Remove the group handling logic for focusing a direct-linked annotation's group as this is handled in the groups service now.
Improve tests.

Dependent on https://github.com/hypothesis/client/pull/1215 and https://github.com/hypothesis/client/pull/1220.
Partial fix for https://github.com/hypothesis/product-backlog/issues/1028 and https://github.com/hypothesis/product-backlog/issues/1016.

### DIFF
--- a/src/sidebar/components/sidebar-content.js
+++ b/src/sidebar/components/sidebar-content.js
@@ -233,16 +233,20 @@ function SidebarContentController(
   // Re-fetch annotations when focused group, logged-in user or connected frames
   // change.
   $scope.$watch(
-    () => [groups.focused(), store.profile().userid, ...store.searchUris()],
-    ([currentGroup], [prevGroup]) => {
-      if (!currentGroup) {
-        // When switching accounts, groups are cleared and so the focused group
+    () => [
+      store.focusedGroupId(),
+      store.profile().userid,
+      ...store.searchUris(),
+    ],
+    ([currentGroupId], [prevGroupId]) => {
+      if (!currentGroupId) {
+        // When switching accounts, groups are cleared and so the focused group id
         // will be null for a brief period of time.
         store.clearSelectedAnnotations();
         return;
       }
 
-      if (!prevGroup || currentGroup.id !== prevGroup.id) {
+      if (!prevGroupId || currentGroupId !== prevGroupId) {
         // The focused group may be changed during loading annotations as a result
         // of switching to the group containing a direct-linked annotation.
         //
--- a/src/sidebar/components/sidebar-content.js
+++ b/src/sidebar/components/sidebar-content.js
@@ -1,38 +1,19 @@
 'use strict';
 
-const SearchClient = require('../search-client');
 const events = require('../events');
 const isThirdPartyService = require('../util/is-third-party-service');
 const tabs = require('../tabs');
-
-/**
- * Returns the group ID of the first annotation in `results` whose
- * ID is `annId`.
- */
-function getGroupID(annId, results) {
-  const annot = results.find(function(annot) {
-    return annot.id === annId;
-  });
-  if (!annot) {
-    return null;
-  }
-  return annot.group;
-}
 
 // @ngInject
 function SidebarContentController(
   $scope,
   analytics,
+  annotations,
   store,
-  annotationMapper,
-  api,
-  features,
   frameSync,
-  groups,
   rootThread,
   settings,
-  streamer,
-  streamFilter
+  streamer
 ) {
   const self = this;
 
@@ -90,61 +71,7 @@ function SidebarContentController(
     }
   }
 
-  const searchClients = [];
-
-  function _resetAnnotations() {
-    annotationMapper.unloadAnnotations(store.savedAnnotations());
-  }
-
-  function _loadAnnotationsFor(uris, group) {
-    const searchClient = new SearchClient(api.search, {
-      // If no group is specified, we are fetching annotations from
-      // all groups in order to find out which group contains the selected
-      // annotation, therefore we need to load all chunks before processing
-      // the results
-      incremental: !!group,
-    });
-    searchClients.push(searchClient);
-    searchClient.on('results', function(results) {
-      if (store.hasSelectedAnnotations()) {
-        // Focus the group containing the selected annotation and filter
-        // annotations to those from this group
-        let groupID = getGroupID(store.getFirstSelectedAnnotationId(), results);
-        if (!groupID) {
-          // If the selected annotation is not available, fall back to
-          // loading annotations for the currently focused group
-          groupID = groups.focused().id;
-        }
-        results = results.filter(function(result) {
-          return result.group === groupID;
-        });
-        groups.focus(groupID);
-      }
-
-      if (results.length) {
-        annotationMapper.loadAnnotations(results);
-      }
-    });
-    searchClient.on('end', function() {
-      // Remove client from list of active search clients.
-      //
-      // $evalAsync is required here because search results are emitted
-      // asynchronously. A better solution would be that the loading state is
-      // tracked as part of the app state.
-      $scope.$evalAsync(function() {
-        searchClients.splice(searchClients.indexOf(searchClient), 1);
-      });
-
-      store.frames().forEach(function(frame) {
-        if (0 <= uris.indexOf(frame.uri)) {
-          store.updateFrameAnnotationFetchStatus(frame.uri, true);
-        }
-      });
-    });
-    searchClient.get({ uri: uris, group: group });
-  }
-
-  this.isLoading = function() {
+  this.isLoading = () => {
     if (
       !store.frames().some(function(frame) {
         return frame.uri;
@@ -153,46 +80,8 @@ function SidebarContentController(
       // The document's URL isn't known so the document must still be loading.
       return true;
     }
-
-    if (searchClients.length > 0) {
-      // We're still waiting for annotation search results from the API.
-      return true;
-    }
-
-    return false;
+    return store.isFetchingAnnotations();
   };
-
-  /**
-   * Load annotations for all URLs associated with `frames`.
-   */
-  function loadAnnotations() {
-    _resetAnnotations();
-
-    searchClients.forEach(function(client) {
-      client.cancel();
-    });
-
-    // If there is no selection, load annotations only for the focused group.
-    //
-    // If there is a selection, we load annotations for all groups, find out
-    // which group the first selected annotation is in and then filter the
-    // results on the client by that group.
-    //
-    // In the common case where the total number of annotations on
-    // a page that are visible to the user is not greater than
-    // the batch size, this saves an extra roundtrip to the server
-    // to fetch the selected annotation in order to determine which group
-    // it is in before fetching the remaining annotations.
-    const group = store.hasSelectedAnnotations() ? null : groups.focused().id;
-
-    const searchUris = store.searchUris();
-    if (searchUris.length > 0) {
-      _loadAnnotationsFor(searchUris, group);
-
-      streamFilter.resetFilter().addClause('/uri', 'one_of', searchUris);
-      streamer.setConfig('filter', { filter: streamFilter.getFilter() });
-    }
-  }
 
   $scope.$on('sidebarOpened', function() {
     analytics.track(analytics.events.SIDEBAR_OPENED);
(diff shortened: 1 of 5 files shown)


src/sidebar/components/sidebar-content.js
```diff
@@ -234,7 +131,11 @@ function SidebarContentController(
   // Re-fetch annotations when focused group, logged-in user or connected frames
   // change.
   $scope.$watch(
-    () => [groups.focused(), store.profile().userid, ...store.searchUris()],
+    () => [
+      store.getState().focusedGroupId,
+      store.profile().userid,
+      ...store.searchUris(),
+    ],
```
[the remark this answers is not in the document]
hmstepanek: Agree-I'll prob address this in a separate PR though as I feel like it's a pretty significant change in itself.

src/sidebar/services/annotations.js
```diff
@@ -0,0 +1,104 @@
...
+  if (!annot) {
+    return null;
+  }
+  return annot.group;
+}
+
+// @ngInject
+function annotations(
+  $scope,
+  annotationMapper,
```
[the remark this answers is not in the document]
hmstepanek: I agree.

src/sidebar/components/sidebar-content.js
```diff
@@ -234,27 +110,33 @@ function SidebarContentController(
   // Re-fetch annotations when focused group, logged-in user or connected frames
   // change.
   $scope.$watch(
-    () => [groups.focused(), store.profile().userid, ...store.searchUris()],
-    ([currentGroup], [prevGroup]) => {
-      if (!currentGroup) {
-        // When switching accounts, groups are cleared and so the focused group
+    () => [
+      store.getState().focusedGroupId,
```
robertknight: There is a `focusedGroupId` selector in the store that you can use instead of accessing the state directly. It currently just returns the relevant state value.

src/sidebar/search-client.js
```diff
@@ -74,11 +74,6 @@ class SearchClient extends EventEmitter {
           return;
         }
         self.emit('error', err);
-      })
-      .then(function() {
-        if (self._canceled) {
-          return;
-        }
```
robertknight: I think you made this change in another PR. Does this one need rebasing?

src/sidebar/services/annotations.js
```diff
@@ -0,0 +1,60 @@
...
+      streamFilter.resetFilter().addClause('/uri', 'one_of', uris);
+      streamer.setConfig('filter', { filter: streamFilter.getFilter() });
+    }
+  }
+
+  function searchAndLoad(uris, groupId) {
+    searchClient = new SearchClient(api.search, {
+      incremental: true,
+    });
+    searchClient.on('results', function(results) {
```
robertknight: For new code prefer arrow functions unless you are creating a method and need to access the value of `this` used to call the function.

src/sidebar/services/test/annotations-test.js
```diff
@@ -0,0 +1,255 @@
...
+        sinon.match({ id: fakeUris[0] + '456' }),
+      ]);
+    });
+
+    it('loads all annotations for a frame with multiple URIs', () => {
+      const uri = 'http://example.com/test.pdf';
+      const fingerprint = 'urn:x-pdf:fingerprint';
+      fakeUris = [uri, fingerprint];
+      const svc = service();
+      // Override the default trames set by the service call above.
```
robertknight:
```suggestion
      // Override the default frames set by the service call above.
```

src/sidebar/services/test/annotations-test.js
```diff
@@ -0,0 +1,255 @@
...
+    });
+
+    it('loads all annotations for all URIs', () => {
+      fakeUris = ['http://example.com', 'http://foobar.com'];
+      const svc = service();
+
+      svc.load(fakeUris, fakeGroupId);
+      assert.calledWith(fakeAnnotationMapper.loadAnnotations, [
+        sinon.match({ id: fakeUris[0] + '123' }),
+      ]);
```
robertknight: There is some repetition in these calls that could be refactored out, eg:

```js
[ /* ids... */ ].forEach(id => {
  assert.calledWith(fakeAnnotationMapper.loadAnnotations, [sinon.match({ id })]);
});
```

src/sidebar/services/test/annotations-test.js
```diff
@@ -0,0 +1,255 @@
...
+      svc.load(fakeUris, fakeGroupId);
+      assert.notCalled(fakeAnnotationMapper.loadAnnotations);
+    });
+
+    it('calls annotationFetchStarted when it starts searching for annotations', () => {
+      const svc = service();
+
+      svc.load(fakeUris, fakeGroupId);
+
+      assert.called(fakeStore.annotationFetchStarted);
```
robertknight: `assert.calledOnce` would be slightly more specific here.

robertknight:
I noticed on this branch that when following a direct link to an annotatio, there is a brief flash of the annotation unavailable message:

Flash of unavailable message

Compare to master (at 680642f2c6ec97de1):

Master no flash of unavailable message

src/sidebar/components/sidebar-content.js
```diff
@@ -91,110 +71,6 @@ function SidebarContentController(
...
-      store.frames().forEach(function(frame) {
-        if (0 <= uris.indexOf(frame.uri)) {
-          store.updateFrameAnnotationFetchStatus(frame.uri, true);
-        }
-      });
-    });
-    searchClient.get({ uri: uris, group: group });
-  }
-
-  this.isLoading = function() {
```
robertknight: There are still a couple of references to this function in the `sidebar-content.html` template.

src/sidebar/components/sidebar-content.js
```diff
@@ -91,110 +71,6 @@ function SidebarContentController(
...
-    if (
-      !store.frames().some(function(frame) {
-        return frame.uri;
```
robertknight: The replacement for `isLoading` is missing this test for whether the document has connected to the sidebar or not. Until the document loads, the client can't start fetching annotations. In terms of the client's display, it still needs to show the loading state though.

The state where the URI is not known is hard to observe with a typical HTML document, but is easier to see if you try and look at a large PDF for example, as the URI isn't reported to the sidebar until the PDF finishes loading.

src/sidebar/components/sidebar-content.js
```diff
@@ -234,27 +122,32 @@ function SidebarContentController(
...
       }
 
-      if (!prevGroup || currentGroup.id !== prevGroup.id) {
+      if (!prevGroupId || currentGroupId !== prevGroupId) {
         // The focused group may be changed during loading annotations as a result
         // of switching to the group containing a direct-linked annotation.
         //
         // In that case, we don't want to trigger reloading annotations again.
-        if (this.isLoading()) {
+        if (store.isFetchingAnnotations()) {
```
hmstepanek: This isn't checking for the document url being loaded now but I think that's ok here. If no document urls are loaded it will clear the annotations which should be empty anyway and even if it's not it will be emptied anyway once the urls are loaded. It will then attempt to load annotations but since no urls are present it will simply exit early.

src/sidebar/services/test/annotations-test.js
```diff
@@ -0,0 +1,254 @@
...
+  let fakeApi;
+  let fakeAnnotationMapper;
+  let fakeStreamer;
+  let fakeStreamFilter;
+
+  let fakeUris;
+  let fakeGroupId;
+
+  beforeEach(() => {
+    searchClients = [];
```
LMS007: I sorta like this here too, feels like it resets a global var rather than having any one test set this and unset it again.
`longRunningSearchClient = false`

src/sidebar/services/test/annotations-test.js
```diff
@@ -0,0 +1,254 @@
...
+
+    it("cancels previously search client if it's still running", () => {
+      const svc = service();
+
+      // Issue a long running load annotations request.
+      longRunningSearchClient = true;
+      svc.load(fakeUris, fakeGroupId);
+      // Issue another load annotations request while the
+      // previous annotation load is still running.
+      longRunningSearchClient = false;
```
LMS007: see prev comment, can be moved to beforeEach

src/sidebar/services/annotations.js
```diff
@@ -0,0 +1,60 @@
...
+   */
+  function load(uris, groupId) {
+    annotationMapper.unloadAnnotations(store.savedAnnotations());
+
+    // Cancel previously running search client.
+    if (searchClient) {
+      searchClient.cancel();
+    }
+
+    if (uris.length > 0) {
```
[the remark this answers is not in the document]
hmstepanek: So here it's waiting for the uri's to load essentially and sometimes when this is called they haven't loaded yet. It's not so much a failure IMO as it is the data is not available to perform a search yet although I suppose it could in theory never be available.

src/sidebar/services/annotations.js
```diff
@@ -0,0 +1,60 @@
...
+
+      store.frames().forEach(function(frame) {
+        if (0 <= uris.indexOf(frame.uri)) {
+          store.updateFrameAnnotationFetchStatus(frame.uri, true);
+        }
+      });
+      store.annotationFetchFinished();
+    });
+    store.annotationFetchStarted();
+    searchClient.get({ uri: uris, group: groupId });
```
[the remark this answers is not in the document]
hmstepanek: It does emit an error [here](https://github.com/hypothesis/client/blob/b5d16912941ca3cea449d0a4fb78fa14e269d6fa/src/sidebar/search-client.js#L76) which emits end after emitting the error I suppose in that scenario maybe we wouldn't want to mark the frames as fetched. I'll have to think through that particular scenario a bit.

src/sidebar/services/test/annotations-test.js
hmstepanek: Ah yes that would be good to have there!

robertknight:
>  I do want to ask if ever intend to convert services to classes if/went ng goes away?

Good question. Semantically AngularJS services are classes, the reason they are not used here is partly because it was written pre-ES2015.

Classes do have some downsides compared to the current functions + closures style (eg. possible issues with `this`, they don't minify as well), but it would be semantically clearer and would be more consistent with other service-like objects in the codebase which tend to use classes.

src/sidebar/services/annotations.js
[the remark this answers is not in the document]
hmstepanek: @robertknight would this error get logged to the console automatically or should we be doing something like that here in an on('error') handler?

src/sidebar/services/annotations.js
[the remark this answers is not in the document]
hmstepanek: @LMS007 @robertknight looking for some feedback/decision here? Do we want to change something or is this ok the way it is?

robertknight:
In testing the current version I noticed a minor issue where the "Show all annotations" button can briefly appear and then change to "Show all notes" if the direct-linked annotation is a note. That issue also occurs on master though, so is not a regression in this PR:

Direct link flash

src/sidebar/services/test/annotations-test.js
```diff
@@ -0,0 +1,254 @@
...
+        ]);
+      });
+    });
+
+    it('updates annotation fetch status for all frames', () => {
+      fakeUris = ['http://example.com', 'http://foobar.com'];
+      const svc = service();
+
+      svc.load(fakeUris, fakeGroupId);
+      assert.calledWith(
```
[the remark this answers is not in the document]
hmstepanek: nope good catch!

src/sidebar/services/test/annotations-test.js
```diff
@@ -0,0 +1,254 @@
...
+      const svc = service();
+
+      svc.load(fakeUris, fakeGroupId);
+
+      [
+        fakeUris[0] + '123',
+        fakeUris[0] + '456',
+        fakeUris[1] + '123',
+        fakeUris[1] + '456',
+      ].forEach(id => {
```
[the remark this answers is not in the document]
hmstepanek: I think this was id because it's id inside the loadAnnotations call but I changed it to uri.

src/sidebar/components/sidebar-content.js
hmstepanek: This block of code is also relevant when the group changes before annotations are loaded from the previous group I believe but I think in that case we would still want to load annotations so yes I think you are correct and this whole block can go now.

hmstepanek:
Ok I've addressed the code review feedback. I'll plan on merging this tomorrow am unless someone wants to take one last peak before then.

PR #1212, opened 2019-06-28
Move real-time update state to store

robertknight:
Move state related to received-but-not-yet-applied annotation updates
("pending updates") and deletions from the `streamer` service to a new
`real-time-updates` module in the store.

This will make it possible for React-based UI to reflect this state and
update when it changes. It also removes some usage of Angular-specific
logic (`$rootScope`) from the streamer service.

Applying pending updates still requires a call to the
`applyPendingUpdates` method of the streamer because that currently
needs to trigger side effects and dispatches Angular events.

 - Move `pendingUpdates` and `pendingDeletions` local variables in
   streamer to `real-time-updates` store module

 - Replace calls to `store.{countPendingUpdates, hasPendingDeletion}`
   with calls to the store instead

 - Call store/streamer methods related to real-time updates directly
   from `<top-bar>` instead of passing it down from `<hypothesis-app>`.

   The `<top-bar>` component is not likely to be used outside the app,
   so there is no benefit to the indirection


src/sidebar/services/test/streamer-test.js
```diff
@@ -122,12 +121,18 @@ describe('Streamer', function() {
...
     fakeStore = {
       annotationExists: sinon.stub().returns(false),
-      isSidebar: sinon.stub().returns(true),
+      clearPendingUpdates: sinon.stub(),
       getState: sinon.stub().returns({
         session: {
           userid: 'jim@hypothes.is',
         },
       }),
+      hasPendingDeletion: sinon.stub().returns(false),
```
hmstepanek: This isn't used.

src/sidebar/store/modules/test/real-time-updates-test.js
```diff
@@ -0,0 +1,162 @@
...
+    const updates = addPendingUpdates(store);
+
+    // Dispatch `ADD_ANNOTATIONS` directly here rather than using
+    // the `addAnnotations` action creator because that has side effects.
+    store.dispatch({ type: 'ADD_ANNOTATIONS', annotations: updates });
+
+    assert.deepEqual(store.pendingUpdateCount(), 0);
+  });
+
+  it('clears pending updates when annotations are removed', () => {
```
hmstepanek: If I comment out the `delete pendingDeletions[ann.id];` line this test still passes. We might want to add some pending deletions as well here.

src/sidebar/templates/top-bar.html
```diff
@@ -34,11 +34,11 @@
     <group-list class="group-list" auth="vm.auth"></group-list>
     <div class="top-bar__expander"></div>
     <a class="top-bar__apply-update-btn"
-       ng-if="vm.pendingUpdateCount > 0"
-       ng-click="vm.onApplyPendingUpdates()"
+       ng-if="vm.pendingUpdateCount() > 0"
+       ng-click="vm.applyPendingUpdates()"
```
hmstepanek: I noticed you changed the name here-just curious what is the naming convention for this kinda thing? Seems like sometimes we use `on` and sometimes we don't.

src/sidebar/store/modules/test/real-time-updates-test.js
robertknight: Good catch. Resolved.

src/sidebar/store/modules/test/real-time-updates-test.js
```diff
@@ -0,0 +1,162 @@
...
+    it("adds pending updates where the focused group matches the annotation's group", () => {
+      addPendingUpdates(store);
+      assert.deepEqual(store.pendingUpdates(), {
+        'updated-ann': { id: 'updated-ann', group: 'group-1' },
+        'created-ann': { id: 'created-ann', group: 'group-1' },
+      });
+    });
+
+    it("does not add pending updates if the focused group does not match the annotation's group", () => {
+      fakeFocusedGroupId.returns('other-group');
```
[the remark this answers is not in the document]
robertknight: Agh. Good catch!

src/sidebar/store/modules/real-time-updates.js
```diff
@@ -0,0 +1,184 @@
...
+  ) {
+    const pendingUpdates = { ...state.pendingUpdates };
+    const pendingDeletions = { ...state.pendingDeletions };
+
+    updatedAnnotations.forEach(ann => {
+      // In the sidebar, only save pending updates for annotations in the
+      // focused group, since we only display annotations from the focused
+      // group and reload all annotations and discard pending updates
+      // when switching groups.
+      if (
```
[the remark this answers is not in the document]
robertknight: There is a test for when `isSidebar` returns `false` ("always adds pending updates in the stream where there is no focused group") but the test for the focused group in the sidebar case was broken as you pointed out.

src/sidebar/services/test/streamer-test.js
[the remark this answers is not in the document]
robertknight: I've removed the unused mocks.

src/sidebar/templates/top-bar.html
robertknight: This file has been removed now since the `<top-bar>` component was ported to Preact.

We generally use `on<Event>` naming for component inputs that are callbacks called in response to some abstract event (eg. "user clicked Sign Up"). In contrast functions / methods created within a component to perform a specific task and that are used as event handlers are generally named after what the function does.

This means that if you see something like `<button onClick={onSignUpClicked}>Sign up</button>` you can see at a glance that the event is propagated up to be handled by the parent, whereas if you see `<button onClick={showSignUpForm}>Sign up</button>` you know the event is locally handled (in this case to show a form).

src/sidebar/store/modules/real-time-updates.js
```diff
@@ -0,0 +1,184 @@
...
+  state => [state.pendingUpdates, state.pendingDeletions],
+  ([pendingUpdates, pendingDeletions]) =>
+    Object.keys(pendingUpdates).length + Object.keys(pendingDeletions).length
+);
+
+/**
+ * Return true if an annotation has been deleted on the server but the deletion
+ * has not yet been applied.
+ */
+function hasPendingDeletion(state, id) {
```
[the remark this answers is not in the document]
robertknight: There is a wider conversation here about the extent to which we encapsulate the structure of state in the store vs exposing it to components. Callers could also use `store.pendingDeletions().hasOwnProperty(<id>)`. I've left it as-is for the moment.

robertknight:
Thanks for catching the missing tests @hmstepanek. I've addressed those and also removed the unused mocks. PTAL.

PR #1215, opened 2019-07-01
Add loading of annotations state to store

hmstepanek:
Track annotation loads in the store by keeping a counter of currently active annotation loads.

### DIFF
--- a/src/sidebar/store/modules/activity.js
+++ b/src/sidebar/store/modules/activity.js
@@ -14,6 +14,10 @@ function init() {
        * The number of API requests that have started and not yet completed.
        */
       activeApiRequests: 0,
+      /**
+       * The number of annotation fetches that have started and not yet completed.
+       */
+      activeAnnotationFetches: 0,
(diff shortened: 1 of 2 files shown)


src/sidebar/store/modules/activity.js
```diff
@@ -64,16 +99,26 @@ function isLoading(state) {
   return state.activity.activeApiRequests > 0;
 }
 
+/**
+ * Return true when annotations are actively loading.
```
robertknight: I think "loading" is ambiguous in this context because it could refer to just fetching annotations from the server, but it could also refer to other work that has to happen once annotations are fetched - sending them to the host page, waiting for them to anchor, and waiting for the results to be received.

hmstepanek: It's issuing an annotation fetch api request, loading the results into the store, and marking the fetch status of the frame as complete. I guess `fetch` would be most appropriate here.

PR #1225, opened 2019-07-02
Fix rendering bug in moderation-banner component

### LINKED ISSUE #1224: Moderation banner shows on non-flagged annotations
### Expected behaviour
As a moderator, the banner should not be shown if the message is not flagged, or we should not arrive in a conflicted state of both hidden and un-flagged.

### Actual behaviour
As a moderator, this annotation (reply) is hidden and has not been flagged. I'm not sure how exactly to arrive at this state since we don't have a UI feature that un-flags an annotation.

The banner should probably not be shown when the annotation is not flagged, or we need to better understand how an annotation can be both un-flagged and hidden?

Screen Shot 2019-07-02 at 12 51 26 PM

Additionally there is a random "0" node injected into the dom, but this may be related to a preact bug or idiosyncrasy with inline conditionals in jsx, but upon code inspection, this element is a total mystery.

Additionally, if a moderator clicks "Unhide" the entire banner goes away.

Chrome

LMS007:
fixes #1224
1. Removes the ng template (forgot to do that form previous conversion of this component)
2. Fixes a small client bug that I think is an issue with preact. 

This does not fix the problem of having hidden annotations with 0 flags. I tried for a long time to reproduce this w/o success. I also feel that the client may not be incorrect in that the mod banner should always show when an annotation is hidden (regardless of the flag count.) Perhaps the issue is more server side with how an annotation got into this state in the first place.

Pre fix UI
Screen Shot 2019-07-02 at 12 51 26 PM

Post fix UI
Screen Shot 2019-07-02 at 2 19 17 PM


robertknight:
This morning I had a look at the official React docs as they are the "source of truth" for expected behaviour. The conversion of `number` values into displayed text is specified [here](https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored). I'm guessing this is a fairly common issue for people to run into because the issue is explicitly mentioned:

> One caveat is that some “falsy” values, such as the 0 number, are still rendered by React. For example, this code will not behave as you might expect because 0 will be printed when props.messages is an empty array

PR #1233, opened 2019-07-08
Move tab counts to store

hmstepanek:
* move tab counts to the store
* remove passing of selectedTab from sidebar-content to child components and instead make direct calls to the store
* move annotation-metadata and tabs to util

This is a partial fix for https://github.com/hypothesis/product-backlog/issues/1028.


src/sidebar/store/modules/selection.js
```diff
@@ -162,9 +163,12 @@ const update = {
   },
 
   ADD_ANNOTATIONS(state, action) {
-    const counts = tabs.counts(action.annotations);
+    const countNotes = arrayUtil.countIf(
```
hmstepanek: Wasn't sure if I should call the selector in the other store module here or just implement the counting here?

src/sidebar/components/test/search-status-bar-test.js
```diff
@@ -26,6 +26,8 @@ describe('SearchStatusBar', () => {
       clearDirectLinkedGroupFetchFailed: sinon.stub(),
```
LMS007: from `selectTab` to `clearSelection,` these actions appear to not be used in the test or required for 11 tests to pass.

src/sidebar/components/search-status-bar.js
```diff
@@ -106,13 +104,6 @@ function SearchStatusBar({
...
 
 SearchStatusBar.propTypes = {
-  selectedTab: propTypes.oneOf([
-    uiConstants.TAB_ANNOTATIONS,
-    uiConstants.TAB_ORPHANS,
-    uiConstants.TAB_NOTES,
-  ]).isRequired,
-  totalAnnotations: propTypes.number.isRequired,
-  totalNotes: propTypes.number.isRequired,
   rootThread: propTypes.object.isRequired,
```
LMS007: above rootThread add

// Injected services.

src/sidebar/tabs.js
```diff
@@ -3,7 +3,6 @@
 // Selectors that calculate the annotation counts displayed in tab headings
```
LMS007: Me wonders if this should this file be in /sidebar/utils>? I don't have a sense of these things yet, but seems like its a static util type of thing. Thoughts?

src/sidebar/components/sidebar-content.js
```diff
@@ -27,13 +27,11 @@ function SidebarContentController(
...
     self.selectedTab = state.selectedTab;
 
-    const counts = tabs.counts(state.annotations);
-
     Object.assign(self, {
-      totalNotes: counts.notes,
-      totalAnnotations: counts.annotations,
-      totalOrphans: counts.orphans,
-      waitingToAnchorAnnotations: counts.anchoring > 0,
+      totalNotes: store.countNotes(),
```
LMS007: @hmstepanek this gets removed once we rebase again and get new selection-tabs component right?

src/sidebar/tabs.js
hmstepanek: Looking at this code it's used in rootThread and it seems like there needs to be some cleanup here. I get the impression there is a bit of translation hell going on here and ultimately this whole file will disappear once it's cleaned up.  For example it's currently going from:
rootThread->tabs->annotation-metadata. rootThread is trying to collect all annotations that belong under a certain tab, tabs is calculating per annotation whether an annotation belongs to a certain tab by first determining which tab an annotation belongs to. Metadata does some checks for attributes and things on the annotation to ultimately determine what tab an annotation belongs under. It kinda feels like these should all be functions in the store and when an annotation is added to the store, maybe it should be added with an attribute that identifies in which tab it belongs and maybe we should have a selector where we say give me all annotations under this tab and it returns those or something instead of this jumping around between (as you indicated) what feel like utility files that aren't in the utils area? Long thoughts to have on a PR and something that probably deserves a larger discussion.

hmstepanek: Speaking of annotation-metadata I'm currently calling into that from the store and after looking at that code I'm wondering if perhaps I should just move that logic into the store right now.

src/sidebar/components/sidebar-content.js
hmstepanek: yes-I'm gonna do that now once I rebase.

src/sidebar/components/test/search-status-bar-test.js
hmstepanek: Ah yes I think those may have been leftover from a previous refactor. Good catch thanks!

src/sidebar/components/search-status-bar.js
```diff
@@ -21,16 +21,14 @@ const countVisibleAnns = annThread => {
...
-  totalAnnotations,
-  totalNotes,
-  rootThread,
-}) {
+function SearchStatusBar({ rootThread }) {
   const storeState = useStore(store => store.getState());
   const clearSelection = useStore(store => store.clearSelection);
   const filterQuery = storeState.filterQuery;
   const filterActive = !!storeState.filterQuery;
+  const countAnnotations = useStore(store => store.countAnnotations());
```
robertknight: These variable names confused me because they read as actions rather than values. `annotationCount` and `noteCount` would be clearer.

src/sidebar/components/search-status-bar.js
```diff
@@ -21,16 +21,14 @@ const countVisibleAnns = annThread => {
...
+  const countNotes = useStore(store => store.countNotes());
+  const selectedTab = storeState.selectedTab;
```
robertknight: It is more efficient to extract only the data you need from the store in a `useStore` call rather than fetching the entire state object. Extracting only the data you need enables `useStore` to limit re-rendering of a component to only when the relevant state has changed.

src/sidebar/store/modules/annotations.js
```diff
@@ -388,6 +390,46 @@ function findAnnotationByID(state, id) {
...
+  }
+);
+
+/**
+ * Returns true if some annotations have not been anchored yet.
+ */
+const isWaitingToAnchorAnnotations = createSelector(
+  state => state.annotations,
+  annotations => {
+    return arrayUtil.countIf(annotations, metadata.isWaitingToAnchor) > 0;
```
robertknight: I'm guessing this was taken from existing code, but this could be written using just standard library functions as `annotations => annotations.some(metadata.isWaitingToAnchor)`

src/sidebar/store/modules/selection.js
robertknight: I think implementing the counting here is fine, and a little more decoupled.

src/sidebar/templates/sidebar-content.html
```diff
@@ -1,6 +1,6 @@
 <selection-tabs
   ng-if="vm.showSelectedTabs()"
-  is-waiting-to-anchor-annotations="vm.waitingToAnchorAnnotations"
+  is-waiting-to-anchor-annotations="vm.isWaitingToAnchorAnnotations"
```
robertknight: 👍for using `is` and `has` prefixes for boolean variables when appropriate.

src/sidebar/tabs.js
robertknight: > maybe it should be added with an attribute that identifies in which tab it belongs

Generally data in the store follows the principle that it only contains data you couldn't derive from other information already in the store. This avoids cache invalidation bugs where a piece of state is not updated when the state it is derived from changes.

In this context, the tab that an annotation appears in is determined from a) whether it refers to the whole document or a specific part of the document and b) whether it was successfully anchored in the document or not (if (a)) is true.

In future I could imagine that we might make other changes to the tabs and the rules for when annotations appear in them.

> Speaking of annotation-metadata I'm currently calling into that from the store and after looking at that code I'm wondering if perhaps I should just move that logic into the store right now.

As per @LMS007's comment, I'd prefer to put annotation-metadata in the `utils/` directory rather than add the functionality to the store. These are simple functions for answering questions about annotation objects, so from a testing + reusability point of view, I think it is better not to couple them to the store.

The primary constraint on the `util/` directory is that the code should be free of dependencies on other parts of the app (eg. `store/`, `components/`, `services/`).

src/sidebar/components/search-status-bar.js
hmstepanek: I looked at this in the search-status-bar pr and i had trouble determining which state the rootThread needed but I just looked at it again and for whatever reason it's clearer now so I'll go ahead and limit the states that this needs now.

src/sidebar/tabs.js
hmstepanek: Ok I'll move the annotation-metadata and the tabs into the util area for now. I still feel like these could use some simplification but that will involve root-thread and I'm definitely not going to attempt that right now.

src/sidebar/store/modules/annotations.js
```diff
@@ -388,6 +390,44 @@ function findAnnotationByID(state, id) {
   return findByID(state.annotations, id);
 }
 
+/**
+ * Return the number of page notes.
+ */
+const noteCount = createSelector(
+  state => state.annotations,
+  annotations => {
+    return arrayUtil.countIf(annotations, metadata.isPageNote);
```
robertknight: We don't strictly require this as it requires some judgement about readability, but if an arrow function's body consists of a single `return` statement it can be replaced by the return expression (`annotations => arrayUtil.countIf(annotations, metadata.isPageNote)`)

src/sidebar/templates/sidebar-content.html
```diff
@@ -1,19 +1,10 @@
...
+  is-loading="vm.isLoading()">
 </selection-tabs>
 
 <search-status-bar
-  class="search-status-bar"
-  ng-if="!vm.isLoading()"
-  selected-tab="vm.selectedTab"
-  total-annotations="vm.totalAnnotations"
-  total-notes="vm.totalNotes">
+  ng-show="!vm.isLoading()">
```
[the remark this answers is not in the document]
hmstepanek: Agh, unintentional and the result of a merge.

src/sidebar/store/modules/annotations.js
hmstepanek: Kinda surprised this wasn't optimized by the formatter.

robertknight:
LGTM. The issues @LMS007 highlighted have also been addressed, so I'm going to merge this.

PR #1238, opened 2019-07-10
Convert draft service to store

LMS007:
More or less this is a straight conversion but I did create a helper class called Draft to encapsulate the object and a few methods it uses directly.

I would like another set of eyes to try this is out in the UI, make sure it "works"


src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Returns the draft changes for an annotation, or null if no draft exists
+ * or the draft is empty.
+ */
+function getDraftIfNotEmpty(state, model) {
+  const draft = getDraft(state, model);
+  if (!draft) {
+    return null;
+  }
+  return draft.isEmpty() ? null : draft;
```
hmstepanek: What cases would the draft exist but be empty? Is that when you open an annotation for editing but haven't made any changes?

src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,177 @@
...
+   * Annotations are matched by ID or local tag.
+   */
+  match(model) {
+    return (
+      (this.model.$tag && model.$tag === this.model.$tag) ||
+      (this.model.id && model.id === this.model.id)
+    );
+  }
+  /**
+   * Return true if draft is empty and can be discarded without losing
```
LMS007: needs some grammar help

src/sidebar/store/modules/drafts.js
robertknight: Yes. IIRC when you click on an annotation to start editing it, the initial state is that you have a draft with no changes. There is no separate "is editing" flag.

LMS007:
added tests

LMS007:
Updated with all tests passing.
Ready for review

src/sidebar/components/test/annotation-test.js
```diff
@@ -54,7 +54,7 @@ function findActionButton(annotationEl, label) {
     : null;
 }
 
-describe('annotation', function() {
+describe('annotation77', function() {
```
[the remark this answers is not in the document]
LMS007: yuup!

src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,176 @@
...
+const actions = util.actionTypes(update);
+
+/* Actions */
+
+/**
+ * Create or update the draft version for a given annotation by
+ * replacing any existing draft or simply creating a new one.
+ */
+function createDraft(annotation, changes) {
+  return function(dispatch) {
```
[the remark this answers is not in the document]
LMS007: Will do, I got back and fourth, still old ES5 habits.

src/sidebar/components/test/annotation-test.js
```diff
@@ -190,12 +189,13 @@ describe('annotation', function() {
...
-          update: sandbox.stub(),
-          remove: sandbox.stub(),
-          get: sandbox.stub().returns(null),
+          // draft store
+          countDrafts: sandbox.stub().returns(0),
+          createDraft: sandbox.stub(),
+          discardAllDrafts: sandbox.stub(),
+          getDraft: sandbox.stub().returns(null),
+          getDraftIfNotEmpty: sandbox.stub().returns(null),
+          removeDraft: sandbox.stub().returns(null),
```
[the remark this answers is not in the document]
LMS007: yes

src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,176 @@
...
+
+const actions = util.actionTypes(update);
+
+/* Actions */
+
+/**
+ * Create or update the draft version for a given annotation by
+ * replacing any existing draft or simply creating a new one.
+ */
+function createDraft(annotation, changes) {
```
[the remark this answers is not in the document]
LMS007: good call, took another look, this makes more sense + more efficient.

LMS007:
Also forgot to remove the old drafts service! done.

src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,194 @@
...
+ */
+class Draft {
+  constructor(annotation, changes) {
+    this.annotation = { id: annotation.id, $tag: annotation.$tag };
+    this.isPrivate = changes.isPrivate;
+    this.tags = changes.tags;
+    this.text = changes.text;
+  }
+  /**
+   * Returns true if `draft` is a draft for a given
```
robertknight:
```suggestion
   * Returns true if this draft matches a given annotation
```

src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,194 @@
...
+    // removes a matching existing draft, then adds
+    const drafts = state.drafts.filter(draft => {
+      return !draft.match(action.draft.annotation);
+    });
+    drafts.push(action.draft); // push ok since its a copy
+    return {
+      drafts,
+    };
+  },
+  ADD_DRAFT: function(state, action) {
```
robertknight: The `ADD_DRAFT` action isn't used any more.

src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,194 @@
...
+  if (!draft) {
+    return null;
+  }
+  return draft.isEmpty() ? null : draft;
+}
+
+/**
+ * Returns a list of draft annotations which have no id.
+ *
+ * @return {Object[]}
```
robertknight:
```suggestion
 * @return {Draft[]}
```

src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,194 @@
...
+  return null;
+}
+
+/**
+ * Returns the draft changes for an annotation, or null if no draft exists
+ * or the draft is empty.
+ *
+ * Since this is a non-primitive object injected into the store, there are
+ * some consequences with serialization. Slack discussion:
+ * https://hypothes-is.slack.com/archives/C1M8NH76X/p1562779754051700?thread_ts=1562776681.050500&cid=C1M8NH76X
```
robertknight: It is true we may need to do some extra work if we need to serialise state in future, but that's not really specific to this particular method, so I'm not sure this comment is useful here.

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -0,0 +1,200 @@
+'use strict';
+
+const immutable = require('seamless-immutable');
+
+const drafts = require('../drafts');
+const { Draft } = require('../drafts');
+const createStore = require('../../create-store');
+
+const fixtures = {
+  draftWithText: immutable({
```
robertknight: I think this code was just copied from the previous tests, but I'll note that I think it would have been possible to wrap the entire fixtures object with `immutable` (ie. `fixtures = immutable({ ... })`)

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -0,0 +1,200 @@
...
+
+describe('Drafts Store', () => {
+  let store;
+
+  beforeEach(() => {
+    store = createStore([drafts]);
+  });
+
+  describe('Draft', () => {
+    it('creates a non empty draft object', () => {
```
robertknight: The structure of the `describe` blocks isn't quite right here, as the `describe` blocks for the methods are outside the one for the `Draft` class. For a class it should typically look something like this:

```js
describe("SomeClassName", () => {
  describe('constructor', () => {

  });
  
  // Test for a specific method
  describe("#someMethod", () => {
  ...
  });

  // Test not about a specific method
  it("should ...", () => {
    
  });
});

describe('someOtherFunction', () => {
  ...
});
```

src/sidebar/store/modules/drafts.js
LMS007: @robertknight `unsavedDrafts` returns an array of annotation objects, but not the entire Draft itself. Perhaps we should change the name of the selector to `unsavedAnnotations`

?

src/sidebar/store/modules/test/drafts-test.js
LMS007: roger

LMS007:
@robertknight 
I changed `unsavedDrafts` to `unsavedAnnotations` as it was returning an object more similar to an annotation but this still may be misleading since its in the draft store. 

We could make this return the actual `Draft` object which inside has the terse annotation object too, but that requires caller updates? Thoughts?

src/sidebar/store/modules/drafts.js
robertknight: Ah, of course. I was just looking at the function name which had confused me. `unsavedAnnotations` is a better name IMO. I notice that the single consumer in `hypothesis-app.js` has an incorrect or misleading parameter name in its callback.

src/sidebar/components/hypothesis-app.js
```diff
@@ -150,30 +149,32 @@ function HypothesisAppController(
...
 
   // Log the user out.
   this.logout = function() {
     if (!promptToLogout()) {
       return;
     }
+
     store.clearGroups();
-    drafts.unsaved().forEach(function(draft) {
+    store.unsavedAnnotations().forEach(function(draft) {
```
robertknight: The callback parameter name (`draft`) was incorrect / misleading before as `draft` here is actually an annotation with no properties other than `$tag` and `id`. I'd suggest renaming to `annotation` or destructuring to `{ id, $tag }`.

src/sidebar/store/modules/drafts.js
```diff
@@ -0,0 +1,183 @@
...
+ *
+ *  A draft consists of:
+ *
+ * 1. `annotation` which is the original annotation object which the
+ *    draft is associated with. If this is just a draft, then this may
+ *    not have an id yet and instead, $tag is used.
+ *
+ * 2. `isPrivate` (boolean), `tags` (array of objects) and `text` (string)
+ *    which are the user's draft changes to the annotation. These are returned
+ *    from the drafts store selector by `drafts.getDraft()`.
```
robertknight: 👍

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -0,0 +1,228 @@
...
+        assert.isFalse(
+          draft.match({
+            $tag: 'fake',
+          })
+        );
+      });
+    });
+  });
+
+  describe('#getIfNotEmpty', () => {
```
robertknight: The function name is `getDraftIfNotEmpty`

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -0,0 +1,228 @@
...
+    it('should remove drafts', () => {
+      store.createDraft(fixtures.annotation, fixtures.draftWithText);
+      assert.isOk(store.getDraft(fixtures.annotation));
+      store.removeDraft(fixtures.annotation);
+      assert.isNotOk(store.getDraft(fixtures.annotation));
+    });
+  });
+
+  describe('#unsavedAnnotations', () => {
+    it('should return drafts for unsaved annotations which have no ID', () => {
```
robertknight:
```suggestion
    it('should return unsaved annotations which have drafts', () => {
```

And a similar adjustment in the test below. The previous description implied that the return value was a `Draft`.

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -0,0 +1,228 @@
...
+      };
+      store.createDraft(fakeAnnotation1, fixtures.draftWithText);
+      store.createDraft(fakeAnnotation2, fixtures.draftWithText);
+      assert.deepEqual(store.unsavedAnnotations(), [
+        fakeAnnotation1,
+        fakeAnnotation2,
+      ]);
+    });
+
+    it('should not return drafts for saved annotations', () => {
```
robertknight:
```suggestion
    it('should not return saved annotations which have drafts', () => {
```

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -0,0 +1,228 @@
...
+      const draft = new Draft(fixtures.annotation, fixtures.draftWithText);
+      assert.deepEqual(draft, {
+        annotation: {
+          ...fixtures.annotation,
+        },
+        ...fixtures.draftWithText,
+      });
+    });
+    describe('#isEmpty', () => {
+      it('creates a non empty draft object', () => {
```
robertknight:
```suggestion
      it('returns false if draft has tags or text', () => {
```

The test description should be written in terms of what the method being tested returns or does, preferably in a way that is easy to relate to end-user behaviour. In this case we disable certain UI elements until an annotation being edited has either some tags or text.

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -0,0 +1,228 @@
...
+        const draft = new Draft(fixtures.annotation, fixtures.draftWithText);
+        assert.isFalse(draft.isEmpty());
+      });
+      it('creates an empty draft object', () => {
```
robertknight:
```suggestion
      it('returns true if draft has no tags or text', () => {
```

src/sidebar/components/hypothesis-app.js
LMS007: good catch

PR #1254, opened 2019-07-17
Convert search status bar to preact

hmstepanek:
Convert search-status-bar to preact. This is a fix for https://github.com/hypothesis/client/issues/1207.


src/sidebar/components/search-status-bar.js
```diff
@@ -1,17 +1,21 @@
...
-  this.TAB_ORPHANS = uiConstants.TAB_ORPHANS;
-
+function SearchStatusBar({
+  selectedTab,
+  totalAnnotations,
+  totalNotes,
+  rootThread,
+}) {
   const thread = () => {
-    return rootThread.thread(store.getState());
```
robertknight: Hooks have to be used at the top-level of the component function and not inside any conditionals or loops. See [this page of the React docs](https://reactjs.org/docs/hooks-overview.html#rules-of-hooks). The reason is because hooks rely on the order in which they are called being consistent across renders in order to remember their state from the last render.

src/sidebar/components/search-status-bar.js
```diff
@@ -23,41 +27,90 @@ function SearchStatusBarController(store, rootThread) {
...
+    const selection = useStore(store => store.getState().selectedAnnotationMap);
     if (!selection) {
       return false;
     }
     return Object.keys(selection).length > 0;
   };
 
-  this.filterQuery = function() {
-    return store.getState().filterQuery;
+  const filterQuery = () => {
```
robertknight: When converting helper methods on controllers from Anguar => React/Preact, any methods that don't take arguments can be replaced by values, ie:

```js
const filterQuery = useStore(store => store.getState().filterQuery)
const filterActive = !!filterQuery
```

In Angular, a method was needed because a component's controller is a class which is instantiated once and the methods are called whenever the data might have changed. In React however the component is a function which is called each time the component is rendered.

src/sidebar/components/search-status-bar.js
```diff
@@ -1,17 +1,21 @@
...
+  totalAnnotations,
+  totalNotes,
+  rootThread,
+}) {
   const thread = () => {
-    return rootThread.thread(store.getState());
+    return rootThread.thread(useStore(store => store.getState()));
   };
 
   const visibleCount = memoize(thread => {
```
robertknight: I've just realized that the recursive use of `visibleCount` made the memoization here basically pointless because any recursive calls will have a different argument. To use memoization here you'd need to do two things:

1. Use the `useMemo` or `useCallback` hooks rather than `memoize`, because the hooks remember state across renders.
2. Separate `visibleCount` into a memoized function which takes a root thread as input and a non-memoized helper which is used for the recursive calls.

It might be helpful here to remove the memoization and get this working directly, then re-add it in a separate PR.

src/sidebar/components/search-status-bar.js
```diff
@@ -23,41 +27,90 @@ function SearchStatusBarController(store, rootThread) {
...
+        </div>
+      )}
+      {!filterActive() && areAllAnnotationsVisible() && (
+        <div className="search-status-bar">
+          <button
+            className="primary-action-btn primary-action-btn--short"
+            onClick={onClearSelection}
+            title="Clear the selection and show all annotations"
+          >
+            {(!selectedTab || selectedTab === uiConstants.TAB_ORPHANS) && (
```
robertknight: Hmm, can `selectedTab` ever be null? If not, I would remove this condition.

src/sidebar/components/search-status-bar.js
```diff
@@ -23,41 +27,90 @@ function SearchStatusBarController(store, rootThread) {
...
+              <span>Show all annotations and notes</span>
+            )}
+            {selectedTab === uiConstants.TAB_ANNOTATIONS && (
+              <span>
```
robertknight: It doesn't look to me like the outer `span` is needed in the DOM as it doesn't have any semantic meaning or custom HTML attributes. If you need a wrapper element for a group of text nodes or elements purely so that the JSX is syntactically valid, you can use [`Fragment`](https://reactjs.org/docs/react-api.html#fragments) instead:

```js
<span><Fragment>item1 item2 item3</Fragment></span>
```

Would render as:

```html
<span>item1 item2 item3</span>
```

Search the existing client code for some example uses.

src/sidebar/components/test/search-status-bar-test.js
```diff
@@ -26,174 +27,178 @@ describe('searchStatusBar', () => {
...
+  afterEach(() => {
+    SearchStatusBar.$imports.$restore();
   });
 
-  describe('filterMatchCount', () => {
-    it('returns the total number of visible annotations or replies', () => {
+  [
+    {
+      description:
+        'reports correct text if 2 annotations match the search fitler',
```
robertknight:
```suggestion
        'reports correct text if 2 annotations match the search filter',
```

And the same with a few tests below.

src/sidebar/components/search-status-bar.js
```diff
@@ -1,63 +1,109 @@
...
+          <button
+            className="primary-action-btn primary-action-btn--short"
+            onClick={onClearSelection}
+            title="Clear the selection and show all annotations"
+          >
+            {selectedTab === uiConstants.TAB_ORPHANS && (
+              <Fragment>Show all annotations and notes</Fragment>
+            )}
+            {selectedTab === uiConstants.TAB_ANNOTATIONS && (
+                <Fragment>Show all annotations
```
robertknight: There is a missing space before the annotation count, and the same with the note count:

Screenshot 2019-07-22 18 12 03

src/sidebar/components/search-status-bar.js
```diff
@@ -1,63 +1,109 @@
...
+    return countVisibleAnns(thread);
+  }, [thread]);
 
-  this.filterMatchCount = function() {
-    return visibleCount(thread());
+  const filterResults = () => {
+    const resultsCount = visibleCount;
+    switch (resultsCount) {
+      case 0:
+        return 'No results for ' + filterQuery;
```
robertknight: There is a presumably unintended change to the display when the result count is 0. On master the filter query is quoted. In this branch it is not:

**Master:**

Screenshot 2019-07-22 19 07 17


**This branch:**

Screenshot 2019-07-22 19 07 22

src/sidebar/components/test/search-status-bar-test.js
```diff
@@ -26,174 +27,172 @@ describe('searchStatusBar', () => {
...
+          id: '2',
+          visible: false,
+          children: [],
+        },
+      ],
+      expectedText: '2 search results',
+    },
+    {
+      description:
+        'reports correct text if 1 annotation matches the search fitler',
```
robertknight:
```suggestion
        'reports correct text if 1 annotation matches the search filter',
```

Typo

src/sidebar/components/test/search-status-bar-test.js
```diff
@@ -26,174 +27,172 @@ describe('searchStatusBar', () => {
...
+          id: '2',
+          visible: false,
+          children: [],
+        },
+      ],
+      expectedText: '1 search result',
+    },
+    {
+      description:
+        'reports correct text if no annotation matches the search fitler',
```
robertknight:
```suggestion
        'reports correct text if no annotation matches the search fitler',
```

Typo

src/sidebar/templates/sidebar-content.html
```diff
@@ -9,7 +9,8 @@
 </selection-tabs>
 
 <search-status-bar
-  ng-show="!vm.isLoading()"
+  class="search-status-bar"
```
robertknight: This class is applied to the top-level `<div>` rendered by the `<search-status-bar>` so it shouldn't be needed here as well. Note that when this template is converted, the Angular components will be replaced by a direct call to `SearchStatusBar` and the `<search-status-bar>` DOM element will disappear.

src/sidebar/components/search-status-bar.js
```diff
@@ -1,63 +1,121 @@
...
+      case 1:
+        return '1 search result';
+      default:
+        return resultsCount + ' search results';
+    }
   };
 
-  this.areAllAnnotationsVisible = function() {
-    if (store.getState().directLinkedGroupFetchFailed) {
+  const areAllAnnotationsVisible = () => {
```
robertknight: Now that I look again, I think the naming here is inverted. Conceptually, we want to show a button to show all annotations/notes only if *not all annotations are visible*, eg. due to there being an active selection. This issue applied to the Angular code.

hmstepanek: Hmmm...yeah you're right.

PR #1255, opened 2019-07-17
Refactor annotation events

### LINKED ISSUE #1236: Refactor annotation events using $rootScope.$on in root-thread.js into the redux store
Doing the preact conversion revealed dependancies on the angular $broadcast mechanism that we should to refactor into a redux action/store. 

These are the obvious dependancies from `new-note-bnt` that was the original req. for this work:

1. BEFORE_ANNOTATION_CREATED
2. deleteNewAndEmptyAnnotations
3. ANNOTATION_DELETED

This refactor could also include the load events

```
  const loadEvents = [
    events.ANNOTATION_CREATED,
    events.ANNOTATION_UPDATED,
    events.ANNOTATIONS_LOADED,
  ];```

### PR DESCRIPTION
**Pull request by LMS007**
This is mostly just moves 3 parts of functionality from root-thread out into appropriate stores and improves the test cases.

fixes #1236

### DIFF
--- a/src/sidebar/services/root-thread.js
+++ b/src/sidebar/services/root-thread.js
@@ -5,7 +5,6 @@ const events = require('../events');
 const memoize = require('../util/memoize');
 const metadata = require('../util/annotation-metadata');
 const tabs = require('../util/tabs');
-const uiConstants = require('../ui-constants');
 
 function truthyKeys(map) {
   return Object.keys(map).filter(function(k) {
@@ -82,18 +81,6 @@ function RootThread($rootScope, store, searchFilter, viewFilter) {
     });
   }
 
-  function deleteNewAndEmptyAnnotations() {
-    store
-      .getState()
-      .annotations.filter(function(ann) {
-        return metadata.isNew(ann) && !store.getDraftIfNotEmpty(ann);
-      })
-      .forEach(function(ann) {
-        store.removeDraft(ann);
-        $rootScope.$broadcast(events.ANNOTATION_DELETED, ann);
-      });
-  }
-
   // Listen for annotations being created or loaded
   // and show them in the UI.
   //
@@ -111,33 +98,14 @@ function RootThread($rootScope, store, searchFilter, viewFilter) {
   });
 
   $rootScope.$on(events.BEFORE_ANNOTATION_CREATED, function(event, ann) {
-    // When a new annotation is created, remove any existing annotations
-    // that are empty.
-    deleteNewAndEmptyAnnotations();
-
-    store.addAnnotations([ann]);
-
-    // If the annotation is of type note or annotation, make sure
-    // the appropriate tab is selected. If it is of type reply, user
-    // stays in the selected tab.
-    if (metadata.isPageNote(ann)) {
-      store.selectTab(uiConstants.TAB_NOTES);
-    } else if (metadata.isAnnotation(ann)) {
-      store.selectTab(uiConstants.TAB_ANNOTATIONS);
-    }
-
-    (ann.references || []).forEach(function(parent) {
-      store.setCollapsed(parent, false);
-    });
+    store.createAnnotation(ann);
   });
 
   // Remove any annotations that are deleted or unloaded
   $rootScope.$on(events.ANNOTATION_DELETED, function(event, annotation) {
     store.removeAnnotations([annotation]);
-    if (annotation.id) {
-      store.removeSelectedAnnotation(annotation.id);
-    }
   });
+
   $rootScope.$on(events.ANNOTATIONS_UNLOADED, function(event, annotations) {
     store.removeAnnotations(annotations);
   });
--- a/src/sidebar/services/test/root-thread-test.js
+++ b/src/sidebar/services/test/root-thread-test.js
@@ -39,6 +39,7 @@ describe('rootThread', function() {
       state: {
         annotations: [],
         expanded: {},
+        drafts: [],
         filterQuery: null,
         focusedAnnotationMap: null,
         forceVisible: {},
@@ -63,6 +64,7 @@ describe('rootThread', function() {
       selectTab: sinon.stub(),
       getDraftIfNotEmpty: sinon.stub().returns(null),
       removeDraft: sinon.stub(),
+      createAnnotation: sinon.stub(),
     };
 
     fakeBuildThread = sinon.stub().returns(fixtures.emptyThread);
@@ -341,6 +343,12 @@ describe('rootThread', function() {
   context('when annotation events occur', function() {
     const annot = annotationFixtures.defaultAnnotation();
 
+    it('creates a new annotation in the store when BEFORE_ANNOTATION_CREATED event occurs', function() {
+      $rootScope.$broadcast(events.BEFORE_ANNOTATION_CREATED, annot);
+      assert.notCalled(fakeStore.removeAnnotations);
+      assert.calledWith(fakeStore.createAnnotation, sinon.match(annot));
+    });
+
     unroll(
       'adds or updates annotations when #event event occurs',
       function(testCase) {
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/sidebar/services/root-thread.js [hunk 1]
```diff
@@ -82,18 +81,6 @@ function RootThread($rootScope, store, searchFilter, viewFilter) {
     });
   }
 
-  function deleteNewAndEmptyAnnotations() {
```
LMS007: Moved this to drafts store, renamed to `deleteNewAndEmptyDrafts`

src/sidebar/services/root-thread.js
```diff
@@ -111,33 +98,14 @@ function RootThread($rootScope, store, searchFilter, viewFilter) {
   });
 
   $rootScope.$on(events.BEFORE_ANNOTATION_CREATED, function(event, ann) {
-    // When a new annotation is created, remove any existing annotations
```
LMS007: created a new action called `createAnnotation` in annotations store

src/sidebar/services/root-thread.js
```diff
@@ -111,33 +98,14 @@ function RootThread($rootScope, store, searchFilter, viewFilter) {
...
-    }
-
-    (ann.references || []).forEach(function(parent) {
-      store.setCollapsed(parent, false);
-    });
+    store.createAnnotation(ann);
   });
 
   // Remove any annotations that are deleted or unloaded
   $rootScope.$on(events.ANNOTATION_DELETED, function(event, annotation) {
```
LMS007: created new action called `removeAndDeselectAnnotation` in annotations store

src/sidebar/services/test/root-thread-test.js
```diff
@@ -350,38 +359,23 @@ describe('rootThread', function() {
...
-      assert.calledWith(fakeStore.setCollapsed, reply.references[0], false);
+    it('removes and de-selects annotations when ANNOTATION_DELETED event occurs', function() {
+      $rootScope.$broadcast(events.ANNOTATION_DELETED, annot);
+      assert.calledWith(
+        fakeStore.removeAndDeselectAnnotation,
+        sinon.match(annot)
+      );
     });
 
-    unroll(
```
LMS007: had to break this up since they use different  asserts now

src/sidebar/services/test/root-thread-test.js
```diff
@@ -395,24 +389,6 @@ describe('rootThread', function() {
         fakeStore.state.annotations.push(existingNewAnnot);
       });
 
-      it('removes drafts for new and empty annotations', function() {
```
LMS007: moves tests to drafts stores

src/sidebar/services/root-thread.js
```diff
@@ -111,33 +98,14 @@ function RootThread($rootScope, store, searchFilter, viewFilter) {
...
-    store.removeAnnotations([annotation]);
```
hmstepanek: Couldn't this just be store.removeAnnotation(annotation)?

src/sidebar/services/root-thread.js
```diff
@@ -111,33 +98,14 @@ function RootThread($rootScope, store, searchFilter, viewFilter) {
...
-    if (metadata.isPageNote(ann)) {
-      store.selectTab(uiConstants.TAB_NOTES);
-    } else if (metadata.isAnnotation(ann)) {
-      store.selectTab(uiConstants.TAB_ANNOTATIONS);
-    }
-
-    (ann.references || []).forEach(function(parent) {
-      store.setCollapsed(parent, false);
-    });
+    store.createAnnotation(ann);
```
hmstepanek: Much cleaner!

src/sidebar/store/modules/annotations.js
```diff
@@ -331,6 +332,48 @@ function hideAnnotation(id) {
   };
 }
 
+/**
+ * De-select and remove an annotation.
+ */
+function removeAndDeselectAnnotation(annotation) {
```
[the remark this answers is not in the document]
hmstepanek: My first thought is, looking at the code where this is actually called it appears to always create a list of 1 annotation so I'm not sure there's a need for a plural version anyway. 

Since this is dispatching a singular action inside the plural store function does this really have the advantage of being more efficient? Based on what you describe above it seems like it would still be calling every store subscriber for every dispatched action anyway although I see your point if it was a plural action that didn't dispatch other singular actions then it would be more efficient.

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -225,4 +225,40 @@ describe('Drafts Store', () => {
       assert.deepEqual(store.unsavedAnnotations(), []);
     });
   });
+
+  describe('#deleteNewAndEmptyDrafts', () => {
+    it('should remove only new or empty drafts', () => {
+      const newAnnotations = [
```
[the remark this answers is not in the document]
hmstepanek: Personally I like having the test description as part of the test params for each test case but there are certainly multiple ways to do it!

[{
  description: '...',
  ann: '...'
}].forEach( (test) => {
  it(test.description, () => {
    ...
  });
});

src/sidebar/services/test/root-thread-test.js
```diff
@@ -341,6 +344,12 @@ describe('rootThread', function() {
   context('when annotation events occur', function() {
     const annot = annotationFixtures.defaultAnnotation();
 
+    it('adds or updates annotations when BEFORE_ANNOTATION_CREATED event occurs', function() {
```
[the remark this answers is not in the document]
hmstepanek: That's fine, I guess my point here was just that I was expecting to see an assertion like addAnnotation or updateAnnotation and instead I saw createAnnotation which doesn't have add or update in the name and then I had to go look at the implementation of createAnnotation to see what it did.

LMS007:
I'm gonna look over the comments here again, but I think I might need to step back from this PR and form a general approach to this conversion.  It seems to me that there are some cases where an action does not really belong to any one store and then multiple reducers need to handle it. BEFORE_ANNOTATION_CREATE may be one such example.

Also, I'll need to get the functionality correct w/o regression or mistake. Thanks for catching this @robertknight

LMS007:
Added another commit here. There was enough changes that I wanted to show the difference between them so its not a fixup commit. I did some more refactoring, cleaned up confusing/compound actions like removeAndDeselectAnnotation and removeSelectedAnnotation,
cleaned up tests, made them simpler and fixed the bugs that @robertknight discovered.

src/sidebar/store/modules/drafts.js
```diff
@@ -96,14 +97,41 @@ function createDraft(annotation, changes) {
...
 
-/** Remove all drafts. */
+/**
+ * Remove any drafts that are empty.
+ *
+ * An empty draft has no text and no reference tags.
+ */
+
+function deleteNewAndEmptyDrafts() {
+  const annotations = require('./annotations');
```
LMS007: This one is beyond me, It won't work imported from the top of the file, I could use some help figuring out why. The other stores import just fine.

robertknight: There is a circular import here where the "drafts" module requires the "annotations" module and vice-versa. CommonJS imports don't support circular imports properly. If we converted this to ES imports that would work. Circular imports are often something to be wary of as it indicates coupling which might be problematic in future. In this context, I think there are a few ways this could be done

1. Refactor the logic so that either 1) drafts knows about annotations OR 2) annotations knows about drafts, but not both.
2. Extract this action into a separate module which knows about annotations and drafts
3. Combine the modules together so there is no circular import to worry about

I think I'd prefer to merge this as-is though so we can move on.

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -225,4 +226,44 @@ describe('Drafts Store', () => {
       assert.deepEqual(store.unsavedAnnotations(), []);
     });
   });
+
+  describe('#deleteNewAndEmptyDrafts2', () => {
```
robertknight:
```suggestion
  describe('#deleteNewAndEmptyDrafts', () => {
```

I think this was a typo

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -225,4 +226,44 @@ describe('Drafts Store', () => {
...
+    unroll(
```
robertknight: We've been avoiding `unroll` in new code recently in favour of just using an array and `forEach`:

```js
[{
  // Object with properties for test case
}].forEach(({ prop1, prop2, ... }) => {
 it(/* description */, () => {
    // Test case
 });
});
```

The rationale was that not using a custom utility saved developers from having to learn an extra thing, and it made it easier to achieve consistency across projects.

src/sidebar/store/modules/test/drafts-test.js
```diff
@@ -225,4 +226,44 @@ describe('Drafts Store', () => {
...
+      },
+      [
+        {
+          key: 'should remove new and empty drafts',
+          annotation: {
+            id: undefined,
+            $tag: 'my_annotation_tag',
+          },
+          draft: fixtures.emptyDraft,
+          assertTest: assert.isNotOk,
```
robertknight: I think it would be preferable here to make the test case property more obviously meaningful, eg. `shouldRemove: true|false` and then use the corresponding assert in the test body.

src/sidebar/store/modules/test/selection-test.js
```diff
@@ -1,15 +1,16 @@
 'use strict';
 
+const annotations = require('../annotations');
 const createStore = require('../../create-store');
 const selection = require('../selection');
 const uiConstants = require('../../../ui-constants');
 
-describe('selection', () => {
+describe('Selection Store', () => {
```
robertknight: We have a number of different naming schemes for the top-level describe block in tests. This was discussed in Slack a while back before you joined, but I would like us to standardize on just a few:

- If the module exports one function/class, then the name of that function/class is fine providing it is reasonably likely to be unique (eg. `MyReactComponent`)
- If the module has a basename that is reasonably likely to be unique, then that's OK (eg. `build-thread`)
- Otherwise, use the path as if you were importing it from `src` eg. 'store/modules/selection'.

Having more consistent naming should make it easier to know what files to look at when looking at a list of test failures.

src/sidebar/services/root-thread.js
```diff
@@ -111,33 +98,14 @@ function RootThread($rootScope, store, searchFilter, viewFilter) {
...
   });
 
   // Remove any annotations that are deleted or unloaded
   $rootScope.$on(events.ANNOTATION_DELETED, function(event, annotation) {
     store.removeAnnotations([annotation]);
```
robertknight: Following this refactoring, it looks like there are no remaining listeners for `ANNOTATION_DELETED` or `ANNOTATIONS_UNLOADED`, so a follow-up refactoring can just replace places that emit this event with a direct call to `store.removeAnnotations`.

src/sidebar/services/test/root-thread-test.js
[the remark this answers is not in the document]
robertknight: Reading this again I get your point. Perhaps "creates a new annotation in the store when `BEFORE_ANNOTATION_CREATED` event occurs" would be clearer.

src/sidebar/store/modules/drafts.js
```diff
@@ -96,14 +97,41 @@ function createDraft(annotation, changes) {
   };
 }
 
-/** Remove all drafts. */
+/**
+ * Remove any drafts that are empty.
+ *
+ * An empty draft has no text and no reference tags.
```
robertknight: Minor nit. We usually refer to the `tags` field of an annotation as just "tags".

PR #1257, opened 2019-07-18
Add react selection-tabs component

### LINKED ISSUE #1208: Migrate SelectionTabs to Preact

LMS007:
Original PR:
https://github.com/hypothesis/client/pull/1222

fixes #1208 

There was some discussion about making the tabs actually own the annotation content as opposed to just a symbolic dummy head (as they are now). I would like to discuss that further but may need to do that in a meeting


src/sidebar/components/selection-tabs.js
```diff
@@ -1,49 +1,212 @@
...
+}) {
+  return (
+    <a
+      className={classnames('selection-tabs__type', {
+        'is-selected': selected,
+      })}
+      onClick={onChangeTab.bind(this, type)}
+      onTouchStart={onChangeTab.bind(this, type)}
+    >
+      <span>
```
robertknight: It doesn't look to me like this `<span>` is necessary, as it has no styling.

src/sidebar/components/selection-tabs.js
```diff
@@ -1,49 +1,212 @@
...
-    selectedTab: '<',
-    totalAnnotations: '<',
-    totalNotes: '<',
-    totalOrphans: '<',
-  },
-  template: require('../templates/selection-tabs.html'),
+/**
+ *  Display name of the tab and annotation count.
+ */
+function Tab({
```
robertknight: 👍 for factoring out a `Tab` component

src/sidebar/components/selection-tabs.js
```diff
@@ -1,49 +1,212 @@
...
+  onChangeTab: propTypes.func.isRequired,
+  /**
+   * Is this tab currently selected?
+   */
+  selected: propTypes.bool.isRequired,
+  /**
+   * The type value for this tab. One of
+   * 'annotation', 'note', or 'orphan'.
+   */
+  type: propTypes.string.isRequired,
```
robertknight: You can use `propTypes.oneOf(['annotation', 'note', 'orphan'])` for more precise type checking here.

src/sidebar/components/selection-tabs.js
```diff
@@ -1,49 +1,212 @@
...
+        )}
+      </span>
+    </a>
+  );
+}
+Tab.propTypes = {
+  /**
+   * Child components.
+   */
+  children: propTypes.node.isRequired,
```
robertknight: Ah, I'd missed the existence of `propTypes.node`. I think there are probably some other components where the `children` prop should use `propTypes.node` but is currently using something else.

src/sidebar/components/selection-tabs.js
```diff
@@ -1,49 +1,212 @@
...
+  }));
+
+  const isThemeClean = settings.theme === 'clean';
+
+  const selectTab = function(type) {
+    store.clearSelectedAnnotations();
+    store.selectTab(type);
+  };
+
+  const showAnnotationsUnavailableMessage = (() => {
```
robertknight: `showAnnotationsUnavailableMessage`, `showNotesUnavailableMessage` and `showSidebarTutorial` can all be values rather than immediately executed functions. The one valid case for using an IIFE that I've seen @lyzadanger use sometimes is when the value is complex to calculate and it is useful to split up into several statements. All of these functions just consist of a single `return` statement though.

src/sidebar/components/selection-tabs.js
```diff
@@ -1,49 +1,212 @@
...
+  const showNotesUnavailableMessage = (() => {
+    return selectedTab === uiConstants.TAB_NOTES && totalNotes === 0;
+  })();
+
+  const showSidebarTutorial = (() => {
+    return sessionUtil.shouldShowSidebarTutorial(session.state);
+  })();
+
+  return (
+    <div>
```
robertknight: This outer `<div>` isn't needed in the DOM as it has no styling, so it can be replaced by `Fragment`, which provides a container for JSX elements that doesn't produce any DOM elements itself.

```js
const { Fragment } = require('preact');

function MyComponent() {
  return (
    <Fragment>{/* Fragment produces no DOM elements */}
     <div>First child</div>
     <div>Second child</div>
    </Fragment>
  );
}
```

src/sidebar/components/selection-tabs.js
```diff
@@ -1,49 +1,212 @@
...
+   */
+  isLoading: propTypes.bool.isRequired,
+  /**
+   * Are there any annotations still waiting to anchor?
+   */
+  isWaitingToAnchorAnnotations: propTypes.bool.isRequired,
+  /**
+   * The currently selected tab (annotations, notes or orphans).
+   */
+  selectedTab: propTypes.string.isRequired,
```
robertknight: This could use `propTypes.oneOf`

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -1,245 +1,249 @@
...
-        selectedTab: 'note',
-        totalAnnotations: '123',
-        totalNotes: '456',
+      const wrapper = createComponent({
+        selectedTab: uiConstants.TAB_NOTES,
       });
-      const newNoteElem = elem[0].querySelectorAll('new-note-btn');
+      assert.equal(wrapper.find(NewNoteBtn).length, 1);
+    });
+    // here
```
robertknight: Comment left in by mistake?

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -1,245 +1,249 @@
...
+        session={fakeSession}
+        settings={fakeSettings}
+        {...defaultProps}
+        {...props}
+      />
+    ).dive();
+  }
+
+  // required for <Tab> rendering
+  function createDeepComponent(props) {
```
robertknight: An alternative to using `mount` here is to find the `Tab` component in the shallow-rendered output of `SelectionTabs` and then call `dive`:

```js
const tab = wrapper.find('Tab').first().dive();
assert.include(tab.text(), 'Annotations');
assert.include(tab.text(), '123');
```

This will avoid deep rendering all components (eg. `NewNoteBtn`).

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -1,245 +1,249 @@
...
-          flagEnabled: sinon.stub().returns(true),
-        },
-        settings: { theme: 'clean' },
-      });
-
-      const elem = util.createDirective(document, 'selectionTabs', {
-        selectedTab: 'annotation',
-        totalAnnotations: '123',
-        totalNotes: '456',
+    it('should not display orphans tab as if there is 0 orphans', function() {
```
robertknight:
```suggestion
    it('should not display orphans tab if there are 0 orphans', function() {
```

Pardon the pedantry 😛

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -1,245 +1,249 @@
...
-        const msg = elem[0].querySelector(
-          '.annotation-unavailable-message__label'
-        );
 
-        assert.include(
-          msg.textContent,
-          'There are no page notes in this group.'
-        );
-        assert.notInclude(msg.textContent, 'Create one by clicking the');
+        const msg = wrapper.find('.annotation-unavailable-message__label');
```
robertknight: If a selector is repeated frequently in tests, I find it can be helpful to extract that into a helper function (eg. `const unavailableMessage = wrapper => wrapper.find(...)`), so there is less work to do to update the tests if you decide to later change the HTML structure.

src/sidebar/components/selection-tabs.js
```diff
@@ -1,49 +1,207 @@
...
+   */
+  children: propTypes.node.isRequired,
+  /**
+   * The total annotations for this tab.
+   */
+  count: propTypes.number.isRequired,
+  /**
+   * Are there any annotations still waiting to anchor?
+   */
+  isWaitingToAnchor: propTypes.bool.isRequired,
```
hmstepanek: @LMS007 I have some changes in https://github.com/hypothesis/client/pull/1233 that add a todo for your PR here. I was planning on getting mine in and then doing those changes here in this one but I don't want to block this PR either.

LMS007:
Added another commit here. There was enough changes that I wanted to show the difference between them so its not a fixup commit. I did some more refactoring, cleaned up confusing/compound actions like removeAndDeselectAnnotation and removeSelectedAnnotation,
cleaned up tests, made them simpler and fixed the bugs that @robertknight discovered.

hmstepanek:
@LMS007 this got merged into a branch instead of master-we need to fix that. I've added this [PR](https://github.com/hypothesis/client/pull/1268) to actually merge it into master. Maybe you can double check it's equivalent and merge it when you get in tomorrow?

PR #1273, opened 2019-07-29
Simplify root thread generation in top-level UI components

robertknight:
Since an Angular digest cycle is triggered whenever Redux store state
changes, we can simplify the top-level component controllers in the UI
by making `vm.rootThread` a method which generates the top-level thread
when called.

Since `rootThread.thread` memoizes based its input, thread recalculation
won't happen more often than necessary. In fact, it should in theory be more efficient
now since the thread-list is only generated when the UI tries to render it, which may
be less frequently than store subscribers are called.

This change should also make it easier to migrate these components to
Preact/React in future.

----

**Testing**

Test that the thread list appears in:

- The sidebar
- The single annotation view (shown when clicking a timestamp in an annotation car)
- The stream (at http://localhost:5000/stream)

### DIFF
--- a/src/sidebar/components/annotation-viewer-content.js
+++ b/src/sidebar/components/annotation-viewer-content.js
@@ -36,15 +36,11 @@ function AnnotationViewerContentController(
   streamFilter,
   annotationMapper
 ) {
-  const self = this;
-
   store.setAppIsSidebar(false);
 
   const id = $routeParams.id;
 
-  store.subscribe(function() {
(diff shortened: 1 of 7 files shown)


robertknight:
> One question I have is since this is dependent on memoize which is currently inside the rootThread implementation would the memoize at some point move to the components and be replaced with useMemo?

That's a possibility, although if we are using the root thread in multiple places we may need to either pass that thread down the tree or create some other way to share the memoized instance, as otherwise we would end up recalculating the thread in each component that uses it.

PR #1279, opened 2019-07-30
Switch default browser for running tests from PhantomJS to headless Chrome

robertknight:
This PR changes the default browser used for running tests from PhantomJS (based on a now pretty old version of WebKit) to headless Chrome. For local development, a version of headless Chrome is installed via [Puppeteer](https://github.com/GoogleChrome/puppeteer), which provides a consistent version across all devs machines. I'm not certain this is essential, but it seems to me like it could avoid problems. On Jenkins, we use an Alpine-based image for consistency with the rest of our projects, but this means we need to use the Chrome package from the Alpine repos instead of the one Puppeteer downloads.

As part of changing the browser, I had to fix some minor issues with existing tests which were causing failures in Chrome, mostly due to various Chrome APIs being stricter about their arguments.

Switching from PhantomJS to Chrome provides many benefits, but most notably:

- Stack traces are more helpful when tests fail
- Tests run a bit faster (due to a faster JS engine + browser APIs)
- When I tried this previously, it fixed an issue where `gulp test-watch` would use a lot of CPU when idle
- It is a more realistic test environment (wrt. behavior of various APIs) that closer matches what the majority of our users are using
- It means that we can load fewer polyfills in tests, which should make them run faster

----

- [x] Make tests pass locally, on Travis and in Jenkins using headless Chrome
- [x] Make sure deployment process still works in Jenkins


.dockerignore
```diff
@@ -0,0 +1,4 @@
+.git/
+build/
+coverage/
+node_modules/
```
robertknight: This file makes building the Docker image locally faster by not sending unnecessary files (_build context_) to the Docker daemon.

src/sidebar/components/test/group-list-item-test.js
```diff
@@ -208,7 +208,7 @@ describe('GroupListItem', () => {
     const onExpand = sinon.stub();
     const wrapper = createGroupListItem(fakeGroup, { onExpand });
     const toggleSubmenu = () => {
-      const dummyEvent = new Event();
+      const dummyEvent = new Event('dummy');
```
robertknight: The _type_ argument to `Event` is required.

src/sidebar/components/test/timestamp-test.js
```diff
@@ -31,7 +31,7 @@ describe('Timestamp', () => {
 
   it('displays a link if an "href" is provided', () => {
     const wrapper = createTimestamp({
-      timestamp: '',
+      timestamp: '2016-06-10',
```
robertknight: This fixes an issue where we ended up passing an invalid argument to the `Date` constructor which throws in Chrome.

src/sidebar/services/test/api-test.js
```diff
@@ -22,7 +22,7 @@ describe('sidebar.services.api', function() {
 
   function defaultBodyForStatus(status) {
     if (status === 204) {
-      return '';
+      return null;
```
robertknight: Chrome complains if a 204 response has content.

Jenkinsfile
```diff
@@ -57,9 +56,18 @@ node {
...
+
+    stage('Setup') {
+      nodeEnv.inside("-e HOME=${workspace}") {
+        sh "yarn install"
+      }
+    }
+
     stage('Test') {
         nodeEnv.inside("-e HOME=${workspace}") {
-            sh 'make test'
```
robertknight: Instead of running only tests in Jenkins, we now run all CI steps. This helps to ensure that all checks have been run on the master branch before it is deployed to prod. `checkformatting` is not particularly important in this context, but it is also pretty quick to run, so I've included it for consistency.

robertknight:
I'm going to merge this now so I can test the end-to-end CI deployment workflow on Jenkins and seek a post-hoc review.

PR #1303, opened 2019-08-12
Fix JS error when searching after creating a page draft

### LINKED ISSUE #1290: JS Error after searching while an annotation draft exists
I'm not sure this is a recent regression or been around for longer, but there is js error when searching after creating a draft note annotation. *This does not happen when creating a highlight draft annotation.*

Steps to reproduce:
- Create a new page note
- Click the search bar and type a search phrase like "test"
- Click enter

Screen Shot 2019-08-05 at 9 47 36 AM

LMS007:
This adjusts view-filter so it can handle draft annotations which can be defined by a missing id.

fixes #1290 

Steps to repro.

1. Create page note draft
2. Search for something
3. See js error in console.

### DIFF
--- a/src/sidebar/services/test/view-filter-test.js
+++ b/src/sidebar/services/test/view-filter-test.js
@@ -74,9 +74,10 @@ describe('sidebar/services/view-filter', () => {
   describe('"any" field', () => {
     it('finds matches in any field', () => {
       const annotations = [
-        { id: 1, text: poem.tiger },
-        { id: 2, user: 'Tiger' },
-        { id: 3, tags: ['Tiger'] },
+        { id: 1, text: poem.tiger, target: [{}] },
+        { id: 4, user: 'lion', target: [{}] },
+        { id: 2, user: 'Tiger', target: [{}] },
+        { id: 3, tags: ['Tiger'], target: [{}] },
       ];
       const filters = { any: { terms: ['Tiger'], operator: 'and' } };
 
@@ -191,6 +192,7 @@ describe('sidebar/services/view-filter', () => {
       const annotation = {
         id: 1,
         updated: isoDateWithAge(50),
+        target: [{}],
       };
       const filters = {
         since: { terms: [100], operator: 'and' },
@@ -205,6 +207,7 @@ describe('sidebar/services/view-filter', () => {
       const annotation = {
         id: 1,
         updated: isoDateWithAge(150),
+        target: [{}],
       };
       const filters = {
         since: { terms: [100], operator: 'and' },
@@ -217,7 +220,11 @@ describe('sidebar/services/view-filter', () => {
   });
 
   it('ignores filters with no terms in the query', () => {
-    const annotation = { id: 1, tags: ['foo'] };
+    const annotation = {
+      id: 1,
+      tags: ['foo'],
+      target: [{}],
+    };
     const filters = {
       any: {
         terms: ['foo'],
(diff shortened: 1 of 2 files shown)


src/sidebar/services/view-filter.js
```diff
@@ -100,9 +100,8 @@ function viewFilter(unicode) {
     quote: {
       autofalse: ann => (ann.references || []).length > 0,
       value(annotation) {
-        if (!annotation.target) {
-          // FIXME: All annotations *must* have a target, so this check should
-          // not be required.
+        if (!annotation.id) {
```
robertknight: Reading this code it wasn't immediately obvious why we're testing for `annotation.id` but the following code doesn't access `annotation.id` at all. Looking at the value of a new not-yet-saved annotation, it seems the issue is that the `target` array is empty, whereas this code previously assumed it was either not present or non-empty if present.

Screenshot 2019-08-13 15 05 18

Related to this, I notice this code partly duplicates some logic from the [annotation component](https://github.com/hypothesis/client/blob/2334426497bcf4c930ba9deae91a5d03ab35616c/src/sidebar/components/annotation.js#L331). In future we could extract that into a helper.

src/sidebar/services/view-filter.js
```diff
@@ -183,7 +182,9 @@ function viewFilter(unicode) {
     const rootFilter = new BinaryOpFilter('and', fieldFilters);
 
     return annotations
-      .filter(ann => rootFilter.matches(ann))
+      .filter(ann => {
+        return rootFilter.matches(ann) && ann.id;
```
robertknight: An alternative way of resolving the issue would be to swap the order of the checks here (`return ann.id && rootFilter.matches(ann)`) so that the logic short-circuits and the filters never have to deal with the possibility of a non-saved annotation.

src/sidebar/services/view-filter.js
```diff
@@ -100,9 +100,9 @@ function viewFilter(unicode) {
     quote: {
       autofalse: ann => (ann.references || []).length > 0,
       value(annotation) {
-        if (!annotation.target) {
-          // FIXME: All annotations *must* have a target, so this check should
-          // not be required.
+        if (annotation.target.length === 0) {
+          // Annotations with no targets are draft page note
+          // annotations -- just ignore them.
```
robertknight: Looking at the coverage report (open `coverage/index.html` and navigate to `sidebar > services > view-filter.js`) the body of this `if` is never called because of the earlier check for whether an annotation has an id.

I had a look in "h" and the backend will currently always produce an annotation with a `target` array that has [exactly one entry](https://github.com/hypothesis/h/blob/0d4cd0c606bb62deb97a8dac2bded60efb95b517/h/presenters/annotation_base.py#L48). It could still be useful to have a test case in case that ever changes in future.

src/sidebar/services/test/view-filter-test.js
```diff
@@ -186,11 +187,18 @@ describe('sidebar/services/view-filter', () => {
     });
   });
 
+  /*
+  Alter conditions to make more sense with cotext
+
+  - allow for short circuit of missing id to prevent the check ojn matches()
+- checking target length makes more sense in the context of code it nests.
+*/
```
robertknight: These comments look like some notes you were writing for yourself that accidentally got left in the PR?

LMS007:
@robertknight awe, switching the condition to short circuit on the id caused the test to miss its target... good catch! This is much better as we can just rip out the entire check and test case in the first place

PR #1307, opened 2019-08-15
Fix JS error when searching after creating a page draft

LMS007:
This adjusts view-filter so it can handle draft annotations which can be defined by a missing id.

### DIFF
--- a/src/sidebar/services/test/view-filter-test.js
+++ b/src/sidebar/services/test/view-filter-test.js
@@ -74,9 +74,10 @@ describe('sidebar/services/view-filter', () => {
   describe('"any" field', () => {
     it('finds matches in any field', () => {
       const annotations = [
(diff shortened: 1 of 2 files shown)


src/sidebar/services/test/view-filter-test.js
```diff
@@ -233,4 +240,38 @@ describe('sidebar/services/view-filter', () => {
 
     assert.deepEqual(result, [1]);
   });
+
+  describe('malformed target object', () => {
+    it('should not fail on annotations without a target object', () => {
```
LMS007: These two tests are a little different in that they dont test the result of the filter, rather they just make sure the code does not produce a js error.  If we remove either of those checks in view-test.js on line 103, these tests will fail. The fitler result itself does not change and is not important to the test.

If there is a better way to do this, please let me know

Issue #1235, opened 2019-07-09
Namespace the store


LMS007:
Currently, the entire state for the store gets flattened to a single reducer. This makes it hard to separate logic and function. We should use separate reducers and state for each module.

1. Prep create-store so it can facilitate namespaces
2. Convert each module to use a namespace

Modules to convert

- [x] activity
- [x] annotations
- [x] direct-links
- [x] drafts
- [x] frames
- [x] groups
- [x] links
- [x] real-time-annotations (depends on #1319)
- [x] session
- [x] viewer

LMS007:
From Rob

> The code that creates these methods from the functions exported by each store module is here: 

https://github.com/hypothesis/client/blob/2ed42cfb45acc930850664c4956aa9f697502343/src/sidebar/store/create-store.js#L47

LMS007:
It may also be helpful to further namespace the actions from the selectors. Its not always clear which is which and this could make the code more readable.

example.

`focusGroup()` vs `focusedGroup()`

The first is an action while the second is just a selector. There is no easy way of knowing which is which without looking them up.

PR #1309, opened 2019-08-19
Add namespace capability to store modules

LMS007:
Each module can now opt into being namespaced by exporting the value for "namespace". The `links` module  is the only module that has been namespaced.

**When using a namespace a few things change:**

- Reducers need not be aware of the namespace. The scope will be set when they are called.
- Namespaced selectors of the module will need to use the namespace because they will be given the root of the store.
- A namespaced module's state keys no longer need to be unique since it can't collide with other state keys.
- A namespaced module's state can return an array rather than an object if the state is a simple list of thing.
- There is a new helper method called getRootState which returns the root of the store which has access to both "base" and all other new namespaces.
- getRootState returns the root of the store, 

**Some behavior changes for Legacy (non-namespaced) modules:**

- Legacy modules are nested in a pseudo namespace called "base" which will be depreciated once all module are namespaced. 
- getState returns the "base" state
- Selectors of legacy modules will not need to use the namespace because they will be given a scoped state of the "base" namespace

The long term plan is to export a namespace for each module at which point we'll do a little deprecation work inside of create-store. 

_Note, this work does not namespace actions or selectors functions themselves -- that work may come later._

#1235

### DIFF
--- a/src/sidebar/store/create-store.js
+++ b/src/sidebar/store/create-store.js
@@ -26,14 +26,63 @@ const { createReducer, bindSelectors } = require('./util');
  * @param [any[]] middleware - List of additional Redux middlewares to use.
  */
 function createStore(modules, initArgs = [], middleware = []) {
-  // Create the initial state and state update function.
-  const initialState = Object.assign(
-    {},
-    ...modules.map(m => m.init(...initArgs))
-  );
-  const reducer = createReducer(...modules.map(m => m.update));
+  // Create the initial state and state update function. The "base"
+  // namespace is reserved for non-namespaced modules which will eventually
+  // be converted over.
 
-  // Create the store.
+  // Namespaced objects for initial states.
+  const initialState = {
+    base: null,
+  };
+  // Namespaced reducers from each module.
+  const allReducers = {
+    base: null,
+  };
+  // Namespaced selectors from each module.
+  const allSelectors = {
+    base: {
+      selectors: {},
+      // Tells the bindSelector method to use local state for these selectors
+      // rather than the top level root state.
+      useLocalState: true,
+    },
+  };
+
+  // Temporary list of non-namespaced modules used for createReducer.
+  const baseModules = [];
+
+  // Iterate over each module and prep each module's:
+  //    1. state
+  //    2. reducers
+  //    3. selectors
+  //
+  // Modules that have no namespace get dumped into the "base" namespace.
+  //
+  modules.forEach(module => {
+    if (module.namespace) {
+      initialState[module.namespace] = module.init(...initArgs);
+      allReducers[module.namespace] = createReducer(module.update);
+      allSelectors[module.namespace] = {
+        selectors: module.selectors,
+      };
+    } else {
+      // No namespace
+      allSelectors.base.selectors = {
+        // Aggregate the selectors into a single "base" map
+        ...allSelectors.base.selectors,
+        ...module.selectors,
+      };
+      initialState.base = {
+        ...initialState.base,
+        ...module.init(...initArgs),
+      };
+      baseModules.push(module);
+    }
+  });
+
+  // Create the base reducer for modules that are not opting in for namespacing
+  allReducers.base = createReducer(...baseModules.map(m => m.update));
+
   const defaultMiddleware = [
     // The `thunk` middleware handles actions which are functions.
     // This is used to implement actions which have side effects or are
@@ -41,13 +90,30 @@ function createStore(modules, initArgs = [], middleware = []) {
     thunk,
   ];
   const enhancer = redux.applyMiddleware(...defaultMiddleware, ...middleware);
-  const store = redux.createStore(reducer, initialState, enhancer);
+
+  // Create the store.
+  const store = redux.createStore(
+    redux.combineReducers(allReducers),
+    initialState,
+    enhancer
+  );
+
+  // Temporary wrapper while we use the "base" namespace. This allows getState
+  // to work as it did before. Under the covers the state is actually
+  // nested inside "base" namespace.
+  const getState = store.getState;
+  store.getState = () => getState().base;
+
+  // Because getState is overridden, we still need a fallback for the root state
+  // for the namespaced modules. They will temporarily use getRootState
+  // until all modules are namespaced and then this will be deprecated.
+  store.getRootState = () => getState();
 
   // Add actions and selectors as methods to the store.
   const actions = Object.assign({}, ...modules.map(m => m.actions));
   const boundActions = redux.bindActionCreators(actions, store.dispatch);
-  const selectors = Object.assign({}, ...modules.map(m => m.selectors));
-  const boundSelectors = bindSelectors(selectors, store.getState);
+  const boundSelectors = bindSelectors(allSelectors, store.getRootState);
+
   Object.assign(store, boundActions, boundSelectors);
 
   return store;
--- a/src/sidebar/store/modules/annotations.js
+++ b/src/sidebar/store/modules/annotations.js
@@ -254,7 +254,7 @@ function addAnnotations(annotations, now) {
 
   return function(dispatch, getState) {
     const added = annotations.filter(function(annot) {
-      return !findByID(getState().annotations, annot.id);
+      return !findByID(getState().base.annotations, annot.id);
     });
 
     dispatch({
@@ -262,7 +262,7 @@ function addAnnotations(annotations, now) {
       annotations: annotations,
     });
 
-    if (!getState().isSidebar) {
+    if (!getState().base.isSidebar) {
       return;
     }
 
@@ -277,7 +277,7 @@ function addAnnotations(annotations, now) {
     if (anchoringIDs.length > 0) {
       setTimeout(() => {
         // Find annotations which haven't yet been anchored in the document.
-        const anns = getState().annotations;
+        const anns = getState().base.annotations;
         const annsStillAnchoring = anchoringIDs
           .map(id => findByID(anns, id))
           .filter(ann => ann && metadata.isWaitingToAnchor(ann));
--- a/src/sidebar/store/test/util-test.js
+++ b/src/sidebar/store/test/util-test.js
@@ -14,8 +14,23 @@ const fixtures = {
       return { tab: action.tab };
     },
   },
-  countAnnotations: function(state) {
-    return state.annotations.length;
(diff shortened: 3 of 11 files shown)


src/sidebar/store/util.js
```diff
@@ -20,42 +20,62 @@ function actionTypes(updateFns) {
...
-      const args = [].slice.apply(arguments);
-      args.unshift(getState());
-      return selector.apply(null, args);
-    };
-    return bound;
-  }, {});
+function bindSelectors(namespaces, getState) {
+  const totalSelectors = {};
+  Object.keys(namespaces).forEach(namespace => {
+    const selectors = namespaces[namespace].selectors;
```
LMS007: todo: add test coverage for this

src/sidebar/store/util.js
```diff
@@ -20,42 +20,62 @@ function actionTypes(updateFns) {
...
-  return (state, action) => {
+  return (state = {}, action) => {
     const fns = actionToUpdateFns[action.type];
     if (!fns) {
       return state;
     }
+    // Some states return an array rather than an object. They need to be
+    // handled differently so we don't cast them to an object.
+    if (Array.isArray(state)) {
+      return [...fns[0](state, action)];
```
LMS007: todo: add test coverage here

LMS007: The array will go away once we move all modules to namespaces. Array types are only used for namespaces modules and they can only have 1 item in the array. Using `[0]` is okay here for now until we remove the array altogether.

src/sidebar/store/modules/annotations.js
```diff
@@ -254,15 +254,15 @@ function addAnnotations(annotations, now) {
 
   return function(dispatch, getState) {
     const added = annotations.filter(function(annot) {
-      return !findByID(getState().annotations, annot.id);
+      return !findByID(getState().base.annotations, annot.id);
```
LMS007: thunk actions are not scoped so they need to drill into the "base" namespace until the module is converted.

src/sidebar/store/util.js
```diff
@@ -20,42 +20,62 @@ function actionTypes(updateFns) {
...
-  }, {});
+function bindSelectors(namespaces, getState) {
+  const totalSelectors = {};
+  Object.keys(namespaces).forEach(namespace => {
+    const selectors = namespaces[namespace].selectors;
+    const scopeSelector = namespaces[namespace].scopeSelector;
+    Object.keys(selectors).forEach(selector => {
+      totalSelectors[selector] = function() {
+        const args = [].slice.apply(arguments);
+        if (scopeSelector) {
```
LMS007: when using scopeSelector, the selector does not need to drill into the top level namespace scope. This is a way to allow legacy modules to work without changes but keep them under the "base" namespace.

Only the legacy (non-namespaced) modules get `scopeSelector=true`. This will eventually be deprecated when all modules are switched over.

src/sidebar/store/test/util-test.js
```diff
@@ -36,6 +51,15 @@ describe('reducer utils', function() {
   });
 
   describe('#createReducer', function() {
+    it('can not return undefined during initialization', function() {
```
robertknight: It wasn't immediately clear to me what this was trying to test. `it("returns an object if input state is undefined and action is unrecognized")` might be a more direct description, with a brief note on when this happens as a comment below.

src/sidebar/store/test/util-test.js
```diff
@@ -109,19 +133,19 @@ describe('reducer utils', function() {
...
-
-      getState.returns({ annotations: annotations.concat([{ id: 2 }]) });
-      assert.equal(bound.countAnnotations(), 2);
+        namespace2: {
+          annotations: [{ id: 1 }],
+        },
+      });
+      const bound = util.bindSelectors(fixtures.selectors, getState);
+      // test namespaced selector
+      assert.equal(bound.countAnnotations1(), 1);
```
robertknight: `countAnnotations1` is a non-namespaced selector rather than a namespaced one. Conversely `countAnnotations2` is a namespaced selector.

src/sidebar/store/util.js
```diff
@@ -20,42 +20,62 @@ function actionTypes(updateFns) {
...
     });
   });
 
-  return (state, action) => {
+  return (state = {}, action) => {
     const fns = actionToUpdateFns[action.type];
     if (!fns) {
       return state;
     }
+    // Some states return an array rather than an object. They need to be
```
robertknight: By "some states return an array" did you mean "some modules use an array as their state"?

src/sidebar/store/util.js
```diff
@@ -20,42 +20,62 @@ function actionTypes(updateFns) {
...
-    const selector = selectors[key];
-    bound[key] = function() {
-      const args = [].slice.apply(arguments);
-      args.unshift(getState());
-      return selector.apply(null, args);
-    };
-    return bound;
-  }, {});
+function bindSelectors(namespaces, getState) {
+  const totalSelectors = {};
```
robertknight: I think `allSelectors` or `combinedSelectors` would be a better name than `totalSelectors` here, since that implies a numeric sum to me.

src/sidebar/store/util.js
robertknight: We'll still need to come up with something for selectors that access state from multiple modules, or some equivalent for what those selectors do.

src/sidebar/store/util.js
```diff
@@ -20,42 +20,62 @@ function actionTypes(updateFns) {
...
+  return totalSelectors;
 }
 
 module.exports = {
-  actionTypes: actionTypes,
-  bindSelectors: bindSelectors,
-  createReducer: createReducer,
+  actionTypes,
+  bindSelectors,
+  createReducer,
```
robertknight: 👍 - Use of object shorthand is preferred in new code.

src/sidebar/store/create-store.js
```diff
@@ -26,28 +29,86 @@ const { createReducer, bindSelectors } = require('./util');
...
-    // This is used to implement actions which have side effects or are
-    // asynchronous (see https://github.com/gaearon/redux-thunk#motivation)
-    thunk,
-  ];
-  const enhancer = redux.applyMiddleware(...defaultMiddleware, ...middleware);
-  const store = redux.createStore(reducer, initialState, enhancer);
+  const store = configureStore({
+    reducer: totalReducers,
+    middleware: [thunk, ...middleware],
+    devTools: process.env.NODE_ENV !== 'production',
```
robertknight: I was about to write that there wasn't much point enabling the devtools as it didn't support applications that run in iframes. It looks like that [might have changed](https://github.com/zalmoxisus/redux-devtools-extension/issues/50) though. I'm happy to add this if it works and is useful, but see caveat about redux-starter-kit below.

LMS007:
- Updated this to have test full coverage again.
- Fixed the issue with link module (I originally slotted that for the first namespaced example, then verted it back, but in that effort, I didn't change a few peripheral lines that causes some errors. 
- `links` module is now namespaced. 
- redux-start-kit is removed and this uses combineReducers directly.

src/sidebar/store/util.js
LMS007: I will make that change

src/sidebar/store/util.js
LMS007: namespaced selectors get the root store so thy will still have access to everything , but paths will change of course.

src/sidebar/store/test/util-test.js
LMS007: The action does not matter here, but perhaps using "fake" is misleading for that action

src/sidebar/store/test/util-test.js
LMS007: good catch!

src/sidebar/store/util.js
LMS007: yes

src/sidebar/store/test/util-test.js
```diff
@@ -36,6 +51,16 @@ describe('reducer utils', function() {
   });
 
   describe('#createReducer', function() {
+    it('returns an object if input state is undefined', function() {
+      // See redux.js:assertReducerShape in the "redux" package.
+      const reducer = util.createReducer(fixtures.update);
+      const initialState = reducer(undefined, {
+        type: 'ADD_ANNOTATIONS',
```
robertknight: I think it might be clearer here to name this `DUMMY_ACTION` or something to indicate that it is an action that the reducer doesn't handle. Internally Redux will dispatch actions with [randomized "type" strings](https://github.com/reduxjs/redux/blob/9c9a4d2a1c62c9dbddcbb05488f8bd77d24c81de/src/utils/actionTypes.js#L16) for various purposes.

src/sidebar/store/test/util-test.js
```diff
@@ -105,23 +130,39 @@ describe('reducer utils', function() {
 
       assert.deepEqual(newState, { firstCounter: 6, secondCounter: 11 });
     });
+
+    it('creates a reducer that can consume a state that returns an array', function() {
```
robertknight:
```suggestion
    it('supports reducer functions that return an array', function() {
```

I had trouble parsing this as "a state" is just an object, it doesn't return anything.

src/sidebar/store/create-store.js
```diff
@@ -26,28 +26,94 @@ const { createReducer, bindSelectors } = require('./util');
...
+  //    3. selectors
+  //
+  // Modules that have no namespace get dumped into the "base" namespace.
+  //
+  modules.forEach(module => {
+    if (module.namespace) {
+      initialState[module.namespace] = module.init(...initArgs);
+      allReducers[module.namespace] = createReducer(module.update);
+      allSelectors[module.namespace] = {
+        selectors: module.selectors,
```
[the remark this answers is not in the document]
LMS007: @robertknight, I intended useLocalState to be a property of the intermediary translation object allSelectors, but not part of any modules (since its temporary). In other words, we would never export that value from any module. It's just hard coded to the "base" set but not any other namespaced module. Do you think that is a issue?

My intention here is that all namespaced module selectors get the top level root state parameter  and would be responsible for drilling down into each namespace they want to access. This allows module selectors to see other states too.

Thoughts on that?

src/sidebar/store/test/util-test.js
LMS007: better :)

PR #1315, opened 2019-08-21
Namespace the frames module

LMS007:
I believe some of these tests were a bit older as defined by a comment 

_"Note that some of the tests in this module pre-date the `createStore` helper and have not been refactored to use it yet."_

I went ahead and refactored these tests to use it so the tests are a bit simpler now.

#1235

### DIFF
--- a/src/sidebar/store/modules/frames.js
+++ b/src/sidebar/store/modules/frames.js
@@ -5,25 +5,21 @@ const { createSelector } = require('reselect');
 const util = require('../util');
 
 function init() {
-  return {
-    // The list of frames connected to the sidebar app
-    frames: [],
-  };
+  // The list of frames connected to the sidebar app
+  return [];
 }
 
 const update = {
   CONNECT_FRAME: function(state, action) {
-    return { frames: state.frames.concat(action.frame) };
+    return [...state, action.frame];
   },
 
   DESTROY_FRAME: function(state, action) {
-    return {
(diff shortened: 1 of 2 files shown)


src/sidebar/store/modules/frames.js
```diff
@@ -5,25 +5,21 @@ const { createSelector } = require('reselect');
 const util = require('../util');
 
 function init() {
-  return {
-    // The list of frames connected to the sidebar app
-    frames: [],
-  };
+  // The list of frames connected to the sidebar app
+  return [];
```
LMS007: This gets refactored as a direct array

src/sidebar/store/modules/test/frames-test.js
```diff
@@ -110,89 +84,83 @@ describe('sidebar/store/modules/frames', function() {
...
-      function(testCase) {
-        let state = init();
-        if (testCase.features) {
-          state.session.features = testCase.features;
-        }
-        testCase.frames.forEach(function(frame) {
-          state = update(state, actions.connectFrame(frame));
-        });
-        assert.deepEqual(selectors.searchUris(state), testCase.searchUris);
+  describe('xxx #searchUris', function() {
```
robertknight: The "xxx" looks like it was left in by mistake.

robertknight: I pushed a tiny change to remove this. Once CI passes, I'll merge.

PR #1317, opened 2019-08-21
Namespace the direct-linked module

LMS007:
#1235


src/sidebar/store/modules/direct-linked.js
```diff
@@ -129,6 +129,7 @@ function clearDirectLinkedIds() {
 
 module.exports = {
   init,
+  namespace: 'directLinked',
```
robertknight: This matches the current name of the module. I think it would make sense to rename the module to `direct-links.js` and the namespace to `directLinks` in future for consistency with "frames", "groups", "annotations" etc.

PR #1320, opened 2019-08-23
Replace legacy Sentry JS SDK with new Sentry SDK

robertknight:
As a step towards re-enabling Sentry error reports from the client in QA and production environments, this PR replaces the legacy `raven-js` SDK with the [new Sentry JavaScript SDK](https://github.com/getsentry/sentry-javascript). The new SDK is designed to be more unified across different languages in terms of its API, features etc. We already use the new Python SDK for h and the lms app.

A breakdown of the changes:

- Replace `raven-js` with `@sentry/browser`. The client has a wrapper module for the SDK which used to be in `src/sidebar/raven.js` and has now been moved to `src/sidebar/util/sentry.js`
- Remove custom logic that is no longer necessary for handling rejected promises. The new SDK already does this for us
- Remove the Angular.js integration for the time being rather than re-implement it with the new SDK. We won't need this long term since we're migrating away from Angular.
- Remove the URL rewriting for the time being. That was used for the browser extension. I think we may need to reintroduce it, but for the moment I just want to get everything working with the embed and to keep the Sentry config as simple as possible
- We now configure the Sentry environment (dev, qa, prod). This is currently based on the h environment since the two are always in sync. The browser extension would need to supply its own `environment` values

One new thing I've added is a basic client-side quota to help protect our overall Sentry quotas. For each client session, we only allow a maximum of 5 errors to be reported to Sentry. Sentry also has its own [server side filtering, quotas and rate limiting](https://docs.sentry.io/accounts/quotas/). Our past experience though was that it was useful to prevent a problematic client (unusual browser, faulty configuration, network issues) from spamming Sentry in the first place. We'll have to see how this works out in production.

**Testing**

If you want to test this manually, you can do so as follows:

1. Check out this branch and the [corresponding h PR](https://github.com/hypothesis/h/pull/5686)
2. Set the `SENTRY_DSN_CLIENT` env var in your h dev environment to the value of the "client-sentry-sdk" DSN from this page: https://sentry.io/settings/hypothesis/projects/client/keys/
3. Modify the client to generate an error when you click something (for example, find a Preact component and modify one of the event handlers to generate an error)
4. In the dev client, trigger an error. You should see the error being reported as coming from "sentryWrapped" in the console.
5. Go to https://sentry.io/organizations/hypothesis/issues/?environment=dev&project=69811 (note that the "dev" environment is selected) and your error should appear.

Note that the stack trace and exact error location will be missing since sourcemaps won't have been uploaded to Sentry for your dev environment. If you want to do that, you'll need to create your own Sentry project and use the Sentry CLI to upload sourcemaps. For QA and production we upload sourcemaps to public URLs which Sentry can fetch automatically.

As an alternative to using the Hypothesis organization on Sentry, you can also create your own organization and create a project within that.


src/sidebar/util/sentry.js
```diff
@@ -0,0 +1,73 @@
...
+const warnOnce = require('../../shared/warn-once');
+
+/**
+ * @typedef SentryConfig
+ * @prop {string} dsn
+ * @prop {string} environment
+ */
+
+let eventsSent = 0;
+const maxEventsToSendPerSession = 5;
```
hmstepanek: How did you decide on 5?

robertknight: I just picked a number that seemed reasonable to me for one session of the client. It means that if multiple errors happen, we'll find out about them, but if there are a cacophony of errors, only the first few will hit Sentry.

My expectation is that we'll adjust the filtering based on what we see happening in production.

robertknight:
> How many duplicate errors were getting sent before we disabled sentry per session (aka: will this limit have an impact at reducing the number of requests).

Unfortunately it was so long ago that we last used Sentry for the client that we don't have recent data to use as a gauge. The number here is just a guess at something that seemed reasonable to me.

> When is the session reset (aka: the counter for the number of requests sent reset).

Whenever a new instance of the client loads. In other words, whenever you reload the page or open a new tab, that is a separate instance of the client.

> Are there alternatives to this implementation that might allow us to reduce that limit across sessions like storing it in the browser's cache-what are the impacts of that?

Sure, we could do something like that using e.g. `localStorage`. I'd like to see what kind of noise we get when this is live in production before adding more logic to the client. One caveat is that browsers with built-in ad-blocking/tracking protection may prevent us from locally storing preferences or other data across sessions, as that would allow us to track a user across different sites even if they weren't logged in.

PR #1321, opened 2019-08-23
Namespace selection module

LMS007:


src/sidebar/services/test/root-thread-test.js
```diff
@@ -165,9 +162,7 @@ describe('rootThread', function() {
     });
 
     it('passes the highlighted set to buildThread()', function() {
-      fakeStore.state = Object.assign({}, fakeStore.state, {
-        highlighted: ['id1', 'id2'],
-      });
+      fakeStore.state.selection.highlighted = ['id1', 'id2'];
```
LMS007: I did this in a number of places. In the same file I see that we set the store directly and we also set it as a copy which was not consistent. Since there is no possible way to mutate the store across test cases Its simpler to just set the values directly rather than copy to new objects .

src/sidebar/services/test/root-thread-test.js
```diff
@@ -194,33 +189,44 @@ describe('rootThread', function() {
       ];
     }
 
-    unroll(
```
LMS007: did some refactoring where I could

src/sidebar/store/modules/annotations.js
```diff
@@ -260,6 +249,7 @@ function addAnnotations(annotations, now) {
     dispatch({
       type: actions.ADD_ANNOTATIONS,
       annotations: annotations,
+      currentAnnotationCount: getState().base.annotations.length,
```
robertknight: So to summarize the pattern here: Where existing reducers rely state from multiple modules, we're either:

- Moving the logic to a different reducer so that each reducer only depends on its own state. This is the preferred approach.
- If that's not possible, instead calculating the required information in the action creator, which may involve turning it into a thunk, and adding that as information to the action as part of the payload. The downside is that e.g. `ADD_ANNOTATIONS` needs to know about the requirements of the selection module and includes additional information beyond that needed to describe what happened. The upside is that it allows the "selection" module reducers to avoid needing to access external state.

I think it would be interesting to know whether the wider Redux community have found alternative patterns/approaches for this problem that could work well.

src/sidebar/services/test/root-thread-test.js
```diff
@@ -255,11 +252,7 @@ describe('rootThread', function() {
 
     it('filter matches only notes when Notes tab is selected', function() {
       fakeBuildThread.reset();
-
-      fakeStore.state = Object.assign({}, fakeStore.state, {
-        selectedTab: uiConstants.TAB_NOTES,
-      });
-
+      fakeStore.state.selection.selectedTab = uiConstants.TAB_NOTES;
```
robertknight: I'm a little worried here that seeing code like this in these tests, where it happens to work fine, might make developers less familiar with Redux think it is OK to write code like this in the actual application, where it would break important invariants.

We use "seamless-immutable" to guard against accidental mutation of state in some tests. It might be a good idea to start making use of it in the actual application in development mode as well.

PR #1335, opened 2019-08-28
Record document URL as "extra" metadata rather than a tag

robertknight:
https://github.com/hypothesis/client/pull/1328 added the URL of the page in which the client was embedded as metadata in crash reports. It was added as a Sentry "tag" which is an indexed key/value pair. It turns out that Sentry has various constraints on the characters that can appear in tags as well as their length. Any tags which don't match these constraints are simply dropped so we lose the information. This turned out to happen quite often for document URLs.

For debugging purposes we want to capture the full document URL unmodified. Therefore capture this as un-indexed "extra" metadata, which has fewer constraints, rather than
an indexed "tag".

In Sentry error reports, the metadata will now show up in the "Additional data" section:

Screenshot 2019-08-28 08 33 48

### DIFF
--- a/src/sidebar/util/sentry.js
+++ b/src/sidebar/util/sentry.js
@@ -50,7 +50,7 @@ function init(config) {
   // via `document.referrer`. More information about the document is available
   // later when frames where the "annotator" code has loaded have connected to
   // the sidebar via `postMessage` RPC messages.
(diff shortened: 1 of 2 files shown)


robertknight:
I'm going to go ahead and merge this because a large percentage of error reports are missing the information at the moment, which is hampering debugging.

PR #1346, opened 2019-09-03
Add docs for building the docs locally

seanh:
This adds documentation for how to build the docs locally.

Previously the `make docs` command required you to have Python 3 and tox
installed system-wide, neither of which are required for any of our
other projects anymore, and this wasn't documented anywhere.

This PR changes `make docs` to use pyenv to install Python 3 and tox, as
all our other Python projects do, and adds a Building the Docs Locally
page to the docs with instructions on how to install pyenv and build the
docs locally.

pyenv and tox are _not_ added to the general client dev install
instructions since they're only needed to build the docs.


docs/developers/docs.rst
```diff
@@ -0,0 +1,47 @@
+Building the Docs Locally
+=========================
+
+The Hypothesis client's documentation is hosted at https://h.readthedocs.io/projects/client/.
+The docs are built using https://readthedocs.org.
```
robertknight:
```suggestion
The docs are built using [Sphinx](https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html) and hosted by https://readthedocs.org.
```

seanh: I'd like to leave this as just a link to RTD's home page, which includes a link to their docs. Technically it's not just Sphinx that our docs use, it's the RTD system, which has its own docs and adds various particular ways of doing things and additional features on top of Sphinx. Markdown and MkDocs support, for example, and GitHub integration, canonical URLs, versioning, their own configuration file format, their dashboard, custom domain support, sub projects, redirects, localisation, etc. The RTD docs link to the reST and Sphinx docs as appropriate, as well as providing their own "Getting started with Sphinx" guide. So I think readthedocs.org is the correct starting page for "How do our docs work?"

PR #1350, opened 2019-09-04
Namespace annotations module

LMS007:
This PR is branched from the `namespace-viewer-module` so that should be merged first.

#1235

### DIFF
--- a/src/sidebar/store/modules/test/annotations-test.js
+++ b/src/sidebar/store/modules/test/annotations-test.js
@@ -7,7 +7,6 @@ const fixtures = require('../../../test/annotation-fixtures');
 const selection = require('../selection');
 const viewer = require('../viewer');
 const uiConstants = require('../../../ui-constants');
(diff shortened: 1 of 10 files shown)


src/sidebar/store/modules/test/annotations-test.js
```diff
@@ -147,14 +168,56 @@ describe('annotations reducer', function() {
...
+      },
+      {
+        // Moderator un-flags annotation
+        wasFlagged: true,
+        nowFlagged: false,
+        oldModeration: { flagCount: 1 },
+        newModeration: { flagCount: 0 },
+      },
+    ].forEach(testCase => {
+      it('updates the flagged status of an annotation', () => {
```
robertknight: I realize this description was taken from existing code. I think it would be helpful here to add some discriminator to the description to indicate which case failed, in case any of these tests fail. I typically do this by putting something in brackets at the end of the string (``it(`does something ${testCase.description}`)``).

PR #1352, opened 2019-09-05
Fix race condition when anchoring in a PDF during initial document load

### LINKED ISSUE #1330: Race condition causes annotations to sometimes fail to anchor in PDFs
**Steps to reproduce**

Go to https://via.hypothes.is/https://www.nsqol.org/wp-content/uploads/2019/02/National-Standards-for-Quality-Online-Teaching.pdf and select the "Public" group

**Expected result**

There is an annotation on the last page (https://hypothes.is/a/LJMxNsjFEem55S9pNtX0Aw). It should always anchor.

**Actual result**

Testing in Chrome 78.0.3887.7, I see the annotation fail to anchor once every few attempts.

Inspecting the store state in the sidebar showed that no anchoring timeout had occurred but instead the annotation had just failed to anchor:

Screenshot 2019-08-27 13 27 48

Adding a breakpoint in the client code in `annotator.bundle.js` where anchoring failures are handled (the `catch` clause in response to a failed `anchor` - [see code](https://github.com/hypothesis/client/blob/4dfebbd6afbd415cd0e7858e0a86ee899f4bed32/src/annotator/guest.coffee#L235)), I got the following exception:

Screenshot 2019-08-27 13 36 05

Which points to `pdfPage` being `undefined` here: https://github.com/hypothesis/client/blob/4dfebbd6afbd415cd0e7858e0a86ee899f4bed32/src/annotator/anchoring/pdf.js#L76

Looking at [the implementation of the `getPageView` method](https://github.com/mozilla/pdf.js/blob/c9a74984ec7eca8a49d33acdf760761e3d6df989/web/pdf_viewer.js#L112) in PDF.js which returns the object that is expected to have a `pdfPage` property, it looks like there are indeed certain states when that method could return undefined. Reading through the implementation, it looks like we should be using `pdfDocument.getPage` whenever we want a reference to a page, which returns a `Promise<PDFPageProxy>` on which we can call `getTextContent` directly.

robertknight:
Fix a race condition that could occur when the client tried to anchor
annotations in a PDF while the PDF was still loading.

PDF.js' `PDFViewer` class initializes its `PDFPageView` objects, as
returned by `PDFViewer.getPageView`, asynchronously. While the page
views are being initialized, `PDFViewer.pagesCount` returns 0 and
`PDFViewer.getPageView` will return a nullish value. When loading has
progressed further, `getPageView` returns a `PDFPageView` but with no
`pdfPage` property. Once the page views are fully ready, a bubbling
"pagesloaded" event is dispatched at the PDF viewer's container DOM
element.

The PDF anchoring code previously assumed that page views were always
immediately available and they always had a `pdfPage` property. If this
was not the case, anchoring would fail and all/many annotations would
appear as orphans in the client. See the associated issue for ways to
reproduce this locally with Via.

Resolve the issue by making the `getPageView` helper which wraps
`PDFViewer.getPageView` async, and block in that function until the page
views are fully ready.

 - Make `getPageView` helper async and block if necessary until page
   views are ready.
 - Add a test to simulate scenarios where page views are not ready yet
 - Change the `quotePositionCache` cache to store page indexes rather
   than `PDFPageView` objects. This means that less code in the file
   needs to know about `PDFPageView` objects, and deal with the fact
   that they may not be immediately available. It also means it should
   be easier to adapt the code to future PDF.js API changes.

Fixes #1330

----

**Testing**

See [this comment](https://github.com/hypothesis/client/issues/1330#issuecomment-528242209) for a way to reproduce the original issue with any PDF locally in Via (I used https://arxiv.org/pdf/1909.01862.pdf).

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -21,7 +21,7 @@ const { TextPositionAnchor, TextQuoteAnchor } = require('./types');
 let pageTextCache = {};
 
 /**
- * 2D map from `[quote][position]` to `{page, anchor}` intended to optimize
+ * 2D map from `[quote][position]` to `{pageIndex, anchor}` intended to optimize
  * re-anchoring of a pair of quote and position selectors if the position
  * selector fails to anchor on its own.
  */
@@ -41,11 +41,35 @@ function getNodeTextLayer(node) {
 /**
  * Returns the view into which a PDF page is drawn.
  *
+ * If called while the PDF document is still loading, this will delay until
+ * the document loading has progressed far enough for a `PDFPageView` and its
+ * associated `PDFPage` to be ready.
+ *
  * @param {number} pageIndex
- * @return {PDFPageView}
+ * @return {Promise<PDFPageView>}
  */
-function getPage(pageIndex) {
-  return PDFViewerApplication.pdfViewer.getPageView(pageIndex);
+async function getPageView(pageIndex) {
+  const pdfViewer = PDFViewerApplication.pdfViewer;
+  let pageView = pdfViewer.getPageView(pageIndex);
+
+  if (!pageView || !pageView.pdfPage) {
+    // If the document is still loading, wait for the `pagesloaded` event.
+    //
+    // Note that loading happens in several stages. Initially the page view
+    // objects do not exist (`pageView` will be nullish), then after the
+    // "pagesinit" event, the page view exists but it does not have a `pdfPage`
+    // property set, then finally after the "pagesloaded" event, it will have
+    // a "pdfPage" property.
+    pageView = await new Promise(resolve => {
+      const onPagesLoaded = () => {
+        document.removeEventListener('pagesloaded', onPagesLoaded);
+        resolve(pdfViewer.getPageView(pageIndex));
+      };
+      document.addEventListener('pagesloaded', onPagesLoaded);
+    });
+  }
+
+  return pageView;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/annotator/anchoring/pdf.js
```diff
@@ -72,11 +96,16 @@ function getPageTextContent(pageIndex) {
...
+  // Fetch the text content for a given page as a string.
+  const getTextContent = async pageIndex => {
+    const pageView = await getPageView(pageIndex);
+    const textContent = await pageView.pdfPage.getTextContent({
       normalizeWhitespace: true,
-    })
-    .then(joinItems);
+    });
+    return joinItems(textContent.items);
+  };
```
robertknight: This step has been rewritten as an async function for readability, and to `await` the result of `getPageView(...)`.

src/annotator/anchoring/pdf.js
```diff
@@ -41,11 +41,35 @@ function getNodeTextLayer(node) {
...
+    // objects do not exist (`pageView` will be nullish), then after the
+    // "pagesinit" event, the page view exists but it does not have a `pdfPage`
+    // property set, then finally after the "pagesloaded" event, it will have
+    // a "pdfPage" property.
+    pageView = await new Promise(resolve => {
+      const onPagesLoaded = () => {
+        document.removeEventListener('pagesloaded', onPagesLoaded);
+        resolve(pdfViewer.getPageView(pageIndex));
+      };
+      document.addEventListener('pagesloaded', onPagesLoaded);
```
LMS007: if this timeouts or breaks, should we bother with error handling or a reject case?

robertknight: If an exception occurs in this callback, the promise will reject. The document loading process could take a long time, and I'm not sure whether it depends on purely local computation or whether it can include network fetches. On this basis, any timeout would need to be generous I think.

PR #1353, opened 2019-09-06
Namespace cleanup

LMS007:
* Branched from `namespace-annotations-module`

- By in large, this is mostly a search and replace with `getRootState` ->  `getState`

- This also removes the logic for lumping the non-namespaced modules into "base" and now any module not having a namespace shows a console warning and it won't be added to the store.

- I also refactored `createReducer` because its no longer ever the case it will need an array of reducers since each duplicate reducer ID is fully namespaced. In other words we call `createReducer` multiple times now rather than taking an array.

- I remove some test logic that dealt with an array parm for createReducer 

- Added a test to specifically test multiple reducers across 2 different modules triggered by a single action.

### DIFF
--- a/src/sidebar/store/create-store.js
+++ b/src/sidebar/store/create-store.js
@@ -26,62 +26,32 @@ const { createReducer, bindSelectors } = require('./util');
  * @param [any[]] middleware - List of additional Redux middlewares to use.
  */
 function createStore(modules, initArgs = [], middleware = []) {
-  // Create the initial state and state update function. The "base"
-  // namespace is reserved for non-namespaced modules which will eventually
-  // be converted over.
+  // Create the initial state and state update function.
 
   // Namespaced objects for initial states.
-  const initialState = {
-    base: null,
-  };
+  const initialState = {};
   // Namespaced reducers from each module.
-  const allReducers = {
-    base: null,
-  };
+  const allReducers = {};
   // Namespaced selectors from each module.
-  const allSelectors = {
-    base: {
-      selectors: {},
-      // Tells the bindSelector method to use local state for these selectors
-      // rather than the top level root state.
-      useLocalState: true,
-    },
-  };
-
-  // Temporary list of non-namespaced modules used for createReducer.
-  const baseModules = [];
+  const allSelectors = {};
 
   // Iterate over each module and prep each module's:
   //    1. state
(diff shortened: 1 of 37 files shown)


src/sidebar/store/create-store.js
```diff
@@ -26,63 +26,33 @@ const { createReducer, bindSelectors } = require('./util');
...
-        // Aggregate the selectors into a single "base" map
-        ...allSelectors.base.selectors,
-        ...module.selectors,
-      };
-      initialState.base = {
-        ...initialState.base,
-        ...module.init(...initArgs),
-      };
-      baseModules.push(module);
+      console.warn('Module namespace value is missing', module);
```
robertknight: The term "module" is perhaps ambiguous if you see only the warning and are not looking at the code. "Store module does not specify a namespace" perhaps?

src/sidebar/test/fake-redux-store2.js
```diff
@@ -0,0 +1,58 @@
+// 'use strict';
```
robertknight: Was this file committed by mistake?

LMS007: yes! good catch

src/sidebar/store/test/util-test.js
```diff
@@ -27,7 +27,7 @@ const fixtures = {
       selectors: {
         countAnnotations2: function(state) {
           // useLocalState does not need namespaced path.
```
robertknight: This comment about `useLocalState` is obsolete now.

src/sidebar/store/test/util-test.js
```diff
@@ -27,7 +27,7 @@ const fixtures = {
       selectors: {
```
robertknight: The `useLocalState` flag is no longer used by the implementation, so the corresponding tests can be removed as well.

src/sidebar/util/test/state-util-test.js
```diff
@@ -18,11 +20,9 @@ describe('state-util', function() {
       return store.getState().val;
     }
 
-    it('should return promise that resolves to a non-null value', function() {
+    it(' should return promise that resolves to a non-null value', function() {
```
robertknight: An extra space was added after `it('`

PR #1367, opened 2019-09-20
Show full test output in CI

robertknight:
This may help to track down some frequent build failures seen on Jenkins
where Karma disconnects from the browser after 20 seconds by showing us
which tests, if any, have just run when the disconnect happens.

See eg. https://jenkins.hypothes.is/job/client/job/dependabot%252Fnpm_and_yarn%252Fpuppeteer-1.20.0/2/console


robertknight:
I'm going to merge this now and try rebasing the failed PRs to see if we learn anything. I tried rebuilding one this morning but it passed after a rebuild.

PR #1379, opened 2019-09-24
Enable Chrome debug logging

robertknight:
This PR [enables logging](https://www.chromium.org/for-testers/enable-logging) for Chrome in Jenkins so that it writes logs to `chrome_debug.log` in the workspace directory.

After a build completes, the files in the workspace can be browsed via the "Workspaces" link on the left hand side in Jenkins. For example https://jenkins.hypothes.is/job/client/job/chrome-debug-logging/2/ws/.

This should help us in tracking down a cause of frequent test execution failures in Jenkins where Chrome disconnects during the configuration process before it starts running any tests.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -427,14 +427,6 @@ function runKarma(baseConfig, opts, done) {
     },
   };
 
-  // Work around a bug in Karma 1.10 which causes console log messages not to
-  // be displayed when using a non-default reporter.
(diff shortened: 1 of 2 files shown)


robertknight:
The build passed and ~150KB of logs were generated at https://jenkins.hypothes.is/job/client/job/chrome-debug-logging/2/execution/node/3/ws/chrome_debug.log. I'm going to merge this and try re-doing some of the other failed builds.

PR #1394, opened 2019-10-03
Set SIDEBAR_APP_URL to a sensible default value in dev

seanh:
Set SIDEBAR_APP_URL to work with your local dev instance of h by default
in dev.

This is in line with the rest of our apps (h, lms, bouncer, via): when
run in dev they default to working with the dev instances of the others
app, not the production ones.

The idea is that, when setting up all the apps in a dev env, you just
have to install each app and run `make dev` and it will work correctly
with the other dev apps. Some apps also require a `make devdata` command
if they need sensitive DB contents and/or envvars.

Don't require developers to have to set lots and lots of envvars in each
individual project.

I also deleted some docs that instructed people to create authclients,
and set envvars, that are already automated in h.

### DIFF
--- a/docs/developers/developing.rst
+++ b/docs/developers/developing.rst
@@ -83,27 +83,12 @@ run:
 
 .. code-block:: sh
 
-   export SIDEBAR_APP_URL=http://localhost:5000/app.html
    make dev
 
-Next, you'll need to create an OAuth client which enables the Hypothesis client
-to request an access token from the service in order to make API calls.
-
-1. Go to http://localhost:5000/admin/oauthclients (you'll need to be logged
-   in to h as an admin user)
-2. Select "Register a new OAuth client"
-3. Choose a name (eg. "Client") and set the redirect URL to
-   http://localhost:5000/app.html. Leave the other settings at their default values.
-4. After creating the client, make a note of the randomly generated client ID.
-
-**In the `hypothesis/h` repository**, set the :envvar:`CLIENT_URL` and
-:envvar:`CLIENT_OAUTH_ID` env vars to tell h where to load the client from and
-what OAuth client to use, before running ``make dev``:
+**In the `hypothesis/h` repository** run:
 
 .. code-block:: sh
 
-   export CLIENT_OAUTH_ID={ OAuth client ID from step above }
-   export CLIENT_URL=http://localhost:3001/hypothesis
    make dev
 
 Once the client and h are running, you can test it out by visiting:
--- a/docs/developers/developing.rst
+++ b/docs/developers/developing.rst
@@ -75,17 +75,15 @@ This takes longer to setup than :ref:`running-from-browser-ext`.
 You should follow these steps if you want to make changes to h as well as to
 the client.
 
-First follow the
-`instructions for setting up a development install of h <http://h.readthedocs.io/en/latest/developing/>`_.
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -324,7 +324,7 @@ function generateBootScript(manifest) {
 
   const defaultSidebarAppUrl = process.env.SIDEBAR_APP_URL
     ? `${process.env.SIDEBAR_APP_URL}`
-    : 'https://hypothes.is/app.html';
+    : 'http://localhost:5000/app.html';
```
robertknight: This will break the production build of the client since it uses the default sidebar app URL. See the "Publish to QA" section of Jenkinsfile for a way to set a non-default sidebar app URL for CI builds.

docs/developers/developing.rst
```diff
@@ -83,27 +83,12 @@ run:
...
-   in to h as an admin user)
-2. Select "Register a new OAuth client"
-3. Choose a name (eg. "Client") and set the redirect URL to
-   http://localhost:5000/app.html. Leave the other settings at their default values.
-4. After creating the client, make a note of the randomly generated client ID.
-
-**In the `hypothesis/h` repository**, set the :envvar:`CLIENT_URL` and
-:envvar:`CLIENT_OAUTH_ID` env vars to tell h where to load the client from and
-what OAuth client to use, before running ``make dev``:
+**In the `hypothesis/h` repository** run:
```
robertknight: What about developers who don't have access to the devdata repo?

gulpfile.js
robertknight: It would be convenient if there was a way to enable SSL easily here. We can figure that out later though. I use SSL with locally signed certs for all of my testing as it is a more convenient way to test across a wide range of browsers and devices because configuring each device to accept a particular custom certificate is typically much easier than disabling mixed content warnings in every (browser, device).

docs/developers/developing.rst
seanh: I don't understand the question. The devdata repo isn't used in the client

gulpfile.js
seanh: In the case of the client, that would be just a case of doing a custom `export SIDEBAR_APP_URL="https://localhost:5000/app.html"` in your shell, rignt?

seanh:
Pushed a commit that I think should fix the production build

docs/developers/developing.rst
```diff
@@ -75,35 +75,18 @@ This takes longer to setup than :ref:`running-from-browser-ext`.
...
-
-**In the `hypothesis/h` repository**, set the :envvar:`CLIENT_URL` and
-:envvar:`CLIENT_OAUTH_ID` env vars to tell h where to load the client from and
-what OAuth client to use, before running ``make dev``:
+Then in a different shell, **in the h repository**, run:
 
 .. code-block:: sh
 
-   export CLIENT_OAUTH_ID={ OAuth client ID from step above }
-   export CLIENT_URL=http://localhost:3001/hypothesis
```
seanh: I moved this stuff about manually creating the client's OAuth client in h and setting these envvars in h into the h docs: https://github.com/hypothesis/h/pull/5738

gulpfile.js
robertknight: Yes, exactly.

PR #1397, opened 2019-10-07
Change SASS processing pipeline to match h, lms projects

robertknight:
Change the SASS processing pipeline to match the h and lms projects. The
purpose of this is cross-project consistency for easier maintenance,
consistent processing of SASS files from the upcoming pattern library
and to enable us to use the new SASS module system [1]

The main changes are:

 - Use Dart Sass (the canonical SASS implementation) instead of node-sass.

   This removes a native dependency which occassionally causes build
   headaches (eg. when the installed Node version changes) and will also enable
   us to make use of the new SASS module system.

 - Use a small utility module, `create-style-bundle.js`, that calls the SASS
   compiler directly instead of using gulp plugins. This simplifies
   understanding the build process by removing dependencies and is
   consistent with the h and lms projects.

   This module was copied from the hypothesis/lms repo and modified to
   add the `postcss-url` URL rewriting plugin that the previous
   gulp-sass based pipeline had.

[1] http://sass.logdown.com/posts/7858341-the-module-system-is-launched

----

For testing, I didn't look over every line of the generated bundle but rather just checked that there were no obvious visual changes in the app and made sure that icon fonts, math display etc. looked the same as before.


robertknight:
I got a verbal approval for this from @LMS007 on the dev sync this afternoon, so I'm going to merge.

PR #1398, opened 2019-10-07
Fix sidebar app URL in production

robertknight:
The `npm publish` step currently rebuilds the client. Since the `SIDEBAR_APP_URL`
env var was not set in the shell used to do this, the default value of
http://localhost:5000/app.html got used instead of
https://hypothes.is/app.html.


robertknight:
The Jenkins build passed, so I'm going to merge this as it is urgent.

PR #1434, opened 2019-10-15
Enable throttling of concurrent builds

robertknight:
Jenkins CI builds of the client project are prone to failing when many
concurrent builds occur. This often happens when many Dependabot PRs are
submitted at once during a weekly dependency update. These builds
succeed most of the time when retried.

Limiting the number of concurrent builds per executor can be done globally,
but it can also be done on a per-project basis using the
throttle-concurrents [1] plugin. Assign a throttling category to client
builds so that this latter approach can be used.

[1] https://plugins.jenkins.io/throttle-concurrents

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,9 @@
 #!groovy
 
+// Enable concurrent builds of this project to be throttled using the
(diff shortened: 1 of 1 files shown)


robertknight:
I'm going to merge this so I can test it out with the batch of open Dependabot PRs that recently failed. If it doesn't work out, this change is easy enough to revert.

PR #1435, opened 2019-10-15
Revert "Improve accessability for toggle buttons"

robertknight:
This reverts commit ab3d902ff465063e70431e6744b1de5a6b0a01f0.

This change caused the client to fail to load on pages where
`window.hypothesisConfig` is not defined, which means all pages that do not embed Hypothesis directly and some that do, if use JSON configuration rather than defining `window.hypothesisConfig`.

Original PR: https://github.com/hypothesis/client/pull/1382


robertknight:
CC @LMS007 and @lyzadanger

robertknight:
I'm merging this now to un-break the `master` branch so we can continue with new deployments.

PR #1436, opened 2019-10-15
Improve accessability for toggle buttons

LMS007:
- Add aria attribute for show / hide side bar
- Add aria attribute for show / hide highlights
- Fix a bug that prevents the correct class from showing on the highlight icon when highlights are initially disabled

### DIFF
--- a/src/annotator/plugin/toolbar.coffee
+++ b/src/annotator/plugin/toolbar.coffee
@@ -1,11 +1,14 @@
 Plugin = require('../plugin')
 $ = require('jquery')
+configFrom = require('../config/index')
+
 
 makeButton = (item) ->
   anchor = $('<button></button>')
   .attr('href', '')
   .attr('title', item.title)
   .attr('name', item.name)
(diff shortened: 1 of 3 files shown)


src/annotator/plugin/toolbar.coffee
```diff
@@ -27,6 +30,10 @@ module.exports = class Toolbar extends Plugin
     else
       $(@element).append @toolbar
 
+    config = configFrom(window);
```
LMS007: old PR for reference

https://github.com/hypothesis/client/commit/d7acf2f3e0bfe5d5f38d1c5491c2ea260b903fbb

`This is generally the new part of this diff. Instead of looking into `window.hypothesisConfig` which in most cases is not even there, it now calls `configFrom`. This does mean that configFrom is called twice now, once from `annotator/index.js` and then again here, but the hope is that this is a better source of truth, but I'm still not sure this is correct.

robertknight:
Rebased on master.

PR #1440, opened 2019-10-18
Improve mobile UX of markdown editor and search input

robertknight:
- Increase the font size of the markdown editor and search input fields when using touch input. This makes them easier to read on mobile and is especially important on iOS as it prevents Safari from zooming in the input field, which is really annoying when the parent website is mobile optimized. It also sometimes causes the sidebar iframe to move in unexpected ways (see screenshot below).
- Make the toolbar buttons larger when using touch input and add space at the bottom of the toolbar to make it easier for the browser to differentiate taps on the buttons and taps on selection handles in the input field
- Fix an issue where the toolbar overflowed on most phone-sized screens. Depending on the input mode this is handled by making the toolbar scrollable (touch input, easy to scroll by swiping) or by wrapping the icons (non-touch input, possibly harder to scroll).

**Before**

Note toolbar is overflowing:

Borked sidebar iframe sometimes seen after tapping input field:

**After**

Note toolbar does not overflow but is scrollable instead:

Search input field showing larger input size without zooming and blue action button in the virtual keyboard:

### DIFF
--- a/src/sidebar/components/markdown-editor.js
+++ b/src/sidebar/components/markdown-editor.js
@@ -263,7 +263,7 @@ function MarkdownEditor({ onEditText = () => {}, text = '' }) {
         />
       ) : (
         <textarea
-          className="form-input form-textarea"
+          className="markdown-editor__input form-input form-textarea"
           ref={input}
           onClick={e => e.stopPropagation()}
           onKeydown={handleKeyDown}
--- a/src/sidebar/components/search-input.js
+++ b/src/sidebar/components/search-input.js
@@ -41,7 +41,12 @@ function SearchInput({ alwaysExpanded, query, onSearch }) {
   }
 
   return (
-    <form className="search-input__form" name="searchForm" onSubmit={onSubmit}>
+    <form
+      action="#"
+      className="search-input__form"
+      name="searchForm"
+      onSubmit={onSubmit}
+    >
       <input
         className={classnames('search-input__input', {
           'is-expanded': alwaysExpanded || query,
--- a/src/styles/sidebar/components/markdown-editor.scss
+++ b/src/styles/sidebar/components/markdown-editor.scss
@@ -4,6 +4,10 @@ $toolbar-border: 0.1em solid $grey-3;
   display: flex;
   flex-direction: row;
 
+  // Toolbar buttons wrap on non-touch devices if they don't fit. We don't use
+  // scrolling because that's less convenient to use with a mouse/touchpad.
+  flex-wrap: wrap;
+
   background-color: white;
   border: $toolbar-border;
   border-bottom: none;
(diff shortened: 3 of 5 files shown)


lyzadanger:
Hi! I was ultimately able to get things running locally, but am running into some usability difficulties.

1. (This is also present on master): When I click on the "edit" icon for an annotation in the sidebar, I'm scrolled/taken to the highlighted annotation content on the host page (not into the editor). I'm not sure what's going on here.

2. If I'm able to navigate back to the sidebar (not always possible as the zoom goes kind of crazy), the "Preview" button is half visible, but cut off:

image

Some of this craziness might be from the viewport zoom of the underlying host page, but I'm tearing my hair out just a bit here!

robertknight:
> If I'm able to navigate back to the sidebar (not always possible as the zoom goes kind of crazy), the "Preview" button is half visible, but cut off:

That's expected. If the toolbar is too narrow to fit all the icons, it becomes scrollable. Try swiping it. We could switch the preview button to an icon to save more space, but I left that out of scope for this first round of improvements.

> I'm scrolled/taken to the highlighted annotation content on the host page (not into the editor). 

That might be an issue with the "Edit" icon being too small and so the browser behaving as if you just clicked on the annotation card?

>  I just noticed that the content of the annotation isn't pre-filled in the editor, either

Do you mean you are editing an annotation that has a body, and the body doesn't appear in the `<textarea>`? That's unexpected. Can you reproduce on your laptop?

robertknight:
When doing the next round of testing of this, it may help to check out https://github.com/hypothesis/h/pull/5758 in h, and set the redirect URL for your OAuth client to `{current_scheme}://{current_host}/app.html`.

lyzadanger:
I have given this another whack. There may be something specific to my setup, but using the client at all on my local with a phone is...I'm going to be honest...basically unusable. I'll try to walk through what I'm seeing. It's hard for me to separate the overall experience from the markdown editor specifically. If I had to summarize the overall grief, I'd say "viewport and zoom".

When I first load local client, I see this:

image

The natural inclination, thus, is to pinch to zoom in closer because it's unreadable. Any pinching and zooming closes the sidebar, leaving me in a confusing situation. If I really futz around, I can zoom in and scroll up to the top right corner and perhaps find the handle to re-open the sidebar:

image

Depending on how zoomed in I am at that point, the sidebar, once opened, will take up a varying amount of the available viewport:

image

OK, I've made it this far...now here's the biggest problem, I think. I tap on the edit icon, and the viewport zooms way out and then zooms in again, to the highlighted text:

image

Now I'm totally lost. Zooming out again closes the sidebar, so I have to go find it again using the onerous zoom-scroll-and-search method above. When I'm finally able to re-open the sidebar, I get a partially cut off interface depending on how zoomed I am:

image

The thought that comes into my head is "OK, maybe I'm doing this wrong," but then again, that doesn't make sense. I'm no dafter than anyone, and I would presume this workflow would make a user throw their phone across the room, potentially (I know I almost did).

Anyway, these surrounding conditions make it hard for me to evaluate the PR at hand here.

lyzadanger:
Ah, that's a good tip overall (for mobile testing). However, I'm still seeing the icons in the editor cut off:

image

robertknight:
> Ah, that's a good tip overall (for mobile testing). However, I'm still seeing the icons in the editor cut off:

Yes, that's expected. There isn't enough space to make the icons large enough to tap easily and fit within the width of a small phone screen. You should find that the toolbar scrolls if you swipe it though. I was hoping it was obvious enough that this was possible. Perhaps some affordance could be added to make this clearer.

robertknight:
Thanks Lyza. I'll merge this as-is though It definitely sounds like it wasn't obvious enough that the toolbar is supposed to be scrollable. I might ping folks on Slack on case they have additional suggestions.

PR #1448, opened 2019-10-21
Remove unroll and convert its use cases to use forEach instead

LMS007:
The diff is long, but straight forward, everything just got a forEach more less exactly how the test case was already written.

Only major change was to moderation-banner-test.js where I added the test description text into the test name as it felt appropriate to do so. Previously it was only in a comment.


src/sidebar/components/test/annotation-test.js
```diff
@@ -917,9 +915,39 @@ describe('annotation', function() {
...
+          expected: true,
+        },
+        {
+          case_: 'the draft is shared and the group is restricted',
+          draft: draftFixtures.shared,
+          group: groupFixtures.restricted,
+          expected: true,
+        },
+      ].forEach(testCase => {
+        it(`returns #expected if ${testCase.case_}`, () => {
```
robertknight: `#expected` here needs to be replaced with `${expected}`

src/sidebar/test/markdown-commands-test.js
```diff
@@ -105,55 +104,47 @@ describe('markdown commands', function() {
         input: '<sel></sel>',
         output: '> <sel></sel>',
       },
-    ];
-
-    unroll(
-      '#tag',
-      function(fixture) {
+    ].forEach(fixture => {
+      it(`${fixture.tag}`, () => {
```
robertknight: ``it(`${fixture.tag}`,`` can be simplified to just `it(fixture.tag,`

src/sidebar/components/test/annotation-test.js
LMS007: good catch!

LMS007:
> Looks like @robertknight gave a solid once-over on this, so it LGTM after his comments are addressed.
> 
> One note I'd make is that it might be nice to have a more fully-worded commit for this work. The commit summary certain explains the change clearly, but the motivation for the change might be nice as a body of the commit. This is a nice-to-have, though, not critical.
> 
> Thanks for taking on this housekeeping task. Always nice to keep our tech debt slim and things moving in the right direction!

noted, will do!

src/sidebar/test/markdown-commands-test.js
LMS007: lol @ myself for this one :)

PR #1449, opened 2019-10-22
Fix PDF anchoring when annotation refers to last text on a page

### LINKED ISSUE #1329: Annotations immediately orphan when selecting text in a PDF that includes last character on page
This issue was created as part of investigating https://github.com/hypothesis/lms/issues/875.

**Steps to reproduce**

1. Go to https://via.hypothes.is/https://arxiv.org/pdf/1908.04683v3.pdf, scroll down to page 5 and select the text beginning at the bottom of the page from "For our experiments" up to and _including_ the page number "5" at the bottom of the page.
2. Click "Annotate"

**Expected result:** The selected text is highlighted and a new annotation appears in the "Annotations" tab in the client
**Actual result:** The selected text is not highlighted and the annotation appears in the "Orphans" tab in the client

After reloading the page, the annotation continues to be an orphan instead of anchoring.

**Notes**

The issue occurs whenever a selection to annotate matches these criteria:

1. The text is unique enough within the document that it does not fuzzily/roughly match any other text
2. The selected text includes the last character in the page's text. More specifically, it needs to extend up to the end of the invisible `div.textLayer` element that PDF.js creates for each page to enable text selection.

robertknight:
This PR adds tests for and fixes an issue where annotations that spanned the last text on a PDF page would fail to anchor.

The problem is due to an issue in a function in the dom-anchor-text-position library (see https://github.com/tilgovi/dom-anchor-text-position/pull/7) which converts a pair of `(start, end)` offsets referring to the [textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) of an element into a DOM `Range`. Specifically it failed when the `end` offset was equal to `node.textContent.length`.

This PR addresses the issue by:
 - Adding a test case at the level of PDF anchoring
 - Adding a new implementation of `(start, end)` => `Range` conversion in `src/annotator/anchoring/text-position` and an associated set of tests

We may want to revert back to using the dom-anchor-text-position implementation once the above PR (or some functional equivalent) is merged, although I suggest we should keep the tests in that case. I have to say though, the code is not particularly complex so there is value in having fewer dependencies involved in PDF annotation.

HTML anchoring still uses the previous implementation of text position => Range conversion. HTML anchoring is in practice not affected by the issue because it is impossible to select the last text in a document and annotate it. If you select the last _visible_ text in a page, there is always other content after that in `document.body.textContent`.

Fixes #1329

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -9,9 +9,9 @@ const seek = require('dom-seek');
 const createNodeIterator = require('dom-node-iterator/polyfill')();
 
 const xpathRange = require('./range');
-const html = require('./html');
 const RenderingStates = require('../pdfjs-rendering-states');
 const { TextPositionAnchor, TextQuoteAnchor } = require('./types');
+const { toRange: textPositionToRange } = require('./text-position');
 
 // Caches for performance.
 
@@ -210,10 +210,9 @@ function findPage(offset) {
  *
  * @param {number} pageIndex - The PDF page index
  * @param {TextPositionAnchor} anchor - Anchor to locate in page
- * @param {Object} options - Options for `anchor.toSelector`
  * @return {Promise<Range>}
  */
-async function anchorByPosition(pageIndex, anchor, options) {
+async function anchorByPosition(pageIndex, anchor) {
   const page = await getPageView(pageIndex);
 
   let renderingDone = false;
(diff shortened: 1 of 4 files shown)


LMS007:
I was not able to test this successfully so I suspect I may have done something wrong.

I just tried to do the happy case here http://www.orimi.com/pdf-test.pdf with a doing a select all + highlight. e.g. `cmd+a`

'range.coffee' is giving me some trouble, but it does appear that I'm executing the correct branch here. I spot checked some changes you did and saw them in my console. Its failing inside of `getLastTextNodeUpTo` which is called from line 197 in `range.coffee`

`r.end = Util.getLastTextNodeUpTo node` where node is `null`

`anchorByPosition` is not called at all.

robertknight:
@LMS007 - I was able to reproduce the issue you described. It is a valid bug but subtly different than the one that I fixed in this PR and it occurs in a different part of the code. See the original issue for steps to reproduce on a specific PDF (https://github.com/hypothesis/client/issues/1329#issue-485766645).

The difference is that in the original issue, the selection ends at the end of the text layer in the page. In the issue you describe, pressing Cmd+a creates a selection which ends outside of the text layer in some other element.

What I would suggest to do here is 1) keep the scope of this PR to issue #1329, 2) open a separate issue to track the problem that happens when pressing Cmd+a in http://www.orimi.com/pdf-test.pdf.

PR #1470, opened 2019-11-04
Change approach to mocking components in unit tests

robertknight:
Shallow rendering with Enzyme's `shallow` function conflates two changes in
rendering behavior:

 1. Only rendering "one level deep" so that a test for a component is
    isolated from the details of any child components

 2. In React (but not Preact) rendering to a data structure instead of
    real DOM nodes. Even though Preact still renders real DOM nodes,
    Enzyme's shallow rendering API removes many of the features of
    `mount` rendering

In practice, we only used shallow rendering for (1) and we have found
that there are use cases where we want (1) but still want to be able to
interact with the real DOM nodes (ie. we don't want (2)). The need to
use different mocking approaches and understand the conflated behaviors
mentioned above has been a source of confusion for developers.

This commit changes the mocking approach to always use a pattern that we
have until now only used selectively. Enzyme's `mount` rendering mode is
always used, but a utility function, `mockImportedComponents` (see `src/sidebar/components/test/mock-imported-components.js`), is used
with our existing import mocking (via babel-plugin-mockable-imports) to
semi-automatically mock all imported components .  This means that we can use
the same approach for all tests, regardless of whether they need to interact with
DOM nodes or not.

The first commit adds the helper function, the second commit changes the existing component tests as follows:

 1. Replace `shallow` with `mount`

 2. Call `ComponentUnderTest.$imports.$mock(mockImportedComponents())`
    in the test setup to mock all imported components, but not helper
    components defined in the same file.

    ```js
    beforeEach(() => {
      ThingBeingTested.$imports.$mock(mockImportedComponents());
      ThingBeingTested.$imports.$mock({
        // Mock other imported functions
      });
    }):
    ```

 3. Call `ComponentUnderTest.$imports.$restore()` in test cleanup, if
    the test was not already doing that:
   
    ```js
    afterEach(() => {
     ThingBeingTested.$imports.$restore();
    });
    ```

 4. Replace uses of (`wrapper.find(SomeChildComponent)`) with
    (`wrapper.find('SomeChildComponent')`. This is necessary because the
    mocked component no longer has the same identity as the original, so
    we look it up by name instead.

### DIFF
--- a/src/sidebar/components/test/selection-tabs-test.js
+++ b/src/sidebar/components/test/selection-tabs-test.js
@@ -1,11 +1,11 @@
 'use strict';
 
-const { shallow, mount } = require('enzyme');
+const { mount } = require('enzyme');
 const { createElement } = require('preact');
 
-const NewNoteBtn = require('../new-note-btn');
 const uiConstants = require('../../ui-constants');
 const SelectionTabs = require('../selection-tabs');
+const mockImportedComponents = require('./mock-imported-components');
 
 describe('SelectionTabs', function() {
   // mock services
@@ -18,23 +18,7 @@ describe('SelectionTabs', function() {
     isLoading: false,
   };
 
-  SelectionTabs.$imports.$mock({
-    '../store/use-store': callback => callback(fakeStore),
-  });
-
   function createComponent(props) {
-    return shallow(
-      <SelectionTabs
-        session={fakeSession}
-        settings={fakeSettings}
-        {...defaultProps}
-        {...props}
-      />
-    ).dive();
-  }
-
-  // required for <Tab> rendering
-  function createDeepComponent(props) {
     return mount(
       <SelectionTabs
         session={fakeSession}
@@ -69,6 +53,15 @@ describe('SelectionTabs', function() {
         },
       }),
     };
+
+    SelectionTabs.$imports.$mock(mockImportedComponents());
+    SelectionTabs.$imports.$mock({
+      '../store/use-store': callback => callback(fakeStore),
+    });
+  });
+
+  afterEach(() => {
+    SelectionTabs.$imports.$restore();
   });
 
   const unavailableMessage = wrapper =>
@@ -76,7 +69,7 @@ describe('SelectionTabs', function() {
 
   context('displays selection tabs and counts', function() {
     it('should display the tabs and counts of annotations and notes', function() {
-      const wrapper = createDeepComponent();
+      const wrapper = createComponent();
       const tabs = wrapper.find('a');
       assert.isTrue(tabs.at(0).contains('Annotations'));
       assert.equal(
@@ -97,7 +90,7 @@ describe('SelectionTabs', function() {
     });
 
     it('should display annotations tab as selected', function() {
-      const wrapper = createDeepComponent();
+      const wrapper = createComponent();
       const aTags = wrapper.find('a');
       assert.isTrue(aTags.at(0).hasClass('is-selected'));
     });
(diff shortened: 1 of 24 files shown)


robertknight:
>  I know you’ve explained things related to this several times before, but could you remind me why this is no longer necessary (p.s. very happy to see it go!)

Previously, when you did `const SomeWidget = require('../some-widget')` in a test and `SomeWidget` used the `withServices` function, then the output of rendering `SomeWidget` would be:

```
<withServices(SomeWidget)>
  <SomeWidget ...>
    ...
  </SomeWidget>
</withServices(SomeWidget)>
```

In other words, `withServices` creates a component that renders its input as its only child, and passes it the props that were passed to the `withServices(...)` wrapper, plus the services.

Since shallow rendering only renders "one level deep", it didn't actually render the _contents_ of `SomeWidget`. Therefore you had to go one level deeper using `dive` to actually render the contents of `SomeWidget`.

With `mount` rendering, the structure of the output is the same, but every component in the file `some-widget.js` is fully rendered, including `SomeWidget`. Only components that are imported are mocked.

> Presumably, if we didn’t want to mock all of the imported components in one swipe, we could manually mock individual components as we did before when we used mount

Yes. You can still use that approach if it makes sense.

> Could the mock-imported-components module live somewhere else?

Currently the convention is that utilities for tests live alongside their tests but don't have a `-test` suffix. For utilities used only by tests in one directory, perhaps they could go in a `util` dir: `src/sidebar/components/test/util/mock-imported-components.js` instead.

I was thinking in future we would probably want to use Preact for UI in `annotator/`, in which case we would want to move this to a more general-purpose location. Perhaps `src/test-utils` or something.

src/sidebar/components/test/top-bar-test.js
```diff
@@ -1,16 +1,12 @@
 'use strict';
 
 const { createElement } = require('preact');
-const { shallow } = require('enzyme');
+const { mount } = require('enzyme');
 
 const uiConstants = require('../../ui-constants');
 
-const GroupList = require('../group-list');
```
LMS007: I see how this really makes thing a lot easier

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -18,23 +18,7 @@ describe('SelectionTabs', function() {
     isLoading: false,
   };
 
-  SelectionTabs.$imports.$mock({
```
LMS007: much nicer!

src/sidebar/components/test/selection-tabs-test.js
```diff
@@ -69,14 +53,23 @@ describe('SelectionTabs', function() {
         },
       }),
     };
+
+    SelectionTabs.$imports.$mock(mockImportedComponents());
+    SelectionTabs.$imports.$mock({
```
LMS007: could mockImportedComponents be modified to take on overrides so we don't have to make multiple calls using 
`SomeComponent.$imports.$mock(...)`

robertknight: That would certainly be possible, although it may end up re-implementing some of `$mock`'s logic. Alternatively `$mock` could be changed to accept multiple arguments or an array so you could use it something like this:

```js
SelectionTabs.$imports.$mock(
  mockImportedComponents(),
  {
    // Other imports here.
  }
)
```

robertknight:
Thanks for the reviews. I'm going to merge this as-is to unblock Lyza. There are two outstanding suggestions I'll note:

- Moving `mock-imported-components.js` to a different location to better separate it from tests
- Providing a more succinct way to apply both automatic mocking plus a set of manual mocks

PR #1479, opened 2019-11-08
Show focus ring on annotator toolbar buttons when using keyboard focus

robertknight:
Apply the pattern used by the sidebar's horizontal toolbar and menus of
showing a focus ring around tab-focused items, but only when they
have been given focus via the keyboard.

**Testing:**

Go to http://localhost:3000 and click on the buttons in the vertical toolbar. They should work the same as before. Now use the keyboard to focus them, you should see a focus ring whereas on master you won't.

Screenshot 2019-11-08 11 23 51

### DIFF
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -1,10 +1,11 @@
 'use strict';
 
-const configFrom = require('./config/index');
-
 const $ = require('jquery');
 
-// Applications
+// Load polyfill for :focus-visible pseudo-class.
+require('focus-visible');
+
+const configFrom = require('./config/index');
 const Guest = require('./guest');
 const Sidebar = require('./sidebar');
 const PdfSidebar = require('./pdf-sidebar');
--- a/src/styles/annotator/annotator.scss
+++ b/src/styles/annotator/annotator.scss
@@ -1,5 +1,6 @@
 @import '../base';
 @import '../mixins/icons';
+@import '../mixins/focus';
 
 @import './adder';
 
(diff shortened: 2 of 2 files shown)


src/annotator/index.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 
-const configFrom = require('./config/index');
-
 const $ = require('jquery');
 
-// Applications
+// Load polyfill for :focus-visible pseudo-class.
+require('focus-visible');
```
lyzadanger: Can you help me understand why this polyfill becomes necessary because of these changes (i.e. why it wasn't necessary before for the `outline-on-keyboard-focus` mixin)?

LMS007:
I second the poly fill question. My only nit here, and its not related to this diff is that the close button is part of the top bar's DOM branch but rendered to look like the side bar so we get this weird hidden effect. Probably due for a HTML refactor at some point.

I think getting focus state and css correct across the whole app is actually pretty hard :\

Screen Shot 2019-11-14 at 3 37 56 PM

src/annotator/index.js
robertknight: The `outline-on-keyboard-focus` mixin relies on the `:focus-visible` selector being natively supported OR [the official polyfill](https://github.com/WICG/focus-visible) being loaded. In practice all browsers use the polyfill currently. That is expected to change in Chrome [soonish](https://chromestatus.com/features/5823526732824576).

lyzadanger:
👍 from me at this point.

robertknight:
Great, thanks Lyza. I'll merge this then.

PR #1496, opened 2019-11-15
Rename the username to userid in selection module

LMS007:
The username was really the userid and was named incorrectly.

e.g. From this
```json
"focus": {
  "user":{
    "username":"acct:45a3f1713ca9df0e52d0c9651f115c@lms.hypothes.is",
    "displayName":"kyle_student3 K3 Student"
    }
  }
}
```
to this

```json
"focus": {
  "user":{
    "userid":"acct:45a3f1713ca9df0e52d0c9651f115c@lms.hypothes.is",  
    "displayName":"kyle_student3 K3 Student"
    }
  }
}
```

Update: This remains backwards compatible with LMS when only the `username` used in the config or an RPC request.

Created a TODO: ticket to rip those parts out once LMS is fixed up
https://github.com/hypothesis/client/issues/1516

### DIFF
--- a/docs/publishers/config.rst
+++ b/docs/publishers/config.rst
@@ -408,10 +408,8 @@ loads.
           user: {
             // required
             username: "foobar1234",
-
             // optional
             displayName: "Foo Bar",
-            authority: "example.com",
           }
         }
       };
--- a/src/sidebar/store/modules/selection.js
+++ b/src/sidebar/store/modules/selection.js
@@ -10,7 +10,7 @@
 
 /**
  * @typedef User
- * @property {string} username - Unique user's username
+ * @property {string} userid - Unique user's id
  * @property {string} displayName - User's display name
  */
 
@@ -441,16 +441,20 @@ function focusModeFocused(state) {
 }
 
 /**
- * Returns the username for a focused user or `null` if no focused user.
+ * Returns the `userid` for a focused user or `null` if no focused user.
  *
- * @return {object|null}
+ * @return {string|null}
  */
-function focusModeUsername(state) {
-  if (
-    state.selection.focusMode.config.user &&
-    state.selection.focusMode.config.user.username
-  ) {
-    return state.selection.focusMode.config.user.username;
+function focusModeUserId(state) {
+  if (state.selection.focusMode.config.user) {
+    if (state.selection.focusMode.config.user.userid) {
+      return state.selection.focusMode.config.user.userid;
+    }
+    if (state.selection.focusMode.config.user.username) {
+      // remove once LMS no longer sends username in RPC or config
+      // https://github.com/hypothesis/client/issues/1516
+      return state.selection.focusMode.config.user.username;
+    }
   }
   return null;
 }
@@ -462,11 +466,11 @@ function focusModeUsername(state) {
  * @return {boolean}
  */
 function focusModeHasUser(state) {
-  return focusModeEnabled(state) && !!focusModeUsername(state);
+  return focusModeEnabled(state) && !!focusModeUserId(state);
 }
 
 /**
- * Returns the display name for a user or the username
+ * Returns the display name for a user or the userid
  * if display name is not present. If both are missing
  * then this returns an empty string.
  *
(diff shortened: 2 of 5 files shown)


src/sidebar/store/modules/selection.js
```diff
@@ -441,12 +441,18 @@ function focusModeFocused(state) {
 }
 
 /**
- * Returns the username for a focused user or `null` if no focused user.
+ * Returns the `userid` for a focused user or `null` if no focused user.
  *
  * @return {object|null}
```
robertknight: The docstring here is out of date as this returns `string|null` rather than `object|null`.

src/sidebar/store/modules/selection.js
```diff
@@ -441,12 +441,18 @@ function focusModeFocused(state) {
...
  */
-function focusModeUsername(state) {
+function focusModeUserId(state) {
   if (
```
robertknight: I suggest this function would be easier to read if `state.selection.focusMode.config.user` were extracted and tested once at the start.

src/sidebar/store/modules/test/selection-test.js
```diff
@@ -256,14 +256,14 @@ describe('sidebar/store/modules/selection', () => {
...
-        [{ focus: { user: { username: 'whatever' } } }]
+        [{ focus: { user: { userid: 'acct:userid@authority' } } }]
       );
       assert.isTrue(store.focusModeHasUser());
     });
     it('should return `false` if focus enabled but `user` object invalid', () => {
       store = createStore(
         [selection],
-        [{ focus: { user: { displayName: 'whatever' } } }] // `username` is required
+        [{ focus: { user: { displayName: 'acct:userid@authority' } } }] // `userid` is required
```
robertknight: The previous value of `displayName` here was meant to be an example of a display name not a username or userid. Something like "Jim Smith" for example.

I try to use "realistic looking" examples for string values in tests as it provides better documentation about how the code is used in the real app.

src/sidebar/store/modules/test/selection-test.js
```diff
@@ -281,12 +281,12 @@ describe('sidebar/store/modules/selection', () => {
       );
       assert.equal(store.focusModeUserPrettyName(), 'FakeDisplayName');
     });
-    it('should the username when displayName is missing', function() {
+    it('should the userid when displayName is missing', function() {
```
robertknight:
```suggestion
    it('returns the userid when displayName is missing', function() {
```

This was an existing typo

src/sidebar/store/modules/test/selection-test.js
```diff
@@ -295,22 +295,38 @@ describe('sidebar/store/modules/selection', () => {
     it('should an return empty string when there is no focus object', function() {
       assert.equal(store.focusModeUserPrettyName(), '');
     });
+    it('should the username when displayName and userid is missing', function() {
```
robertknight:
```suggestion
    it('returns the username when displayName and userid is missing', function() {
```

src/sidebar/store/modules/test/selection-test.js
```diff
@@ -295,22 +295,38 @@ describe('sidebar/store/modules/selection', () => {
...
+      // remove once LMS no longer sends username in RPC or config
```
robertknight: If we have a temporary change which should be removed when a specific PR or GitHub issue is resolved, whether that is in our own code or some third party library, it is a useful practice to link to that issue.

docs/publishers/config.rst
```diff
@@ -407,11 +407,9 @@ loads.
         focus: {
           user: {
             // required
-            username: "foobar1234",
-
+            userid: "acct:foobar1234@authority",
```
robertknight: I note this is a breaking change from the previous API. The "focus" mode is marked as beta so that's OK, but it is possible that some third-party might run into this.

LMS007: Okay, I won't make this change this round, make a separate ticket to clean this up later.

src/sidebar/store/modules/test/selection-test.js
LMS007: oops!

PR #1504, opened 2019-11-18
Update core-js to current version

robertknight:
This package provides polyfills for new ES/DOM APIs for use in IE 11.
This was held back to a v2.x version due to a conflict with fetch-mock
that is now resolved.

- [x] Test this locally with IE 11 (eg. using Sauce Labs) and verify everything still works


robertknight:
Ah, great. Thanks for the IE testing Kyle :)

PR #1509, opened 2019-11-19
Warn in JSON-RPC server if message is dropped by origin filter

robertknight:
Address some confusion I encountered while testing
https://github.com/hypothesis/lms/pull/1157/ where JSON-RPC messages
from the lms app to the client would be silently dropped if the
`rpcAllowedOrigins` config was not set correctly.

Log a console warning when this happens to make the problem more
obvious.  Before warning though, filter out non JSON-RPC messages
because there are various other sources of "message" events which the
client is likely to receive and which we don't want to result in warning
spam.

### DIFF
--- a/src/sidebar/test/cross-origin-rpc-test.js
+++ b/src/sidebar/test/cross-origin-rpc-test.js
@@ -6,6 +6,7 @@ describe('crossOriginRPC', function() {
   describe('server', function() {
     let addedListener; // The postMessage() listener that the server adds.
     let fakeStore;
+    let fakeWarnOnce;
     let fakeWindow;
     let settings;
     let source;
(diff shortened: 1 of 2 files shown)


src/sidebar/test/cross-origin-rpc-test.js
```diff
@@ -114,6 +123,18 @@ describe('crossOriginRPC', function() {
...
 
+    [{}, null, { jsonrpc: '1.0' }].forEach(invalidMessage => {
+      it('ignores non JSON-RPC messages', () => {
+        const settings = { rpcAllowedOrigins: [] };
+        crossOriginRPC.server.start(fakeStore, settings, fakeWindow);
+
+        postMessage({ data: invalidMessage, origin: 'https://foo.com' });
+
+        assert.notCalled(fakeWarnOnce);
+        assert.isTrue(source.postMessage.notCalled);
```
LMS007: `assert.notCalled(source.postMessage);`
to match prev line

PR #1521, opened 2019-11-21
Convert SASS files to SASS modules

robertknight:
**Depends on https://github.com/hypothesis/client/pull/1520**

Migrate .scss files to SASS modules using [sass-migrator](https://github.com/sass/migrator):

```
sass-migrator module --load-path=node_modules/ --migrate-deps src/styles/annotator/annotator.scss src/styles/sidebar/sidebar.scss
```

This was previously done for the lms app in https://github.com/hypothesis/lms/pull/1057.

For future reference the module system is described in detail on [the SASS blog](http://sass.logdown.com/posts/7858341-the-module-system-is-launched). The primary motivation is to make navigation / refactoring / general maintenance easier, and it will also avoid the need for us to prefix the mixins/variables that the upcoming pattern library defines.


robertknight:
Thanks for the quick review. I've fixed those conflicts. I'll get this merged now to avoid merge conflicts with other work.

PR #1541, opened 2019-11-26
Revert "Improve compression of JavaScript bundles"

robertknight:
Revert hypothesis/client#1538. On QA I saw an error when logging out:

logout-error

Additionally the source maps are broken. Clicking on the source links next to stack trace frames does not go to the right location.


robertknight:
Update: It turns out that the error was *not* caused by the original PR (see https://github.com/hypothesis/client/issues/1542) but the sourcemap breakage is a real issue.

PR #1545, opened 2019-11-28
Improve minification of JS bundles

robertknight:
This is a second attempt at https://github.com/hypothesis/client/pull/1538 which fixes the issues from the original PR where sourcemaps were broken.

There were two reasons for this:
 - `scripts/gulp/minify-stream.js` configured terser to _generate_ source maps, but not to _consume_ source maps from the input code. This was fixed by adding `content: 'inline'` to read inline source maps from the input.
 - The existing code to change the name of the global function that bundles use to import modules from earlier bundles did not update sourcemaps. This was fixed by [changing the approach](https://github.com/hypothesis/client/commit/d90346c9b7636bcea709735cb0c2f00c2600ad26) to a different method that won't break sourcemaps.

The first commit in this branch is the original PR. The next three commits fix the above issues. The final commit makes it easier to test this change locally by enabling testing of a production build with the local dev server as follows:

1. Checkout this branch
2. Run `yarn install`
3. Run `env NODE_ENV=production gulp watch`
4. Once the assets are built, check that the app works. If you introduce a bug in the code which causes it to throw an error, the stack traces in the browser should point to the correct locations. Note that function/class names are expected to be lost in production builds


robertknight:
Thanks!

PR #1558, opened 2019-12-04
Create TagEditor component

LMS007:
Reworked this PR https://github.com/hypothesis/client/pull/1508. This now only covers the editor and the read-only portion has been pieced out into TagList here https://github.com/hypothesis/client/pull/1555

tags

**Safari**
I was not able to land reasonably behavior in Safari so I created a separate ticket to work on that
https://github.com/hypothesis/client/issues/1564

I suspect it may come down to moving away form datalist and building our own component. Safari really packs in some terrible behavior here that is not at all related to the DOM but feels more like a re-branded built in autocomplete from yesteryear.


src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,202 @@
...
+   */
+  const removeTag = tag => {
+    const newTagList = [...tagList]; // make a copy
+    const index = newTagList.indexOf(tag);
+    newTagList.splice(index, 1);
+    saveTags(newTagList);
+  };
+
+  /**
+   * Adds a new local tag equal to the value of the input field only
```
lyzadanger: Can you make this comment more clear? Also I think you can omit the bit about duplicate tags because you have a comment for it within the function.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,202 @@
...
+      setSuggestions(suggestions);
+    } else if (e.type === 'keyup') {
+      // Assume the user selected an option from <datalist>
+      // either by clicking on it with the mouse or pressing
+      // the 'Enter' key.
+      addTag();
     }
   };
+
+  const buildSuggestionsList = () => {
```
lyzadanger: Maybe turn this into an IIFE and assign to a `const` e.g. `suggestionList`?

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,202 @@
...
+    // Only return a datalist if there is something to render in the list. This
+    // avoid a bug in Firefox where the datalist rendered does not correspond to the
+    // actual list of <option>s in the dom. Wiping out the entire <datalist> tag seems
+    // to force Firefox's hand.
+    if (suggestionElements.length > 0) {
+      return (
+        <datalist
+          id={`tag-editor-datalist-${annotation.id}`}
+          className="tag-editor__suggestions"
+          aria-label="Annotation suggestions"
```
lyzadanger: Perhaps "Suggested tags for annotation" (in any case, the word "tags" should be in here)?

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,202 @@
...
-  },
-  template: require('../templates/tag-editor.html'),
+/**
+ * @typedef Tag
+ * @param tag {string} - The tag text
+ */
+
+TagEditor.propTypes = {
+  /* Annotation that owns the tags. */
+  annotation: propTypes.object,
```
lyzadanger: None of these are marked as `isRequired`. How does the component behave if any are missing? Seems like they might all need to be required?

src/styles/sidebar/components/tag-editor.scss
```diff
@@ -0,0 +1,45 @@
...
+    padding: 0 0.5em;
+  }
+
+  &__delete {
+    @include buttons.action-button;
+    appearance: none;
+    border: 1px solid var.$grey-3;
+    border-radius: 0 2px 2px 0;
+    padding: 0 0.5em;
+    &:hover,
```
lyzadanger: This is the only button in the app with a border on hover—perhaps just background color/icon darker on hover?

src/styles/mixins/selectors.scss
```diff
@@ -0,0 +1,6 @@
+@mixin hover-fade {
+  transition: 0.4s ease-out;
```
lyzadanger: We talked about this on a call and decided to back this off a bit (make it a little faster), and in a subsequent PR as it is applied to one of the basic button mixins.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+
+  const suggestionsList = (() => {
+    const suggestionElements = suggestions.map((suggestion, index) => (
+      <option key={index} value={suggestion} />
+    ));
+
+    // Only return a datalist if there is something to render in the list. This
+    // avoid a bug in Firefox where the datalist rendered does not correspond to the
+    // actual list of <option>s in the dom. Wiping out the entire <datalist> tag seems
+    // to force Firefox's hand.
```
lyzadanger: Is there a link we could put here to information about the relevant bug in FF?

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+
+/**
+ * Component to edit annotation's tags.
+ */
+function TagEditor({ annotation, onEditTags, tags: tagsService, tagList }) {
+  const inputEl = useRef(null);
+  const [suggestions, setSuggestions] = useState([]);
+
+  /**
+   * Handle changes to this annotation's tags
```
lyzadanger: This is so much clearer! Thanks!

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+    newTagList.splice(index, 1);
+    updateTags(newTagList);
   };
 
-  this.$onChanges = function(changes) {
-    if (changes.tags) {
-      this.tagList = changes.tags.currentValue.map(function(tag) {
-        return { text: tag };
+  /**
+   * Adds a tag equal to the value of the input field, and
```
lyzadanger: Good comment! Perhaps amend to `Add a tag to the annotation equal...` (i.e. specify the tag is getting added to an annotation).

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+              >
+                <SvgIcon name="cancel" />
+              </button>
+            </li>
+          );
+        })}
+      </ul>
+      <input
+        list={`tag-editor-datalist-${annotation.id}`}
+        onBlur={handleKeyUp}
```
lyzadanger: I still think the naming of the event-handling function is a tad funky given that it also handles `blur`s, but it's a minor thing.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+   *
+   *  @param {Array<Tag>} - Array of tags to save
+   */
+  onEditTags: propTypes.func.isRequired,
+
+  /* The list of editable tags as strings. */
+  tagList: propTypes.array.isRequired,
+
+  /** Services */
+  tags: propTypes.object,
```
lyzadanger: Nit: these are out of alphabetical order.

lyzadanger: I'd presume that both `tags` and `serviceUrl` are _required_, yes?

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
-    angular.mock.module('app', {
-      tags: fakeTags,
+    TagEditor.$imports.$mock(mockImportedComponents());
+  });
+
+  afterEach(() => {
+    TagEditor.$imports.$restore();
+  });
+
+  it('adds appropriate tag values to the elements', () => {
```
lyzadanger: Do you mean "attributes" instead of "tag values"?

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+      assert.isTrue(tag.hasClass('tag-editor__tag-item'));
+      assert.equal(tag.text(), fakeTags[i]);
+      assert.equal(tag.prop('aria-label'), `Tag: ${fakeTags[i]}`);
     });
   });
 
-  it('converts tags to the form expected by ng-tags-input', function() {
-    const element = util.createDirective(document, 'tag-editor', {
-      tags: ['foo', 'bar'],
+  it('creates the component with a unique id', () => {
```
lyzadanger: Can you explain what this test is testing?

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+
+  it('calls fakeTagsService.filter with the value from the input field', () => {
+    const wrapper = createComponent();
+    wrapper.find('input').instance().value = 'tag3';
+    // simulate `keyup` to populate suggestions list
+    wrapper.find('input').simulate('keyup', { key: 'any' });
+    assert.isTrue(fakeTagsService.filter.calledWith('tag3'));
+  });
+
+  it('generates a datalist set equal to the array value returned from fakeTagsService.filter ', () => {
```
lyzadanger: This test tests for the presence of two elements, but does not test that there are no additional values.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+  });
+
+  it('clears the suggestions when adding a new tag', () => {
+    const wrapper = createComponent();
+    wrapper.find('input').instance().value = 'tag3';
+    // simulate `keyup` to populate suggestions list
+    wrapper.find('input').simulate('keyup', { key: 'any' });
+    assert.equal(wrapper.find('datalist option').length, 2);
+    wrapper.find('input').simulate('keyup'); // no key supplied
+    assert.equal(wrapper.find('datalist option').length, 0);
```
lyzadanger: This here looks like the test that the name of the test refers to (that the list is of length 0)....

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+
+    assert.isTrue(
```
lyzadanger: This feels tacked on :)

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+        { text: 'tag1' },
+        { text: 'tag2' },
+        { text: 'tag3' },
+      ])
+    );
+  });
+
+  it('does not render duplicate suggestions', () => {
+    // `tag3` supplied in the `tagList` will be a duplicate value relative
+    // with the fakeTagsService.filter result above.
```
lyzadanger: Although it's redundant, it might be helpful to explicitly (re-)set the return value of the `fakeTagsService.filter` method here for ease-of-test-reading reasons (and you wouldn't need this comment). I do this sometimes even though it's silly, code-wise...

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+    const wrapper = createComponent({
+      editMode: true,
+      tagList: ['tag1', 'tag2', 'tag3'],
     });
-    assert.deepEqual(element.ctrl.tagList, [{ text: 'foo' }, { text: 'bar' }]);
+    // simulate `keyup` to populate suggestions list
```
lyzadanger: I just noticed this—if there's a `keyup`, even if the input field is totally empty, we'll populate the suggestions list? This might be okay, it's just something I want to understand.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+      assert.isTrue(
+        fakeTagsService.store.calledWith(tagList.map(tag => ({ text: tag })))
+      );
+      // called the onEditTags callback prop
+      assert.isTrue(fakeOnEditTags.calledWith({ tags: tagList }));
+      // clears out the suggestions
+      assert.equal(wrapper.find('datalist option').length, 0);
+      // assert the input value is cleared out
+      assert.equal(wrapper.find('input').instance().value, '');
+      // note: focus not tested
```
lyzadanger: Is this because you can't or because you chose not to :) ?

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+    };
+    /**
+     * Helper function to assert that a tag was correctly not added
```
lyzadanger: "was correctly not added" makes me smile. Perhaps "...that a tag was not added" will suffice? It has the word "assert" earlier in the comment, which indicates that the result would be expected.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+     */
+    const addTagsFail = () => {
+      assert.isTrue(fakeTagsService.store.notCalled);
+      assert.isTrue(fakeOnEditTags.notCalled);
+    };
+
+    it('adds a tag from the input field', () => {
```
lyzadanger: I wonder if this test name could be expanded to indicate _when_ this adding of tag is expected to happen.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+      const wrapper = createComponent();
+      wrapper.find('input').instance().value = 'tag3';
+
+      wrapper.find('input').simulate('keyup'); // simulates a selection
+      addTagsSuccess(wrapper, ['tag1', 'tag2', 'tag3']);
+    });
+
+    it('populate the datalist, then adds a tag from the input field', () => {
```
lyzadanger: Again, when :) ?

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+
+      // simulate `keyup` to populate suggestions list
+      wrapper.find('input').simulate('keyup', { key: 'any' });
+      assert.equal(wrapper.find('datalist option').length, 2);
+
+      wrapper.find('input').simulate('keyup'); // simulates a selection
+      addTagsSuccess(wrapper, ['tag1', 'tag2', 'tag3']);
+    });
+
+    it('clears out the <datalist> element after adding a tag', () => {
```
lyzadanger: Good test name here, though!

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+      // clears out the suggestions
+      assert.equal(wrapper.find('datalist option').length, 0);
+      // assert the input value is cleared out
+      assert.equal(wrapper.find('input').instance().value, '');
+      // note: focus not tested
+    };
+    /**
+     * Helper function to assert that a tag was correctly not added
+     */
+    const addTagsFail = () => {
```
lyzadanger: I wonder if this is more of an assertion that a tag didn't get added, versus that adding a tag _failed_, per se. Just a thought...

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
   });
 
-  describe('when tags are changed', function() {
-    let element;
-    let onEditTags;
+  context('when adding tags', () => {
+    /**
+     * Helper function to assert that a tag was correctly added
+     */
+    const addTagsSuccess = (wrapper, tagList) => {
```
lyzadanger: I wonder if the name of this function should reflect that there are embedded assertions. I don't know that we have a convention for this, offhand.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+
+    it('should not add a tag if the input is empty', () => {
+      const wrapper = createComponent();
+      wrapper.find('input').instance().value = '';
+
+      wrapper.find('input').simulate('keyup'); // simulates a selection
+      addTagsFail();
+    });
+
+    it('should not add a tag if the input is blank space', () => {
```
lyzadanger: Good attention to detail here.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+    });
 
-    beforeEach(function() {
-      onEditTags = sinon.stub();
-      element = util.createDirective(document, 'tag-editor', {
-        onEditTags: { args: ['tags'], callback: onEditTags },
-        tags: ['foo'],
-      });
-      element.ctrl.onTagsChanged();
+    it('adds a tag via keypress `Enter`', () => {
```
lyzadanger: Low priority, but I feel like these three tests could be rolled into one with test-case objects and a `forEach`. But they're OK as-is.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
+      assert.isTrue(
+        fakeTagsService.store.calledWith(tagList.map(tag => ({ text: tag })))
+      );
+      // called the onEditTags callback prop
+      assert.isTrue(fakeOnEditTags.calledWith({ tags: tagList }));
+    };
+
+    it('removes `tag1` when clicking its delete button', () => {
+      const wrapper = createComponent();
+      assert.equal(wrapper.find('.tag-editor__edit').length, 2);
```
lyzadanger: This might warrant a comment that there are two of these elements because of the `fakeTagList` value

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,257 @@
...
-      const element = util.createDirective(document, 'tag-editor', {
-        tags: [],
-      });
-      element.ctrl.autocomplete('query');
-      assert.calledWith(fakeTags.filter, 'query');
+  context('when removing tags', () => {
+    /**
+     * Helper function to assert that a tag was correctly added
+     */
+    const removeTagsSuccess = tagList => {
```
lyzadanger: This helper function is only used once, so for ease of human test scanning, it might be better just unrolled into the test.

lyzadanger:
When testing this locally (Chrome 78.0.3904.108 (Official Build) (64-bit)), I ran into some behaviors that I wanted to talk through as at least some of them seem like they may be not quite what we're intending here.

1. Entire suggestion list displays as soon as you click into the tag-editor field in some cases

Here is a video demonstrating what I mean: if you go into the tag editor field, start typing, and then exit and come back, as soon as you click into the editor field again, the entire list of all tag suggestions, ever, shows up.

tag-editor-immediate-tags

I noticed in the code that we'll generate a suggested-tag list even if the `input` field is empty—I wonder if we should always require at least one character first.

2. Full suggested tags list is sometimes incomplete

If I, say, click into the tag-editor field and type "pr" (in this case, matches no stored tags) and then backspace to get to an empty field again, the list of suggested tags is incomplete:

incomplete-tag-list

(I wonder if fixing 1 above such that suggested tags don't display if field is totally empty would fix this problem).

3. It's possible to defeat the duplication-checking

Here's a screenshot of `ten` still showing up in the suggested tags list after it was added:

still-showing-duplicate

4. (Minor) The styling of the little arrow in an active state is a little funky:

selected-arrow-styling

I think much of this could be cleaned up by not showing a suggested-tags list when there are no characters in the input field.

Thoughts?

LMS007:
- 1 and 2 fixed 
- 3 not able to repro, perhaps since fixes to 1 and 2
- 4 I just hid the arrow, I don't think I can change that styling and if we really want one I would rather do our own with a bit more control over look/behavior.
- 5. Can't do anything about this unless we make our own autocomplete and do-away with datalist.

robertknight:
Just a thought - the datalist bugs in Safari and Firefox appear to relate to what happens when the set of options changes while the datalist is already open. Since the datalist filters matches itself as the user types, perhaps for the first iteration we could just avoid doing any filtering in the `TagEditor` component itself as the user types. Instead just create `<option>`s for not-yet-used tags when the input is about to receive focus. Or even simpler, just include all suggestions, even duplicates of current tags, in the `<datalist>`. This will still give us a useful (for the user) & accessible baseline that we can improve from.

lyzadanger:
Problem 1 does seem to be fixed.

I can still reproduce 2 in Chrome: If I focus the field, type the letter `c` and hit delete/backspace, I get:

Screen Shot 2019-12-10 at 9 13 03 AM

I noticed another issue in this round of testing. If I do any scrolling in the sidebar, things get unfortunate; I end up with situations like the following:

Screen Shot 2019-12-10 at 9 11 30 AM

Screen Shot 2019-12-10 at 9 11 52 AM

LMS007:
There is really not much we can do to override much of this behavior. Removing the filtering on our end and just allowing datalist to do it seems to work better per Robs suggestion. For example, it will now no longer open the list up again when you delete back to an empty string; however, it will open the entire list again if you click the component after that point. The detached scrolling issue seems to also be its behavior and I can't seem to get it to work like a normal element that scrolls with its parent.  

Does it on this example too :(
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

I think this pass is about as good as datalist can get but if we want our old behavior we'll have to build out a custom auto complete to replace the datalist. I'm happy to start on that work ? A benefit of that is we will also get substring highlighting.

robertknight:
> I think this pass is about as good as datalist can get but if we want our old behavior we'll have to build out a custom auto complete to replace the datalist. I'm happy to start on that work ? A benefit of that is we will also get substring highlighting.

I'd be happy for you to explore this. I think it would be useful to get a "V2" implementation based on `<datalist>` but relying on the browser's own filtering merged first though. That will tick two boxes:

- It will give us more accessible baseline to build from and a usable autocomplete on mobile (I've only tested on iOS, where `<datalist>` options integrate with the on-screen keyboard. I haven't checked how this works on Android. Can you check?)
- This will mean that all the interior components of an annotation are converted to Preact. This will unblock conversion of components higher up the tree

Depending on whether a custom component can integrate with the native UI on mobile in the same way as a `<datalist>`, it might make sense to switch implementations based on the environment.

src/sidebar/components/tag-editor.js
LMS007: I think this is left over from previous efforts; I dont think we need to bother with the blur event at all

src/sidebar/components/test/tag-editor-test.js
LMS007: How about we prefix it with assert?

src/sidebar/components/test/tag-editor-test.js
LMS007: fixed

src/sidebar/components/test/tag-editor-test.js
LMS007: I agree

src/sidebar/components/tag-editor.js
LMS007: Don't need this anymore after not changing datalist options around.

LMS007:
Discussion here, Dan gave it the okay.
https://hypothes-is.slack.com/archives/C07NXBDNW/p1576103466008500

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,187 @@
...
+      addTag();
+      // preventDefault stops the delimiter from being
+      // added to the input field
+      e.preventDefault();
     }
   };
+  /**
+   *  As the user types, change the suggestions query
+   *  based on the value of inputEl.
+   */
```
robertknight: This comment needs updating or removing to reflect the removal of suggestion filtering.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,187 @@
...
+  };
+
+  const suggestionsList = () => {
+    return (
+      <datalist
+        id={`tag-editor-datalist-${annotation.id}`}
+        className="tag-editor__suggestions"
+        aria-label="Annotation suggestions"
+      >
+        {' '}
```
robertknight: Was the `{' '}` added by mistake here (by Prettier?)?

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,187 @@
...
+        {suggestions.map((suggestion, index) => (
+          <option key={index} value={suggestion} />
```
robertknight: When I tested in Firefox I think I found that it didn't sort suggestions alphabetically, whereas Safari and Chrome did. If that is the case, it would be helpful if we sorted them before creating the options list.

src/styles/sidebar/sidebar.scss
```diff
@@ -61,6 +61,7 @@
 @use './components/svg-icon';
 @use './components/spinner';
 @use './components/tag-list';
+@use './components/tag-editor';
```
robertknight: Nit: This should go above `tag-list` as `tag-editor` < `tag-list`.

src/styles/sidebar/components/annotation.scss
```diff
@@ -61,6 +61,13 @@
   padding-top: 0.5em;
 }
 
+.annotation-link-more {
+  display: flex;
+  flex-flow: row-reverse;
+  padding-bottom: 5px;
+  padding-top: 5px;
+}
```
robertknight: This partially duplicates another CSS selector which appears on line 49.

src/styles/sidebar/components/tag-editor.scss
```diff
@@ -0,0 +1,45 @@
+@use "../../mixins/forms";
+@use "../../mixins/buttons";
+@use "../../variables" as var;
+
+.tag-editor {
+  &__input {
+    &::-webkit-calendar-picker-indicator {
+      // Turn off the default arrow icon
```
robertknight: Hmm ... do we really want to disable this affordance?

src/styles/sidebar/components/tag-editor.scss
```diff
@@ -0,0 +1,45 @@
...
+    color: var.$grey-6;
+    background: var.$grey-1;
+    border: 1px solid var.$grey-3;
+    border-radius: 2px 0 0 2px;
+    border-right-width: 0;
+    padding: 0 0.5em;
+  }
+
+  &__delete {
+    @include buttons.action-button;
```
robertknight: I'm not sure about re-using this mixin here because the other style properties need to fit with it. There is a possibility that someone could change the `action-button` mixin and not realize that it is used for tags as well as buttons.

src/sidebar/components/tag-editor.js
LMS007: Yes

src/styles/sidebar/components/annotation.scss
LMS007: I think this was merge conflict typo I made

src/styles/sidebar/components/tag-editor.scss
LMS007: We don't have control over this styling too much, it just shows on hover somewhat abruptly and and Lyza has some concerns about that. I'll remove it so we can see it again.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,201 @@
...
+      if (value.length !== 0) {
+        // Only change suggestions if input is non-empty
+        const suggestions = removeDuplicates(tagsService.filter(value));
+        setSuggestions(suggestions);
+      } else {
+        setSuggestions([]);
+      }
+    } else if (e.type === 'keyup') {
+      // Assume the user selected an option from <datalist>
+      // either by clicking on it with the mouse or pressing
```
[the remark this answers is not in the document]
LMS007: I see... onInput() handles both an `Event` type an `InputEvent` type. Only one has the '.inputType' prop! ahh such a face palm, but I'm glad you discovered that as I didn't notice that before.  This is much better and almost an exact replacement for my key up handler, but after implementing this I found a shortcoming. If you happen to type out the exact tag that is also a suggestion then there could be a problem.

tabs5

When you type out a tag, you have 4 ways of adding that tag.

1. press Enter w/o selecting the option
2. press comma
3. press down arrow once to highlight the item in the options, then press Enter
4. click the option with mouse.

Only case 1 and 2 work. Both 3 and 4 fail because onInput() is never called. Only the keyup even fires in this case. Now I can try to tease out this exact case only for keyUp, but its messy with so many event handlers and still think there are differences between the browsers. I know that the keyup event in FF reports a key "Enter" when you press enter on the datalist, but not in Chrome, for example. I'll have to flush that out a bit. I have a new diff, but too tried to post  atm, I will sync up with you in the morning.

LMS007:
Made the suggestions to the best I could. I detailed out one pitfall with `onInput` where if you type the same string as an option it won't trigger the event. We can indeed hack around this in Chrome with a second check for `onKeyUp`, however that check will not work in FF or Safari because that event is not triggered in either browser so I don't think its worth having a function for 1 browser only. 

The webkit-calendar-picker-indicator is a little clunky and I did not really attempt tp re-style, nor do I want to in this PR. I don't have a strong opinion on keeping it or hiding it. It won't show up on hover (at least in Chrome) until the options are populated. Those options are hidden until a user starts typing which may be a little non-standard for a datalist. The key use case is I wanted to be able to hide the options when a user cleared all he input as opposed to showing all the options when the input was cleared (e.g. an empty string). Also, I'm not sure we can control any transitions with the webkit-calendar-picker-indicator. It might be best to leave this alone and put efforts into a custom datalist if we want further customization.
 
Safari has another behavior difference. When you type some input, then delete it, then both FF and Chrome re-hide the datalist (mentioned above) since the options are no longer in the DOM. But Safari does not obey the DOM here.  Nothing we can do... other than **not** hiding the options at all in which case the list is going to show up as soon as you click into the input. --Not sure we want that, but its probably more standardized.

LMS007:
FYI: Planning to squash the commits here

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+    return removeDuplicates(tagsService.filter(''), tagList);
+  }, [tagsService, tagList]);
+
+  /**
+   * Handle changes to this annotation's tags
+   */
+  const updateTags = tagList => {
+    // update suggested tags list via service
+    tagsService.store(tagList.map(tag => ({ text: tag })));
+    // invoke callback with updated tags
```
robertknight: I think the next line of code is obvious enough you could leave this comment out.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+      return;
+    }
+    if (tagList.indexOf(value) >= 0) {
+      // don't add duplicate tag
+      return;
+    }
+    updateTags([...tagList, value]);
+    setShowSuggestions(false);
+
+    // clear the input field and maintain focus
```
robertknight: I think the next two lines are obvious enough that you could leave this comment out.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+      // There are two ways to arrive here, either click an item with the mouse
+      // or use keyboard navigation and press 'Enter'.
+      //
+      // If the input value typed already exactly matches the option selected
+      // then this Event won't fire and a user would have to press 'Enter' a second
+      // time to trigger the handleKeyPress callback above to add the tag.
+      addTag();
+    } else if (inputEl.current.value.length === 0) {
+      // If the user deleted input, hide suggestions. This has
+      // no effect in Safari and the list will stay open.
```
robertknight: Can you add a link to the GitHub issue where we are tracking known issues with Safari. In general, if you find a need to either work around a browser bug or just call attention to the fact that it exists, it is very helpful if the code includes a link to a relevant bug report so that future readers can check whether the workaround is still relevant or whether they need to be aware of it when making future changes.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+      setShowSuggestions(false);
+    }
+  };
+
+  const suggestionsList = () => {
+    return (
+      <datalist
+        id={`tag-editor-datalist-${annotation.id}`}
```
robertknight: The `id` field is not set on annotations which have not yet been saved. You could use the `$tag` property as a fallback (`annotation.id || annotation.$tag`). Another option would simply be generating a new identifier on mount with something like:

```js
let prevDatalistId = 0;

function TagEditor({ ... }) {
  ...
  const [datalistId] = useState(() => {
    ++datalistId;
    return datalistId;
  }
  ...
}
```

This has the theoretical advantage, albeit probably not important in practice, that there won't be a conflict if there are multiple `TagEditor` instances mounted which refer to the same annotation.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
+        className="tag-editor__suggestions"
+        aria-label="Annotation suggestions"
+      >
+        {showSuggestions &&
+          suggestions.map((suggestion, index) => (
+            <option key={index} value={suggestion} />
```
robertknight: I think you'd want to use the suggestion itself as the key, since the index of an existing suggestion can change if an update results in a new value being inserted earlier in the list.

src/sidebar/components/tag-editor.js
```diff
@@ -1,35 +1,197 @@
...
-  },
-  template: require('../templates/tag-editor.html'),
+/**
+ * @typedef Tag
+ * @param tag {string} - The tag text
+ */
+
+TagEditor.propTypes = {
+  /* Annotation that owns the tags. */
+  annotation: propTypes.object.isRequired,
```
robertknight: It looks to me like the tag editor doesn't use any properties from the annotation other than its ID, so if `TagEditor` just generated its own ID it could avoid this dependency.

src/sidebar/components/test/tag-editor-test.js
robertknight: The `<datalist>` element needs to use unique ID amongst all the existing `<datalist>`s on the page and the `<input>` needs to refer to that in order to associate the two form controls.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,290 @@
...
+    const wrapper = createComponent();
+    assert.equal(
+      wrapper.find('input').prop('list'),
+      'tag-editor-datalist-annotation_id'
+    );
+  });
+
+  it('generates a ordered datalist containing the array values returned from fakeTagsService.filter ', () => {
+    const wrapper = createComponent();
+    // simulate `keyup` to populate suggestions list
```
robertknight: This comment is out of date. It could just be removed I think.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,63 +1,290 @@
...
-        tags: ['foo'],
-      });
-      element.ctrl.onTagsChanged();
+  it('does not clear the suggestions when deleting only part of the input', () => {
+    const wrapper = createComponent();
+    wrapper.find('input').instance().value = 'tag3';
+    typeInput(wrapper);
+    assert.equal(wrapper.find('datalist option').length, 2);
+    wrapper.find('input').instance().value = 't'; // non-empty input remains
+    wrapper.find('input').simulate('keyup', { key: 'deleteContentBackward' });
```
robertknight: `deleteContentBackward` is a value of the [`InputEvent.inputType`](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/inputType) property rather than [`KeyEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values)

robertknight:
I found a possible solution for the issue in Safari on macOS and iOS:

```diff
diff --git a/src/sidebar/components/tag-editor.js b/src/sidebar/components/tag-editor.js
index 4f6ca548..1739610c 100644
--- a/src/sidebar/components/tag-editor.js
+++ b/src/sidebar/components/tag-editor.js
@@ -87,6 +87,17 @@ function TagEditor({ annotation, onEditTags, tags: tagsService, tagList }) {
     }
   };
 
+  const handleKeyUp = e => {
+    // Safari on macOS and iOS have an issue where pressing "Enter" in an
+    // input when its value exactly matches a suggestion in the associated <datalist>
+    // does not generate a "keypress" event. Therefore we catch the subsequent
+    // "keyup" event instead.
+    if (e.key === 'Enter') {
+      // nb. `addTag` will do nothing if the "keypress" event was already handled.
+      addTag();
+    }
+  };
+
   const handleOnInput = e => {
     if (e.inputType === 'insertText') {
       // Show the suggestions if the user types something into the field
@@ -158,6 +169,7 @@ function TagEditor({ annotation, onEditTags, tags: tagsService, tagList }) {
         list={`tag-editor-datalist-${annotation.id}`}
         onInput={handleOnInput}
         onKeyPress={handleKeyPress}
+        onKeyUp={handleKeyUp}
         ref={inputEl}
         placeholder="Add tags..."
         className="tag-editor__input"
```

Catching enter in the keydown or keypress events is preferable as it feels faster than if the app has to wait for the user to release the key. Nevertheless, the behavior seemed reasonable to me when I tested in Chrome, Firefox, and Safari (macOS and iOS).

Do you think this could work?

src/sidebar/components/test/tag-editor-test.js
LMS007: That was a mistake, intended to change that too

src/sidebar/components/tag-editor.js
LMS007: I like the generated id best.

LMS007:
I made your suggestions but I see some concerns, I'll try and outline them as best I can.

There are 5 ways to select a tag that I think are important to discuss.

1. type the whole new tag and press enter (this tag would have not been in the datalist)
2. type a partial tag, key down and press enter and from datalist.
3. type the whole tag and have it match a suggestion and press enter w/o keying down into the datalist.
4. type a partial tag and select it from the datalist with the mouse.
5. type a whole tag and select it from the datalist with the mouse.

I'm still noticing some differences between Chrome and Safari and I'm not sure how to land this.

**In Safari.**

1. Works with an issue, it leaves the datalist open in spite if the options clearing. Seems that Safari will not dismiss the datalist until the user actually clicks on it or away from it, or presses enter with an option selected. Chrome on the other hand will clear it out when the DOM changes. Its possible I could send a click outside event after adding a tag? I'll experiment with that today.

2. Work perfect.

3. Works but suffers from the same issue as 1.

4. Works perfect

5. Does not work and no keyUp event is fired.

_Data list stays open_ 
tags_s


**In Chrome**
1. Works perfect
2. Works perfect 
3. Works perfect
4. Works perfect
5. Does not work, but we could get this working by checking for an `unknown` key in onKeyUp. Where is with Safari, we don't get a keyUp event.

robertknight:
Hi @LMS007,

Thanks for the update. To check my understanding about scenario (5) and what I can see locally: If I type a whole tag which matches a suggestion, and then select the suggestion from the datalist, then the suggestion is not automatically added as a tag but instead focus remains in the input and I need to press Enter again. That seems like only a very minor annoyance to me, if users even notice at all.

src/styles/sidebar/components/tag-editor.scss
LMS007: I think we can remove action button, but Id still like to base this off the base button mixin because after all, it is a button

PR #1567, opened 2019-12-05
Add charset token to Content-Type header for client files deployed to S3

### LINKED ISSUE #1547: Sidebar icons fail to render if annotated page does not set UTF-8 charset
When Hypothesis is activated on a page which does not use `<meta charset="UTF-8">` or some other other supported method of setting the charset to UTF-8, icons in the sidebar's vertical toolbar fail to render:

Screenshot 2019-11-29 16 25 37

These icons come from the h.woff icon font. The `icomoon.css` file in the source tree uses hex encodings to specify the character values to use. The [compiled icomoon.css file](https://cdn.hypothes.is/hypothesis/1.243.0/build/styles/icomoon.css?036138) however uses UTF-8 character values. Therefore if the page is not parsed using UTF-8 encoding, the icons will fail to load. The icomoon.css file itself is served from cdn.hypothes.is without a charset. I'm not sure whether specifying a charset in the `Content-Type` header for icomoon.css would fix the issue or if it has to be set for the page itself.

[Slack thread](https://hypothesis-open.slack.com/archives/C6Y375TT3/p1574743138122700).

robertknight:
Add a `charset=UTF-8` token to the `Content-Type` header of client files
deployed to S3. This ensures that characters in JS / CSS / SVG files used in
the host page are interpreted correctly if the host page does not set a
charset itself or sets a non-UTF-8 charset.

Fixes #1547

----

I have tested this locally as follows (note the "1.0.0-dummy-version" string comes from `package.json` and is overridden for normal releases):

```shellsession
$ export AWS_ACCESS_KEY_ID=<credentials for cdn.hypothes.is>
$ export AWS_SECRET_ACCESS_KEY=<credentials for cdn.hypothes.is>
$ env NODE_ENV=production gulp build
$ ./scripts/deploy-to-s3.js --bucket cdn.hypothes.is --tag testing

$ curl -I 'https://cdn.hypothes.is/hypothesis/1.0.0-dummy-version/build/styles/annotator.css'
HTTP/2 200
date: Thu, 05 Dec 2019 15:15:05 GMT
content-type: text/css; charset=UTF-8
...

$ curl -I 'https://cdn.hypothes.is/hypothesis/1.0.0-dummy-version/build/fonts/h.woff'
HTTP/2 200
date: Thu, 05 Dec 2019 15:15:47 GMT
content-type: font/woff
...
```

### DIFF
--- a/scripts/deploy-to-s3.js
+++ b/scripts/deploy-to-s3.js
@@ -17,7 +17,7 @@ const TEXT_MIME_TYPES = {
   '.md': 'text/markdown',
   '.js': 'application/javascript',
   '.json': 'application/json',
-  '.map': 'application/octet-stream',
+  '.map': 'application/json',
   '.svg': 'image/svg+xml',
 };
 
--- a/scripts/deploy-to-s3.js
+++ b/scripts/deploy-to-s3.js
@@ -12,13 +12,16 @@ const AWS = require('aws-sdk');
 /**
  * File extension / mime type associations for file types we actually use.
  */
-const MIME_TYPES = {
+const TEXT_MIME_TYPES = {
   '.css': 'text/css',
   '.md': 'text/markdown',
   '.js': 'application/javascript',
   '.json': 'application/json',
   '.map': 'application/octet-stream',
   '.svg': 'image/svg+xml',
+};
+
+const BINARY_MIME_TYPES = {
   '.woff': 'font/woff',
   '.woff2': 'font/woff2',
 };
@@ -33,9 +36,16 @@ function contentTypeFromFilename(path) {
     return 'text/plain';
   }
 
-  if (MIME_TYPES[extension]) {
-    return MIME_TYPES[extension];
+  let mimeType = TEXT_MIME_TYPES[extension];
+  if (mimeType) {
+    return mimeType + '; charset=UTF-8';
   }
+
+  mimeType = BINARY_MIME_TYPES[extension];
+  if (mimeType) {
+    return mimeType;
+  }
+
   throw new Error(`Unable to look up Content-Type for ${path}`);
 }
 


scripts/deploy-to-s3.js
```diff
@@ -33,9 +36,16 @@ function contentTypeFromFilename(path) {
     return 'text/plain';
   }
 
-  if (MIME_TYPES[extension]) {
-    return MIME_TYPES[extension];
+  let mimeType = TEXT_MIME_TYPES[extension];
+  if (mimeType) {
+    return mimeType + '; charset=UTF-8';
```
LMS007: I think `application/octet-stream` is a binary file too? Does that need to be in this list, or if not maybe we can just leave a single mapping and do this?
```
const MIME_TYPES = {	const TEXT_MIME_TYPES = {
  '.css': 'text/css',	  '.css': 'text/css; charset=UTF-8 ',
  '.md': 'text/markdown',	  '.md': 'text/markdown; charset=UTF-8',
  '.js': 'application/javascript',	  '.js': 'application/javascript; charset=UTF-8',
  '.json': 'application/json',	  '.json': 'application/json; charset=UTF-8',
  '.map': 'application/octet-stream',	  '.map': 'application/octet-stream; charset=UTF-8',
  '.svg': 'image/svg+xml',	  '.svg': 'image/svg+xml; charset=UTF-8',
  '.woff': 'font/woff',	  '.woff': 'font/woff',
  '.woff2': 'font/woff2',	  '.woff2': 'font/woff2',
};
```

robertknight: > I think application/octet-stream is a binary file too? 

The `application/octet-stream` mime type is indeed used as a "generic mime type" for binary files. In the case of `.map` though, it is a sourcemap, which is [actually a JSON object](https://sourcemaps.info/spec.html#h.qz3o9nc69um5) (see for example: https://cdn.hypothes.is/hypothesis/1.210.0/build/scripts/annotator.bundle.js). I think I took the mime type from [this Stack Overflow answer](https://stackoverflow.com/questions/19911929/what-mime-type-should-i-use-for-javascript-source-map-files). Perhaps it would make sense to copy the Google CDN and change the type to `application/json` for obvious-ness.

robertknight: Okay, I've pushed a change to set the mime type for `.map` files to `application/json`. As the Stack Overflow answer notes, devtools don't care but it means browsers will render the file in a more helpful way if you access them directly.

robertknight:
I got a verbal approval from @LMS007 in the dev sync this afternoon so I'm going to merge this and test on QA.

PR #1571, opened 2019-12-05
Add standardized hover transition

LMS007:
- Use effects.hover-fade where appropriate for clickable things
- Remove dark border hover color from tag item

There is no point in making a GIF of this change because it's too fast to be captured that way so you'll need to pull and run locally to see changes.

**The purpose of this diff is 2 fold.**
1. Add a standardized way to set transition timing across all clickable things.
2. Apply that standardization to many of the clickable things. (I don't except to cover them all with one pass)

This does more of a heavy handed transition where it does not specifically call out `color` or `background-color` or `opacity` or anything, but rather asserts that any css property change will be transitioned in 100ms when the mouse is hovering over the element and then revert in 250ms there after. I wanted to keep these somewhat fast because this is a major UI change and I don't want it to feel too much different or slow, but mostly just more polished.

Also removed that dark border on TagList items which didn't match anything. I realize this diverges from the new design but we can circle back to this on a second pass when we do more uniform changes across the UI. It just felt like it stood out too much right now.

### DIFF
--- a/src/styles/annotator/annotator.scss
+++ b/src/styles/annotator/annotator.scss
@@ -4,6 +4,7 @@
 @use '../mixins/focus';
 @use '../mixins/reset';
 @use '../mixins/shadow';
+@use '../mixins/effects';
 
 @use './adder';
 @use './bucket-bar';
@@ -110,9 +111,9 @@ var.$base-font-size: 14px;
   }
 
   .annotator-frame-button {
+    @include effects.hover-fade;
     @include focus.outline-on-keyboard-focus;
 
-    transition: background-color 0.25s;
     @include shadow.smallshadow;
     background: var.$white;
     border: solid 1px var.$gray-lighter;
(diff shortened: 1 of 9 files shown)


src/styles/mixins/effects.scss
```diff
@@ -0,0 +1,9 @@
+/**
```
robertknight: `effects` is perhaps ambiguous. Perhaps transitions.scss or animations.scss might be more obvious?

src/styles/annotator/annotator.scss
```diff
@@ -4,6 +4,7 @@
 @use '../mixins/focus';
 @use '../mixins/reset';
 @use '../mixins/shadow';
+@use '../mixins/effects';
```
robertknight: Tiny nit - can you sort these?

src/styles/mixins/effects.scss
LMS007: Sure, I like animations. We may end up with some keyframe stuff eventually too.

PR #1582, opened 2019-12-09
Remove unused `gulp-{if, postcss, sourcemaps}` dependencies

robertknight:
These are no longer used since `scripts/gulp/create-style-bundle.js` invokes SASS and postcss directly.


robertknight:
I went ahead and merged this since I am just removing some unused build dependencies.

PR #1593, opened 2019-12-11
Don't filter annotations when showing "All Students"

seanh:
Fixes https://github.com/hypothesis/lms/issues/1259


seanh:
I think it does fix the bug. Whether people would want to refactor/redesign the code differently in the future with more time I honestly don't know but that would be refactoring not bug fixing

PR #1605, opened 2019-12-18
Use mobile-friendly font size for all input fields

robertknight:
Match other input fields in the sidebar in using a larger font size for
the tag editor input on mobile/touch inputs. This also prevents iOS
Safari from zooming into the input when focused.

**Update:** In response to a [suggestion](https://github.com/hypothesis/client/pull/1605#discussion_r359425564) I did some refactoring of our input styles to make all input fields use the `form-input` mixin for their styles and moved the font-size and focus outline adjustments into the `form-input` mixin. The end result is that all input fields in the sidebar now have the same focus outline and use the same larger font size when using touch input. The one visual change on desktop is that the search input's focus ring is consistent with other input fields in the app:

Screenshot 2019-12-19 12 18 49

### DIFF
--- a/src/styles/forms.scss
+++ b/src/styles/forms.scss
@@ -5,15 +5,4 @@
   @include forms.form-input;
   width: 100%;
   display: block;
-
-  &:focus,
-  &.js-focus {
-    @include forms.form-input-focus;
-  }
-}
-
-.form-textarea {
-  min-height: 8em;
-  max-width: 100%;
-  resize: vertical;
 }
--- a/src/styles/sidebar/components/tag-editor.scss
+++ b/src/styles/sidebar/components/tag-editor.scss
@@ -6,13 +6,6 @@
   &__input {
     @include forms.form-input;
     width: 100%;
-    &:focus {
-      @include forms.form-input-focus;
-    }
-
-    @media (pointer: coarse) {
-      font-size: var.$touch-input-font-size;
-    }
   }
 
   &__tag-list {
--- a/src/styles/sidebar/components/tag-editor.scss
+++ b/src/styles/sidebar/components/tag-editor.scss
@@ -9,6 +9,10 @@
     &:focus {
       @include forms.form-input-focus;
     }
+
+    @media (pointer: coarse) {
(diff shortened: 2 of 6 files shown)


src/styles/sidebar/components/tag-editor.scss
```diff
@@ -9,6 +9,10 @@
     &:focus {
       @include forms.form-input-focus;
     }
+
+    @media (pointer: coarse) {
```
LMS007: Can this just be part of @include forms.form-input ?

robertknight:
Thanks for the suggestion @LMS007. I have moved this font size adjustment into the `form-input` mixin. At the same time I did some refactoring so that all input fields now use this mixin, which means that the customized `:focus` styles can also go there as well. This expands the scope of this PR a bit but it should make it easier to adopt any pattern library changes as well.

src/styles/forms.scss
```diff
@@ -5,15 +5,4 @@
   @include forms.form-input;
   width: 100%;
   display: block;
-
-  &:focus,
-  &.js-focus {
-    @include forms.form-input-focus;
```
robertknight: The `:focus` selectors are now part of the `form-input` mixin.

src/styles/forms.scss
```diff
@@ -5,15 +5,4 @@
...
-  }
-}
-
-.form-textarea {
-  min-height: 8em;
-  max-width: 100%;
-  resize: vertical;
```
robertknight: `form-textarea` is no longer used.

src/styles/sidebar/components/tag-editor.scss
```diff
@@ -6,9 +6,6 @@
   &__input {
     @include forms.form-input;
     width: 100%;
-    &:focus {
```
robertknight: This is now part of the `form-input` mixin.

LMS007:
I think this is change is good and points us in the right direction for getting everything consolidated. I did look on my phone (android/chrome), but I found that testing these with just the mobile emulator in chrome was a lot easier.

Issue #1606, opened 2019-12-18
Tag editor freezes in Chrome on macOS if there are a large number of suggestions present


robertknight:
**Steps to reproduce**

1. Apply the following diff to the client code (I tested against the v1.261.0 tag)

```diff
diff --git a/src/sidebar/services/tags.js b/src/sidebar/services/tags.js
index a897251d..8024bf2b 100644
--- a/src/sidebar/services/tags.js
+++ b/src/sidebar/services/tags.js
@@ -19,6 +19,11 @@ function tags(localStorage) {
   const TAGS_LIST_KEY = 'hypothesis.user.tags.list';
   const TAGS_MAP_KEY = 'hypothesis.user.tags.map';
 
+  const testTags = [];
+  for (let i = 0; i < 1000; i++) {
+    testTags.push(`test-tag-${i}`);
+  }
+
   /**
    * Return a list of tag suggestions matching `query`.
    *
@@ -26,7 +31,7 @@ function tags(localStorage) {
    * @return {Tag[]} List of matching tags
    */
   function filter(query) {
-    const savedTags = localStorage.getObject(TAGS_LIST_KEY) || [];
+    const savedTags = testTags;
 
     return savedTags.filter(e => {
       return e.toLowerCase().indexOf(query.toLowerCase()) !== -1;
```
2. In Chrome, go to a page with the dev client loaded (eg. http://localhost:3000) and create an annotation or page note
3. Start typing in the tag editor input field

**Expected result**

Typing is smooth. Only suggestions matching typed text are shown.

**Actual result**

Chrome browser UI freezes. The whole browser, not just the active tab, is unresponsive.

**Notes**

- Tested with Chrome Version 80.0.3987.16 (Official Build) dev (64-bit) on macOS 10.15.2
- The issue is not observed in Safari or Firefox on macOS
- I don't know if the problem is macOS-specific or also affects Windows and/or Linux

robertknight:
Related upstream issues:

1. https://bugs.chromium.org/p/chromium/issues/detail?id=789187 (When the datalist becomes to long, it will not display anymore)
2. https://bugs.chromium.org/p/chromium/issues/detail?id=721655 (Speed up dynamic updates to options in a datalist)
[...]

PR #1611, opened 2019-12-20
Fix datalist performance in Chrome

LMS007:
When tag suggestions exceeds just a few hundred results, a noticeable performance is observed in Chrome. 

I arbitrarily selected 20 results as a reasonable limit, but it may be best to match firefox / safari which window the result to 6 or so at a time?

I left some test code in so you can easily test 1000 suggestions. this should be fast in every browser now.

fixed https://github.com/hypothesis/client/issues/1606

### DIFF
--- a/src/sidebar/components/tag-editor.js
+++ b/src/sidebar/components/tag-editor.js
@@ -17,23 +17,38 @@ let datalistIdCounter = 0;
 function TagEditor({ onEditTags, tags: tagsService, tagList }) {
   const inputEl = useRef(null);
   const [showSuggestions, setShowSuggestions] = useState(false);
+  const [suggestionsFiltered, setSuggestionsFiltered] = useState([]); // For Chrome only
   const [datalistId] = useState(() => {
     ++datalistIdCounter;
     return `tag-editor-datalist-${datalistIdCounter}`;
   });
 
+  const isChromeUA = !!window.navigator.userAgent.match(/Chrome\/[.0-9]*/);
+  // Suggestion limiter to speed up the performance in Chrome
+  // See https://github.com/hypothesis/client/issues/1606
+  const chromeSuggestionsLimit = 20;
+
+  /**
+   * Helper function that returns a list of suggestions less any
+   * results also found from the duplicates list.
+   *
+   * @param {Array<string>} suggestions - Original list of suggestions
+   * @param {Array<string>} duplicates - Items to be removed from the result
+   * @return {Array<string>}
+   */
+  const removeDuplicates = (suggestions, duplicates) => {
+    const suggestionsSet = [];
+    suggestions.forEach(suggestion => {
+      if (duplicates.indexOf(suggestion) < 0) {
+        suggestionsSet.push(suggestion);
+      }
+    });
+    return suggestionsSet.sort();
+  };
+
   // List of suggestions returned from the tagsService
   const suggestions = useMemo(() => {
     // Remove any repeated suggestions that are already tags.
-    const removeDuplicates = (suggestions, tags) => {
-      const suggestionsSet = [];
-      suggestions.forEach(suggestion => {
-        if (tags.indexOf(suggestion) < 0) {
-          suggestionsSet.push(suggestion);
-        }
-      });
-      return suggestionsSet.sort();
-    };
     // Call filter with an empty string to return all suggestions
     return removeDuplicates(tagsService.filter(''), tagList);
   }, [tagsService, tagList]);
@@ -94,9 +109,30 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
   };
 
   const handleOnInput = e => {
-    if (e.inputType === 'insertText') {
-      // Show the suggestions if the user types something into the field
-      setShowSuggestions(true);
+    if (
+      e.inputType === 'insertText' ||
+      e.inputType === 'deleteContentBackward'
+    ) {
+      if (inputEl.current.value.length === 0) {
+        // If the user deleted input, hide suggestions. This has
+        // no effect in Safari and the list will stay open.
+        setShowSuggestions(false);
+      } else {
+        if (isChromeUA) {
+          // Filter down the suggestions list for Chrome to subvert a performance
+          // issue. Returning too many suggestions causes a noticeable delay when
+          // rendering the datalist.
+          // See https://github.com/hypothesis/client/issues/1606
+          setSuggestionsFiltered(
+            removeDuplicates(
+              tagsService.filter(inputEl.current.value, chromeSuggestionsLimit),
+              tagList
+            )
+          );
+        }
+        // Show the suggestions if the user types something into the field
+        setShowSuggestions(true);
+      }
     } else if (
       e.inputType === undefined ||
       e.inputType === 'insertReplacementText'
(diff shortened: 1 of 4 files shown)


src/sidebar/components/tag-editor.js
```diff
@@ -17,23 +17,38 @@ let datalistIdCounter = 0;
...
+
+  /**
+   * Helper function that returns a list of suggestions less any
+   * results also found from the duplicates list.
+   *
+   * @param {Array<string>} suggestions - Original list of suggestions
+   * @param {Array<string>} duplicates - Items to be removed from the result
+   * @return {Array<string>}
+   */
+  const removeDuplicates = (suggestions, duplicates) => {
```
LMS007: same function as before, just pulled it out so I could use it twice.

src/sidebar/components/tag-editor.js
```diff
@@ -94,9 +109,30 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
 
   const handleOnInput = e => {
-    if (e.inputType === 'insertText') {
-      // Show the suggestions if the user types something into the field
-      setShowSuggestions(true);
+    if (
+      e.inputType === 'insertText' ||
+      e.inputType === 'deleteContentBackward'
+    ) {
+      if (inputEl.current.value.length === 0) {
```
LMS007: This was moved from below because I need to capture `deleteContentBackward` here now and call filter() again if there is input.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -307,4 +307,37 @@ describe('TagEditor', function() {
       assert.isTrue(fakeOnEditTags.calledWith({ tags: ['tag2'] }));
     });
   });
+
+  describe('filter on text input based on user agent', () => {
+    const originalUA = navigator.userAgent;
+
+    afterEach(function() {
+      navigator.__defineGetter__('userAgent', () => {
```
LMS007: I could not figure out a better way to mock this :\

src/sidebar/services/tags.js
```diff
@@ -19,17 +19,32 @@ function tags(localStorage) {
...
    *
    * @param {string} query
+   * @param {number} limit - Optional limit of the results.
    * @return {Tag[]} List of matching tags
    */
-  function filter(query) {
+  function filter(query, limit = null) {
     const savedTags = localStorage.getObject(TAGS_LIST_KEY) || [];
-
+    //const savedTags = testTags;
```
LMS007: uncomment this line to test 1000 suggestions

src/sidebar/services/tags.js
```diff
@@ -19,17 +19,32 @@ function tags(localStorage) {
...
+    let resultCount = 0;
     return savedTags.filter(e => {
-      return e.toLowerCase().indexOf(query.toLowerCase()) !== -1;
+      if (e.toLowerCase().indexOf(query.toLowerCase()) !== -1) {
+        if (!limit || resultCount < limit) {
```
robertknight: `!limit` will be true here if it is equal to `0`, which means that `tags.filter(query, 0)` could unexpectedly return a non-empty list.

I wasn't sure if it was safe to assume that `filter` is called with each element in the array once in any particular order, but it turns out that the JS spec does specify the iteration order ([spec link](https://tc39.es/ecma262/#sec-array.prototype.filter)).

src/sidebar/components/test/tag-editor-test.js
robertknight: You can use `sinon.stub`:

```js
// Stub property.
let stub = sinon.stub(navigator, 'userAgent');

// Define getter for stub that returns a fixed string.
stub.get(() => 'Chrome/1.0')

// Remove the stub
stub.restore()
```

Alternatively you could add a test seam to the component - add a `userAgent` prop which is default-initialized to `navigator.userAgent`.

src/sidebar/components/tag-editor.js
```diff
@@ -17,23 +17,38 @@ let datalistIdCounter = 0;
 function TagEditor({ onEditTags, tags: tagsService, tagList }) {
   const inputEl = useRef(null);
   const [showSuggestions, setShowSuggestions] = useState(false);
+  const [suggestionsFiltered, setSuggestionsFiltered] = useState([]); // For Chrome only
```
robertknight: `suggestionsFiltered` sounds like a boolean to me (as in "_are_ the suggestions filtered?"). `filteredSuggestions` would have sounded more like a list.

LMS007: Its been a week so my memory is hazy with this, but I'm pretty sure I trie to stub `navigator` this w/o success, but I'll give it a try again

LMS007: Its been a week so my memory is hazy with this, but I'm pretty sure I trie to stub `navigator` this w/o success, but I'll give it a try again

src/sidebar/services/tags.js
LMS007: I think I was just assuming a limit of 0 made no sense so I didn't bother with an extra condition. e.g. a limit of 0 would just allow the entire list, but I can see how that is confusing.

src/sidebar/components/tag-editor.js
robertknight: I think this reply was meant for a different comment.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -307,4 +307,34 @@ describe('TagEditor', function() {
       assert.isTrue(fakeOnEditTags.calledWith({ tags: ['tag2'] }));
     });
   });
+
+  describe('filter on text input based on user agent', () => {
+    let stub = sinon.stub(navigator, 'userAgent');
```
robertknight: This stub needs to happen in a `beforeEach` block which is executed before each test rather than in the top level of the `describe` block which is executed once at test setup time.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -307,4 +307,34 @@ describe('TagEditor', function() {
...
+      stub.get(() => 'Chrome/1.0');
+      wrapper.find('input').instance().value = 'tag';
+      typeInput(wrapper);
+
+      assert.isTrue(fakeTagsService.filter.calledTwice);
+      assert.isTrue(fakeTagsService.filter.calledWith('tag', 20));
+    });
+
+    it('does not call filter when changing input when browser is not Chrome', () => {
+      let stub = sinon.stub(navigator, 'userAgent');
```
robertknight: This stub is never undone because you are using a new local `stub` variable here. If the `stub = sinon.stub(...)` call happens in a `beforeEach` block then `navigator.userAgent` will already be stubbed when this function runs and you can just call `stub.get(...)`.

src/sidebar/services/tags.js
```diff
@@ -23,13 +23,23 @@ function tags(localStorage) {
    * Return a list of tag suggestions matching `query`.
    *
    * @param {string} query
+   * @param {number} limit - Optional limit of the results.
    * @return {Tag[]} List of matching tags
    */
-  function filter(query) {
+  function filter(query, limit = 0) {
+    const limitNumber = Number(limit); // ensures we are dealing with a number
```
robertknight: I think I understand what you're trying to do here. In general however I think that it is often preferable to use a different type of value (`null` or `undefined`) to indicate that a value is not set rather than a magic value of the same type (`0`). The reason being that it reduces the chance of subtle bugs where the special value (`0`) is manipulated without checking it first. If this happens with `null` it will crash loudly. It also avoids scenarios where the behavior of the function suddenly changes if a programatically generated value happens to equal the special value.

My preference here would be to use `null` as before to indicate "no limit" but test for that value specifically below (ie. `limit === null`).

src/sidebar/components/test/tag-editor-test.js
LMS007: I think thats a typo I meant to delete, thanks

Issue #1570, opened 2019-12-05
Remove keyboard focus halo from annotation-publish buttons


LMS007:
The keyboard active/focus halo should only be for tab/keyboard navigation and not for clicked items.

![active](https://user-images.githubusercontent.com/3939074/70281225-58645300-176f-11ea-83e3-a0fd7f79ca7f.gif)

PR #1628, opened 2019-12-30
Fix focus halo on primary action button mixin

LMS007:
Simple fix + some cleanup

The primary reason for this PR is to fix https://github.com/hypothesis/client/issues/1570. This is achieved by adding the `focus.outline-on-keyboard-focus` mixin to the `@mixin primary-action-btn`

I also noticed that `primary-action-btn.scss` was not used anywhere in the code (I think).

Notice the "Post to Public" button does not show halo when clicking anymore

halo


LMS007:
> other buttons that have this mixin applied though.

Maybe we need an indicator, but perhaps a slightly different one than the bright halo thing for that case? I do think no matter what that having a single place in the code to control this is obviously important and this helps move us in that direction.

PR #1649, opened 2020-01-07
Convert most `require`s to ES `import` syntax

robertknight:
This PR converts most CommonJS/Node `require` imports to ES module `import` syntax using the [convert-to-es-modules](https://github.com/hypothesis/frontend-toolkit/tree/master/packages/convert-to-es-modules) script from the frontend-toolkit repo. The immediate motivation is that it means we'll be using the same module system across the client and lms projects, which will be convenient when sharing UI components and utilities between them.

There will be other benefits now and in the future in that ES imports are more amenable to static analysis. For example "Go to definition"/"Find references" using the ALE plugin in Vim (and I would guess VS Code or Atom, since they all use the same language server) works better.

The first few commits make some small changes that make it easier for the automated script to do its work. The last commit is the result of running the code on all files in `src/` except for `src/karma.config.js`.

The script only processes `require` calls which appear as top-level statements or variable initializers. This means that there are still some `require` calls which are invoked conditionally inside functions that are not touched by this PR.

The script preserves the existing sort order of imports, although it only has two categories for grouping (local imports vs imports from node_modules). We might use automated formatting to standardize the ordering/grouping in future as was recently done for the backend projects.

After this PR lands, I will move to the next step which is migrating `module.exports` to ES `export` syntax.

### DIFF
--- a/src/annotator/adder.js
+++ b/src/annotator/adder.js
@@ -1,6 +1,6 @@
-const classnames = require('classnames');
+import classnames from 'classnames';
 
-const template = require('./adder.html');
+import template from './adder.html';
 
 const ANNOTATE_BTN_SELECTOR = '.js-annotate-btn';
 const HIGHLIGHT_BTN_SELECTOR = '.js-highlight-btn';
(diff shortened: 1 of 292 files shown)


robertknight:
I'm going to split this into smaller PRs that are easier to review.

PR #1653, opened 2020-01-09
Add custom datalist component

LMS007:
- Rework tag-editor to use new custom datalist
- Remove tag-editor’s Chrome hack fix to limit datalist to 20
- useElementShouldClose now works on Preact component as well as native components
datalist

Also fixes https://github.com/hypothesis/client/issues/1690

### DIFF
--- a/src/sidebar/components/datalist.js
+++ b/src/sidebar/components/datalist.js
@@ -1,0 +1,73 @@
+const { createElement } = require('preact');
+const classnames = require('classnames');
+const propTypes = require('prop-types');
+const { useMemo } = require('preact/hooks');
+
+/**
+ * Custom datalist component
+ */
+
+function DataList({
+  activeItem = -1,
+  list,
+  listFormatter = item => item,
+  onSelectItem,
+  open = false,
+}) {
+  const items = useMemo(() => {
+    return list.map((item, index) => {
+      return (
+        <li
+          key={`datalist-${index}`}
+          className={classnames({
+            'is-selected': activeItem === index,
+          })}
+          onClick={() => {
+            onSelectItem(item);
+          }}
+        >
+          {listFormatter(item, index)}
+        </li>
+      );
+    });
+  }, [activeItem, list, listFormatter, onSelectItem]);
+
+  return (
+    <div className="datalist">
+      {list.length && open && (
+        <div className="datalist__items">
+          <span className="datalist__arrow-down"/>
+          <ul className="datalist__ul">{items}</ul>
+        </div>
+      )}
+    </div>
+  );
+}
+
+DataList.propTypes = {
+  /**
+   * The activeItem is highlighted.
+   */
+  activeItem: propTypes.number,
+  /**
+   * The list of items to render. This can be a simple list of
+   * strings or a list of objects when used with listFormatter.
+   */
+  list: propTypes.array.isRequired,
+  /**
+   * An optional formatter to render each item inside an <li> tag
+   */
+  listFormatter: propTypes.func,
+  /**
+   * Callback when an item is clicked with the mouse.
+   *  (item, index) => {}
+   */
+  onSelectItem: propTypes.func.isRequired,
+  /**
+   * Is the list open or closed?
+   */
+
+  open: propTypes.boolean,
+};
+
+module.exports = DataList;
--- a/src/sidebar/components/datalist.js
+++ b/src/sidebar/components/datalist.js
@@ -4,10 +4,10 @@ const propTypes = require('prop-types');
 const { useMemo } = require('preact/hooks');
 
 /**
- * Custom datalist component
+ * Custom datalist component. Use this in conjunction with an input field.
  */
 
-function DataList({
+function Datalist({
   activeItem = -1,
   list,
   listFormatter = item => item,
@@ -33,10 +33,10 @@ function DataList({
   }, [activeItem, list, listFormatter, onSelectItem]);
 
   return (
-    <div className="datalist">
+    <div className={classnames({}, 'datalist')}>
       {list.length && open && (
         <div className="datalist__items">
-          <span className="datalist__arrow-down"/>
+          <span className="datalist__arrow-down" />
           <ul className="datalist__ul">{items}</ul>
         </div>
       )}
@@ -44,7 +44,7 @@ function DataList({
   );
 }
 
-DataList.propTypes = {
+Datalist.propTypes = {
   /**
    * The activeItem is highlighted.
    */
@@ -56,6 +56,8 @@ DataList.propTypes = {
   list: propTypes.array.isRequired,
   /**
    * An optional formatter to render each item inside an <li> tag
+   * This is useful if the list is an array of objects rather than
+   * just strings.
    */
   listFormatter: propTypes.func,
   /**
@@ -66,8 +68,7 @@ DataList.propTypes = {
   /**
    * Is the list open or closed?
    */
-
-  open: propTypes.boolean,
+  open: propTypes.bool,
 };
 
-module.exports = DataList;
+module.exports = Datalist;
--- a/src/sidebar/components/datalist.js
+++ b/src/sidebar/components/datalist.js
@@ -36,7 +36,6 @@ function Datalist({
     <div className="datalist">
       {list.length > 0 && open && (
         <div className="datalist__items">
-          <span className="datalist__arrow-down" />
           <ul className="datalist__ul">{items}</ul>
         </div>
       )}
--- a/src/sidebar/components/datalist.js
+++ b/src/sidebar/components/datalist.js
@@ -33,8 +33,8 @@ function Datalist({
   }, [activeItem, list, listFormatter, onSelectItem]);
 
   return (
-    <div className={classnames({}, 'datalist')}>
-      {list.length && open && (
+    <div className="datalist">
+      {list.length > 0 && open && (
         <div className="datalist__items">
           <span className="datalist__arrow-down" />
           <ul className="datalist__ul">{items}</ul>
--- a/src/sidebar/components/datalist.js
+++ b/src/sidebar/components/datalist.js
@@ -9,6 +9,8 @@ const { useMemo } = require('preact/hooks');
 
 function Datalist({
   activeItem = -1,
+  id,
+  itemPrefixId,
   list,
   listFormatter = item => item,
   onSelectItem,
@@ -16,27 +18,42 @@ function Datalist({
 }) {
   const items = useMemo(() => {
     return list.map((item, index) => {
+      // only add an id if itemPrefixId is passed
+      const props = itemPrefixId ? { id: `${itemPrefixId}${index}` } : {};
+
       return (
         <li
           key={`datalist-${index}`}
+          role="option"
+          aria-selected={activeItem === index}
           className={classnames({
             'is-selected': activeItem === index,
           })}
           onClick={() => {
             onSelectItem(item);
           }}
+          {...props}
         >
           {listFormatter(item, index)}
         </li>
       );
     });
-  }, [activeItem, list, listFormatter, onSelectItem]);
+  }, [activeItem, itemPrefixId, list, listFormatter, onSelectItem]);
 
+  const props = id ? { id } : {}; // only add the id if its passed
   return (
     <div className="datalist">
       {list.length > 0 && open && (
         <div className="datalist__items">
-          <ul className="datalist__ul">{items}</ul>
+          <ul
+            tabIndex="-1"
+            aria-label="Suggestions"
+            role="listbox"
+            className="datalist__ul"
+            {...props}
+          >
+            {items}
+          </ul>
         </div>
       )}
     </div>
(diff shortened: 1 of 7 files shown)


src/sidebar/components/hooks/use-element-should-close.js
```diff
@@ -42,7 +42,11 @@ function useElementShouldClose(closeableEl, isOpen, handleClose) {
       document.body,
       'focus',
       event => {
-        if (!closeableEl.current.contains(event.target)) {
+        // preact element or just a dom element?
+        const node = closeableEl.current.base
```
LMS007: This was not working out of the box on a top level preact component. From my research the "base" object is a solid test to check

src/sidebar/components/tag-editor.js
```diff
@@ -1,30 +1,28 @@
...
 const propTypes = require('prop-types');
-const { useMemo, useRef, useState } = require('preact/hooks');
+const { useRef, useState } = require('preact/hooks');
 
+const useElementShouldClose = require('./hooks/use-element-should-close');
 const { withServices } = require('../util/service-context');
-const SvgIcon = require('./svg-icon');
 
-// Global counter used to create a unique id for each instance of a TagEditor
-let datalistIdCounter = 0;
```
LMS007: The chrome fix is all removed now

src/sidebar/components/tag-editor.js
```diff
@@ -87,96 +95,90 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
-  const handleKeyPress = e => {
-    if (e.key === 'Enter' || e.key === ',') {
-      addTag();
-      // preventDefault stops the delimiter from being
-      // added to the input field
-      e.preventDefault();
-    }
-  };
-
   const handleOnInput = e => {
```
LMS007: handleOnInput is now very simple and is only checking for changes to the text value in the input field, the other stuff is now handled all in keyDown because we have to have keyDown handler to check for up/down key navigation anyway and it just made more sense to move the "Enter" AND "," checking there to add a tag.

LMS007: handleOnInput is now very simple and is only checking for changes to the text value in the input field, the other stuff is now handled all in handleKeyDown because we have to have a keyDown handler to check for up/down key selection navigation anyway, and it just made more sense to move the "Enter" AND "," delimiter checking there to add a new tag.

src/sidebar/components/tag-editor.js
```diff
@@ -87,96 +95,90 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
+      updateSuggestions();
+    }
+  };
+
+  /**
+   *  Callback when the user clicked one of the items in the suggestions list.
+   *  This will add a new tag.
+   */
+  const handleSelect = item => {
+    if (item) {
```
LMS007: we could get away w/o this condition, but addTag does not check for empty objects, I could go either way on this

src/sidebar/components/tag-editor.js
```diff
@@ -206,17 +208,25 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
-        type="text"
-      />
-      {suggestionsList()}
+      <span ref={closeWrapperRef}>
+        <input
+          onInput={handleOnInput}
+          onKeyDown={handleKeyDown}
+          onFocus={handleFocus}
+          ref={inputEl}
+          placeholder="Add new tags"
```
LMS007: I changed this value because when testing a11y it read this out with the word "ellipsis" and that sucks! I could not figure out how to get it to not read this so I figured I would just make it work for screen readers in the first place.

src/sidebar/components/tag-editor.js
```diff
@@ -206,17 +208,25 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
-        onInput={handleOnInput}
-        onKeyPress={handleKeyPress}
-        onKeyUp={handleKeyUp}
-        ref={inputEl}
-        placeholder="Add tags..."
-        className="tag-editor__input"
-        type="text"
-      />
-      {suggestionsList()}
+      <span ref={closeWrapperRef}>
```
LMS007: I needed an extra node here for the close ref so it would work outside both the datalist and the input

src/sidebar/components/tag-editor.js
robertknight: I guess if you wanted to have the visual and verbal text be different you'd need to set an `aria-label` property. The solution you've gone for here seems fine to me though.

src/styles/sidebar/components/datalist.scss
```diff
@@ -0,0 +1,41 @@
...
+
+.datalist {
+  color: var.$white;
+}
+
+.datalist__items {
+  position: absolute;
+  width: 100%;
+  padding: 0;
+  background-color: var.$grey-mid;
```
robertknight: I guess you are using a dark theme locally and have gone with colors to match that? So this fits in with whatever the user's configuration is locally, I was going to suggest that the default colors assume a light theme and then we use `prefers-color-scheme` media queries in CSS [to adapt to Dark Mode](https://css-tricks.com/dark-modes-with-css/). Does that sound reasonable?

We can do this adaptation on a per-component basis and potentially in future we could also use CSS variables for application-wide color changes.

src/sidebar/components/datalist.js
```diff
@@ -0,0 +1,73 @@
...
+    return list.map((item, index) => {
+      return (
+        <li
+          key={`datalist-${index}`}
+          className={classnames({
+            'is-selected': activeItem === index,
+          })}
+          onClick={() => {
+            onSelectItem(item);
+          }}
```
robertknight: We'll need appropriate ARIA properties to associate the datalist with the input and to indicate the currently selected item in the datalist. I don't know the exact details of the preferred pattern for that. Can you look into it?

src/sidebar/components/hooks/use-element-should-close.js
robertknight: I'm not sure if the `base` property of Preact components is considered part of its public API. I've asked on their Slack channel for confirmation.

In any case, as long as you're passing a ref to a DOM element in, I don't think this should be necessary. If you do think it is, can you clarify why?

src/sidebar/components/tag-editor.js
jaredpdesigns: I agree with @robertknight in terms of replacing the `placeholder` with something that doesn’t conflict with a11y issues.

src/styles/sidebar/components/datalist.scss
jaredpdesigns: @robertknight @LMS007 I could certainly help provide appropriate CSS for dark-theming—it’s already built into the Pattern Library stuff however

src/sidebar/components/datalist.js
LMS007: Sorry, didn't add too many details to this PR last night. I did want to mention I struggled a bit with this as it would end up reading everything in the list and got crazy when I had 50 items. I also observed that the original `<datalist>` was not being read by at least Chrome Vox so I would like to know what the aural  behavior should be here. I'll dig into this more!

src/styles/sidebar/components/datalist.scss
LMS007: Yes, I originally had all that in the Datalist, but moved it out as I needed the keyboard events to work off the input, but I think I can change the props around and move the logic for the carousel navigation into the DL component

LMS007: @jaredpdesigns , yes I think two styles here would be good, a light and dark theme. Pretty much the only css is a bit for the little arrow when the list is open and then the highlight colors for keyboard navigation and a slightly lighter color for mouse hover. But I'm open to anything here.

I also don't have any transitions. IF we added a fade transition, it gets . a bit more complicated with state and re-flow, and I could do that on a second PR.

LMS007:
- Added all the necessary tests (so far)
- Noticed one issue with hiding the list when the input is empty and fixed that
- I did not touch a11y _yet_, I added some suggestions in slack here https://hypothes-is.slack.com/archives/C1M8NH76X/p1578680573071300 to be further discussed

I did take a look at moving the up/down navigation into Datalist, but struggled because the events are triggered from the input and passing those events seems less clean than just passing which index is active. Furthermore, the parent component needs to know which index is active because it also listens for `Enter` key and needs to pick that out of the original list. It just seemed more correct to keep everything it the parent. Does that does make tag-editor heavy weight and datalist light, but I think that implementation proved to be easier at least the way I was thinking about it.


TODO: css theming.

LMS007:
The new css is designed to look more like the Menu component, but with some small changes. I decided not to use an actual icon but keep the css triangle as that was a bit more flexible and simple. I also just kept the triangle a solid color as that was easier to deal with.

Hover color in dark mode lightens the background, rather than darkest it. I think that is inverse of how our primary buttons work, but I find that more intuitive and it is how the original  datalist works too.

There is still not changes to a11y. I think about the only useful thing we could do here is my suggestion in slack #frontend to make the input value bind to the datalist selected value when navigating down down the list. In that case, a reader would inform the user what the option is, but I have not implemented that here.

datalist2-light

datalist2

jaredpdesigns:
@LMS007 I played around with styling a pseudo-element for the triangle as a CSS-only solution and like the results: https://codepen.io/jaredpdesigns/pen/WNbKERb?editors=1100#0

Pertinent CSS below:
``` scss
&:before {
  background-color: inherit;
  border: inherit;
  clip-path: polygon(0 0, 100% 0, 0% 100%, 0% 100%);
  content: "";
  display: inline-block;
  height: 0.5rem;
  left: 0.5rem;
  position: absolute;
  top: -0.3125rem;
  transform: rotate(45deg);
  width: 0.5rem;
}
```

LMS007:
Great suggestion, I like this css a lot more


Screen Shot 2020-01-14 at 5 05 10 PM

LMS007:
Added accessibility based on w3 model. "Combobox with Listbox Popup Examples"
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

This adds a few links that require a unique id both on the ul and then again on the li tags so there was a number of additional props needed to be passed down to the Datalist.

I tested this with ChromeVox and Narrator and it does readout the results on keyboard navigation which is what we wanted to achieve. I think this implementation is still a bit unique in that we clear out the suggestion on enter and I'm not sure how obvious that will be with a screen reader.  Other then that, I think we are a lot closer to a final product here.

src/sidebar/components/datalist.js
```diff
@@ -0,0 +1,100 @@
...
+
+Datalist.propTypes = {
+  /**
+   * The activeItem is highlighted.
+   */
+  activeItem: propTypes.number,
+  /**
+   * Optional unique HTML attribute id.
+   */
+  id: propTypes.string,
```
LMS007: In an effort to make Datalist somewhat decoupled from TagEditor, I decided to make id and itemPrefixId optional

src/styles/sidebar/components/datalist.scss
```diff
@@ -0,0 +1,75 @@
...
+    font-size: var.$touch-input-font-size;
+  }
+}
+
+.datalist__ul {
+  list-style: none;
+  padding: 0;
+  > li {
+    padding: 0.25em 1.5em 0.25em 1em;
+    border-left: 4px transparent solid;
```
robertknight: On mobile the list items are quite short vertically if you are trying to tap them. Can you add a `@media (pointer: coarse) { ... }` query to make sure that the items are at least `var.$touch-target-size` in height on mobile?

src/sidebar/components/datalist.js
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Custom datalist component. Use this in conjunction with an input field.
+ */
+
+function Datalist({
+  activeItem = -1,
+  id,
+  itemPrefixId,
+  list,
+  listFormatter = item => item,
```
robertknight: Note that since default arguments are evaluated on every function call in JavaScript, `listFormatter` will be different every time if not passed as a prop, this will invalidate the `useMemo` dependencies below. A way to avoid that is to hoist the value to a top-level constant:

```js
const defaultListFormatter = item => item;

function Datalist({
...,
listFormatter = defaultListFormatter,
...
})
```

src/sidebar/components/test/datalist-test.js
```diff
@@ -0,0 +1,165 @@
+import { mount } from 'enzyme';
+import { createElement } from 'preact';
+
+import Datalist from '../datalist';
+import { $imports } from '../tag-editor';
```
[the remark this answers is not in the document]
LMS007: good catch!

LMS007:
After doing some exploring about that cmd+option space issue, I decided to pivot the combobox model to one that fits our UI a bit better.

https://hypothes-is.slack.com/archives/C1M8NH76X/p1579814401056000

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -1,18 +1,22 @@
...
 describe('TagEditor', function() {
+  let container;
   let fakeTags = ['tag1', 'tag2'];
   let fakeTagsService;
   let fakeServiceUrl;
   let fakeOnEditTags;
 
   function createComponent(props) {
+    container = document.createElement('div');
+    document.body.appendChild(container);
```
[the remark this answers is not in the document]
LMS007: I think I overlooked this, I will use an array

robertknight:
This is superseded by https://github.com/hypothesis/client/pull/1712 and https://github.com/hypothesis/client/pull/1714. On that basis I guess we can close this now @LMS007?

PR #1655, opened 2020-01-09
Modify a mocking idiom that doesn't work with ES modules

robertknight:
We have an idiom in many tests where we import the "default export" from
a module and then access the `$imports` property on that
function/class/object in order to mock or unmock imports. For example:

```js
const Widget = require('../widget');

beforeEach(() => {
  Widget.$imports.$mock(...);
});

afterEach(() => {
  Widget.$imports.$restore();
});
```

This won't work when the module under test is converted to an ES module
because the `$imports` object will become a separate export from the
module rather than a property of the default export.

This PR changes the above to:

```js
const Widget = require('../widget');
const { $imports } = require('../widget');

beforeEach(() => {
  $imports.$mock(...);
});

afterEach(() => {
  $imports.$restore();
});
```

This applies to any test where:

1. The module under test has `module.exports = <some function or class>`
2. We are using `$imports.$mock(...)` and `$imports.$restore(...)` in the test

### DIFF
--- a/src/annotator/config/test/index-test.js
+++ b/src/annotator/config/test/index-test.js
@@ -1,4 +1,5 @@
 const configFrom = require('../index');
+const { $imports } = require('../index');
 
 describe('annotator.config.index', function() {
   let fakeSettingsFrom;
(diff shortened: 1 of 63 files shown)


robertknight:
> I see you also caught a few places we were not restoring or restoring for no reason

Indeed. It didn't happen in too many places but its an easy enough mistake to make at present.

PR #1656, opened 2020-01-09
Normalize module imports to an ES module-friendly format

robertknight:
This PR prepares for converting the client to ES module syntax by normalizing some of our existing imports into a form that is easier to map to ES module imports.

1. Modify several imports to conform to one of two forms that easily maps to ES imports:

```js
// Maps to `import foo from "foo"` or `import * as foo from "foo"` depending
// on what module `./foo` exports.
const foo = require('./foo'); 
// Maps to `import { bar, baz } from "foo"`
const { bar, baz } = require('./foo'); 
```

2. Move some imports to the top level of the file (eg. not in a conditional or function) because ES module imports must appear at the top level. This change mostly affects `src/sidebar/index.js`.
3. I changed the way that some store modules import functions from other store modules to a version that will work after the conversion. My plan is that store modules will default-export a single object rather than having multiple exports
4. When testing out ES modules locally I ran into a conflict between the Babel plugin that handles `async` functions and the Babel plugin that handles `// @ngInject` comments in `src/sidebar/services/oauth-auth.js`. I haven't yet determined exactly why the conflict only happens after converting to ES imports, but since it only happened with this one module I've worked around it for now by replacing the `// @ngInject` comment with the code that would normally be generated for it (`auth.$inject = ...`)

### DIFF
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -4,8 +4,7 @@ const { fetchConfig } = require('./util/fetch-config');
 const serviceConfig = require('./service-config');
 const { jsonConfigsFrom } = require('../shared/settings');
 const crossOriginRPC = require('./cross-origin-rpc.js');
-
-let sentry;
+const sentry = require('./util/sentry');
 
 // Read settings rendered into sidebar app HTML by service/extension.
 const appConfig = jsonConfigsFrom(document);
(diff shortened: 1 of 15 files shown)


src/sidebar/index.js
```diff
@@ -110,6 +109,76 @@ function sendPageView(analytics) {
   analytics.sendPageView();
 }
 
+// Preact UI components that are wrapped for use within Angular templates.
```
LMS007: Much more readable

PR #1682, opened 2020-01-15
Convert text highlighter to JS

robertknight:
As part of several ongoing/upcoming tasks we are going to be touching the code that highlights annotated text in the DOM. In preparation this PR simplifies the existing code by removing an unused abstraction layer over different highlighter implementations and converting the actual highlighter implementation in src/annotator/highlighter/dom-wrap-highlighter/index.coffee and associated tests from CoffeeScript to JS.

This PR also eliminates a console warning about a `overlay_highlighter` feature flag when the client loads.


src/annotator/highlighter.js
```diff
@@ -0,0 +1,75 @@
...
+  return nodes
+    .wrap(hl)
+    .parent()
+    .toArray();
+}
+
+/**
+ * Remove highlights from a range previously highlighted with `highlightRange`.
+ *
+ * @param highlights - The highlight elements returned by `highlightRange`
```
LMS007: perhaps `{HTMLElement[]} highlights`

PR #1683, opened 2020-01-15
Fix useElementShouldClose so it may work on preact components

LMS007:
The diff is a little hard to follow, but all I did for the test.js file was wrap the entire existing tests in a `forEach` with 2 items so that now all the tests are run twice with:

1. createComponent
2. createCompoundComponent

Where the difference is one attaches useElementShouldClose to the `<div>` and the second one attaches it to `<FakeCustomComponent>`

I confirmed all the new tests fail w/o the `getCurrentNode` fix applied, but then pass when that is used.

In terms of how to test for a regular html node vs preact component. I did some searching and it seems that the existence of "base" prop is a pretty safe bet, but I'm open to other ideas

### DIFF
--- a/src/sidebar/components/hooks/use-element-should-close.js
+++ b/src/sidebar/components/hooks/use-element-should-close.js
@@ -24,6 +24,21 @@ export default function useElementShouldClose(
   isOpen,
   handleClose
 ) {
+  /**
+   *  Helper to return the underlying node object whether
+   *  `closeableEl` is attached to an HTMLNode or Preact component.
+   *
+   *  @param {Preact ref} closeableEl
+   *  @returns {HTMLNode}
+   */
+
+  const getCurrentNode = closeableEl => {
+    // if base is present, assume its a preact component
+    return closeableEl.current.base
+      ? closeableEl.current.base
+      : closeableEl.current;
+  };
+
   useEffect(() => {
     if (!isOpen) {
       return () => {};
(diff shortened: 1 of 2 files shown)


src/sidebar/components/hooks/use-element-should-close.js
```diff
@@ -24,6 +24,21 @@ export default function useElementShouldClose(
...
+  /**
+   *  Helper to return the underlying node object whether
+   *  `closeableEl` is attached to an HTMLNode or Preact component.
+   *
+   *  @param {Preact ref} closeableEl
+   *  @returns {HTMLNode}
+   */
+
+  const getCurrentNode = closeableEl => {
+    // if base is present, assume its a preact component
```
robertknight: The `base` property is undocumented, although I have been told by the devs that it is considered a public API ([Slack comment](https://preact.slack.com/archives/C3NMBSJKH/p1578589918070400)). Since this is a Preact-specific API (as opposed to being part of the core APIs that Preact/React share), I suggest adding a short note about what exactly `base` refers to (the first rendered DOM child node of the component).

LMS007:
@robertknight I believe throwing a `new Error('...')` inside of that listener callback handler causes some problems because its async relative to the dispatched mouse events that could trigger it

I can't figure out how to wrap a try/catch around a test that can actually catch that exception. So it ends up being an uncaught exception and the test fails no matter what.

My other idea is to just console.error, but I wanted to double check you with you.

robertknight:
> I can't figure out how to wrap a try/catch around a test that can actually catch that exception. So it ends up being an uncaught exception and the test fails no matter what.

When did you want to throw an exception? If some of the criteria I mentioned are not met? If the error is functioning as an assert (against programmer error) I think you can afford not to test it.

PR #1704, opened 2020-01-21
Move `export` keyword to declaration site

robertknight:
Improve code style consistency by always putting the `export` keyword
at the site of the var/function/class declaration where possible.


robertknight:
> So observable.js is the only outlier now?

Yes. I suppose that could be written using `export ... from ...` syntax.

Issue #1629, opened 2019-12-30
Keyboard focus on menu arrow should work


LMS007:
Tabbing to the menu arrow button should show the focus ring, but the button has a 0x0 size. This does not allow for the blue halo on keyboard navigation in spite of the mixin being used

```
.menu__toggle {
  @include focus.outline-on-keyboard-focus;
```

![focus](https://user-images.githubusercontent.com/3939074/71601568-d4e91800-2b08-11ea-8ed2-461dd8f4a5b2.gif)

![Screen Shot 2019-12-30 at 1 33 42 PM](https://user-images.githubusercontent.com/3939074/71601611-06fa7a00-2b09-11ea-99f7-d45d7b5fe231.png)

robertknight:
Good spot! I just noticed this when trying to tab around the interface just now.

PR #1711, opened 2020-01-24
Fix Menu tab focus on keyboard navigation

LMS007:
This fix was a bit tricky land. My goal was to be non-invasive and make what we had work so it did not require large changes to other parts of the code that used <Menu>. In short, the keyboard navigation now correctly shows the focus state of the open menu button for publishing an annotation.

fixes https://github.com/hypothesis/client/issues/1629
halo

### DIFF
--- a/src/sidebar/components/annotation-publish-control.js
+++ b/src/sidebar/components/annotation-publish-control.js
@@ -58,9 +58,7 @@ function AnnotationPublishControl({
       <div
         className="annotation-publish-control__btn-dropdown-arrow-indicator"
         style={applyTheme(themeProps, settings)}
-      >
-        <div>▼</div>
-      </div>
+      />
     </div>
   );
 
--- a/src/sidebar/components/menu.js
+++ b/src/sidebar/components/menu.js
@@ -126,14 +126,19 @@ export default function Menu({
         onClick={toggleMenu}
         title={title}
       >
-        {label}
-        {menuIndicator && (
-          <span
-            className={classnames('menu__toggle-arrow', isOpen && 'is-open')}
-          >
-            <SvgIcon name="expand-menu" className="menu__toggle-icon" />
-          </span>
-        )}
+        <span
+          // wrapper is needed to serve as the flex layout for the label and indicator content.
+          className="menu__toggle-wrapper"
+        >
+          {label}
+          {menuIndicator && (
+            <span
+              className={classnames('menu__toggle-arrow', isOpen && 'is-open')}
+            >
+              <SvgIcon name="expand-menu" className="menu__toggle-icon" />
+            </span>
+          )}
+        </span>
       </button>
       {isOpen && (
         <Fragment>
--- a/src/styles/sidebar/components/annotation-publish-control.scss
+++ b/src/styles/sidebar/components/annotation-publish-control.scss
@@ -16,6 +16,7 @@
     $arrow-indicator-width: 26px;
 
     height: $height;
+    display: flex;
     position: relative;
     margin-right: 1em;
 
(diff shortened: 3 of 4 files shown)


src/sidebar/components/annotation-publish-control.js
```diff
@@ -58,9 +58,7 @@ function AnnotationPublishControl({
       <div
         className="annotation-publish-control__btn-dropdown-arrow-indicator"
         style={applyTheme(themeProps, settings)}
-      >
-        <div>▼</div>
```
LMS007: I noticed that ChromeVox screen reader was reading out this char as "down triangle" and in Mac voice over it was "Down pointy black point button" or something like that, quite funny! So I moved it to css ::after so  now is correctly reads out "Change annotation sharing setting" along with menu's name as its collapsed state.

src/sidebar/components/menu.js
```diff
@@ -126,14 +126,16 @@ export default function Menu({
...
       >
-        {label}
-        {menuIndicator && (
-          <span
-            className={classnames('menu__toggle-arrow', isOpen && 'is-open')}
-          >
-            <SvgIcon name="expand-menu" className="menu__toggle-icon" />
-          </span>
-        )}
+        <span className="menu__toggle-wrapper">
```
LMS007: This was the magic sauce, I needed to move display: flex off `.menu__toggle` so it would behave like a flex item, not a flex container. But we need still need a flex lay out for both the label and menuIndicator stuff so wrapping it in another container was required.

lyzadanger: Given that this explanation is crucial for understanding the changes in this PR, it might be a good idea to put some commenting around the CSS for this wrapper `span` to elucidate future devs, or create a commit body with the same information so that it will be available in the future.

src/styles/sidebar/components/annotation-publish-control.scss
```diff
@@ -38,22 +39,20 @@
...
-      padding-right: $arrow-indicator-width + 8px;
+      border-top-right-radius: 0;
+      border-bottom-right-radius: 0;
     }
 
     // dropdown arrow which reveals the button's associated menu
     // when clicked
     &-dropdown-arrow {
-      position: absolute;
+      display: flex;
```
LMS007: changed from absolute position to a flex layout that centers everything

src/styles/sidebar/components/annotation-publish-control.scss
```diff
@@ -90,15 +89,17 @@
...
       &-indicator {
         color: $text-color;
-        position: absolute;
         left: 0px;
         right: 0px;
         top: 0px;
         bottom: 0px;
         line-height: $height;
         text-align: center;
+        width: 100%;
```
LMS007: Had to add this to make sure it filled the parent container

robertknight:
Hi @LMS007,

Can you explain why the previous behavior was happening and, at a high level, why this change resolves it?

src/sidebar/components/menu.js
LMS007: Good idea, I'll add a comment in code, its a bit harder to fid commit logs

LMS007:
@robertknight 

I believe what was happening is as you change focus, the button.menu__toggle tag was rendering at 0,0 size because its content was position: absolute.  Specifically these tags were "absolute"

1. .annotation-publish-control__btn-dropdown-arrow
2. .annotation-publish-control__btn-dropdown-arrow-indicator

I did leave position:absolute on the separator so it would not take up space and affect the centering of the arrow, but this PR changes the layout of those two tags and allows <button> to have a size.

What it looked like before:

Screen Shot 2020-01-24 at 9 56 16 AM


@lyzadanger  I'm gonna make a separate ticket for the <button> to <Button> fix if thats okay with you?

src/sidebar/components/menu.js
LMS007: I should be a bit more clear, it was part of the magic sauce! W/o this change, the button didn't fill the parent container and match the height of the publish button. It was actually not even apparent until I started scaling the browser with the CMD+ and I noticed it was off by 1 pixel in height. This gif I made shows the problem more dramatically as I increase the height from 35 or 65 
layout

lyzadanger:
> @lyzadanger I'm gonna make a separate ticket for the
> to fix if thats okay with you?

Ticket already exists: https://github.com/hypothesis/client/issues/1659

Issue #1690, opened 2020-01-17
Can't add new tags in Edge v44


LMS007:
When typing a tab in edge, it adds the tag on the keypress (or perhaps keydown) event and commits it to the tag list. This means that only 1 char tags can be added. I have not testing this is newer edge versions, but 44 is still being maintained by MS.

PR #1714, opened 2020-01-24
Improved tag editor (Part 2/2)

LMS007:
- This updates the TagEditor to use the new Datalist component and also adds the a11y combobox model found here https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
- Also fixes https://github.com/hypothesis/client/issues/1690 (Can't add new tags in Edge v44)

Relevant code is pulled from https://github.com/hypothesis/client/pull/1653 verbatim

### DIFF
--- a/src/sidebar/components/tag-editor.js
+++ b/src/sidebar/components/tag-editor.js
@@ -1,31 +1,38 @@
 import { createElement } from 'preact';
-import { useMemo, useRef, useState } from 'preact/hooks';
+import { useRef, useState } from 'preact/hooks';
 import propTypes from 'prop-types';
 
 import { withServices } from '../util/service-context';
 
+import AutocompleteList from './autocomplete-list';
 import SvgIcon from './svg-icon';
+import useElementShouldClose from './hooks/use-element-should-close';
 
 // Global counter used to create a unique id for each instance of a TagEditor
-let datalistIdCounter = 0;
+let tagEditorIdCounter = 0;
 
 /**
  * Component to edit annotation's tags.
+ *
+ * Component accessibility is modeled after "Combobox with Listbox Popup Examples" found here:
+ * https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
  */
 
 function TagEditor({ onEditTags, tags: tagsService, tagList }) {
   const inputEl = useRef(null);
-  const [showSuggestions, setShowSuggestions] = useState(false);
-  const [suggestionsFiltered, setSuggestionsFiltered] = useState([]); // For Chrome only
-  const [datalistId] = useState(() => {
-    ++datalistIdCounter;
-    return `tag-editor-datalist-${datalistIdCounter}`;
+  const [suggestions, setSuggestions] = useState([]);
+  const [activeItem, setActiveItem] = useState(-1); // -1 is unselected
+  const [suggestionsListOpen, setSuggestionsListOpen] = useState(false);
+  const [tagEditorId] = useState(() => {
+    ++tagEditorIdCounter;
+    return `tag-editor-${tagEditorIdCounter}`;
   });
 
-  const isChromeUA = !!window.navigator.userAgent.match(/Chrome\/[.0-9]*/);
-  // Suggestion limiter to speed up the performance in Chrome
-  // See https://github.com/hypothesis/client/issues/1606
-  const chromeSuggestionsLimit = 20;
+  // Set up callback to monitor outside click events to close the AutocompleteList
+  const closeWrapperRef = useRef();
+  useElementShouldClose(closeWrapperRef, suggestionsListOpen, () => {
+    setSuggestionsListOpen(false);
+  });
 
   /**
    * Helper function that returns a list of suggestions less any
@@ -45,12 +52,25 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
     return suggestionsSet.sort();
   };
 
-  // List of suggestions returned from the tagsService
-  const suggestions = useMemo(() => {
-    // Remove any repeated suggestions that are already tags.
-    // Call filter with an empty string to return all suggestions
-    return removeDuplicates(tagsService.filter(''), tagList);
-  }, [tagsService, tagList]);
+  /**
+   * Get a list of suggestions returned from the tagsService
+   * reset the activeItem and open the AutocompleteList
+   */
+  const updateSuggestions = () => {
+    const value = inputEl.current.value.trim();
+    if (value === '') {
+      // If there is no input, just hide the suggestions
+      setSuggestionsListOpen(false);
+    } else {
+      // Call filter() with a query value to return all matching suggestions.
+      const suggestions = tagsService.filter(value);
+      // Remove any repeated suggestions that are already tags
+      // and set those to state.
+      setSuggestions(removeDuplicates(suggestions, tagList));
+      setSuggestionsListOpen(true);
+    }
+    setActiveItem(-1);
+  };
 
   /**
    * Handle changes to this annotation's tags
@@ -77,8 +97,8 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
    * Adds a tag to the annotation equal to the value of the input field
    * and then clears out the suggestions list and the input field.
    */
-  const addTag = () => {
-    const value = inputEl.current.value.trim();
+  const addTag = newTag => {
+    const value = newTag.trim();
     if (value.length === 0) {
       // don't add an empty tag
       return;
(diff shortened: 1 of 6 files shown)


src/sidebar/components/tag-editor.js
```diff
@@ -88,98 +106,97 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
+        changeSelectedItem(1);
+        e.preventDefault();
+        break;
+      case 'Enter':
+      case ',':
+        if (activeItem === -1) {
+          // nothing selected, just add the typed text
+          addTag(inputEl.current.value);
+        } else {
+          addTag(suggestions[activeItem]);
```
LMS007: Looks like this is getting flagged as not covered in tests. I'll take a closer look, might need to add 1 test here.

src/sidebar/components/tag-editor.js
```diff
@@ -1,31 +1,38 @@
...
+  const [tagEditorId] = useState(() => {
+    ++tagEditorIdCounter;
+    return `tag-editor-${tagEditorIdCounter}`;
   });
 
-  const isChromeUA = !!window.navigator.userAgent.match(/Chrome\/[.0-9]*/);
-  // Suggestion limiter to speed up the performance in Chrome
-  // See https://github.com/hypothesis/client/issues/1606
-  const chromeSuggestionsLimit = 20;
+  // Set up callback to monitor outside click events to lose the Datalist
```
robertknight:
```suggestion
  // Set up callback to monitor outside click events to close the Datalist
```

src/sidebar/components/tag-editor.js
```diff
@@ -1,31 +1,38 @@
...
 function TagEditor({ onEditTags, tags: tagsService, tagList }) {
   const inputEl = useRef(null);
-  const [showSuggestions, setShowSuggestions] = useState(false);
-  const [suggestionsFiltered, setSuggestionsFiltered] = useState([]); // For Chrome only
-  const [datalistId] = useState(() => {
-    ++datalistIdCounter;
-    return `tag-editor-datalist-${datalistIdCounter}`;
+  const [suggestions, setSuggestions] = useState([]);
+  const [activeItem, setActiveItem] = useState(-1); // -1 is unselected
+  const [isOpen, setOpen] = useState(false);
```
robertknight: I find `isOpen` slightly ambiguous in this context. `suggestionsListOpen` perhaps?

src/sidebar/components/tag-editor.js
```diff
@@ -88,98 +106,97 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
+      case 'ArrowUp':
+        changeSelectedItem(-1);
+        e.preventDefault();
+        break;
+      case 'ArrowDown':
+        changeSelectedItem(1);
+        e.preventDefault();
+        break;
+      case 'Enter':
+      case ',':
```
robertknight: This is something for a follow-up PR, but do you think we should do anything in response to the Tab key? I saw this was discussed in #support as a change in behavior from the previous implementation.

src/sidebar/components/tag-editor.js
```diff
@@ -207,17 +224,37 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
+        <input
+          onInput={handleOnInput}
+          onKeyDown={handleKeyDown}
+          aria-controls={`${tagEditorId}-datalist`}
+          onFocus={handleFocus}
+          ref={inputEl}
+          placeholder="Add new tags"
+          className="tag-editor__input"
+          type="text"
+          autoComplete="off"
```
robertknight: The capitalization of `autoComplete` here confused me but it turns out that at least in React (I haven't tested whether Preact accepts both), [it is required](https://stackoverflow.com/questions/37503656/react-doesnt-render-autocomplete-off/37503673).

LMS007: I don't know if it's required or not, but I was getting Chrome prompts from my previously typed tags and I can't see people wanting to use that since thats exactly what the suggestions list does already and the UI looks totally different. I just want it disabled.

src/sidebar/components/tag-editor.js
```diff
@@ -1,31 +1,38 @@
...
 
 // Global counter used to create a unique id for each instance of a TagEditor
-let datalistIdCounter = 0;
+let tagEditorIdCounter = 0;
 
 /**
  * Component to edit annotation's tags.
+ *
+ * Component accessibility is modeled after "Combobox with Listbox Popup Examples" found here:
+ * https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
```
robertknight: 👍 - Let's make sure we add similar comments on any other complex widgets we implement from these guidelines.

src/sidebar/components/tag-editor.js
```diff
@@ -45,12 +52,23 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
+   * reset the activeItem and open the Datalist
+   */
+  const updateSuggestions = () => {
+    const value = inputEl.current.value.trim();
+    if (value === '') {
+      // If there is no input, just hide the suggestions
+      setSuggestionsListOpen(false);
+    } else {
+      // Remove any repeated suggestions that are already tags.
+      // Call filter with an empty string to return all suggestions
```
robertknight: The "Call filter" comment refers to just a small part of the next line, which slightly confused me. It would help to separate out the `tagsService.filter` call onto a separate line which the comment can refer to on its own.

src/sidebar/components/tag-editor.js
```diff
@@ -88,98 +106,97 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
-        {showSuggestions &&
-          optionsList.map(suggestion => (
-            <option key={suggestion} value={suggestion} />
-          ))}
-      </datalist>
-    );
+  /**
+   * Keydown handler for keyboard navigation of the suggestions list
+   * and when the user presses "Enter" or ","" to add a new typed item not
+   * found in the suggestions list
```
robertknight: These comments describing the intended behavior from the user's point of view are helpful.

src/sidebar/components/tag-editor.js
```diff
@@ -88,98 +106,97 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
+        }
+        e.preventDefault();
+        break;
+    }
   };
 
+  // Only add the aria-activedescendant prop when there is an item selected.
+  // -1 value means no item selected.
+  const activeDescendant =
+    activeItem >= 0 ? `${tagEditorId}-datalist-item-${activeItem}` : false;
```
robertknight: In testing I found that VoiceOver with Safari and Chrome did not read the suggestion when pressing the down/up arrow for the first time to open the list, but did read subsequent suggestions after that. I changed the false branch here so that the `aria-activedescendant` attribute is set when there is no active item, but with an empty/missing value, and that resolved the problem.

```suggestion
    activeItem >= 0 ? `${tagEditorId}-datalist-item-${activeItem}` : "";
```

src/sidebar/components/tag-editor.js
```diff
@@ -207,17 +224,38 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
-      {suggestionsList()}
+      <span
+        id={tagEditorId}
+        className="tag-editor__combobox-wrapper"
+        ref={closeWrapperRef}
+        role="combobox"
+        aria-controls={`${tagEditorId}-datalist`}
+        aria-expanded={suggestionsListOpen.toString()}
+        aria-owns={`${tagEditorId}-datalist`}
+        aria-haspopup="true"
```
robertknight: The value of `aria-haspopup` here should match the role of the popup container (`"list"` in this case) in ARIA 1.1. See https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup. The value "true" is treated as if it were "menu" for backwards compatibility.

src/sidebar/components/tag-editor.js
```diff
@@ -207,17 +224,38 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
-        className="tag-editor__input"
-        type="text"
-      />
-      {suggestionsList()}
+      <span
+        id={tagEditorId}
+        className="tag-editor__combobox-wrapper"
+        ref={closeWrapperRef}
+        role="combobox"
+        aria-controls={`${tagEditorId}-datalist`}
```
robertknight: The `aria-controls` attribute should be set on the input field rather than the wrapper, per https://www.w3.org/TR/wai-aria-1.1/#combobox and the examples at https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -66,118 +95,166 @@ describe('TagEditor', function() {
     });
   });
 
-  it("creates a `list` prop on the input that matches the datalist's `id`", () => {
+  it('generates an ordered datalist containing the array values returned from filter()', () => {
```
robertknight: It wasn't immediately obvious to me what `filter()` refers to here. I presume you mean the tags service?

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -66,118 +95,166 @@ describe('TagEditor', function() {
...
+    it('does not render duplicate suggestions', () => {
+      // `tag3` supplied in the `tagList` will be a duplicate value relative
+      // with the fakeTagsService.filter result above.
+      const wrapper = createComponent({
+        editMode: true,
+        tagList: ['tag1', 'tag2', 'tag3'],
+      });
+      wrapper.find('input').instance().value = 'non-empty';
+      typeInput(wrapper);
+      assert.equal(wrapper.find('Datalist').prop('list').length, 1);
```
robertknight: This could be compressed into one assert with `assert.deepEqual(wrapper.find('Datalist').prop('list'), ['tag4'])`. The advantage of using `assert.deepEqual` to check both the length and contents of an array or object vs. a series of assets is that you get a more useful error if the check fails.

src/sidebar/components/tag-editor.js
```diff
@@ -1,31 +1,38 @@
 import { createElement } from 'preact';
-import { useMemo, useRef, useState } from 'preact/hooks';
+import { useRef, useState } from 'preact/hooks';
 import propTypes from 'prop-types';
 
 import { withServices } from '../util/service-context';
 
+import Datalist from './datalist';
```
robertknight: Re-reading this code makes me wonder whether `Datalist` is still the right name for this component. I don't think it is a completely drop-in substitute for `<datalist>` any more is it (at least, not without making changes to the component containing the input field?). Perhaps we ought to rename it `Listbox` (to match the a11y role name) or `AutocompleteList`.

LMS007:
> Hi Kyle,
> 
> The code looks great. I did find a couple of minor a11y issues where ARIA attributes that should be set on the `<input>` according to the ARIA 1.1 combobox [examples](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html) and [spec](https://www.w3.org/TR/wai-aria-1.1/#combobox) were set on the wrapper element instead. Changing this caused VoiceOver to read out suggestions as "_tag name_, text completion selected" rather than the more generic "_tag name_, clickable".
> 
> I also found an issue where it is necessary, at least on macOS with VoiceOver, to make sure that the `aria-activedescendant` attribute is always present even if no suggestion is selected, to ensure that the first suggestion is read out when the user presses the down/up arrows after the list first appears.
> 
> Although not introduced in this PR, it is good to see support for dark mode being introduced. I think we should make a plan to support that fully in the client in future.
> 
> Once the ARIA issues mentioned above are resolved, this is good to merge from my perspective. Please take another look over the commit messages before merging in case anything in them needs. updating.

The only aria attr I could find that looked incorrect was `aria-controls` and that needed to be on the input rather than its wrapper. Is that what are you suggesting? I moved that, but I'm not getting the same readout from VoiceOver that you are both from the w3.org example and the TagEditor. Maybe my settings are different?

I also did not experience the first suggestion not being not read out because aria-activedescendant was missing, but I'll set the value to the string "false" when its on the -1 index in any case.

src/sidebar/components/tag-editor.js
LMS007: Good idea!

src/sidebar/components/tag-editor.js
LMS007: Interested, I totally do not experience the same behavior, but I'll make that change anyway, It would be good to know why we're seeing a difference :\

src/sidebar/components/tag-editor.js
LMS007: The role of the popup container is "listbox" I'm assuming that is what you mean. I'll make that change

src/sidebar/components/tag-editor.js
LMS007: yep, I saw that! good catch

src/sidebar/components/test/tag-editor-test.js
LMS007: 👍

src/sidebar/components/tag-editor.js
LMS007: I like 
AutocompleteList

LMS007:
- Renamed Datalist to AutocompleteList
- Added last few fixes suggestion in PR and cleaned up commit
- Rebased off master

src/styles/sidebar/sidebar.scss
```diff
@@ -37,7 +37,7 @@
 @use './components/annotation-thread';
 @use './components/annotation-user';
 @use './components/button';
-@use './components/datalist';
+@use './components/autocomplete-list';
```
robertknight: This should be moved up above the "button" import here after the name change.

src/sidebar/components/tag-editor.js
```diff
@@ -88,98 +108,99 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
...
+          addTag(inputEl.current.value);
+        } else {
+          addTag(suggestions[activeItem]);
+        }
+        e.preventDefault();
+        break;
+    }
   };
 
+  // Only add the aria-activedescendant prop when there is an item selected.
```
robertknight: This comment is not quite accurate any more, since we are _always_ adding the prop, but setting it to an empty string if there is no active item.

Issue #1716, opened 2020-01-25
Unable to align text directionality to RTL


nadavkav:
We are using Hebrew to write annotations, and we are unable to set text alignment to "right" and also unable to set text directionality to RTL.

![image](https://user-images.githubusercontent.com/231168/73119819-f60a2080-3f6f-11ea-8703-ef6f9150ea18.png)

Please see if you can add these buttons to the annotation editor.

robertknight:
Great, thanks for confirming! We'll look into integrating this change into the Hypothesis client (and therefore also the Hypothesis LMS app) soon.

robertknight:
This issue has been resolved as per the above comments in v1.298.0 of the client (go to Help => About this version). That is available now in the bookmarklet, LMS integration, and at https://via.hypothes.is. It will appear in the browser extension shortly.

Please let me know if you encounter any issues with it.

PR #1717, opened 2020-01-27
Improve handling of text in RTL languages

robertknight:
In response to #1716, improve display of user-controlled text in Right-To-Left (RTL) languages (eg. Hebrew, Arabic) by applying the `dir="auto"` attribute to:
    
- Input fields (annotation body editor, search bar)
- The rendered annotation body's container element
- The rendered annotation quote (ie. the `<blockquote>` element)
    
This will cause these elements to switch their "base" direction for text layout [1] to RTL or LTR automatically depending on the content [2]. Sequences of characters are rendered in an appropriate order for the specific characters, but the _base_ direction is important when handling text that mixes characters with different directionality amongst other things (see [1] for examples).
    
Fixes https://github.com/hypothesis/client/issues/1716
    
[1] https://www.w3.org/International/articles/inline-bidi-markup/uba-basics
[2] https://www.w3.org/International/questions/qa-html-dir

### DIFF
--- a/src/sidebar/components/annotation-quote.js
+++ b/src/sidebar/components/annotation-quote.js
@@ -26,6 +26,7 @@ function AnnotationQuote({ annotation, settings = {} }) {
       >
         <blockquote
           className="annotation-quote__quote"
+          dir="auto"
           style={applyTheme(['selectionFontFamily'], settings)}
         >
           {quote(annotation)}
--- a/src/sidebar/components/markdown-editor.js
+++ b/src/sidebar/components/markdown-editor.js
@@ -283,6 +283,7 @@ export default function MarkdownEditor({ onEditText = () => {}, text = '' }) {
       ) : (
         <textarea
           className="markdown-editor__input"
+          dir="auto"
           ref={input}
           onClick={e => e.stopPropagation()}
           onKeydown={handleKeyDown}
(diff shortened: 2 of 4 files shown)


robertknight:
To see the effects of this, compare the results on master vs this branch of:

- Pasting a snippet of English text vs Hebrew text into an empty annotation body edit field. The textfield direction should switch depending on the directionality of the first character
- Rendering of an annotation body that combines LTR and RTL characters (eg. "<span dir="auto">בן כמה רוברט?</span>" which incorrectly renders as "בן כמה רוברט?" on master)
- Annotating a snippet of text that mixes LTR and RTL characters

robertknight:
> Rob, this seems like it could affect the tag editor too right?

Yes, indeed. It might be best to avoid making that change in this PR to avoid a conflict with your one. Do you want to add it to your tag editor PR?

LMS007:
> > Rob, this seems like it could affect the tag editor too right?
> 
> Yes, indeed. It might be best to avoid making that change in this PR to avoid a conflict with your one. Do you want to add it to your tag editor or created a separate PR for it after the tag editor update is merged?

I'll add it in mine current PR, thanks

robertknight:
I'm going to go ahead and merge this, we can revert easily if there are unexpected issues.

judell:
❤️

PR #1740, opened 2020-02-01
Improve a11y testing for AutocompleteList and TagEditor

LMS007:
This diff is a bit ugly because I moved some of the tests around, but there are far less changes in the test file than what the diff indicates, sorry about that. I wanted to move all the a11y tests and related a11y things to the end of the test file.

**Here is what this covers**

- Adds axe tests for both TagEditor and AutocompleteList
- I reset activeItem to -1 after adding a tag. This was not being done before, but I'm now seeing a problem with the screen reader because it still thinks the active index is set. This was not an issue previously because of how the AutocompleteList was being shown/hidden via DOM, where as now, its via `display:none` (read below for explanation)
- Some minor fixes ups to get those tests to pass (as follows)
  - grey-5 is very very slightly darker, I decided to make this across the whole app because we use that against the grey-0 background often enough. This meets the 4.5:1 contrast req.
  - Needed to add `aria-label` - I was allowing placeholder to be the label, but a11y lint didn't like that, probably best to just add the label. I decided to call it "Add tag field" but I'm open to other suggestions. I liked the idea of adding the "field" suffix to reinforce to the user what the thing is.
  - Needed to do full DOM testing of the suggestions list in tests. This is one of those components where you can't just allow the mockImportedComponents to mock everything so I'm importing AutocompleteList in the TagEditor test and its getting fully rendered, otherwise it won't pass axe tests. I suspect we may encounter this again.
AutocompleteList can't be done with an if condition on the items anymore. This is one of those sad moments where react/preact is too good for its own good. I really like the idea of hiding DOM elements but it breaks the a11y tests when the id counterparts of the aria tags are not in the DOM! So, I followed what they did on w3.org and added a hidden style with `display:none`

@jaredpdesigns I would use some feedback on the grey-5, the change is so small its not even worth a screen shot, but I wanted to make sure it fits with the rest of our grays and / or do we want to approach this contrast thing more methodically?
```
// Interim color variable for migration purposes.
// Used as very-subtle background color for form fields. $grey-1 is too dark.
$grey-0: #fafafa;
$grey-1: #f2f2f2;
$grey-2: #ececec;
$grey-3: #dbdbdb;
$grey-4: #a6a6a6;
// Interim color variable for migration purposes, as the step between `$grey-4`
// and `$grey-5` is large. Represents `base-semi` in proposed future palette,
// minus blue tint.
$grey-semi: #9c9c9c;
$grey-5: #737373;
// Interim color variable for migration purposes, as the step between `$grey-5`
// and `$grey-6` is large. Represents `base-mid` in proposed future palette,
// minus blue tint.
$grey-mid: #595959;
$grey-6: #3f3f3f;
$grey-7: #202020;
$black: #000 !default;```

### DIFF
--- a/src/sidebar/components/test/tag-editor-test.js
+++ b/src/sidebar/components/test/tag-editor-test.js
@@ -2,6 +2,7 @@ import { mount } from 'enzyme';
 import { createElement } from 'preact';
 import { act } from 'preact/test-utils';
 
+import AutocompleteList from '../autocomplete-list';
 import TagEditor from '../tag-editor';
 import { $imports } from '../tag-editor';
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/sidebar/components/test/tag-editor-test.js [hunk 1]
```diff
@@ -408,11 +351,98 @@ describe('TagEditor', function() {
...
+      // initial aria-activedescendant value is "" when index is -1
+      assert.equal(wrapper.find('input').prop('aria-activedescendant'), '');
+      // 2 suggestions: ['tag3', 'tag4'];
+      navigateDown(wrapper); // press down once
+      checkAttributes(wrapper);
+      navigateDown(wrapper); // press down again once
+      checkAttributes(wrapper);
+    });
+
+    it(
```
LMS007: From here down is the  only change in this section

src/styles/variables.scss
```diff
@@ -20,7 +20,7 @@ $grey-4: #a6a6a6;
 // minus blue tint.
 $grey-semi: #9c9c9c;
 
-$grey-5: #767676;
+$grey-5: #737373;
```
LMS007: Meets 4.5:1 with `grey-0`

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -6,6 +6,7 @@ import TagEditor from '../tag-editor';
 import { $imports } from '../tag-editor';
 
 import { checkAccessibility } from './accessibility';
+import AutocompleteList from '../autocomplete-list';
```
robertknight: Can you group this with the other imports from the `src/sidebar/components`, rather than `src/sidebar/components/test` directory above? Imports are grouped in three categories: 1) Vendor/third-party imports, 2) Imports from other directories than the current module, 3) Imports from the same directory as the current module.

nb. It is still on my TODO list to automate this.

src/sidebar/components/test/tag-editor-test.js
```diff
@@ -81,6 +82,10 @@ describe('TagEditor', function() {
       store: sinon.stub(),
     };
     $imports.$mock(mockImportedComponents());
+    // create a full dom tree for a11y testing
+    $imports.$mock({
+      './autocomplete-list': AutocompleteList,
```
robertknight: This works, although it also affects all tests, whereas we really only need to make this change for the accessibility tests. To do that, you can put this logic in the `content` function for a11y tests, or the `beforeEach` function if the a11y tests are grouped in a `describe` block.

Using `$imports.$mock` to undo a mock like this works, although if you update the `babel-plugin-mockable-imports` package to the latest version there is [an easier/more general method](https://github.com/robertknight/babel-plugin-mockable-imports#restoring-specific-mocks):

```js
$imports.$restore({
 './autocomplete-list': true,
});
```

LMS007:
> You could mock the autocomplete

Yes, it was struggling the validate when the list was closed. I suppose we could just skip that test, but I hesitated because the w3.org combobox examples did in fact use display:none rather than removing the DOM, so I figured let's be safe and model it closer to the canonical examples.


Then after I made that change, it was no longer sufficient to not clear out the selected index after adding a tag because the screen reader was still telling the user that "blah blah is currently selected" when in fact its not. This was not a problem previously because we didn't have the DOM node rendered after adding a tag. (e.g. list was removed)

src/sidebar/components/test/tag-editor-test.js
LMS007: @robertknight oh I missed that one.

In general it looks like we have 3 groups. Is my understanding correct?

1. boiler plate imports from third party libs

2. imports from files we need to test against

3. imports from test utils


Is this correct?

src/sidebar/components/test/tag-editor-test.js
LMS007: awe, good call.

PR #1753, opened 2020-02-05
wip - Menu item refactor

LMS007:
I would like to get some feedback about this reactor. I needed to restructure this component because we were running into static validation issues where div's act like buttons or links.

_previous structure for a button_
```

<div class="menu-item> // acts like a  <button>
  <div class="menu-item__action> 
    <div class="menu-item__icon-container"/> // optional left icon or spacer 
    <span class="menu-item__label/> // button label
    <div class="menu-item__icon-container"/> // optional right icon
   </div>  
   <div class="menu-item__toggle"/> // optional right toggle button
</div>
```

_previous structure for a link_
```

<div class="menu-item> // is focusable, but is not the actual link
  <div class="menu-item__action> 
    <div class="menu-item__icon-container"/> // optional left icon or spacer 
    <a class="menu-item__label/>
    <div class="menu-item__icon-container"/> // optional right icon
   </div>  
   <div class="menu-item__toggle"/> // optional right toggle button
</div>
```

The main issue here is the parent `menu-item` div handles all the events, focus state and can't have aria attributes. Also a small detail is that you can click on the right most icon on the button item and it will work, but it won't for the link item. The changes proposed here fix all of this.

**proposed new structure** 

```
<button class="menu-item__action> 
   <span class="menu-item__icon-container"/> // optional left icon or spacer 
   <span class="menu-item__label/> // content 
   <span class="menu-item__icon-container"/> // optional right icon
   <Button class="menu-item__toggle"/> // optional right toggle button
</button>
```

```
<a class="menu-item__action> 
    <span class="menu-item__icon-container"/> // optional left icon or spacer 
    <span class="menu-item__label/> // content 
    <span class="menu-item__icon-container"/> // optional right icon
    <Button class="menu-item__toggle"/> // optional right toggle button
</a>
```

The restructure moves the semantic tags to the top level so we can attach the right aria attributes and events without lint errors. However, there are still issues. The fact I'm nesting buttons inside buttons is not valid HTML, and I will explore more solutions tomorrow, but this ends up looking the best and nicely wraps the hover effect around the nested button. 

menu

You'll notice that that toggle button's active state shows red and thats just the native behavior from the `<Button>` component and I did not bother overriding that at this time. I also made the background a bit darker on the toggle buttons so they would show up as a different color when opened up compared to the row itself.

I still think there is some room to trim a bit more here, and I think I would like to clean up some class names. I did away with `menu-item` and just moved `menu-item__action` to the parent level. I may re-purpose that class into `menu-item` and do away with `menu-item__action`, or I may add it back if we need to un-nest the button because I'll need a wrapper again.

Thinking ahead a little, there is also the notion of `aria-checked` on the button itself. This also corresponds to a unique class that renders the red bar on the left of only the selected item. I think this would then be best as a "radio" `role` because that's how it operates.

There are still some long term details to consider if we do a further refactor and try to remove focus state and make this behavior like a canonical menu component. I'm not sure how that would work with embedded radio buttons which need focus state. Something to think about later on and not related to this PR.

### DIFF
--- a/src/sidebar/components/menu-item.js
+++ b/src/sidebar/components/menu-item.js
@@ -4,6 +4,7 @@ import propTypes from 'prop-types';
 
 import { onActivate } from '../util/on-activate';
 
+import Button from './button';
 import Slider from './slider';
 import SvgIcon from './svg-icon';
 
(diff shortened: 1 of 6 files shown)


src/sidebar/components/menu-item.js
```diff
@@ -56,57 +57,74 @@ export default function MenuItem({
...
           'is-expanded': isExpanded,
           'is-selected': isSelected,
         })}
-        role="menuitem"
+        href={href}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        {hasLeftIcon && (
+          <div className="menu-item__icon-container">{leftIcon}</div>
```
LMS007: change to span

src/sidebar/components/menu-item.js
```diff
@@ -56,57 +57,74 @@ export default function MenuItem({
...
+        )}
+        <span className={labelClass}>{label}</span>
+        {hasRightIcon && (
+          <div className="menu-item__icon-container">{rightIcon}</div>
```
LMS007: change to span

src/sidebar/components/menu-item.js
```diff
@@ -56,57 +57,74 @@ export default function MenuItem({
...
-              {label}
-            </a>
-          )}
-          {!href && <span className={labelClass}>{label}</span>}
-          {hasRightIcon && (
-            <div className="menu-item__icon-container">{rightIcon}</div>
-          )}
-        </div>
+        {hasLeftIcon && (
+          <div className="menu-item__icon-container">{leftIcon}</div>
```
LMS007: change to span

src/sidebar/components/menu-item.js
```diff
@@ -56,57 +57,74 @@ export default function MenuItem({
...
+        )}
+        <span className={labelClass}>{label}</span>
+        {hasRightIcon && (
+          <div className="menu-item__icon-container">{rightIcon}</div>
```
LMS007: change to span

src/sidebar/components/menu-item.js
LMS007: change to span

src/sidebar/components/menu-item.js
```diff
@@ -56,57 +57,74 @@ export default function MenuItem({
...
+        )}
+
         {typeof isSubmenuVisible === 'boolean' && (
-          <div
+          <Button
```
LMS007: nesting issue
`<button><Button/></button>`

src/sidebar/components/menu-item.js
```diff
@@ -56,57 +57,74 @@ export default function MenuItem({
...
+          <div className="menu-item__icon-container">{rightIcon}</div>
+        )}
+      </a>
+    );
+  } else {
+    // button item
+    //
+    menuItem = (
+      // eslint-disable-next-line jsx-a11y/role-supports-aria-props
+      <button
```
LMS007: not using the Button component here because we need to have some complex inner children.

src/sidebar/components/menu-item.js
```diff
@@ -56,57 +57,74 @@ export default function MenuItem({
...
+        <span className={labelClass}>{label}</span>
+        {hasRightIcon && (
+          <div className="menu-item__icon-container">{rightIcon}</div>
+        )}
+      </a>
+    );
+  } else {
+    // button item
+    //
+    menuItem = (
```
LMS007: Probably will slice this out into its other component, maybe at the top of this file.

src/sidebar/components/menu-item.js
```diff
@@ -56,57 +57,74 @@ export default function MenuItem({
...
-      {/* eslint-disable-next-line jsx-a11y/role-supports-aria-props */}
-      <div
-        aria-checked={isSelected}
-        className={classnames('menu-item', {
+  let menuItem = null;
+
+  if (href) {
+    // link item
+    //
+    menuItem = (
```
LMS007: Probably will slice this out into its other component, maybe at the top of this file.

src/styles/sidebar/components/menu-item.scss
```diff
@@ -157,3 +157,58 @@ $menu-item-padding: 10px;
 .menu-item__submenu {
   border-bottom: solid 1px var.$grey-2;
 }
+
+// new css below, todo: refactor and cleanup
```
LMS007: CSS is WIP

The css will move up and replace some of the older css above, I wanted to keep all the new rules separate until we're locked into a design.

robertknight:
Thanks for the detailed PR description. I think I understand what you are getting. However...

> I needed to restructure this component because we were running into static validation issues where div's act like buttons or links.

The ARIA attribute validation checks that ARIA attributes are legal for a particular _role_ rather than a particular _element type_. In many cases, the element type (button, link etc.) implies a particular _role_ by default. In this case however we're using a generic `div` but assigning the `menuitem` role to it. The linter is complaining that the [menuitem role](https://www.w3.org/TR/wai-aria/#menuitem) that doesn't support `aria-checked`, not that it is a `div`. There are however [menuitemcheckbox](https://www.w3.org/TR/wai-aria/#menuitemcheckbox) and [menuitemradio](https://www.w3.org/TR/wai-aria/#menuitemradio) roles which do support `aria-checked`. `menuitemradio` looks like the appropriate choice here.

It looks to me like the simplest solution is to change the menu item's role depending on whether the menu item is checkable.

One thing that using an actual `button` _would_ do for us is to provide handling of the Space key "for free", however we still need to handle this ourselves for links because `menuitem`s are supposed [to be activated when pressing space](https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-12).

At a high level, I think the main issue with this control at the moment is that it is described, correctly, as a menu, but it doesn't support all of the standard keyboard interactions and focus behavior. In particular:

- Focus doesn't jump to the first item when the menu opens and return to the toggle button when it closes
- Up/Left and Down/Right arrows do not select the next/previous menu items

PR #1756, opened 2020-02-05
Fix a11y attribute issues in MenuItem component

LMS007:
I addressed the lint issues and assigned the logical aria roles to the items when they needed one. Note, in the case of links, there is no role because we don't use the `<div>` as the "thing" that serves as the link. There is simply a nested <a> that is focusable (as was prior to this PR.)

Fix various a11y lint issues with aria props
 - Topmenu items have role=“radio” since they have a selected state
 - Submenu items that are buttons have role=“button” attribute
 - Submenu items that are links have no role attribute (as there is an <a> tag)
 - Add appropriate aria attribute tests

### DIFF
--- a/src/sidebar/components/menu-item.js
+++ b/src/sidebar/components/menu-item.js
@@ -55,13 +55,15 @@ export default function MenuItem({
   }
   const leftIcon = isSubmenuItem ? null : renderedIcon;
   const rightIcon = isSubmenuItem ? renderedIcon : null;
+  // Only add `aria-checked` attribute if `isSubmenuItem` is false. Ensure that the value
+  // is a "true" or "false" string
+  const ariaChecked = !isSubmenuItem
+    ? { 'aria-checked': (!!isSelected).toString() }
+    : {};
 
   return (
     <Fragment>
-      {/* FIXME-A11Y */}
-      {/* eslint-disable-next-line jsx-a11y/role-supports-aria-props */}
       <div
-        aria-checked={isSelected}
         className={classnames('menu-item', {
           'menu-item--submenu': isSubmenuItem,
           'is-disabled': isDisabled,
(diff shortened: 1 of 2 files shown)


src/sidebar/components/menu-item.js
```diff
@@ -55,23 +55,33 @@ export default function MenuItem({
   }
   const leftIcon = isSubmenuItem ? null : renderedIcon;
   const rightIcon = isSubmenuItem ? renderedIcon : null;
+  // Only add `aria-checked` attribute if `isSubmenuItem` is false. Ensure that the value
+  // is a "true" or "false" string
+  const ariaChecked = !isSubmenuItem
```
LMS007: This got a bit tricky but what I wanted to end up with there is only adding the prop when it had the role="radio"
and I wanted the value to be a string, not a bool. I added corresponding tests for each case

src/sidebar/components/menu-item.js
```diff
@@ -55,23 +55,33 @@ export default function MenuItem({
...
-        <div className="menu-item__action">
+        <div
+          className="menu-item__action"
+          // When `isSubmenuItem` is false, treat these items as "radio" buttons since they can be
+          // "selected". Otherwise this is treated as a "button". Note, If there is no `onClick`
+          // event, then no role will be added in any case.
+          {...ariaChecked}
+          // onActivate will add the `role` attribute
+          {...(onClick &&
+            onActivate(isSubmenuItem ? 'button' : 'radio', onClick))}
```
LMS007: Toplevel items get "radio"
Submenu items get "button" or nothing (in the case of links)

src/sidebar/components/menu-item.js
```diff
@@ -55,23 +55,33 @@ export default function MenuItem({
...
           'is-disabled': isDisabled,
           'is-expanded': isExpanded,
           'is-selected': isSelected,
         })}
         role="menuitem"
-        {...(onClick && onActivate('menuitem', onClick))}
       >
-        <div className="menu-item__action">
+        <div
+          className="menu-item__action"
```
LMS007: Basically I shifted the action stuff from the parent <div> to this <div> and this helps square away the aria attributes and role.

LMS007:
further discussion here
https://hypothes-is.slack.com/archives/C1M8NH76X/p1581402527176000

PR #1757, opened 2020-02-06
Fix a11y lint issues in menu.js

LMS007:
- Disable lint rules: "no-static-element-interactions"  and "click-events-have-key-events"  on the menu wrapper. It is acceptable to allow click event handling on static elements when its used to catch event propagation

- Add a -1 tabindex on menu__content for lint purposes

relates to https://github.com/hypothesis/client/issues/1727

### DIFF
--- a/src/sidebar/components/menu.js
+++ b/src/sidebar/components/menu.js
@@ -105,7 +105,7 @@ export default function Menu({
   };
 
   return (
-    // FIXME-A11Y
+    // See https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md#case-the-event-handler-is-only-being-used-to-capture-bubbled-events
     // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events
     <div
       className="menu"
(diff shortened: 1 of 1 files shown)


src/sidebar/components/menu.js
```diff
@@ -145,15 +145,14 @@ export default function Menu({
...
-          {/* FIXME-A11Y */}
-          {/* eslint-disable-next-line jsx-a11y/interactive-supports-focus */}
           <div
             className={classnames(
               'menu__content',
               `menu__content--align-${align}`,
               contentClass
             )}
             role="menu"
+            tabIndex="-1"
```
robertknight: The lint error is unambiguous that this element must be focusable, although I notice that the various [menu examples](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html) that I could find don't seem to do this. When a menu is opened, focus is supposed to be placed on the first menu item, and returned to the button when it is closed. It makes sense that the menu items needs to be programatically focusable. I'm not sure about the "menu" itself.

LMS007: I'm not sure we do that with the focus, but I'll add it to he spike for the menu, it would be good to get this stuff right.

PR #1775, opened 2020-02-10
Highlight a11y (1/2) - Rename highlights CSS styles and move to separate module

robertknight:
This PR does some refactoring of the highlight CSS styles in preparation for changes to them to make highlights visible to assistive technologies. My current plan is to follow the example of Google Docs by using CSS `::before` and `::after` to add visually hidden "comment start" and "comment end" notices around highlighted text.

 - Remove the vestigial "annotator-" prefix for the hypothesis highlight
   CSS rules and replace with "hypothesis-".

 - Move the highlights styles to a separate module for consistency with
   the sidebar styles, where each component's styles goes in its own
   module and also to simplify the entry point (annotator.scss) module

### DIFF
--- a/src/annotator/guest.coffee
+++ b/src/annotator/guest.coffee
@@ -22,6 +22,13 @@ animationPromise = (fn) ->
       catch error
         reject(error)
 
+# A selector which matches elements added to the DOM by Hypothesis (eg. for
+# highlights and annotation UI).
+#
+# We can simplify this once all classes are converted from an "annotator-"
+# prefix to a "hypothesis-" prefix.
+IGNORE_SELECTOR = '[class^="annotator-"],[class^="hypothesis-"]'
+
 module.exports = class Guest extends Delegator
   SHOW_HIGHLIGHTS_CLASS = 'hypothesis-highlights-always-on'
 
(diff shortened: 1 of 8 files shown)


src/annotator/guest.coffee
```diff
@@ -181,7 +181,7 @@ module.exports = class Guest extends Delegator
     this.selections.unsubscribe()
     @adder.remove()
 
-    @element.find('.annotator-hl').each ->
+    @element.find('.hypothesis-highlight').each ->
       $(this).contents().insertBefore(this)
       $(this).remove()
```
LMS007: Line 229 and 
Line 329

```
        ignoreSelector: '[class^="annotator-"]'
```

Will that be affected by this? Looks like its got a pattern to ignore "annotator-" but will it also need "hypothesis-" now?

src/styles/annotator/highlights.scss
```diff
@@ -0,0 +1,24 @@
+@use '../variables' as var;
+
+.hypothesis-highlights-always-on {
```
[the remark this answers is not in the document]
robertknight: The `hypothesis-highlights-always-on` class is a bit special. It is added to / removed from the root element of the document when you toggle the visibility of highlights.

The general change here is from `annotator-$component` to `hypothesis-$component`. The prefix helps to avoid conflicts with the page's CSS and is helpful when you're looking at elements in the dev tools to know which things come from Hypothesis.

This comment reminded me that I need to surface this convention change more widely. I'll post a note in #eng-frontend.

src/annotator/guest.coffee
robertknight: Good catch! - I'll update this. We still have CSS classes that use the "annotator-" prefix so we'll need to match both for the time being.

robertknight:
I discussed this with Kyle in a dev sync earlier this week and got approval to merge after addressing the comments above.

PR #1776, opened 2020-02-10
Highlights a11y (2/2) - Make highlights "visible" to screen readers

robertknight:
**Depends on https://github.com/hypothesis/client/pull/1775**

Use CSS `::before` and `::after` to add "comment start" and "commend. end" notices around highlighted text which screen readers will read but which are not visible.

This pattern follows the example of how highlights are presented in Google Docs and also [this blog post](https://developer.paciellogroup.com/blog/2017/12/short-note-on-making-your-mark-more-accessible/).

This will get quite audibly "noisy" if there are a lot of overlapping highlights, but I think it is better than not having the highlights be visible to screen reader users at all.

This change doesn't fix the problem of being able to activate highlighted text and navigate to the corresponding annotation in the sidebar, that will come separately.

### DIFF
--- a/src/styles/annotator/highlights.scss
+++ b/src/styles/annotator/highlights.scss
@@ -1,11 +1,34 @@
 @use '../variables' as var;
 
+@mixin screen-reader-only {
+  // Hide this content visually, but without preventing screen readers from
+  // reading it.
+  opacity: 0;
+
+  // Avoid this content affecting the visual layout.
+  position: absolute;
+}
+
 // `hypothesis-highlights-always-on` is a class that is toggled on the root
 // of the annotated document when highlights are enabled/disabled.
 .hypothesis-highlights-always-on {
   .hypothesis-highlight {
     background-color: var.$highlight-color;
     cursor: pointer;
+
+    // Make highlights visible to screen readers.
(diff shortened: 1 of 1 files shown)


LMS007:
This implementation seems reasonable but I had a hard time trying to get screen readers to read this out. I tried the mac voice over and chromevox and maybe I'm confused how to drill into an annotation in the first place. Any tips on that? Rob, I'll also give this a try on my windows PC later tonight for thoroughness

robertknight:
I did a demo with Kyle earlier this week where I showed how VoiceOver reads the "comment start" and "comment end" markers in Safari and Chrome. I did some more testing this morning and found that I got different results depending on the navigation method:

- When reading the whole page (VO+A) the comment start/end markers were read
- When reading item by item (VO+Left/Right arrows) the comment start/end markers were read
- When navigating line by line (Up/Down arrows) the comment start/end markers were *not* read.

We'll need to do some testing with other common assistive technology / OS combinations to see how they fair.

If this method doesn't work out, we might want to try switching from using `::before` and `::after` content markers to hidden inline elements adjacent to the text instead.

robertknight:
(I got approval to merge this on the aforementioned demo. We'll test more and revise as needed afterwards)

PR #1778, opened 2020-02-12
Enable debugging info for JSX elements in dev builds and tests

robertknight:
Enable the "development" [1] option for the `@babel/preset-react` preset
to enable more helpful warnings/errors from Preact in development
builds. For example, see the stack trace in this warning I triggered by removing an effect's dependency list:

Screenshot 2020-02-12 15 29 25

Note that the configuration added here _replaces_ rather than _augments_
the `@babel/preset-react` config for other builds (ie. when `NODE_ENV` !== "development"), so other keys like
"createElement" also need to be specified in both places.

See https://babeljs.io/docs/en/babel-preset-react#development and https://babeljs.io/docs/en/6.26.3/babelrc#env-option.

Footnote: On a technical level, what this change does is cause some additional metadata about the _location_ in the source code of JSX expressions to be passed in the `createElement(...)` calls that the JSX expressions are converted into.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -13,5 +13,15 @@
     "angularjs-annotate",
     "transform-async-to-promises"
   ],
-  "ignore": ["**/vendor/*"]
+  "ignore": ["**/vendor/*"],
+  "env": {
+    "development": {
+      "presets": [
+        ["@babel/preset-react", {
+          "development": true,
+          "pragma": "createElement"
+        }]
+      ]
+    }
+  }
 }


LMS007:
Looks like a very helpful tool to add :)

robertknight:
> This seems useful and nice... any chance it could get obnoxiously chatter-y?

🤷‍♂ - I didn't notice any issues in my brief testing, but let me know if you see it causing an issue.

PR #1779, opened 2020-02-12
Fix AutocompleteList aria-selecterd prop

LMS007:
Minor fix here. The `aria-selected` prop shall be “true” or “false” on items in the list. Previously this was truthy so it was omitted from the DOM when false.  Undefined has a different meaning than aria-selected=“false”. I don't think this has any visible changes, even for the screen readers that I tested, but I wanted to get the semantics correct so we have proper examples in our code base.

### DIFF
--- a/src/sidebar/components/autocomplete-list.js
+++ b/src/sidebar/components/autocomplete-list.js
@@ -34,7 +34,7 @@ export default function AutocompleteList({
         <li
           key={`autocomplete-list-${index}`}
           role="option"
-          aria-selected={activeItem === index}
+          aria-selected={(activeItem === index).toString()}
           className={classnames(
             {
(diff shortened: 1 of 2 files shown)


robertknight:
I encountered an issue today where Safari was not reading out the suggestion when I only had one item in the completion list, but did read the suggestion when navigating between multiple candidates. I wonder if this will resolve it.

LMS007:
I gave this a try on this branch, does appear to work at least now.

LMS007:
> LGTM. We might want to look at introducing a runtime check against passing booleans as aria prop values. It might be worth checking whether React (not Preact) has any special handling of `aria-` props. If it does I could file an upstream issue against Preact about doing the same.

I'll do that

PR #1786, opened 2020-02-13
Fix a11y lint on SelectionTabs component

LMS007:
Straight forward conversion. I would like to get opinion on using the <Button> component here, but again, the css was a bit easier to match without it.

### DIFF
--- a/src/styles/sidebar/components/selection-tabs.scss
+++ b/src/styles/sidebar/components/selection-tabs.scss
@@ -1,4 +1,5 @@
 @use "../../mixins/focus";
+@use "../../mixins/buttons";
 @use "../../variables" as var;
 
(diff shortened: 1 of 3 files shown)


src/styles/sidebar/components/selection-tabs.scss
```diff
@@ -28,6 +30,10 @@
   min-height: 18px;
 
   user-select: none;
+
+  &:hover {
+    color: var.$link-color-hover;
```
robertknight: This preserves the existing behavior, although we might want to consider dropping it entirely as non-link actions elsewhere in the app generally don't have this hover effect.

LMS007: agreed

Issue #1794, opened 2020-02-14
Set appropriate language hints for the sidebar HTML page


robertknight:
## Problem

The sidebar HTML page, which is actually part of the h and browser-extension rather than client repos, does not currently specify the primary language of the UI. This causes several problems:

- Automated WCAG compliance checkers complain about the lack of `lang` attribute on the `<html>` element in the sidebar's HTML document
- When annotating a document that uses a non-English language (eg. https://lemonde.fr) the screen reader can end up reading UI elements with a voice appropriate for the language of the host document rather than the language of the UI (currently always English).

## Discussion

Solving this isn't quite as simple as slapping a `lang="en"` tag on the HTML page because that will apply to _all_ content in the sidebar, including annotation quotes and bodies. Having those read out with an English voice when they are not English would be, I imagine, quite annoying.

The ideal behavior would be that the UI is read with a voice for English (or whatever language it is in, when we add translation in future) but user-generated content is read in a voice appropriate for that content. There is a way to tell the browser that the language of an element is unknown (`lang=""`) and ideally, the browser would detect the language itself. Chrome does have an experimental option to auto-detect the language in that scenario, which produces the desired result:

<img width="913" alt="Screenshot 2020-02-14 15 22 25" src="https://user-images.githubusercontent.com/2458/74544558-3f320c80-4f3f-11ea-86fa-b549423f30a2.png">

Chrome (without this option) and Safari don't do this. I'm unclear as yet whether `lang=""` makes them default to the language of the host page (probably OK), the language of the sidebar's HTML document (not so good) or the user's default language (probably OK).

Twitter seems to do server-side processing of tweets to determine the language and it sets `lang="<code>"` appropriately on each tweet. We could do the same, but this will obviously require a fair amount of work.

Some simple approaches that we could potentially use to start with:

- Set `lang="en"` on the sidebar, `lang=""` on user-generated content and leave it up to the browser to do something appropriate or for the user to configure how unspecified languages are handled.
- Set `lang="en"` on the sidebar but for annotation quotes and bodies, use the language of the host page, or unknown if the host page doesn't specify. The assumption here is that most annotations are in the same language as the document
- Anything else?

LMS007:
This is not easy to come up with a set of rules to apply.  Let me see if I can understand this more specifically.

Problem 1.
lang value for sidebar UI

Problem 2. 
lang value for annotations.

[...]

PR #1795, opened 2020-02-17
Add `lang=""` to user-generated/chosen content in sidebar UI

robertknight:
Add a `lang=""` attribute to the annotation quote, body and tag list in
the UI which display content chosen or generated by the user, to
indicate to the user agent that the language is unknown and may be
different than the client UI, which is currently always English.

Part of https://github.com/hypothesis/client/issues/1794.

### DIFF
--- a/src/sidebar/components/tag-list.js
+++ b/src/sidebar/components/tag-list.js
@@ -32,6 +32,7 @@ function TagList({ annotation, serviceUrl, settings, tags }) {
             <a
               className="tag-list__link"
               href={createTagSearchURL(tag)}
+              lang=""
               target="_blank"
               rel="noopener noreferrer"
               aria-label={`Tag: ${tag}`}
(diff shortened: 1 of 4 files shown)


src/sidebar/components/tag-list.js
```diff
@@ -32,6 +32,7 @@ function TagList({ annotation, serviceUrl, settings, tags }) {
             <a
```
LMS007: Might be wise to add it to the tag editor's AutocompleteList too. Or even a just blanket the prop on all AutocompleteList components. (we are only using it once of course) Perhaps on the `<div class="autocomplete-list">` tag?

robertknight: Hmm ... it depends on where the suggestions for an autocomplete list are coming from. For the moment I added the `lang=""` attribute just to suggestions rendered by the tag editor when it uses `AutocompleteList`, since we know they are definitely user-supplied (as opposed to coming from some source where we know the language).

LMS007: Yes, thinking about this more, its highly likely the user added all these tags anyway, perhaps we let it be whatever the UI is.

PR #1796, opened 2020-02-17
Add accessible label to annotation body markdown editor

robertknight:
Visually I think the purpose of this field is obvious without an explicit label
but it may not be clear to assistive tech users.

### DIFF
--- a/src/sidebar/components/test/markdown-editor-test.js
+++ b/src/sidebar/components/test/markdown-editor-test.js
@@ -100,13 +100,14 @@ describe('MarkdownEditor', () => {
     describe(`"${command}" toolbar command`, () => {
       it('applies formatting when toolbar button is clicked', () => {
         const onEditText = sinon.stub();
-        const wrapper = createComponent({ text: 'test', onEditText });
+        const text = 'toolbar command test';
+        const wrapper = createComponent({ text, onEditText });
         const button = wrapper.find(
           `ToolbarButton[title="${command}"] > button`
         );
         const input = wrapper.find('textarea').getDOMNode();
         input.selectionStart = 0;
-        input.selectionEnd = 4;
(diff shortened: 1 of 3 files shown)


src/sidebar/components/test/markdown-editor-test.js
```diff
@@ -96,9 +100,7 @@ describe('MarkdownEditor', () => {
     describe(`"${command}" toolbar command`, () => {
       it('applies formatting when toolbar button is clicked', () => {
         const onEditText = sinon.stub();
-        const wrapper = mount(
-          <MarkdownEditor text="test" onEditText={onEditText} />
-        );
+        const wrapper = createComponent({ text: 'test', onEditText });
```
LMS007: don't need `text: 'test'`

src/sidebar/components/test/markdown-editor-test.js
```diff
@@ -122,9 +124,7 @@ describe('MarkdownEditor', () => {
       if (key) {
         it('applies formatting when shortcut key is pressed', () => {
           const onEditText = sinon.stub();
-          const wrapper = mount(
-            <MarkdownEditor text="test" onEditText={onEditText} />
-          );
+          const wrapper = createComponent({ text: 'test', onEditText });
```
LMS007: same

src/sidebar/components/test/markdown-editor-test.js
robertknight: Ah, I had actually left this intentionally to make it easier to see that the test inputs and assertions corresponded. I have pushed a change to make this more obvious by using a variable rather than repeating the `'test'` string.

robertknight:
I made a change to address the feedback (see last commit). Thanks!

PR #1812, opened 2020-02-20
Enable AnnotationQuote a11y test

robertknight:
All of the issues causing this test to fail have been fixed.


robertknight:
I saw Lyza fixed this in https://github.com/hypothesis/client/pull/1811/.

PR #1832, opened 2020-02-25
Stop loading icon font CSS in sidebar app

robertknight:
All uses of the icon font in the sidebar app have now been replaced by
`SvgIcon`.

### DIFF
--- a/src/boot/boot.js
+++ b/src/boot/boot.js
@@ -121,7 +121,6 @@ function bootSidebarApp(doc, config) {
 
     'styles/angular-csp.css',
     'styles/angular-toastr.css',
-    'styles/icomoon.css',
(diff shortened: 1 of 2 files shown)


lyzadanger:
@robertknight There are no icon-font icons left in the `client` part of the application, but what about the annotator? e.g.

image

robertknight:
> @robertknight There are no icon-font icons left in the client part of the application, but what about the annotator? e.g.

That should work the same as before. If you look in `boot/boot.js` you can see there is still a reference to `icomoon.css`. I don't think the icon font works in the devtools though.

robertknight:
Gah, I failed to update the tests for `boot/boot.js`...

PR #1870, opened 2020-03-02
Refactor fetch-config.js and allow requestConfigFromFrame to be an object

LMS007:
- Rename fetchConfig to fetchConfigLegacy
- Allow the requestConfigFromFrame variable from hypothesisConfig to be either a string containing the origin url (legacy),
or an object containing the origin url + the ancestor level value of the parent frame. This is used to directly find the
frame which to send RPC requests too so we don't have to guess and create erroneous console errors.
- Rename fetchConfig to fetchConfigLegacy (now deprecated)
- Add fetchConfigRpc to replace fetchConfigLegacy
- Add fetchConfigEmbed for the simple hash fragment case
- Refactor fetchConfig to decide how it will fetch the host config (one of 3-ways)
 1. simple case from url hash fragment
 2. legacy RPC call from (depreciated)
 3. RPC call directly to parent frame.
-------------

Effectively this diff should have no visible change to the client. In order to see it working with the new config var, changes to via and lms need to be made. I have locally shimmed in those changes and verified that it does work and the console error is gone. Expect more PR's to follow this one, then the plan is to circle back to the client and remove some of the @depricated code in this diff.

### DIFF
--- a/src/sidebar/util/fetch-config.js
+++ b/src/sidebar/util/fetch-config.js
@@ -1,8 +1,10 @@
 import getApiUrl from '../get-api-url';
 import hostConfig from '../host-config';
-
 import * as postMessageJsonRpc from './postmessage-json-rpc';
 
+/**
+ * @deprecated
+ */
 function ancestors(window_) {
   if (window_ === window_.top) {
     return [];
@@ -19,6 +21,27 @@ function ancestors(window_) {
 }
 
 /**
+ * Returns the global embedder ancestor frame.
+ *
+ * @param {number} levels - Number of ancestors levels to ascend.
+ * @param {Window=} window
+ * @return {Window}
+ */
+function getAncestorFrame(levels, window_ = window) {
+  let ancestorWindow = window_;
+  for (let i = 0; i < levels; i++) {
+    if (ancestorWindow === ancestorWindow.top) {
+      throw new Error(
+        'The target parent frame has exceeded the ancestor tree. Try reducing the `requestConfigFromFrame.ancestorLevel` value in the `hypothesisConfig`'
+      );
+    }
+    ancestorWindow = ancestorWindow.parent;
+  }
+  return ancestorWindow;
+}
+
+/**
+ * @deprecated
  * Fetch client configuration from an ancestor frame.
  *
  * @param {string} origin - The origin of the frame to fetch config from.
@@ -47,6 +70,7 @@ function fetchConfigFromAncestorFrame(origin, window_ = window) {
 }
 
 /**
+ * @deprecated
  * Merge client configuration from h service with config fetched from
  * embedding frame.
  *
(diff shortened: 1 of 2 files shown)


src/sidebar/index.js
```diff
@@ -347,13 +355,54 @@ function startAngularApp(config) {
...
+//
+// There are 3 ways to get the config to launch the app.
+//  Case 1: [Classic embed] - From the query string of the embedder frame.
+//  Case 2a: [RPC with unknown parent] - From a ancestor parent frame that passes it down via RPC. (deprecated)
+//  Case 2b: [RPC with known parent] - From a ancestor parent frame that passes it down via RPC.
+
+const hostPageConfig = hostConfig(window);
+const requestConfigFromFrame = hostPageConfig.requestConfigFromFrame;
+const configFrameAncestorLevel = hostPageConfig.configFrameAncestorLevel;
+
```
robertknight: I would prefer that the complete logic for fetching the config could be pushed down into `fetch-config.js` and the logic in `src/sidebar/index.js` is kept as simple as possible. This will make it easier to write tests for this functionality and also help to avoid  `src/sidebar/index.js` growing into a huge module which is difficult to follow.

src/annotator/config/index.js
```diff
@@ -31,6 +31,9 @@ export default function configFrom(window_) {
     }),
     query: settings.query,
     requestConfigFromFrame: settings.hostPageSetting('requestConfigFromFrame'),
+    configFrameAncestorLevel: settings.hostPageSetting(
+      'configFrameAncestorLevel'
+    ),
```
robertknight: I think it would be preferable to enable `requestConfigFromFrame` to be an object with `origin` and `ancestorLevel` properties OR a string. My rationale for this is because it is easier to communicate in documentation that there is an optional field which, if specified, is an object with two required properties, rather than two separate optional fields where, if one is specified, the other should be as well.

Note that this is separate from the change to use two separate query params in Via to specify the parent frame. In that context I think using separate params is reasonable since there isn't a "natural" way to encode nesting in query strings.

src/annotator/config/index.js
```diff
@@ -31,6 +31,9 @@ export default function configFrom(window_) {
     }),
     query: settings.query,
     requestConfigFromFrame: settings.hostPageSetting('requestConfigFromFrame'),
+    configFrameAncestorLevel: settings.hostPageSetting(
```
robertknight: I think this change is left over from the first iteration of this PR and no longer applicable?

src/annotator/config/test/index-test.js
```diff
@@ -91,6 +91,7 @@ describe('annotator.config.index', function() {
     'branding',
     'openSidebar',
     'requestConfigFromFrame',
+    'configFrameAncestorLevel',
```
robertknight: I think this change is left over from the first iteration of this PR and no longer applicable?

src/annotator/config/test/index-test.js
```diff
@@ -99,6 +100,7 @@ describe('annotator.config.index', function() {
         branding: 'BRANDING_SETTING',
         openSidebar: 'OPEN_SIDEBAR_SETTING',
         requestConfigFromFrame: 'https://embedder.com',
+        configFrameAncestorLevel: 1,
```
robertknight: I think this change is left over from the first iteration of this PR and no longer applicable?

src/sidebar/host-config.js
```diff
@@ -43,6 +43,9 @@ export default function hostPageConfig(window) {
     // Fetch config from a parent frame.
     'requestConfigFromFrame',
 
+    // Number of ancestor frames that the parent frame resides above this sidebar
+    'configFrameAncestorLevel',
```
robertknight: I think this change is left over from the first iteration of this PR and no longer applicable?

src/sidebar/test/host-config-test.js
```diff
@@ -16,6 +16,7 @@ describe('hostPageConfig', function() {
       appType: 'bookmarklet',
       openSidebar: true,
       requestConfigFromFrame: 'https://embedder.com',
+      configFrameAncestorLevel: 1,
```
robertknight: I think this change is left over from the first iteration of this PR and no longer applicable?

src/sidebar/util/fetch-config.js
```diff
@@ -75,3 +97,87 @@ export function fetchConfig(appConfig, window_ = window) {
     return mergedConfig;
   });
 }
+
+/**
+ * Merge client configuration from h service with config fetched from
+ * an embedding frame synchronously.
+ *
+ * Use this method to retrieve the config synchronously from the embedding
+ * frame's query string. See tests for more details.
```
robertknight: In the common case, configuration is passed from the parent frame to the child via a hash fragment rather than query string. To avoid this comment getting out of date, you could just omit mentioning _how_ the configuration is passed here.

I think the comment is possibly confusing here because "embedding frame" sounds like it refers to the parent document, whereas it actually refers to the _sidebar frame_.

src/sidebar/util/fetch-config.js
```diff
@@ -75,3 +97,87 @@ export function fetchConfig(appConfig, window_ = window) {
...
+    ...appConfig,
+    ...hostPageConfig,
+  };
+  mergedConfig.apiUrl = getApiUrl(mergedConfig);
+  return mergedConfig;
+}
+
+/**
+ * Merge client configuration from h service with config fetched from
+ * an embedding frame asynchronously.
```
robertknight:
```suggestion
 * a parent frame asynchronously.
```

The frame supplying the config might not be the frame that directly embeds the client, and isn't in the lms case.

src/sidebar/util/fetch-config.js
```diff
@@ -75,3 +97,87 @@ export function fetchConfig(appConfig, window_ = window) {
...
+export async function fetchConfig(appConfig, window_ = window) {
+  const hostPageConfig = hostConfig(window);
+  const requestConfigFromFrame = hostPageConfig.requestConfigFromFrame;
+
+  if (!requestConfigFromFrame) {
+    // Directly embed: just get the config.
+    return fetchConfigEmbed(appConfig, hostPageConfig);
+  }
+  if (typeof requestConfigFromFrame === 'string') {
+    // Legacy: send RPC to all parents to find config. (note will cause js error(s))
```
robertknight: Can you expand on this "will cause js error(s)" comment for the benefit of other developers in future? Something like "Browsers may display errors in the console when messages are  sent to frames that don't match the origin filter".

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -147,17 +146,90 @@ describe('sidebar.util.fetch-config', () => {
...
+        fakeJsonRpc.call.resolves({});
+        fakeHostConfig.returns({
+          requestConfigFromFrame: {
+            origin: 'https://embedder.com',
+            ancestorLevel: 2,
+          },
+        });
+      });
+      describe('getAncestorFrame', () => {
+        it('makes an rcp request to `requestConfig` ', async () => {
```
robertknight:
```suggestion
        it('makes an RPC request to `requestConfig` ', async () => {
```

src/sidebar/util/fetch-config.js
```diff
@@ -19,6 +21,25 @@ function ancestors(window_) {
...
+ *
+ * @param {number} levels - Number of ancestors levels to ascend.
+ * @param {Window=} window
+ * @return {Window}
+ */
+function getAncestorFrame(levels, window_ = window) {
+  let ancestorWindow = window_;
+  for (let i = 0; i < levels; i++) {
+    if (ancestorWindow === ancestorWindow.top) {
+      break;
```
robertknight: Hmm, I wonder if this should throw rather than simply stopping if the number of levels is greater than the depth of the frame tree, since this very likely indicates incorrect configuration.

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -147,17 +146,95 @@ describe('sidebar.util.fetch-config', () => {
...
+          apiUrl: fakeApiUrl(),
+        });
+      });
+
+      it('rejects if fetching config fails` ', async () => {
+        fakeJsonRpc.call.rejects(new Error('Nope'));
+        const appConfig = { foo: 'bar', appType: 'via' };
+
+        const result = fetchConfig(appConfig, fakeWindow);
+        return assertPromiseIsRejected(result, 'Nope');
```
LMS007: Will remove `assertPromiseIsRejected` in another PR

src/sidebar/index.js
```diff
@@ -4,6 +4,7 @@ import crossOriginRPC from './cross-origin-rpc.js';
 import addAnalytics from './ga';
 import disableOpenerForExternalLinks from './util/disable-opener-for-external-links';
 import { fetchConfig } from './util/fetch-config';
+
```
LMS007: todo: could remove changes to this file

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -147,17 +146,95 @@ describe('sidebar.util.fetch-config', () => {
...
+        it('throws an error when target ancestor exceeds top window', async () => {
+          fakeHostConfig.returns({
+            requestConfigFromFrame: {
+              origin: 'https://embedder.com',
+              ancestorLevel: 10, // The top window is only 2 levels high
+            },
+          });
+          try {
+            await fetchConfig({}, fakeWindow);
+          } catch (e) {
```
robertknight: This test could pass incorrectly if `fetchConfig` failed to throw here

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -23,61 +27,58 @@ describe('sidebar.util.fetch-config', () => {
     fakeJsonRpc.call.throws(new Error('call() response not set'));
 
     // Setup fake window hierarchy.
-    const fakeTopWindow = { parent: null, top: null };
+    fakeTopWindow = { parent: null, top: null, index: 2 };
```
robertknight: It wasn’t immediately clear to me what the `index` property was since real windows don’t have this. I assume this is just for use in test asserts? Maybe call it `testId` or something similar to make that more obvious?

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -147,17 +146,95 @@ describe('sidebar.util.fetch-config', () => {
...
+      beforeEach(() => {
+        fakeJsonRpc.call.resolves({});
+        fakeHostConfig.returns({
+          requestConfigFromFrame: {
+            origin: 'https://embedder.com',
+            ancestorLevel: 2,
+          },
+        });
+      });
+      describe('getAncestorFrame', () => {
```
robertknight: `getAncestorFrame` is an implementation detail rather than an exported API method so it doesn’t need to be mentioned in the tests.

LMS007: awe.. that was left over from refactoring, forgot to remove it!

src/sidebar/util/test/fetch-config-test.js
LMS007: correct, index was a terrible name

PR #1873, opened 2020-03-04
Replace two dependencies that are only used once

robertknight:
Following on from https://github.com/hypothesis/client/pull/1872, this PR reduces the size of the sidebar bundle a bit more by removing a couple of dependencies that we can easily do without.

- Use of `lodash.get` to take a dotted API method path (`annotations.search`) and extract the relevant value from a nested object (`{ annotations: { ..., search: { ... }, ... }`)
- Use of `node-uuid` to generate a random identifier when setting up the WebSocket connection (note I've verified that h doesn't care whether the identifier is a UUID or not)

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
     "lodash.debounce": "^4.0.3",
     "loose-envify": "^1.4.0",
     "mocha": "^7.0.0",
-    "node-uuid": "^1.4.3",
     "npm-packlist": "^2.0.1",
     "postcss": "^7.0.13",
     "postcss-url": "^8.0.0",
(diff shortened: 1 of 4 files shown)


robertknight:
> My only question would be to make sure that the generateHexString util function is indeed what we want, and that we don't want to make of UUIDs in general...

At present it doesn't matter, so long as two clients are sufficiently unlikely to generate the same ID. If there was a need for clients to generate globally unique identifiers as part of a public API then we probably should recommend version 4 UUIDs. The generated hex string here is the same length as the UUID (32 hex digits).

PR #1885, opened 2020-03-07
Add sendReadyToReceive method to fetch-config.js

LMS007:
This is a necessary request back to the parent app to tell it that the client’s RCP server is ready to receive incoming RCP requests. Without message, the parent APP does not know when to start sending RPC requests down from to the client. If they are sent too soon, they risk timeout. _I noticed the timeout happened after making some of the requests async._

Screen Shot 2020-03-13 at 10 04 53 AM

The thing we need to do is send this RPC request after `startAngularApp` is called (which in turn starts the RCP server)

I was not initially sure how to fit this request in with the current structure but decided on just adding a stand alone method that takes care of everything. The downside is this is a bit redundant with `fetchConfig` in that it also has to call `getAncestorFrame` and also check the `requestConfigFromFrame` variable -- both of which are done already with `fetchConfig`, but this is a minor bit of overhead to making the api clean.

The other thing I was not sure about is where this method should live. fetch-config is not really the right name anymore of everything in that file, but I didn't really want to make a new file. We could rename fetch-confg to rcp-config-requests or something similar? Open to ideas.

If this does get checked in right now, it won't "hurt" anything, but it will mean that there will be an error in the console due to unknown method. The changes on LMS side are not straightforward and will break the client without this PR (we're at another cat vs mouse junction), so I propose we stub out `readyToReceive` as a no-op registered method in LMS first if we want to avoid the error.

PR's on the LMS side:
https://github.com/hypothesis/lms/pull/1539
https://github.com/hypothesis/lms/pull/1540

### DIFF
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -5,7 +5,7 @@ import { jsonConfigsFrom } from '../shared/settings';
 import crossOriginRPC from './cross-origin-rpc.js';
 import addAnalytics from './ga';
 import disableOpenerForExternalLinks from './util/disable-opener-for-external-links';
-import { fetchConfig } from './util/fetch-config';
+import { fetchConfig, sendReadyToReceive } from './util/fetch-config';
 import * as sentry from './util/sentry';
 
 // Read settings rendered into sidebar app HTML by service/extension.
@@ -359,6 +359,7 @@ function startAngularApp(config) {
 fetchConfig(appConfig)
   .then(config => {
     startAngularApp(config);
+    sendReadyToReceive();
   })
   .catch(err => {
     // Report error. This will be the only notice that the user gets because the
--- a/src/sidebar/util/fetch-config.js
+++ b/src/sidebar/util/fetch-config.js
@@ -180,3 +180,33 @@ export async function fetchConfig(appConfig, window_ = window) {
     );
   }
 }
+
+/**
+ * Makes an RPC request `readyToReceive` to the ancestor window to
+ * inform that it is ready to receive incoming RPC requests. This
+ * function must only be called after the RPC server is started.
+ *
+ * e.g.
+ * crossOriginRPC.server.start
+ *
+ * @param {Window} window_ - Test seam.
+ */
+export function sendReadyToReceive(window_ = window) {
+  const requestConfigFromFrame = hostConfig(window).requestConfigFromFrame;
+  // Does it get the config from an ancestor frame?
+  if (
+    requestConfigFromFrame &&
+    typeof requestConfigFromFrame.ancestorLevel === 'number' &&
+    typeof requestConfigFromFrame.origin === 'string'
+  ) {
+    const parentFrame = getAncestorFrame(
+      requestConfigFromFrame.ancestorLevel,
+      window_
+    );
+    postMessageJsonRpc.call(
+      parentFrame,
+      requestConfigFromFrame.origin,
+      'readyToReceive'
+    );
+  }
+}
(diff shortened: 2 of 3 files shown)


robertknight:
> Without message, the parent APP does not know when to start sending RPC requests down from to the client.

Since the client will always make a request to the ancestor frame to fetch configuration as the first thing it does when it starts, perhaps the parent frame could simply keep track of whether it has received this request or not. If so, that would allow us to simplify the protocol by not adding a separate "ready to receive notification". Can you see a reason that wouldn't work?

LMS007:
> > Without message, the parent APP does not know when to start sending RPC requests down from to the client.
> 
> Since the client will always make a request to the ancestor frame to fetch configuration as the first thing it does when it starts, perhaps the parent frame could simply keep track of whether it has received this request or not. If so, that would allow us to simplify the protocol by not adding a separate "ready to receive notification". Can you see a reason that wouldn't work?

I think thats just it, it does fetch first, but it not ready at that point. It can only receive RCP's requests as a response to its own outgoing client request, not unsolicited incoming requests to its server. I did not seriously look into moving the `crossOriginRPC.server.start` out of the angular start procedure as that may be another way to achieve this but I suspect it may be more difficult.

The order as I understand it is.

1. Line 349: `fetchConfig` // now we have the LMS's frame
2. Line: 342: `startAngularApp(config);`
3. Line 361: `.run(crossOriginRPC.server.start);` // now the RCP server is ready
4. (tell LMS we're ready)

robertknight:
> I think thats just it, it does fetch first, but it not ready at that point. It can only receive RCP's requests as a response to its own outgoing client request, not unsolicited incoming requests to its server.

Hmm, what entity are you referring to by "it not ready"? My thinking was that when the client sends a configuration request, that tells the LMS frontend that the Hypothesis client is ready to receive future notifications from the server.

To take a concrete example, supposing the LMS app is in grading mode and the user switches from one user to another:

- If the LMS app has received the configuration request from the client, it sends the focused user message directly to the client
- If the LMS app has not yet received the configuration request from the client, it waits until the client has fetched configuration before sending that message

In fact, it looks to me like that should already be happening because the `getSidebarWindow` function returns a `Promise` which only resolves once the sidebar is connected:

https://github.com/hypothesis/lms/blob/4bec8b7e2a9cbdfcbce6d6f6382ce3ac4c4f5ccc/lms/static/scripts/frontend_apps/components/LMSGrader.js#L31

Am I missing something here?

LMS007:
> Hmm, what entity are you referring to by "it not ready"? My thinking was that when the client sends a configuration request, that tells the LMS frontend that the Hypothesis client is ready to receive future notifications from the server.

That is not correct. The client server is not ready. It has to get the config back, then start the RCP server and then it's ready. There is a little delay between when it receives the config response call back and when it spins up its own RPC server. If any new RCP message come at that time, then they timeout.

robertknight:
> The client server is not ready. It has to get the config back, then start the RCP server and then it's ready. There is a little delay between when it receives the config response call back and when it spins up its own RPC server. If any new RCP message comes at that time, then they timeout.

Ah! - I get it. Re-reading your comment above that makes sense, I'm not sure why it didn't click last week. Sorry for the delay on this.

The `crossOriginRPC` service currently depends on the store and settings (although only for the `rpcAllowedOrigins` key, which actually always comes from h), so you're right that you can't easily start it before the store is initialized. However, I can suggest an alternate possibility which doesn't require adding a new RPC call:  Add some mechanism that sets up a "message" event listener early in the lifecycle of the application, before fetching configuration, which buffers messages in a queue. When the `crossOriginRPC` service is started and all the methods are registered, then you can drain the queue. This way you won't miss any messages sent in between the configuration request and the server being started.

Issue #1710, opened 2020-01-24
use-element-should-close should handle key press "Esc"


LMS007:
When looking at the menu for posting a new annotation with Mac narrator turned on, it prompts user that the escape key will close the menu, but it does not.

The only way to close the menu with the keyboard is to tab away from it but it might be nice to also support the "Esc" key. I believe Mac OS narrator makes this suggestion because role=
menuitem"

key events could be handled inside of use-element-should-close.js

![Screen Shot 2020-01-23 at 4 37 22 PM](https://user-images.githubusercontent.com/3939074/73035907-ab1ccb80-3dfe-11ea-8846-c08f4a5eb119.png)

PR #1897, opened 2020-03-11
Fix useElementShouldClose to work with Escape key

LMS007:
‘keypress’ is not fired for “Escape” key, so we need to listen for ‘keydown’. 

This is needed for a number of a11y components such as dialogs. So its good to fix this sooner than later. I specifically need this now because I'm porting `useElementShouldClose` to LMS for an a11y dialog change but I need `useElementShouldClose` for the Escape feature --which is actually broken, hence, this PR :)

fixed https://github.com/hypothesis/client/issues/1710

### DIFF
--- a/src/sidebar/components/hooks/test/use-element-should-close-test.js
+++ b/src/sidebar/components/hooks/test/use-element-should-close-test.js
@@ -12,7 +12,7 @@ describe('hooks.useElementShouldClose', () => {
   const events = [
     new Event('mousedown'),
     new Event('click'),
-    ((e = new Event('keypress')), (e.key = 'Escape'), e),
+    ((e = new Event('keydown')), (e.key = 'Escape'), e),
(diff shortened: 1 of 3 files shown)


robertknight:
I confirm this change fixes pressing Esc to close menus in the client in Chrome and Firefox. It looks like Safari _does_ generate `keypress` events for the Escape key however and closing menus with Esc did work there. According to MDN, the [`keypress` event](https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event) is deprecated, so we shouldn't ever listen for it anyway.

I wonder if there is a lint check we could use to catch this for JSX components at least.

LMS007:
Yes, we have some more keypress events, I'll create a ticket and checklist for those
(also I need to change line 22 in the docs in this diff to say "keydown" instead of "keypress")

PR #1901, opened 2020-03-12
Add keyboard shortcuts for the Annotate/Highlight ("adder") toolbar

robertknight:
This PR makes it easier to create annotations and highlights using the keyboard. It adds keyboard shortcuts for the "Annotate" and "Highlight" buttons which are "a" and "h" respectively. My thinking is that since the toolbar is only shown when there is a non-empty selection of static text (ie. not in an input field), we can use plain keys, with no modifiers, as the shortcut keys. I've left open the possibility to change this in future.

I plan to extend this in future by adding an additional toolbar action which selects the annotations associated with the selected text, ie. as if you had clicked on the highlighted text, and a corresponding shortcut. This will provide a solution to the problem of how to view annotations associated with a highlighted piece of text using only keyboard interaction.

The PR is split into two commits. The first adds a set of utilities in `src/shared/shortcut.js` that make it easy to create document-level shortcuts that are associated with a component. The second modifies `src/annotator/components/adder-toolbar.js` to use this to register shortcuts for the "Annotate" and "Highlight" buttons. When reviewing this, it will probably be easier to start by looking at the changes to `adder-toolbar.js`.

To make the shortcuts discoverable, I added titles to these buttons. I expect we'll also need to write up a document in the knowledge base.

### DIFF
--- a/src/shared/shortcut.js
+++ b/src/shared/shortcut.js
@@ -30,8 +30,10 @@ export function matchShortcut(event, shortcut) {
     const modifierFlag = modifiers[part];
     if (modifierFlag) {
       requiredModifiers |= modifierFlag;
-    } else {
+    } else if (requiredKey === null) {
       requiredKey = part;
+    } else {
+      throw new Error('Multiple non-modifier keys specified');
     }
   }
 
--- a/src/shared/shortcut.js
+++ b/src/shared/shortcut.js
@@ -1,0 +1,109 @@
+import { useEffect } from 'preact/hooks';
+
+// Bit flags indicating modifiers required by a shortcut or pressed in a key event.
+const modifiers = {
+  alt: 1,
+  ctrl: 2,
+  meta: 4,
+  shift: 8,
+};
+
+/**
+ * Match a `shortcut` key sequence against a keydown event.
+ *
+ * A shortcut key sequence is a string of "+"-separated keyboard modifiers and
+ * keys. The list may contain zero or more modifiers and must contain exactly
+ * one non-modifier key. The key and modifier names are case-insensitive.
+ * For example "Ctrl+Enter", "shift+a". Key names are matched against `KeyboardEvent.key`.
+ *
+ * @param {KeyboardEvent} event
+ * @param {string} shortcut
+ * @return {boolean}
+ */
+export function matchShortcut(event, shortcut) {
+  const parts = shortcut.split('+').map(p => p.toLowerCase());
+
+  let requiredModifiers = 0;
+  let requiredKey = null;
+
+  for (let part of parts) {
+    const modifierFlag = modifiers[part];
+    if (modifierFlag) {
+      requiredModifiers |= modifierFlag;
+    } else {
+      requiredKey = part;
+    }
+  }
+
+  if (!requiredKey) {
+    throw new Error(`Invalid shortcut: ${shortcut}`);
+  }
+
+  const actualModifiers =
+    (event.ctrlKey && modifiers.ctrl) |
+    (event.metaKey && modifiers.meta) |
+    (event.altKey && modifiers.alt) |
+    (event.shiftKey && modifiers.shift);
+
+  return (
+    actualModifiers === requiredModifiers &&
+    event.key.toLowerCase() === requiredKey
+  );
+}
+
+/**
+ * @typedef ShortcutOptions
+ * @prop {HTMLElement} rootElement -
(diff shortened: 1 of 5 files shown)


src/shared/test/shortcut-test.js
```diff
@@ -0,0 +1,156 @@
...
+    ['ctrl', 'META'].forEach(shortcut => {
+      it('should throw an error if no non-modifier key is specified', () => {
+        assert.throws(() => {
+          matchShortcut(createEvent('a'), shortcut);
+        });
+      });
+    });
+  });
+
+  describe('installShortcut', () => {
```
LMS007: Might be good to also have a test where it installs it on a custom element other than document.body, unless I missed that?

robertknight: You're right. I've added a test for this.

robertknight:
> I also (for the life of me) cant figure out how to highlight with a keyboard only! So I had to test with some mouse use :)

Do you mean "figure out how to _select_ text with a keyboard only"? That will depend on the OS and browser, but we're assuming for the purposes of this change that the user has figured that out how to do that with whatever OS + browser they use.

In Safari on macOS you can do this when VoiceOver is active by using the arrow keys to move the VoiceOver cursor (the black rectangle) to the start of the text you want to select and then using Shift+arrow keys to set the end point of the selection. Left/Right arrow keys move the cursor/selection by character. Up/down arrow keys move the cursor/selection by line. I don't yet know how to make that work in Chrome though.

src/shared/test/shortcut-test.js
```diff
@@ -0,0 +1,168 @@
...
+  return {
+    key,
+    ctrlKey: !!ctrl,
+    altKey: !!alt,
+    shiftKey: !!shift,
+    metaKey: !!meta,
+  };
+}
+
+describe('shared/shortcut', () => {
```
LMS007: I think its worth having 1 or 2 tests with multiple modifier keys

`(createEvent('a', { ctrl: true, shift: true })`

src/shared/shortcut.js
```diff
@@ -0,0 +1,113 @@
...
+ * A shortcut key sequence is a string of "+"-separated keyboard modifiers and
+ * keys. The list may contain zero or more modifiers and must contain exactly
+ * one non-modifier key. The key and modifier names are case-insensitive.
+ * For example "Ctrl+Enter", "shift+a". Key names are matched against `KeyboardEvent.key`.
+ *
+ * @param {KeyboardEvent} event
+ * @param {string} shortcut
+ * @return {boolean}
+ */
+export function matchShortcut(event, shortcut) {
```
LMS007: matchShortcut(createEvent('a', { ctrl: true }), 'ctrl+a+b'));  returns `false`
matchShortcut(createEvent('a', { ctrl: true }), 'ctrl+b+a'));  returns `true`

Do you see an issue with mal formatted strings like these two? One idea is to throw an exception if more than 1 key is detected.

src/shared/shortcut.js
```diff
@@ -0,0 +1,113 @@
...
+      onPress(event);
+    }
+  };
+  rootElement.addEventListener('keydown', onKeydown);
+  return () => rootElement.removeEventListener('keydown', onKeydown);
+}
+
+/**
+ * An effect hook that installs a shortcut using `installShortcut` and removes
+ * it when the component is unmounted.
```
LMS007: The note about removing is helpful!

src/shared/test/shortcut-test.js
```diff
@@ -0,0 +1,168 @@
...
+    it('should remove the shortcut when component is unmounted', () => {
+      const onClick = sinon.stub();
+
+      act(() => {
+        render(<Button onClick={onClick} />, container);
+        render(null, container);
+      });
+
+      triggerShortcut({ key: 'a' });
+    });
```
LMS007: Does this test need an assert?

src/shared/shortcut.js
robertknight: Specifying multiple keys in the call to `matchShortcut` would be incorrect usage, so throwing an exception makes sense to me.

src/shared/test/shortcut-test.js
robertknight: Ah, indeed it does. I'll add one.

robertknight:
Thanks for the feedback. I've added additional tests and made specifying multiple non-modifier keys trigger an exception per suggestions.

Issue #1899, opened 2020-03-12
Replace keypress event listeners with keydown


LMS007:
The `keypress` event is deprecated and does not work the same across all browsers. We should replace keypress with keydown in all instances

- [ ] use-element-should-close.js: (Just rename removeKeypressListener to removeKeydownListener)
- [ ] menu.js:
- [ ] menu-test.js:
- [ ] on-activate

PR #1904, opened 2020-03-12
Replace all keypress events with the keydown event

LMS007:
keypress is deprecated
https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event

I have manually tested each event to make sure it still fires
fixes https://github.com/hypothesis/client/issues/1899

### DIFF
--- a/src/sidebar/components/markdown-editor.js
+++ b/src/sidebar/components/markdown-editor.js
@@ -293,7 +293,7 @@ export default function MarkdownEditor({
           dir="auto"
           ref={input}
           onClick={e => e.stopPropagation()}
-          onKeydown={handleKeyDown}
+          onKeyDown={handleKeyDown}
           onInput={e => onEditText({ text: e.target.value })}
           value={text}
         />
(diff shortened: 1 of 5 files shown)


src/sidebar/components/markdown-editor.js
```diff
@@ -293,7 +293,7 @@ export default function MarkdownEditor({
           dir="auto"
           ref={input}
           onClick={e => e.stopPropagation()}
-          onKeydown={handleKeyDown}
+          onKeyDown={handleKeyDown}
```
robertknight: I wasn't sure whether the case mattered here. After a little research I discovered that. known event names are [case-insensitive in Preact](https://github.com/preactjs/preact/blob/33a8c15d78fb9db2f3f090b2781b466b6b86042d/src/diff/props.js#L98). From a quick skim through the React repo and docs, it seems like this casing (`onKeyDown` rather than `onKeydown` as it was before) is more idiomatic.

LMS007: Thanks for looking that up, my hunch was to keep it consistent

PR #1906, opened 2020-03-13
Replace Angular router

### LINKED ISSUE #1878: Replace angular-route in client
As part of the migration away from Angular JS we need to replace angular-route with an alternative means of switching between the 3 routes that the client has (sidebar, single annotation page, stream) based on the URL.

I don't know how actively the "stream" view is actually used, we might be able to drop it if the visitor count is low enough. However the single annotation page is still useful and linked to from the client.

robertknight:
Implement a replacement for the Angular JS router which will enable the
root components of the app to be more easily migrated to Preact in future.
Since the app only has three routes with very simple parameters, we
don't yet need a fully-fledged router implementation but can get by with
some pretty basic URL parsing.

 - Introduce a store module which holds the currently active route
   (`null`, "sidebar", "annotation" or "stream") and any parameters of
   that route.

   This state replaces the `isSidebar` boolean in the `viewer` module,
   since whether the app is the sidebar can be determined by checking
   the active route name.

 - Add a `router` service which handles synchronizing the store route
   with the current URL (via the `sync` method or in response to a
   `popstate` DOM event) or updating the URL and active route in the
   store (via the `navigate` method).

 - Modify the `<hypothesis-app>` component to fetch the active route
   from the store and use it to conditionally render the appropriate
   content component. This replaces the Angular router configuration
   that used to be in `src/sidebar/index.js`

 - Change the places that used to read or update the current route to
   read the active route and route params from the store (`store.{route,
   routeParams}`) and switch the current route by using the `router` service
   (`router.navigate`)

 - Remove the unused angular-route package

Fixes #1878

----

**Testing**

1. Check that the sidebar loads as before on a page with the client embedded
2. Check that links to the HTML page for a single annotation (the timestamps on annotation cards) still work.
3. Check that entering a search query on the single annotation page and pressing Enter navigates to `/stream?q=<query>` and updates the content accordingly
4. Check that http://localhost:5000/stream works (this view has many problems, but check that it at least does the same as before) and that it updates when changing the search query and pressing Enter
5. Check that the content and URL update when pressing the Back/Forward buttons after a navigation in steps 3/4

### DIFF
--- a/src/sidebar/components/annotation-viewer-content.js
+++ b/src/sidebar/components/annotation-viewer-content.js
@@ -35,7 +35,7 @@ function AnnotationViewerContentController(
 ) {
   store.clearAnnotations();
 
-  const id = store.routeParams().id;
+  const annotationId = store.routeParams().id;
 
   this.rootThread = () => rootThread.thread(store.getState());
 
@@ -43,7 +43,7 @@ function AnnotationViewerContentController(
     store.setCollapsed(id, collapsed);
   };
 
-  this.ready = fetchThread(api, id).then(function(annots) {
+  this.ready = fetchThread(api, annotationId).then(function(annots) {
     annotationMapper.loadAnnotations(annots);
 
     const topLevelAnnot = annots.filter(function(annot) {
@@ -64,8 +64,8 @@ function AnnotationViewerContentController(
       store.setCollapsed(annot.id, false);
     });
 
-    if (topLevelAnnot.id !== id) {
-      store.highlightAnnotations([id]);
+    if (topLevelAnnot.id !== annotationId) {
+      store.highlightAnnotations([annotationId]);
     }
   });
 }
--- a/src/sidebar/components/annotation-viewer-content.js
+++ b/src/sidebar/components/annotation-viewer-content.js
@@ -26,7 +26,6 @@ function fetchThread(api, id) {
 
 // @ngInject
 function AnnotationViewerContentController(
-  $routeParams,
   store,
   api,
   rootThread,
@@ -34,9 +33,9 @@ function AnnotationViewerContentController(
   streamFilter,
   annotationMapper
 ) {
-  store.setAppIsSidebar(false);
+  store.clearAnnotations();
 
-  const id = $routeParams.id;
+  const id = store.routeParams().id;
 
   this.rootThread = () => rootThread.thread(store.getState());
 
--- a/src/sidebar/components/stream-content.js
+++ b/src/sidebar/components/stream-content.js
@@ -1,16 +1,12 @@
 // @ngInject
 function StreamContentController(
   $scope,
-  $route,
-  $routeParams,
   annotationMapper,
   store,
   api,
   rootThread,
   searchFilter
 ) {
-  store.setAppIsSidebar(false);
-
   /** `offset` parameter for the next search API call. */
   let offset = 0;
 
@@ -19,6 +15,8 @@ function StreamContentController(
     offset += result.rows.length;
     annotationMapper.loadAnnotations(result.rows, result.replies);
   };
+
+  const currentQuery = () => store.routeParams().q;
 
   /**
    * Fetch the next `limit` annotations starting from `offset` from the API.
@@ -30,7 +28,7 @@ function StreamContentController(
         offset: offset,
         limit: limit,
       },
-      searchFilter.toObject($routeParams.q)
+      searchFilter.toObject(currentQuery())
     );
 
     api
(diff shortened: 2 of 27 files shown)


src/sidebar/components/annotation-viewer-content.js
```diff
@@ -26,17 +26,16 @@ function fetchThread(api, id) {
...
-  $routeParams,
   store,
   api,
   rootThread,
   streamer,
   streamFilter,
   annotationMapper
 ) {
-  store.setAppIsSidebar(false);
+  store.clearAnnotations();
```
robertknight: This ensures that annotations are cleared if switching from the stream back to the annotations view via the browser "Back" button.

lyzadanger:
I've been giving the code a review this morning; I wanted to note that when I pull this branch and try to build, I get a build error; there is still a reference to `angular-route` in [the `vendor-bundles` gulp script](https://github.com/hypothesis/client/blob/6ec410105baab259b7d2fd6eca060feffa0955b5/scripts/gulp/vendor-bundles.js#L12)

When I remove that reference, things build OK. I do notice that I'm having some woes with Websockets in my local (Chrome):

image

I'll keep working on the code review.

src/sidebar/services/router.js
```diff
@@ -0,0 +1,85 @@
+import * as queryString from 'query-string';
+
+/**
+ * A service that manages the association between the route and route parameters
+ * implied by the URL and the corresponding route state in the store.
+ */
+// @ngInject
+export default function router($window, store) {
```
lyzadanger: This is using angular's `$window`, injected here. Do you foresee that we'll simply inject `window` later (using a non-ng injection mechanism)?

robertknight: Maybe. I think even in a post-Angular world it may still be useful to have a `$window` service to provide a test seam that is easier to mock than stubbing methods on the real `window` in tests, like some of our utilities have `window_` arguments.

src/sidebar/components/stream-content.js
```diff
@@ -41,21 +39,27 @@ function StreamContentController(
...
+
+    // Fetch initial batch of annotations.
+    offset = 0;
+    fetch(20);
+  }
+
+  let lastQuery = currentQuery();
+  const unsubscribe = store.subscribe(() => {
+    const query = currentQuery();
+    if (query !== lastQuery) {
```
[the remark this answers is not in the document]
robertknight: We'll generate an API call to the search API with no filter in that case, so all annotations visible to the user are returned.

src/sidebar/index.js
```diff
@@ -103,6 +71,17 @@ function setupApi(api, streamer) {
   api.setClientId(streamer.clientId);
 }
 
+/**
+ * Perform the initial fetch of groups and user profile and then set the initial
+ * route to match the current URL.
+ */
+// @ngInject
+function setupRoute(api, groups, session, router) {
```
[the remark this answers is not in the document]
robertknight: Indeed it doesn't. I've removed that.

src/sidebar/services/router.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Generate a URL for a given route.
+   */
+  function routeUrl(name, params = {}) {
+    let url;
+    const queryParams = { ...params };
+
+    if (name === 'annotation') {
```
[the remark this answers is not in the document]
robertknight: Done. I didn't intend this code to be that extensible because I figure if we're adding a bunch more routes in future then we'll probably add a proper router to map between URL patterns and route names/params.

src/sidebar/services/router.js
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Return the name and parameters of the current route.
+   */
+  function currentRoute() {
+    const path = $window.location.pathname;
+    const pathSegments = path.slice(1).split('/');
+    const params = queryString.parse($window.location.search);
+
+    let route;
+    if (pathSegments[0] === 'a') {
```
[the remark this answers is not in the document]
robertknight: Done

robertknight:
Hi @lyzadanger - I have fixed the WebSocket issue, removed the remaining `angular-route` reference and addressed the other comments you raised. PTAL.

There is one issue I didn't address here which is how existing annotations get cleared when switching between routes. Currently each "content" component is responsible for doing that. I think this logic should be moved elsewhere, but I left that as an exercise for after this initial change lands.

robertknight:
After spending some time looking into this I've found two separate issues with the WebSocket:

1. The first issue, in this PR, is that when the client is loaded on a page where the sidebar is configured to open automatically AND the user is logged in, then `streamer.connect()` and `streamer.reconnect()` both get called on startup. This is wasteful as we're creating a WS connection only to tear it down and reconnect a moment later. However it _shouldn't_ result in an error.
2. The second issue, which already exists, is that a _normal_ WebSocket disconnection initiated by the client when logging in or logging out, results in a spurious error in the browser console. This is actually [an issue in h](https://github.com/hypothesis/h/issues/5969), but it is much easier to [work around](https://github.com/hypothesis/client/pull/1941) in the client for the time being.

On this branch I'll get issue (1) fixed before merging, but leave issue (2) for a separate PR.

robertknight:
I found out why a [WebSocket reconnect](https://github.com/hypothesis/client/pull/1906#issuecomment-602567716) was happening on this branch but not on master. The issue is that, on this branch, the route was being set too early before the initial profile and group fetch had completed. This was happening in Safari and Chrome, but not Firefox. The reason is because Safari and Chrome emit a "popstate" event on page load, causing a route update sooner than expected.

I have added a fix for this in https://github.com/hypothesis/client/pull/1906/commits/d8ac7666607e43eca6829a37eee10e9c5e74973d so that the behavior matches master.

I couldn't reproduce the "popstate" issue in a trivial test case yet, so I'm still digging into exactly when Safari/Chrome do/do not deliver this event on page load. It is documented as an expected behavior on MDN / several Stack Overflow posts though.

src/sidebar/services/router.js
```diff
@@ -61,13 +61,31 @@ export default function router($window, store) {
...
    * URL.
+   *
+   * The first call to this method also registers a listener for future back/forwards
+   * navigation in the browser.
    */
   function sync() {
     const { route, params } = currentRoute();
     store.changeRoute(route, params);
+
+    // Setup listener for back/forward navigation. We do this in `sync()` to avoid
```
lyzadanger: Extreme pedantry alert: _Setup_ is the noun form; I think you want _set up_ here

src/sidebar/services/router.js
```diff
@@ -61,13 +61,31 @@ export default function router($window, store) {
...
+    // the route being changed by eg. a "popstate" emitted by the browser on
```
lyzadanger: You say "e.g." here, but could this happen by any other mechanism (than `popstate`)?

robertknight: Oh, I meant "eg." as in "this is an example of when `popstate` might be emitted before `sync()` is called".

src/sidebar/services/router.js
robertknight: Pedantry appreciated :)

Issue #1854, opened 2020-02-28
Markdown editor to forgo tabstops for keyboard navigation


LMS007:
Per WAI-ARIA standard, we should not use a tab stops for each item in the markdown editor, but instead use a `role="toolbar"` with correctly implement keyboard navigation and aria attributes.

![Screenshot_2](https://user-images.githubusercontent.com/3939074/75595628-06675c80-5a42-11ea-8c4f-c735a3aa12b9.jpg)

We can use this example as a model.
https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html

Relates to
https://gist.github.com/robertknight/a2303f74f38a50736c20fafb7804e49e#annotations:ShkRVlmGEeqJ1ccSlkY2CQ

PR #1910, opened 2020-03-15
Add keyboard navigation to MarkdownEditor to improve a11y

LMS007:
- `arrow left`, `arrow right`, `home` and `end` keys are all supported and modeled after https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html

- Add roving tabIndex value for all toolbar elements.

mark

fixes: https://github.com/hypothesis/client/issues/1854

### DIFF
--- a/src/sidebar/components/markdown-editor.js
+++ b/src/sidebar/components/markdown-editor.js
@@ -1,5 +1,5 @@
 import classnames from 'classnames';
-import { createElement } from 'preact';
+import { createElement, createRef } from 'preact';
 import { useEffect, useRef, useState } from 'preact/hooks';
 import propTypes from 'prop-types';
 
@@ -93,12 +93,14 @@ function handleToolbarCommand(command, inputEl) {
 }
 
 function ToolbarButton({
+  buttonRef,
   disabled = false,
   iconName,
   label = null,
   onClick,
   shortcutKey,
-  title,
+  tabIndex,
+  title = '',
 }) {
   let tooltip = title;
   if (shortcutKey) {
@@ -116,6 +118,8 @@ function ToolbarButton({
       disabled={disabled}
       onClick={onClick}
       title={tooltip}
+      tabIndex={tabIndex}
+      ref={buttonRef}
     >
       {iconName && (
         <SvgIcon
@@ -129,26 +133,142 @@ function ToolbarButton({
 }
 
 ToolbarButton.propTypes = {
+  buttonRef: propTypes.object.isRequired,
   disabled: propTypes.bool,
   iconName: propTypes.string,
   label: propTypes.string,
-  onClick: propTypes.func,
+  onClick: propTypes.func.isRequired,
   shortcutKey: propTypes.string,
+  tabIndex: propTypes.number.isRequired,
   title: propTypes.string,
 };
 
+/**
+ * @typedef {string} ButtonID
+ * A unique string that can be one of one of:
+ * [bold, italic, quote, link, image, math, numlist, list, preview, help]
+ */
+
+/**
+ * An array of toolbar elements with a roving tab stop. Left and right
+ * array keys can be used to change focus of the elements. Home and end
+ * keys will navigate to the first and last elements respectively.
+ *
+ * Canonical example
+ * https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html
+ */
 function Toolbar({ isPreviewing, onCommand, onTogglePreview }) {
+  const buttonIds = {
+    // Ordered buttons
+    bold: 0,
+    italic: 1,
+    quote: 2,
+    link: 3,
+    image: 4,
+    math: 5,
+    numlist: 6,
+    list: 7,
+    help: 8,
+    preview: 9,
+
+    // Total button count
+    maxId: 10,
+  };
+
+  // Keep track of a roving index. The active roving tabIndex
+  // is set to 0, and all other elements are set to -1.
+  const [rovingElement, setRovingElement] = useState(0);
+
+  // An array of refs
+  const buttonRefs = useRef([]).current;
+  if (buttonRefs.length === 0) {
+    // Initialize buttonRefs on first render only
+    for (let i = 0; i <= buttonIds.maxId; i++) {
+      buttonRefs.push(createRef());
+    }
+  }
+
+  /**
+   * Sets the element to be both focused and the active roving index.
+   *
+   * @param {number} index - Ordered index that matches the element
+   */
+  const setFocusedElement = index => {
+    setRovingElement(index);
+    buttonRefs[index].current.focus();
+  };
+
+  /**
+   * Handles left and right arrow navigation as well as home and end
+   * keys so the user may navigate the toolbar without multiple tab stops.
+   */
+  const handleKeyDown = e => {
+    let lowerLimit = 0;
+    const upperLimit = buttonIds.maxId - 1;
+    if (isPreviewing) {
+      // When isPreviewing is true, only allow navigation of
+      // the last 2 items.
+      lowerLimit = buttonIds.help;
+    }
+    let newFocusedElement = null;
+    switch (e.key) {
+      case 'ArrowLeft':
+        if (rovingElement <= lowerLimit) {
+          newFocusedElement = upperLimit;
+        } else {
+          newFocusedElement = rovingElement - 1;
+        }
+        break;
+      case 'ArrowRight':
+        if (rovingElement >= upperLimit) {
+          newFocusedElement = lowerLimit;
+        } else {
+          newFocusedElement = rovingElement + 1;
+        }
+        break;
+      case 'Home':
+        newFocusedElement = lowerLimit;
+        break;
+      case 'End':
+        newFocusedElement = upperLimit;
+        break;
+    }
+    if (newFocusedElement !== null) {
+      setFocusedElement(newFocusedElement);
+      e.preventDefault();
+    }
+  };
+
+  /**
+   * Returns the tab index value for a given element.
+   * Each element should be set to -1 unless its the
+   * active roving index, in which case it will be 0.
+   *
+   * @param {ButtonID} id
+   * @return {number} index - An index from `buttonIds`
+   */
+  const getTabIndex = index => {
+    if (rovingElement === index) {
+      return 0;
+    } else {
+      return -1;
+    }
+  };
+
   return (
     <div
       className="markdown-editor__toolbar"
       role="toolbar"
       aria-label="Markdown editor toolbar"
+      onKeyDown={handleKeyDown}
     >
       <ToolbarButton
         disabled={isPreviewing}
         iconName="format-bold"
         onClick={() => onCommand('bold')}
         shortcutKey={SHORTCUT_KEYS.bold}
+        buttonRef={buttonRefs[buttonIds.bold]}
+        tabIndex={getTabIndex(buttonIds.bold)}
         title="Bold"
       />
       <ToolbarButton
@@ -156,6 +276,8 @@ function Toolbar({ isPreviewing, onCommand, onTogglePreview }) {
         iconName="format-italic"
         onClick={() => onCommand('italic')}
         shortcutKey={SHORTCUT_KEYS.italic}
+        buttonRef={buttonRefs[buttonIds.italic]}
+        tabIndex={getTabIndex(buttonIds.italic)}
         title="Italic"
       />
       <ToolbarButton
@@ -163,6 +285,8 @@ function Toolbar({ isPreviewing, onCommand, onTogglePreview }) {
         iconName="format-quote"
         onClick={() => onCommand('quote')}
         shortcutKey={SHORTCUT_KEYS.quote}
+        buttonRef={buttonRefs[buttonIds.quote]}
+        tabIndex={getTabIndex(buttonIds.quote)}
         title="Quote"
       />
       <ToolbarButton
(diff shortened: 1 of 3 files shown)


src/sidebar/components/markdown-editor.js
```diff
@@ -134,69 +136,180 @@ ToolbarButton.propTypes = {
...
+  const maxElementIndex = 9;
+
+  // Keep track of a roving index. The active roving tabIndex
+  // is set to 0, and all other elements are set to -1.
+  const [rovingElement, setRovingElement] = useState(0);
+
+  // An array of refs. Note we can't use `useRef` in a loop.
+  const elementsRefs = [];
+  for (let i = 0; i <= maxElementIndex; i++) {
+    elementsRefs.push(createRef());
```
LMS007: I could use some feedback on this. I think this is the "right" way to do it, but this is the first time I needed to have hooks, but in a loop. So instead of using useRef which won't work, I create an array of new refs.

src/sidebar/components/markdown-editor.js
```diff
@@ -134,69 +136,180 @@ ToolbarButton.propTypes = {
...
       role="toolbar"
       aria-label="Markdown editor toolbar"
+      onKeyDown={handleKeyDown}
     >
       <ToolbarButton
         disabled={isPreviewing}
         iconName="format-bold"
         onClick={() => onCommand('bold')}
         shortcutKey={SHORTCUT_KEYS.bold}
+        ref={elementsRefs[0]}
```
LMS007: The simple dumb way of doing this is to put the link-age directly into the JSX like this as opposed to some sort of array of state that we reference. I think this way ends up being a bit more simple, but we  need to be careful about the index values because they are all hard coded here. e.g. if one changes, we need to change everyone after that.

src/styles/sidebar/components/markdown-editor.scss
```diff
@@ -1,3 +1,4 @@
+@use "../../mixins/buttons";
```
LMS007: This fixes an unrelated issue. When you click the tool buttons with the mouse, the blue focus ring shows up. This fixes that. I think it might be the last place in our code where we have that?  A further improvement could be to revisit this when we convert more things to use `<Button>`

src/sidebar/components/markdown-editor.js
```diff
@@ -134,69 +136,180 @@ ToolbarButton.propTypes = {
...
+   * Set the element to be focused and set it to be
+   * the active roving index.
+   *
+   * @param {number} index - Ordered index that matches the element
+   */
+  const setFocusedElement = index => {
+    setRovingElement(index);
+    try {
+      // preact component DOM elements are under 'base'
+      elementsRefs[index].current.base.focus();
```
robertknight: One approach to avoid needing the component case would be to pass the ref down to `ToolbarButton` via a different prop (eg. `buttonRef`) which `ToolbarButton` then passes as the `ref` directly to the `<button>`.

```js
<ToolbarButton buttonRef={elementsRefs...}/>

function ToolbarButton({ buttonRef, ... }) {
  return <button ref={buttonRef} .../>;
}
```

This avoids coupling to a Preact-specific API. I've seen some discussion (don't have a link ATM, but can try to find it) that in future in React/Preact it might be possible to access the `ref` prop directly from within the component so that it ends up referring to whatever the component wants it to refer to (eg. the `<button>`) rather than the component itself.

src/sidebar/components/test/markdown-editor-test.js
```diff
@@ -301,6 +305,303 @@ describe('MarkdownEditor', () => {
...
+      // on elements shall be "-1".
+      assert.equal(
+        wrapper.find('ToolbarButton[tabIndex="0"]').length +
+          wrapper.find('a[tabIndex="0"]').length,
+        1
+      );
+    };
+
+    context('when `isPreviewing` is false', () => {
+      it('changes focus to the second button when pressing `right` arrow', () => {
```
robertknight: There is quite a lot of repetition in these tests. Perhaps some of this could be factored out to make it easier to follow at a high level. For example, extracting a helper for checking if the focused button's label matches some string.

src/styles/sidebar/components/markdown-editor.scss
robertknight: Some of the specific styles (eg. `border: none`) may no longer be needed with this change.

src/sidebar/components/markdown-editor.js
```diff
@@ -134,69 +136,180 @@ ToolbarButton.propTypes = {
...
+ * Canonical example
+ * https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html
+ */
 function Toolbar({ isPreviewing, onCommand, onTogglePreview }) {
+  // Total number of elements in the toolbar minus 1.
+  const maxElementIndex = 9;
+
+  // Keep track of a roving index. The active roving tabIndex
+  // is set to 0, and all other elements are set to -1.
+  const [rovingElement, setRovingElement] = useState(0);
```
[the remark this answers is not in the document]
LMS007: ya, that was my reason for "element" :\

okay: )

src/sidebar/components/markdown-editor.js
LMS007: Yes, I was going to do that, then forgot! That will indeed be better.

src/sidebar/components/markdown-editor.js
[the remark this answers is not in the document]
LMS007: Doing a mapping like `buttonIds` was one my initial thoughts, but it does make keeping track of the roving index more difficult.

So for example, maybe the roving index is the key into the buttonIds? But its not obvious which one is next or prev.  The simple js object does not preserve order in every case and even if it did in this, its probably not a good standard to assume order so I would want to use the `Map` object or keep track of a secondary ordering array -- more complex.  Which is why I dumped the index values directly into the JSX -- simple and dumb, but works. Then I wrote that test that asserts the order in case it was ever tampered.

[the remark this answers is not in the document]
robertknight: >  The simple js object does not preserve order in every case and even if it did in this, its probably not a good standard to assume order so I would want to use the Map object or keep track of a secondary ordering array -- more complex

By "does not preserve order" are you referring to the iteration order over keys in an object? It used to be the case that this was theoretically implementation-dependent. In practice all browsers ended up making it deterministic because code relied on it and this order was [made part of the ES6 spec](https://2ality.com/2015/10/property-traversal-order-es6.html#traversing-the-own-keys-of-an-object). Incidentally the same is true of Python dicts, but that only changed much more recently.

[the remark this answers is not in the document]
LMS007: Awe, didn't know that about ES6 changes. Also, I Think I was overthinking this as the values in that buttonsId serves as the order, duh, and the rovingIndex can remain largely the same as its implemented.

src/sidebar/components/markdown-editor.js
LMS007: ya, good catch, clearly that was an issue, I suppose we could even use a useState here, but perhaps wrapping it in a ref is slightly more appropriate.

src/sidebar/components/markdown-editor.js
LMS007: I did a bit more indirection if thats okay?, but same thing otherwise.

src/styles/sidebar/components/markdown-editor.scss
LMS007: good catrch

src/sidebar/components/test/markdown-editor-test.js
LMS007: I'll update this PR again later tonight, I don't want to delay anymore on this, but I think you're right. I did remove a few tests I felt were un-needed but added 2 more than seemed logically appropriate. I also removed the `commanFn` indirection as it was the same for each test in the set.

src/sidebar/components/markdown-editor.js
```diff
@@ -129,74 +133,213 @@ function ToolbarButton({
...
+    }
+  }
+
+  /**
+   * Returns the unique button ref associated with the id.
+   *
+   * @param {ButtonID} id
+   * @return {Ref}
+   */
+  const getButtonRef = id => {
```
robertknight: Ah, I was thinking that you would pass the index value rather than the ID directly, ie. using `getButtonRef(buttonIds.button)`. In other words, the `buttonIds` object serves purely to group a bunch of number constants together.

One advantage of such a change is that it would be easier for a developer or static analysis tool to verify that `buttonIds.<some property name>` is a valid property name.

src/sidebar/components/markdown-editor.js
```diff
@@ -129,74 +133,213 @@ function ToolbarButton({
...
+
+  /**
+   * Returns the tab index value for a given element.
+   * Each element should be set to -1 unless its the
+   * active roving index, in which case it will be 0.
+   *
+   * @param {ButtonID} id
+   * @return {number} `tabIndex` value
+   */
+  const getTabIndex = id => {
```
robertknight: As above, I'd suggest a little indirection can be avoided here by passing the numeric ID directly to `getTabIndex`, eg. `getTabIndex(buttonIds.bold)`.

src/sidebar/components/markdown-editor.js
```diff
@@ -129,74 +133,213 @@ function ToolbarButton({
...
+        e.preventDefault();
+        break;
+      case 'Home':
+        setFocusedElement(lowerLimit);
+        e.preventDefault();
+        break;
+      case 'End':
+        setFocusedElement(upperLimit);
+        e.preventDefault();
+        break;
```
robertknight: There is a little bit of duplication between the branches here. One pattern I quite like when faced with a long switch or `if/else` chain where each branch executes the same logic but with different values is to do something like the following:

```js
let newFocusedElement = null;
switch (e.key) {
  // Branches set `newFocusedElement` as appropriate.
}
if (newFocusedElement !== null) {
  setFocusedElement(newFocusedElement);
  e.preventDefault();
}
```

The advantage is that it makes it easier to see that each branch is supposed to be doing the same thing and avoids forgetting to put eg. `e.preventDefault()` in one branch, or if you decide to add `e.stopPropagation()` as well, that would now be easier to add.

src/sidebar/components/test/markdown-editor-test.js
```diff
@@ -301,6 +305,215 @@ describe('MarkdownEditor', () => {
...
+        // preview is the last button
+        matchesFocusedText('Preview');
+        testRovingIndex();
+      });
+
+      it('changes focus circularly to the right', () => {
+        pressKey('ArrowLeft'); // move to the end node
+        pressKey('ArrowRight'); // move back to the start
+        assert.isTrue(
+          document.activeElement.getAttribute('title').indexOf('Bold') >= 0
```
robertknight: This could use `matchesFocusedTitle`

src/sidebar/components/test/markdown-editor-test.js
```diff
@@ -301,6 +305,215 @@ describe('MarkdownEditor', () => {
...
+            resultFn: () => {
+              matchesFocusedText('Preview');
+            },
+          },
+          {
+            resultFn: () => {
+              // back to the start
+              matchesFocusedTitle('Bold');
+            },
+          },
```
robertknight: Just as a suggestion, now that this is simplified you could streamline the array a bit further by making each item an object with `title` or `text` properties and then calling `matchesFocusedTitle` or `matchesFocusedText` accordingly:

```js
[...,
  { text: 'Preview' },
  { title: 'Bold' }
].forEach(({ text, title )) => {
...
}
```

The advantage is just that it saves a few lines and makes the whole test quicker to read.

src/sidebar/components/test/markdown-editor-test.js
```diff
@@ -301,6 +305,215 @@ describe('MarkdownEditor', () => {
...
+        const previewButton = wrapper
+          .find('button')
+          .filterWhere(el => el.text() === 'Write');
+        previewButton.simulate('focus');
+        pressKey('Home');
+      });
+
+      it('changes focus to the last element when pressing `end`', () => {
+        pressKey('End'); // move to the end node
+        assert.isTrue(document.activeElement.innerHTML.indexOf('Write') >= 0);
```
robertknight: This could use `matchesFocusedText`

src/sidebar/components/test/markdown-editor-test.js
```diff
@@ -301,6 +305,215 @@ describe('MarkdownEditor', () => {
...
+        testRovingIndex();
+      });
+
+      it('changes focus to the first element when pressing `home`', () => {
+        pressKey('ArrowRight'); // move focus off first button
+        pressKey('Home');
+        assert.isTrue(
```
robertknight: This could use `matchesFocusedTitle`

src/sidebar/components/markdown-editor.js
LMS007: gotcha! sorry, I missed that last round

src/sidebar/components/markdown-editor.js
LMS007: roger that

src/sidebar/components/markdown-editor.js
LMS007: right, if the keys change, its a syntax error rather than a less obvious logical error.

src/sidebar/components/test/markdown-editor-test.js
LMS007: great suggestion, thanks

PR #1931, opened 2020-03-18
Add button to adder toolbar to show annotations for selection

robertknight:
Add a button to the adder toolbar to show the existing annotations that
relate to the selected text, along with a keyboard shortcut (the "s"
key) to trigger it. This button only appears if the selected text overlaps at least one highlight.

This provides a way to filter annotations in the sidebar to those that
correspond to a particular highlight or region of text using only the
keyboard (we assume that the user already has some method of making the text selection in the first place). It might also be useful for other users in case the highlight is hard to click (eg. if the highlight has a small target area due to font size, text length etc.)

_Edit: Final design implementing Lyza's suggestions:_

adder-show-button-revised

----

_Edit: Notes on initial prototype:_

**Note: The layout, icon and label here are not polished. They are just meant to illustrate the idea.**

Screenshot 2020-03-18 09 19 00

At this point I'm looking for feedback on the design:

- Is this the right approach to providing a keyboard-accessible way to filter annotations to those that correspond to a particular highlight?
- Should this have an icon in the toolbar or just a keyboard shortcut?
- If there is an icon, what should the icon be?
- If there is an icon, what should the label be?
- A specific query I have with screen reader usage is how the user will navigate back from the sidebar to the page once they have read and maybe replied to the annotations.  Maybe it will suffice to rely on the screen reader's existing shortcuts and navigation facilities?

### DIFF
--- a/src/annotator/adder.js
+++ b/src/annotator/adder.js
@@ -52,6 +52,14 @@ function nearestPositionedAncestor(el) {
 }
 
 /**
+ * @typedef AdderOptions
+ * @prop {() => any} onAnnotate - Callback invoked when "Annotate" button is clicked
+ * @prop {() => any} onHighlight - Callback invoked when "Highlight" button is clicked
+ * @prop {(annotations: Object[]) => any} onShowAnnotations -
+ *   Callback invoked when  "Show" button is clicked
+ */
+
+/**
  * Container for the 'adder' toolbar which provides controls for the user to
  * annotate and highlight the selected text.
  *
@@ -66,8 +74,8 @@ export class Adder {
    *
    * The adder is initially hidden.
    *
-   * @param {Element} container - The DOM element into which the adder will be created
-   * @param {Object} options - Options object specifying `onAnnotate` and `onHighlight`
+   * @param {HTMLElement} container - The DOM element into which the adder will be created
+   * @param {AdderOptions} options - Options object specifying `onAnnotate` and `onHighlight`
    *        event handlers.
    */
   constructor(container, options) {
@@ -98,6 +106,17 @@ export class Adder {
     this._arrowDirection = 'up';
     this._onAnnotate = options.onAnnotate;
     this._onHighlight = options.onHighlight;
+    this._onShowAnnotations = options.onShowAnnotations;
+
+    /**
+     * Annotation objects associated with the current selection. If non-empty,
+     * a "Show" button appears in the toolbar. Clicking the button calls the
+     * `onShowAnnotations` callback with the current value of `annotationsForSelection`.
+     *
+     * @type {Object[]}
+     */
+    this.annotationsForSelection = [];
+
     this._render();
   }
 
@@ -198,15 +217,18 @@ export class Adder {
       switch (command) {
         case 'annotate':
           this._onAnnotate();
+          this.hide();
           break;
         case 'highlight':
           this._onHighlight();
+          this.hide();
+          break;
+        case 'show':
+          this._onShowAnnotations(this.annotationsForSelection);
           break;
         default:
           break;
       }
-
-      this.hide();
     };
 
     render(
@@ -214,6 +236,7 @@ export class Adder {
         isVisible={this._isVisible}
         arrowDirection={this._arrowDirection}
         onCommand={handleCommand}
+        annotationCount={this.annotationsForSelection.length}
       />,
       this._shadowRoot
     );
--- a/src/annotator/components/adder-toolbar.js
+++ b/src/annotator/components/adder-toolbar.js
@@ -5,7 +5,7 @@ import propTypes from 'prop-types';
 import { useShortcut } from '../../shared/shortcut';
 import SvgIcon from '../../shared/components/svg-icon';
 
-function ToolbarButton({ icon, label, onClick, shortcut }) {
+function ToolbarButton({ badgeCount, icon, label, onClick, shortcut }) {
   useShortcut(shortcut, onClick);
 
   const title = shortcut ? `${label} (${shortcut})` : label;
@@ -16,14 +16,18 @@ function ToolbarButton({ icon, label, onClick, shortcut }) {
       onClick={onClick}
       title={title}
     >
-      <SvgIcon name={icon} />
+      {icon && <SvgIcon name={icon} />}
+      {typeof badgeCount === 'number' && (
+        <span className="annotator-adder-actions__badge">{badgeCount}</span>
+      )}
       <span className="annotator-adder-actions__label">{label}</span>
     </button>
   );
 }
 
 ToolbarButton.propTypes = {
-  icon: propTypes.string.isRequired,
+  badgeCount: propTypes.number,
+  icon: propTypes.string,
   label: propTypes.string.isRequired,
   onClick: propTypes.func.isRequired,
   shortcut: propTypes.string,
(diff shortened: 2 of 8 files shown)


lyzadanger:
**Disclaimer: The following images are rough, and do not represent a true visual design.**

Given that the new adder-toolbar option to view the relevant annotations is contextual (i.e. not always present), I think it could be helpful to "separate" it visually from the other, always-present, creation-related options. In these examples, that is handled by a vertical separator.

Given that we have the number of matching annotations that apply at-hand, I think it would be nice to surface that number, as a badge or similar, on whatever icon we choose to render here.

The current toolbar buttons are using the icon font, which limits me in ability to prototype what this could look like, so please forgive the jankiness of these mockups. Even the colors ain't right. What I'm trying to illustrate here is the notion of number-badging options.

Screen Shot 2020-03-24 at 9 44 24 AM

In this first version above, I certainly wouldn't actually advise using the same underlying annotation icon; we'd at least want to flip it on its x axis for visual differentiation.

image

In this second option, the "icon" (totally faked here) is not "quite right", but the idea is that it communicates a "list" of things (it is based on Jared's "list" icon in the pattern library). The badge is too small here; it would need work.

Another riff:

image

robertknight:
Thanks for the suggestions Lyza. They look good to me. I'll work on something based on that.

robertknight:
I have updated this WIP PR based on [Lyza's suggestions](https://github.com/hypothesis/client/pull/1931#issuecomment-603260484) above. For the visual design I've gone with something based on the first screenshot in that comment because it allows more space for the badge count which makes it easier to read. I have also bumped up the font size of the labels from 11 to 13px and tweaked the margins a little for readability.

Screenshot 2020-03-25 10 38 09

When the selected text has no highlighted areas, it looks like this:

Screenshot 2020-03-25 10 38 01

robertknight:
@lyzadanger asked to see what it looks like when there are double-digit numbers of annotations, so here is a screenshot of that:

Screenshot 2020-03-26 10 12 35

Even hundreds looks fine, although we might just want to show "99+" to avoid needing to deal with that:

Screenshot 2020-03-26 10 14 17

lyzadanger:
I just pulled this branch again to play around with some visual tweaks and I noticed what might be a bug. Here's a screenshot of the toolbar saying there are 4 matching annotations, when there are in fact 2: 

image

Is it counting something it shouldn't? Things in other groups, e.g.?

robertknight:
> I just pulled this branch again to play around with some visual tweaks and I noticed what might be a bug. Here's a screenshot of the toolbar saying there are 4 matching annotations, when there are in fact 2:

Ah. This happens when a single annotation's highlight has to be split into multiple `<hypothesis-highlight>` elements. I've resolved this so that the count refers to the number of unique annotations referred to by highlights. I also opened a [separate PR](https://github.com/hypothesis/client/pull/1965) to address one case where we unnecessarily create multiple `<hypothesis-highlight>` elements.

lyzadanger:
While I am with you on the need for the badge count to be clearly readable, I feel like the combination of `13px` and `bold` cause the number to look "visually heavy" in comparison to the other icons (especially the annotate icon):

image

Here is at `12px`, still `bold`, and removing the `1px` top padding rule:

image

I think it's still readable, and just a little less weight-imbalanced. What think you?

lyzadanger:
Oop. Just noticed a console error on this branch: 
image

robertknight:
This is ready for review now. To recap changes from the initial prototype:

- The "View annotations" button was renamed to "Show" and the icon was replaced with a count of annotations
- The shortcut key was changed from "v" to "s" to match the new button label
- Tests added, styles polished

src/annotator/adder.js
```diff
@@ -198,22 +217,26 @@ export class Adder {
...
         case 'annotate':
           this._onAnnotate();
+          this.hide();
           break;
         case 'highlight':
           this._onHighlight();
+          this.hide();
+          break;
+        case 'show':
+          this._onShowAnnotations(this.annotationsForSelection);
```
lyzadanger, line 227: Because the `show` option doesn't subsequently `hide` like the other two commands, it leaves a visible focus ring on the adder: 

Screen Shot 2020-04-03 at 9 02 03 AM

Is this intentional?

src/annotator/components/adder-toolbar.js
```diff
@@ -16,14 +16,18 @@ function ToolbarButton({ icon, label, onClick, shortcut }) {
       onClick={onClick}
       title={title}
     >
-      <SvgIcon name={icon} />
+      {icon && <SvgIcon name={icon} />}
+      {typeof badgeCount === 'number' && (
```
lyzadanger, line 20: Given that `propTypes` defines this prop as being a  `number`, can you help me understand the need for the `typeof` check here? Sorry if I missed it in past discussion.

src/annotator/range-util.js
```diff
@@ -129,3 +129,35 @@ export function selectionFocusRect(selection) {
     return textBoxes[textBoxes.length - 1];
   }
 }
+
+/**
+ * Retrieve a set of items associated with nodes in a given range.
+ *
+ * @template T
```
lyzadanger, line 136: Could you be so kind as to remind me where the documentation is for `@typedef`-enabled jsdoc (or whatever this combination is; I've forgotten!) so I can understand the `@template` tag? Google is not my friend this morning.

src/annotator/range-util.js
```diff
@@ -129,3 +129,35 @@ export function selectionFocusRect(selection) {
     return textBoxes[textBoxes.length - 1];
   }
 }
+
+/**
+ * Retrieve a set of items associated with nodes in a given range.
```
lyzadanger, line 134: Sorry, I've lost the plot here a bit. What is an "item?" Perhaps this summary doc string could be expanded a little?

src/styles/annotator/adder.scss
```diff
@@ -97,20 +97,12 @@ $adder-transition-duration: 80ms;
...
-  color: var.$grey-semi !important;
-}
-
 .annotator-adder-actions__button {
   box-shadow: none;
   font-family: h;
   font-size: 18px;
-  background: transparent !important;
-  color: var.$grey-mid !important;
+  background: transparent;
```
lyzadanger, line 104: This is a nit, but it's nice to have layout rules before others, especially in selectors that have lots of rules.

src/annotator/components/adder-toolbar.js
robertknight: The prop is optional, so it may be `undefined`. I could have set a default value of 0 and used `{badgeCount > 0 && ...}`.

src/annotator/range-util.js
robertknight: https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html

src/annotator/range-util.js
robertknight: "Item" can be any data that the caller which to associate with / compute from a node, as determined by the callback.

src/styles/annotator/adder.scss
robertknight: We don't have any kind of consistency or even internal guidelines around property ordering at present. Ideally we would use a tool (maybe [stylelint](https://github.com/stylelint/stylelint)) or auto-formatter to enforce as appropriate, but even some kind of agreed convention would be a first step.

src/annotator/adder.js
robertknight: No. I'll add the same mixin that we use to avoid unwanted focus rings after mouse (not keyboard) input on other buttons.

PR #1933, opened 2020-03-19
Tune virtual thread list debouncing

robertknight:
When scrolling through a long list of annotations quickly, the sidebar could display blank space for a long period of time. This was happening because the debounced update function waited for a period of at least 20ms with no "scroll" events, and a gap of that length could take some time to appear if there was a queue of scroll events being delivered.

Fix the issue by adding an upper bound of 100ms (via a `maxWait` parameter to `debounce`) after a scroll event before recalculating the list of visible threads.

Also make sure to cancel any pending updates after the `detach` method is called.

**Master:**

scrolling-master

**This branch:**

scrolling-tuned

### DIFF
--- a/src/sidebar/test/virtual-thread-list-test.js
+++ b/src/sidebar/test/virtual-thread-list-test.js
@@ -36,8 +36,17 @@ describe('VirtualThreadList', function() {
   }
 
   beforeEach(() => {
+    const fakeDebounce = callback => {
+      const debounced = () => {
+        // Update synchronously instead of really debouncing.
+        callback();
+      };
+      debounced.cancel = sinon.stub();
+      return debounced;
+    };
+
     $imports.$mock({
-      'lodash.debounce': fn => fn,
+      'lodash.debounce': fakeDebounce,
     });
   });
 
--- a/src/sidebar/virtual-thread-list.js
+++ b/src/sidebar/virtual-thread-list.js
@@ -34,8 +34,6 @@ export default class VirtualThreadList extends EventEmitter {
   constructor($scope, window_, rootThread, options) {
     super();
 
-    const self = this;
-
     this._rootThread = rootThread;
 
     this._options = Object.assign({}, options);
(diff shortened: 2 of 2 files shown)


src/sidebar/test/virtual-thread-list-test.js
```diff
@@ -36,8 +36,17 @@ describe('VirtualThreadList', function() {
...
 
   beforeEach(() => {
+    const fakeDebounce = callback => {
+      const debounced = () => {
+        // Update synchronously instead of really debouncing.
+        callback();
+      };
+      debounced.cancel = sinon.stub();
+      return debounced;
+    };
```
robertknight: This change is just about adding a `cancel` method to the result of the `debounce` mock.

src/sidebar/virtual-thread-list.js
```diff
@@ -46,16 +44,22 @@ export default class VirtualThreadList extends EventEmitter {
...
-      self.calculateVisibleThreads();
-      $scope.$digest();
-    }, 20);
+    const debouncedUpdate = debounce(
+      () => {
+        this.calculateVisibleThreads();
+        $scope.$digest();
+      },
+      10,
+      { maxWait: 100 }
```
LMS007: wondering if `leading=true` could make it even snappier

src/sidebar/virtual-thread-list.js
```diff
@@ -46,16 +44,22 @@ export default class VirtualThreadList extends EventEmitter {
...
+    );
+
     this.scrollRoot.addEventListener('scroll', debouncedUpdate);
     this.window.addEventListener('resize', debouncedUpdate);
 
     this._detach = function() {
       this.scrollRoot.removeEventListener('scroll', debouncedUpdate);
       this.window.removeEventListener('resize', debouncedUpdate);
+      debouncedUpdate.cancel();
```
LMS007: good catch!

robertknight:
> It seems like it was waiting until the scroll events stopped before firing again, in other words it was getting too many events and was waiting for a pause of 20ms 

Right. My understanding is that if you do lots of rapid scrolling, you'd get a series of `"scroll"` events being delivered one after another with maybe 1-5ms between them, such that it would take a long time (much more than 20ms) before there was a large enough gap between individual events for `debounce` to register the trailing edge of a sequence of calls and trigger the update.

The reason for using _trailing edge_ rather than _leading edge_ triggering is because we want the calculated state to reflect the _final_ scroll position (after the last "scroll" event) rather than the position after the first "scroll" event. The change here forces a refresh after 100ms even if we haven't seen the trailing edge yet.

PR #1934, opened 2020-03-19
Call requestGroups to get groups filter

### LINKED ISSUE #1908: Get LMS groups from the LMS app's requestGroups method
Part of the LMS sections feature: https://github.com/hypothesis/lms/milestone/10

Depends on https://github.com/hypothesis/lms/issues/1555

Change the client to retrieve the list of groups from the LMS app's new `'requestGroups'` method (https://github.com/hypothesis/lms/issues/1555), instead of getting them from the config returned by the `'requestConfig'` method.

## Background

See <https://github.com/hypothesis/lms/issues/1555> for background information on postMessage-JSON-RPC communications between the Hypothesis client and the LMS app (under _Background on postMessage-JSON-RPC_).

See <https://github.com/hypothesis/lms/milestone/10> for background on the Canvas sections feature.

## Problem

In the future the LMS app is going to have to make requests to Canvas's sections API before it can generate the Hypothesis groups for an assignment. Specifically, the LMS frontend is going to have to [call a new LMS backend sync API](https://github.com/hypothesis/lms/milestone/14) and receive the groups from this sync API's response. This would lead to a significant slow down in the assignment launch user experience for all LMS users, if the LMS app had to wait until it had this sync API response before it could inject the Via iframe and begin loading Via and the client.

We believe this will make all assignment launches **about 1.3 seconds slower** than they already are (and they're pretty slow already). It's the difference between sequences 1 and 4 in this sequence diagram:

sections_times

## Solution

We're going to:

1. **Separate issue:** add a new `'requestGroups'` method to the LMS app's JSON-RPC server, that returns the list of groups. That's this ticket: https://github.com/hypothesis/lms/issues/1555

2. **This issue:** change the client to send _two_ JSON-RPC requests to the LMS app on boot up, not one:

   1. The `'requestConfig'` request that the client already sends
   2. A new `'requestGroups'` request

   The `'requestGroups'` request can be sent at the same time as the `'requestConfig'` request, but the `'requestGroups'` _response_ may not come back as quickly because the LMS app has to wait for its sync API response before it can send the `'requestGroups'` response to the client. The client:

   1. Must retrieve the groups from the `'requestGroups'` response, not from the `'requestConfig'` response. The client mustn't read the groups from the `'requestConfig'` response at all anymore
   2. Must be able to wait a long time for the `'requestGroups'` response, and then continue loading once it finally does receive the groups
   3. Should be able to do as much of its loading as possible using only the `'requestConfig'` response, without the groups

3. **Separate issue:** remove the groups from the config returned by the LMS app's `'requestConfig'` method, now that the client no longer retrieves the groups from this config. That's this issue: https://github.com/hypothesis/lms/issues/1556

Related to https://github.com/hypothesis/client/issues/1793

LMS007:
Proposal for fetching groups async. 

**Update**
Added a function called `mutateAsyncGroups`  which may mutate the config object if groups filter exists. I considered making this function not mutate the object but that seemed more difficult to later merge back into the config because its a deep confg value. I see no side effects here to mutating the config object because its all internal to the fetch-config process anyway.

`mutateAsyncGroups` has 3 cases for a `groups` value:

1.  Missing or falsely `groups`: _Don't do anything_
2. `groups` exist: _Return a resolved promise_
3. `groups` is equal to '$rcp:requestGroups': _return a promise that fetches the filter array via RPC_

The gist here is that if there is a `groups` filter, the groups service can treat it as a promise in that case so we don't have to have a condition in the groups service to see if it's an instanceof Promise.  Its just always undefined or a Promise

**Array of services**
Because services is an array, I just send the service index back as a parameter of the `requestGroups` call. The method in LMS does not accept params nor does the RPC server so that would have to be filled in later if we wanted to handle more than one service, but the idea is this would work if there was more than 1 service later one.

fixes #1908

### DIFF
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -25,6 +25,7 @@ export default function groups(
   serviceUrl,
   session,
   settings,
+  toastMessenger,
   auth
 ) {
   const svc = serviceConfig(settings);
@@ -87,7 +88,7 @@ export default function groups(
    * @param {string|null} directLinkedGroupId
    * @return {Group[]}
    */
-  function filterGroups(
+  async function filterGroups(
     groups,
     isLoggedIn,
     directLinkedAnnotationGroupId,
@@ -106,14 +107,20 @@ export default function groups(
         directLinkedGroupId = null;
       }
     }
-
     // If service groups are specified only return those.
     // If a service group doesn't exist in the list of groups don't return it.
     if (svc && svc.groups) {
-      const focusedGroups = groups.filter(
-        g => svc.groups.includes(g.id) || svc.groups.includes(g.groupid)
-      );
-      return focusedGroups;
+      try {
+        // The groups may not be available if they are being fetched via RPC.
+        // See mutateAsyncGroups in fetch-config.js
+        const focusedGroups = await svc.groups;
+        const filteredGroups = groups.filter(
+          g => focusedGroups.includes(g.id) || focusedGroups.includes(g.groupid)
+        );
+        return filteredGroups;
+      } catch (e) {
+        toastMessenger.error(e.message);
+      }
     }
 
     // Logged-in users always see the "Public" group.
@@ -295,7 +302,7 @@ export default function groups(
     // Step 4. Combine all the groups into a single list and set additional
     // metadata on them that will be used elsewhere in the app.
     const isLoggedIn = token !== null;
-    const groups = filterGroups(
+    const groups = await filterGroups(
       combineGroups(myGroups, featuredGroups, documentUri),
       isLoggedIn,
       directLinkedAnnotationGroupId,
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -111,8 +111,7 @@ export default function groups(
     // If a service group doesn't exist in the list of groups don't return it.
     if (svc && svc.groups) {
       try {
-        // The groups may not be available if they are being fetched via RPC.
-        // See mutateAsyncGroups in fetch-config.js
+        // The groups may not be immediately available if they are being fetched via RPC.
         const focusedGroups = await svc.groups;
         const filteredGroups = groups.filter(
           g => focusedGroups.includes(g.id) || focusedGroups.includes(g.groupid)
@@ -120,6 +119,8 @@ export default function groups(
         return filteredGroups;
       } catch (e) {
         toastMessenger.error(e.message);
+        // Don't show any groups if we catch an error
+        return [];
       }
     }
 
--- a/src/sidebar/util/fetch-config.js
+++ b/src/sidebar/util/fetch-config.js
@@ -132,9 +132,45 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
     [],
     3000
   );
-  return fetchConfigEmbed(appConfig, remoteConfig);
+  // Closure for the RPC call for mutateAsyncGroups to scope parentFrame
+  // and origin variables.
+  const rpcCall = (method, args = [], timeout = 3000) =>
+    postMessageJsonRpc.call(parentFrame, origin, method, args, timeout);
+  const mergedConfig = fetchConfigEmbed(appConfig, remoteConfig);
+  return mutateAsyncGroups(mergedConfig, rpcCall);
 }
 
+/**
+ * Potentially mutates the `groups` config object(s) to be a promise that
+ * resolves right away if the `groups` value exists in the original
+ * config, or returns a promise that resolves with a secondary RPC
+ * call to `requestGroups` when the `groups` value is '$rpc:requestGroups'
+ * If the `groups` value is missing or falsely then it won't be mutated.
+ *
+ * This allows the app to start with an incomplete config and then.
+ * lazily fill in the `groups` value(s) later when its ready. This helps
+ * speed up the loading process.
+ *
+ * @param {Object} config - The configuration object to mutate. This should
+ *  already have the `services` value
+ * @param {function} rpcCall - RPC method
+ *  (method, args, timeout) => Promise
+ * @return {Object} - The mutated settings
+ */
+
+async function mutateAsyncGroups(config, rpcCall) {
+  if (Array.isArray(config.services)) {
+    config.services.forEach((service, index) => {
+      if (service.groups === '$rpc:requestGroups') {
+        // The `groups` need to be fetched from a secondary RPC call
+        service.groups = rpcCall('requestGroups', [index], 5000).catch(() => {
+          throw new Error('Unable to fetch groups');
+        });
+      }
+    });
+  }
+  return config;
+}
 /**
  * Fetch the host configuration and merge it with the app configuration from h.
  *
--- a/src/sidebar/util/fetch-config.js
+++ b/src/sidebar/util/fetch-config.js
@@ -132,12 +132,11 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
     [],
     3000
   );
-  // Closure for the RPC call for mutateAsyncGroups to scope parentFrame
-  // and origin variables.
+  // Closure for the RPC call to scope parentFrame and origin variables.
   const rpcCall = (method, args = [], timeout = 3000) =>
     postMessageJsonRpc.call(parentFrame, origin, method, args, timeout);
   const mergedConfig = fetchConfigEmbed(appConfig, remoteConfig);
-  return mutateAsyncGroups(mergedConfig, rpcCall);
+  return fetchGroupsAsync(mergedConfig, rpcCall);
 }
 
 /**
@@ -147,9 +146,9 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
  * call to `requestGroups` when the `groups` value is '$rpc:requestGroups'
  * If the `groups` value is missing or falsely then it won't be mutated.
  *
- * This allows the app to start with an incomplete config and then.
- * lazily fill in the `groups` value(s) later when its ready. This helps
- * speed up the loading process.
+ * This allows the app to start with an incomplete config and then lazily
+ * fill in the `groups` value(s) later when its ready. This helps speed
+ * up the loading process.
  *
  * @param {Object} config - The configuration object to mutate. This should
  *  already have the `services` value
@@ -157,8 +156,7 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
  *  (method, args, timeout) => Promise
  * @return {Object} - The mutated settings
  */
-
-async function mutateAsyncGroups(config, rpcCall) {
+async function fetchGroupsAsync(config, rpcCall) {
   if (Array.isArray(config.services)) {
     config.services.forEach((service, index) => {
       if (service.groups === '$rpc:requestGroups') {
--- a/src/sidebar/util/test/fetch-config-test.js
+++ b/src/sidebar/util/test/fetch-config-test.js
@@ -178,12 +178,14 @@ describe('sidebar.util.fetch-config', () => {
 
       it('makes an RPC request to `requestConfig` ', async () => {
         await fetchConfig({}, fakeWindow);
-        fakeJsonRpc.call.calledWithExactly(
-          fakeTopWindow,
-          'https://embedder.com',
-          'requestConfig',
-          [],
-          3000
+        assert.isTrue(
+          fakeJsonRpc.call.calledWithExactly(
+            fakeTopWindow,
+            'https://embedder.com',
+            'requestConfig',
+            [],
+            3000
+          )
         );
       });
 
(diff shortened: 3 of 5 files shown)


src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,58 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
...
+ *  already have the `services` value.
+ * @param {Window} parentFrame - Frame to send call to.
+ * @param {string} origin - Origin filter for `window.postMessage` call.
+ * @return {Object} - The mutated settings.
+ */
+
+async function mutateAsyncGroups(config, parentFrame, origin) {
+  if (Array.isArray(config.services)) {
+    config.services.forEach((service, index) => {
+      if (service.groups === '$rcp:requestGroups') {
```
robertknight: Presumably the idea here is that the host frame specifies the name of the RPC method call to use via `$rpc:<methodName>`? That seems a reasonable choice, however if you're going to do that, the `postMessageJsonRpc` call below needs to actually use that method name.

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,58 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
...
+        // The `groups` need to be fetched from a secondary RCP call
+        service.groups = new Promise((resolve, reject) => {
```
robertknight: Since you're calling a Promise-returning method, you don't need to create a Promise here. You can just call `postMessageJsonRpc.call` and add a `catch(...)` handler:

```js
service.groups = postMessaeJsonRpc.call(...).catch(() => {
// Return fallback value or otherwise handle error on failure.
});
```

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,58 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
...
+          postMessageJsonRpc
+            .call(
+              parentFrame,
+              origin,
+              'requestGroups',
+              [index], // Only supports one service for now
+              5000
```
robertknight: It wasn't immediately obvious to me looking at this that `5000` is a timeout value. That's an issue with the API design, but a comment would clarify (ie. `..., [index], 5000 /* timeout */`).

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,58 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
...
+              resolve(serviceGroupsRCP);
+            })
+            .catch(() => {
+              reject('Unable to fetch groups');
+            });
+        });
+      } else if (service.groups) {
+        // The `groups` already exist so just wrap them in a promise so
+        // they can be treated identical do the RPC case above
+        service.groups = Promise.resolve(service.groups);
```
robertknight: Making `service.groups` always be a Promise for consistency seems reasonable. I will note that you can call `await` on a value, so you could actually leave `service.groups` as an array here and I think it would work just the same.

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,58 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
     [],
     3000
   );
-  return fetchConfigEmbed(appConfig, remoteConfig);
+  const mergedConfig = fetchConfigEmbed(appConfig, remoteConfig);
+  return mutateAsyncGroups(mergedConfig, parentFrame, origin);
```
robertknight: Just a thought - rather than passing down all the params for an RPC call to `mutateAsyncGroup`, it might be useful to package all that up into a closure, eg:

```js
const rpcCall = (method, args = [], timeout = 3000) => postMessageJsonRpc.call(
  parentFrame, origin, method, args, timeout
);

const remoteConfig = await rpcCall('requestConfig');
const mergedConfig = fetchConfigEmbed(appConfig, remoteConfig);
mutateAsyncGroups(mergedConfig, rpcCall);
```

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -219,7 +219,7 @@ describe('sidebar.util.fetch-config', () => {
         }
       });
 
-      it('creates a merged config when the rpc requests returns the host config` ', async () => {
+      it.skip('creates a merged config when the rpc requests returns the host config` ', async () => {
```
robertknight: I presume this is skipped because this is work in progress?

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,58 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
...
 }
 
+/**
+ * Potentially mutates the `groups` config object(s) to be a promise that
+ * resolves right away if the `groups` value exists in the original
+ * config, or returns a promise that resolves with a secondary RCP
```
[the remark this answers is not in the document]
LMS007: :(

src/sidebar/util/fetch-config.js
LMS007: awe right!

src/sidebar/util/fetch-config.js
LMS007: sure!

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -178,12 +178,14 @@ describe('sidebar.util.fetch-config', () => {
 
       it('makes an RPC request to `requestConfig` ', async () => {
         await fetchConfig({}, fakeWindow);
-        fakeJsonRpc.call.calledWithExactly(
-          fakeTopWindow,
-          'https://embedder.com',
-          'requestConfig',
-          [],
-          3000
+        assert.isTrue(
```
LMS007: Unrelated to this PR. I noticed this test was not calling assert

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -237,6 +239,52 @@ describe('sidebar.util.fetch-config', () => {
...
+        fakeJsonRpc.call.onSecondCall().resolves(['group1', 'group2']); // requestGroups
+        const result = await fetchConfig(appConfig, fakeWindow);
+        assert.deepEqual(await result.services[0].groups, ['group1', 'group2']);
+        assert.isTrue(
+          fakeJsonRpc.call.getCall(1).calledWithExactly(
+            fakeTopWindow,
+            'https://embedder.com',
+            'requestGroups',
+            [0], // passes service index to requestGroups
+            5000
```
LMS007, line 268: Still need to confirm the right timeout value here

src/sidebar/services/groups.js
```diff
@@ -106,14 +107,20 @@ export default function groups(
...
+      try {
+        // The groups may not be available if they are being fetched via RPC.
+        // See mutateAsyncGroups in fetch-config.js
+        const focusedGroups = await svc.groups;
+        const filteredGroups = groups.filter(
+          g => focusedGroups.includes(g.id) || focusedGroups.includes(g.groupid)
+        );
+        return filteredGroups;
+      } catch (e) {
+        toastMessenger.error(e.message);
```
robertknight, line 122: Since we're not returning anything here, which groups end up getting shown in the LMS context? Presumably all the groups for the different courses that the user is a member of?

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,45 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
...
 }
 
+/**
+ * Potentially mutates the `groups` config object(s) to be a promise that
+ * resolves right away if the `groups` value exists in the original
+ * config, or returns a promise that resolves with a secondary RPC
+ * call to `requestGroups` when the `groups` value is '$rpc:requestGroups'
+ * If the `groups` value is missing or falsely then it won't be mutated.
+ *
+ * This allows the app to start with an incomplete config and then.
```
robertknight, line 150:
```suggestion
 * This allows the app to start with an incomplete config and then
```

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,45 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
...
+ * lazily fill in the `groups` value(s) later when its ready. This helps
+ * speed up the loading process.
+ *
+ * @param {Object} config - The configuration object to mutate. This should
+ *  already have the `services` value
+ * @param {function} rpcCall - RPC method
+ *  (method, args, timeout) => Promise
+ * @return {Object} - The mutated settings
+ */
+
```
robertknight, line 160: Can you remove this extra new line. IIRC this can cause some tools not to associate the JSDoc comment and the function correctly.

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,45 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
...
+async function mutateAsyncGroups(config, rpcCall) {
```
robertknight, line 161: I suggest `fetchGroupsAsync` might be a better name. To me the fact that this function makes an RPC call to fetch groups seems more significant than the fact that it mutates its argument as opposed to returning a copy.

src/sidebar/services/groups.js
LMS007: check out extra-config-params branch in via3, then it should work, I'll try it out too and see if we should do something like filter to nothing perhaps?

LMS007: Oh also, need to force the group value to `$rpc:requestGroups` or just comment out that condition to see the RPC request work.

LMS007:
> Hi Kyle,
> 
> The client code looks good, although I had a suggestion regarding a function name. What happens from the user's point of view if a timeout occurs? Is there an LMS branch that this can be tested with? I can see that an error is displayed to the user, but it looks like we'll still end up showing some groups to the user, which may not be the one that corresponds to the current course. If that is the case, it would probably be better not to show any groups at all and instead just make sure that the error tells the user to refresh the page.

So I think its very reasonable to show no groups when an error happens, I propose we just do that.

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -237,6 +239,52 @@ describe('sidebar.util.fetch-config', () => {
...
       });
+
+      it('returns the `groups` array with the initial host config request', async () => {
+        const appConfig = {
+          services: [{ groups: ['group1', 'group2'] }],
+          appType: 'via',
+        };
+        fakeJsonRpc.call.onFirstCall().resolves({ foo: 'baz' }); // host config
+        const result = await fetchConfig(appConfig, fakeWindow);
+        assert.deepEqual(await result.services[0].groups, ['group1', 'group2']);
```
robertknight, line 250: Is there a need to `await` the groups array here? It should be a plain array in this case shouldn't it?

src/sidebar/util/test/fetch-config-test.js
```diff
@@ -237,6 +239,52 @@ describe('sidebar.util.fetch-config', () => {
...
+      });
+
+      it("creates a merged config where `groups` returns a promise when its initial value is '$rpc:requestGroups'", async () => {
```
robertknight, line 253:
```suggestion
      it("creates a merged config where `groups` is a promise when its initial value is '$rpc:requestGroups'", async () => {
```

src/sidebar/services/groups.js
```diff
@@ -106,14 +107,22 @@ export default function groups(
...
-
     // If service groups are specified only return those.
     // If a service group doesn't exist in the list of groups don't return it.
     if (svc && svc.groups) {
-      const focusedGroups = groups.filter(
-        g => svc.groups.includes(g.id) || svc.groups.includes(g.groupid)
-      );
-      return focusedGroups;
+      try {
+        // The groups may not be available if they are being fetched via RPC.
```
robertknight, line 114: I think it would be clearer to say that groups may not be _immediately_ available, as opposed to simply _not available_.

src/sidebar/util/fetch-config.js
```diff
@@ -132,9 +132,44 @@ async function fetchConfigRpc(appConfig, parentFrame, origin) {
     [],
     3000
   );
-  return fetchConfigEmbed(appConfig, remoteConfig);
+  // Closure for the RPC call for mutateAsyncGroups to scope parentFrame
```
robertknight, line 135:
```suggestion
  // Closure for the RPC call for mutateAsyncGroups to scope parentFrame
```

The name `mutateAsyncGroups` is out of date. Maybe just remove this comment since I think it is easy enough to understand what it does by reading the code.

src/sidebar/util/test/fetch-config-test.js
robertknight: Any update on this? 5 seconds seems like a reasonable value to start with, but we may need to update it if we learn that it can take longer.

src/sidebar/services/groups.js
```diff
@@ -106,14 +107,22 @@ export default function groups(
...
+        // See mutateAsyncGroups in fetch-config.js
```
robertknight, line 115: The reference to `mutateAsyncGroups` is obsolete. I would just shorten this to "See fetch-config.js" and avoid mentioning implementation details as that is less likely to get out of date.

PR #1936, opened 2020-03-20
Remove browserify-shim

robertknight:
This package was originally used for the purpose of making scripts that were not written as CommonJS modules (eg. that didn't use `require`/`module.export` but instead relied on global variables) work with Browserify.

The affected libraries have all evolved to the point where this is no longer necessary however.

### DIFF
--- a/src/sidebar/util/test/wrap-react-component-test.js
+++ b/src/sidebar/util/test/wrap-react-component-test.js
@@ -242,7 +242,7 @@ describe('wrapReactComponent', () => {
     const element = createDirective(document, 'parent');
     assert.isNull(element[0].querySelector('.click-indicator'));
 
-    const btn = element.find('button');
+    const btn = element.find('button')[0];
     btn.click();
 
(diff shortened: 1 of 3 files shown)


src/sidebar/util/test/wrap-react-component-test.js
```diff
@@ -242,7 +242,7 @@ describe('wrapReactComponent', () => {
     const element = createDirective(document, 'parent');
     assert.isNull(element[0].querySelector('.click-indicator'));
 
-    const btn = element.find('button');
+    const btn = element.find('button')[0];
```
robertknight: This change is needed because importing "angular" in tests no longer auto-imports "jquery", so [jqLite](https://docs.angularjs.org/api/ng/function/angular.element#angularjs-s-jqlite) gets used instead. jqLite doesn't provide a `click()` method, so we get the underlying `HTMLButtonElement` and call `click()` on that.

Note there is no difference in the actual application, because we weren't loading the full "jquery" library in the sidebar app anyway.

PR #1940, opened 2020-03-23
Remove "raf" dependency

robertknight:
This is a polyfill for `requestAnimationFrame`, which is [natively supported](https://caniuse.com/#search=requestanimationframe) by all browsers we support.

The client was already broken if `requestAnimationFrame` was not available for any reason because some other dependencies (eg. Preact) already make use of it directly in the sidebar.

### DIFF
--- a/src/annotator/test/guest-test.coffee
+++ b/src/annotator/test/guest-test.coffee
@@ -11,9 +11,6 @@ Guest = require('../guest')
 rangeUtil = null
 selections = null
 
-raf = sinon.stub().yields()
-raf['@noCallThru'] = true
-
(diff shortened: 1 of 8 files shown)


src/annotator/test/guest-test.coffee
```diff
@@ -4,18 +4,13 @@ Plugin = require('../plugin')
 
 Delegator = require('../delegator')
 $ = require('jquery')
-Delegator['@noCallThru'] = true
```
robertknight: This `@noCallThru` property doesn't do anything any more. It is left over from when we used proxyquire. I'm just removing it here as a piece of clean-up.

PR #1961, opened 2020-03-25
Add preStart method to cross-origin-rcp module

LMS007:
When called, allows any incoming rpc request to wait in a queue and be proceeded a short time later when the actual start() method is called. This prevents potential timeouts from incoming RPC requests from LMS (or similar) that arrive after we tipped off the parent frame with a fetchConfig outgoing request but before start() is called.

After refactoring to remove the self-executing function (yesterday's attempt), testing got much more simple.  I also refactored a bit use a fake `emitter` as was done in postmessage-json-rpc-test so its a bit more consistent and less boiler plate code.

no_errors

fixes https://github.com/hypothesis/client/pull/1885

### DIFF
--- a/src/sidebar/cross-origin-rpc.js
+++ b/src/sidebar/cross-origin-rpc.js
@@ -1,5 +1,8 @@
 import warnOnce from '../shared/warn-once';
 
+// Array to keep track of pre-start requests
+let preStartQueue = [];
+
 /**
  * Return the mapped methods that can be called remotely via this server.
  *
@@ -48,7 +51,18 @@ function isJsonRpcMessage(data) {
 function start(store, settings, $window) {
   const methods = registeredMethods(store);
 
-  $window.addEventListener('message', function receiveMessage(event) {
+  // Process the pre-start incoming RPC requests
+  preStartQueue.forEach(event => {
+    receiveMessage(event);
+  });
+  // Clear the queue and remove the preStartMessageListener
+  preStartQueue = [];
+  window.removeEventListener('message', preStartMessageListener);
+
+  // Start listening to new RPC requests
+  $window.addEventListener('message', receiveMessage);
+
+  function receiveMessage(event) {
     let allowedOrigins = settings.rpcAllowedOrigins || [];
 
     if (!isJsonRpcMessage(event.data)) {
@@ -67,7 +81,7 @@ function start(store, settings, $window) {
     let jsonRpcRequest = event.data;
 
     event.source.postMessage(jsonRpcResponse(jsonRpcRequest), event.origin);
-  });
+  }
 
   /** Return a JSON-RPC response to the given JSON-RPC request object. */
   function jsonRpcResponse(request) {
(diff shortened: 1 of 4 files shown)


src/sidebar/cross-origin-rpc.js
```diff
@@ -48,7 +51,18 @@ function isJsonRpcMessage(data) {
...
-  $window.addEventListener('message', function receiveMessage(event) {
+  // Process the pre-start incoming RPC requests
+  preStartQueue.forEach(event => {
+    receiveMessage(event);
+  });
+  // Clear the queue and remove the preStartMessageListener
+  preStartQueue = [];
+  window.removeEventListener('message', preStartMessageListener);
+
+  // Start listening to new RCP requests
```
robertknight:
```suggestion
  // Start listening to new RPC requests
```

I noticed this easily-made typo in a few other places. It might be helpful to install an add-on for your editor to catch this?

src/sidebar/cross-origin-rpc.js
```diff
@@ -93,8 +107,34 @@ function start(store, settings, $window) {
   }
 }
 
+/**
+ * Queues all incoming RPC requests so they can be handled later.
+ *
+ * @param {object} event - RPC event payload
```
robertknight:
```suggestion
 * @param {MessageEvent} event - RPC event payload
```

Since we expect this function to be called with the `event` from a "message" event on a window, you can be more specific about the type. See also the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent).

src/sidebar/cross-origin-rpc.js
```diff
@@ -93,8 +107,34 @@ function start(store, settings, $window) {
...
+ *
+ * Why?
+ *
+ * This allows the client to fetch its config from the parent app frame and potentially
+ * receive new unsolicited incoming requests that the parent app may send before this
+ * RCP server starts.
+ *
+ * @param {Window} $window - Test seam
+ */
+function preStart($window = window) {
```
robertknight: `$window` is a built-in feature in Angular, but outside of Angular code the preferred convention for this kind of test seam (for DOM objects like `window` or `document`) is to add an underscore after the name (eg. `window_ = window`).

src/sidebar/cross-origin-rpc.js
```diff
@@ -93,8 +107,34 @@ function start(store, settings, $window) {
...
+  $window.addEventListener('message', preStartMessageListener);
+}
+
 export default {
```
robertknight: It looks like I didn't adjust the export style to be idiomatic ES modules exports properly here. I think it would make more sense for this module to export `startServer` and `preStartServer` functions directly rather than default-exporting some object.

src/sidebar/test/cross-origin-rpc-test.js
```diff
@@ -1,34 +1,36 @@
...
 import crossOriginRPC from '../cross-origin-rpc';
 import { $imports } from '../cross-origin-rpc';
 
+class FakeWindow {
+  constructor() {
+    this.emitter = new EventEmitter();
+    this.addEventListener = this.emitter.on.bind(this.emitter);
+    this.removeEventListener = this.emitter.off.bind(this.emitter);
+  }
+}
```
robertknight: This is fine. One thing I learned this morning that in some modern browsers it is actually possible to directly create instances of the browser's equivalent of `EventEmitter`, [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/EventTarget).

src/sidebar/test/cross-origin-rpc-test.js
```diff
@@ -212,5 +203,71 @@ describe('crossOriginRPC', function() {
...
+              id: 42,
+              result: 'ok',
+            },
+            'https://allowed1.com'
+          )
+        );
+      });
+
+      it('responds to multiple incoming requests that arrive before the server starts', function() {
+        const payload = id => ({
```
robertknight: Maybe `messageEvent` might be slightly clearer than `payload` in terms of what kind of object the result is supposed to mimic?

src/sidebar/cross-origin-rpc.js
LMS007: Thanks, was wondering about that

src/sidebar/test/cross-origin-rpc-test.js
LMS007: good to know

PR #1962, opened 2020-03-26
Add translation map for specific host-config values

LMS007:
Coerces specific values provided into types that the client assumes it would be in.
This is part of the prep work to support via3 which will eventually only export strings values in the config.

Fixes https://github.com/hypothesis/client/issues/1957

### DIFF
--- a/src/shared/test/type-coercions-test.js
+++ b/src/shared/test/type-coercions-test.js
@@ -1,0 +1,151 @@
+import { toBoolean, toInteger, toObject, toString } from '../type-coercions';
+
+describe('shared/type-coercions', () => {
+  describe('toBoolean', () => {
+    [
+      {
+        value: true,
+        result: true,
+      },
+      {
+        value: 'true',
+        result: true,
+      },
+      {
+        value: 'any',
+        result: true,
+      },
+      {
+        value: 'false',
+        result: false,
+      },
+      {
+        value: ' false',
+        result: false,
+      },
+      {
+        value: 'False',
+        result: false,
+      },
+      {
+        value: 'FALSE',
+        result: false,
+      },
+      {
+        value: '',
+        result: false,
+      },
+      {
+        value: '1',
+        result: true,
+      },
+      {
+        value: '0',
+        result: false,
+      },
+      {
+        value: 1,
+        result: true,
+      },
+      {
+        value: null,
+        result: false,
+      },
+      {
+        value: undefined,
+        result: false,
+      },
+    ].forEach(test => {
+      it('coerces the values appropriately', () => {
+        assert.equal(toBoolean(test.value), test.result);
+      });
+    });
+  });
+
+  describe('toInteger', () => {
+    [
+      {
+        value: '1',
+        result: 1,
+      },
+      {
+        value: '0',
+        result: 0,
+      },
+      {
+        value: 1,
+        result: 1,
+      },
+      {
+        value: 1.1,
+        result: 1,
+      },
+      {
+        value: 'a',
+        result: NaN,
+      },
+    ].forEach(test => {
+      it('coerces the values appropriately', () => {
+        assert.deepEqual(toInteger(test.value), test.result);
+      });
+    });
+  });
+
+  describe('toObject', () => {
+    [
+      {
+        value: { a: 'a', b: { c: ['c'] } },
+        result: { a: 'a', b: { c: ['c'] } },
+      },
+      {
+        value: 1,
+        result: {},
+      },
+      {
+        value: null,
+        result: {},
+      },
+      {
+        value: undefined,
+        result: {},
+      },
+    ].forEach(test => {
+      it('coerces the values appropriately', () => {
+        assert.deepEqual(toObject(test.value), test.result);
+      });
+    });
+  });
+
+  describe('toString', () => {
+    [
+      {
+        value: 'a',
+        result: 'a',
+      },
+      {
+        value: 1,
+        result: '1',
+      },
+      {
+        value: null,
+        result: '',
+      },
+      {
+        value: undefined,
+        result: '',
+      },
+      {
+        value: {
+          // In a rare case where its an object with custom
+          // toString value that is not a function.
+          toString: false,
+        },
+        result: '',
+      },
+    ].forEach(test => {
+      it('coerces the values appropriately', () => {
+        assert.equal(toString(test.value), test.result);
+      });
+    });
+  });
+});
--- a/src/shared/type-coercions.js
+++ b/src/shared/type-coercions.js
@@ -1,0 +1,69 @@
+/**
+ * Type conversion methods that coerce incoming configuration values to an
+ * expected type or format that other parts of the UI may make assumptions
+ * on. This is needed for incoming configuration values that are otherwise
+ * not sanitized.
+ *
+ * Note that if the values passed are plain javascript values (such as ones
+ * produced from JSON.parse), then these methods do not throw errors.
+ */
+
+/**
+ * Returns a boolean
+ *
+ * @param {any} value - initial value
+ * @return {boolean}
+ */
+export function toBoolean(value) {
+  if (typeof value === 'string') {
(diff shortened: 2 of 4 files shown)


src/sidebar/test/host-config-test.js
```diff
@@ -54,4 +54,96 @@ describe('hostPageConfig', function() {
...
+        },
+      },
+      {
+        value: {
+          origin: 'origin',
+          ancestorLevel: 'foo',
+        },
+        result: {
+          origin: 'origin',
+          ancestorLevel: NaN,
```
LMS007: Note,  this will also throw an error later in `fetch-config`

src/sidebar/host-config.js
```diff
@@ -50,13 +50,45 @@ export default function hostPageConfig(window) {
     'usernameUrl',
   ];
 
+  // Coerces specific config values that we expect to be of specific types
+  // but could be provided in another.
+  const translations = {
```
robertknight: I think "coercions" or "casts" would be a preferable term here as it is more specific than "translation".

src/sidebar/host-config.js
```diff
@@ -50,13 +50,45 @@ export default function hostPageConfig(window) {
...
+      } else {
+        return {
+          ...value,
+          ancestorLevel: Math.round(Number(value.ancestorLevel)),
+        };
+      }
+    },
+    // Ensures openSidebar is boolean
+    openSidebar: value => {
+      if (value === 'false') {
```
jon-betts: Could we do the equivalent of `value.lower()` here to catch title case 'False'? This is a value which could easily be emitted by Python.

src/sidebar/host-config.js
```diff
@@ -50,13 +50,45 @@ export default function hostPageConfig(window) {
...
+        return false;
+      }
+      const numericalVal = Number(value);
+      if (!isNaN(numericalVal)) {
+        return Boolean(numericalVal);
+      }
+      return true;
+    },
```
jon-betts: It might be nice to create a `toBool(value)` function and then assign `openSidebar: toBool`. That way the boolean convertion could be tested separately from the sidebar-iness

src/sidebar/host-config.js
LMS007: Will do

src/sidebar/util/type-coercions.js
```diff
@@ -0,0 +1,74 @@
...
+      // "false", "False", " false", "FALSE" all return false
+      return false;
+    }
+  }
+  const numericalVal = Number(value);
+  if (!isNaN(numericalVal)) {
+    return Boolean(numericalVal);
+  }
+  // Any non numerical or falsely string should return true, otherwise
+  // return false
```
robertknight: There is a commented out line of code here.

src/sidebar/util/type-coercions.js
```diff
@@ -0,0 +1,74 @@
...
+ * @return {string}
+ */
+function toString(value) {
+  if (value && value.toString) {
+    return value.toString();
+  }
+  return '';
+}
+
+module.exports = {
```
robertknight: Can you change this module to use ES exports (ie. `export function ...` rather than `module.exports = `).

src/sidebar/util/type-coercions.js
```diff
@@ -0,0 +1,74 @@
...
+  if (typeof value === 'object' && value !== null) {
+    return value;
+  }
+  // Don't attempt to fix the values, just ensure type correctness
+  return {};
+}
+
+/**
+ * Returns the value as a string or an empty string if the
+ * value undefined or null.
```
robertknight: This function will also return an empty string if the value is `0` or `false` as well as undefined/null. That seems like it would make sense, as it will translate these values into a falsey-string.

src/sidebar/util/type-coercions.js
```diff
@@ -0,0 +1,74 @@
+/**
+ * Type conversion methods that gently coerce values that other parts
```
robertknight: Given that these functions seem quite generic, I think we should be more specific that they are for coercing incoming configuration values to an expected type or format.

src/sidebar/host-config.js
```diff
@@ -50,13 +56,33 @@ export default function hostPageConfig(window) {
     'usernameUrl',
   ];
 
+  const coercions = {
```
robertknight: As a follow-up to this PR, I think it would make sense to merge this object with the `paramWhiteList` above, such that there is an object whose keys are allowed parameters and functions are coercion or validation functions.

src/sidebar/util/type-coercions.js
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Returns the value as a string or an empty string if the
+ * value undefined or null.
+ *
+ * @param {any} value - initial value
+ * @return {string}
+ */
+function toString(value) {
+  if (value && value.toString) {
+    return value.toString();
```
robertknight: This could throw if `toString` is not a function. Assuming the incoming value is produced by `JSON.parse`, this could throw if given `{ toString: false }`.

```js
if (!value) {
  return '';
}
try {
  return String(value);
} catch (e) {
  return '';
}
```

src/sidebar/host-config.js
robertknight: I'd suggest adding a comment here explaining:

1) Why we are coercing values (both the use case where "Via" passes through values as strings and where a publisher accidentally provides the wrong type)
2) A note about why only these two specific params are handled at the moment, but that there is an intent to expand this to other params as well.

src/sidebar/util/type-coercions.js
```diff
@@ -0,0 +1,74 @@
+/**
+ * Type conversion methods that gently coerce values that other parts
+ * of the UI may make assumptions on. This is needed for incoming parameters
+ * that are otherwise not sanitized. Note that these methods do not throw 
+ * errors or act like validators with mal-formed values.
```
robertknight: The "do not throw errors" part assumes that the values are "plain JavaScript values", ie. values that could be produced by `JSON.parse`. This is true in the case of configuration parsed in `host-config.js`. I think it would be worth noting that assumption though.

src/sidebar/util/test/type-coercions-test.js
```diff
@@ -0,0 +1,139 @@
+import { toBoolean, toInteger, toObject, toString } from '../type-coercions';
+
+describe('sidebar/type-coercions', () => {
+  describe('#toBoolean', () => {
```
robertknight:
```suggestion
  describe('toBoolean', () => {
```

The `#` prefix comes from JSDoc and indicates an instance method or property. If you're just referring to a non-method function, it should be omitted.

src/sidebar/util/type-coercions.js
LMS007: haha, thats the end of my sentence, wow that ended up looking wrong!

src/sidebar/host-config.js
LMS007: good idea

LMS007: https://github.com/hypothesis/client/issues/1968

src/shared/test/type-coercions-test.js
```diff
@@ -0,0 +1,151 @@
+import { toBoolean, toInteger, toObject, toString } from '../type-coercions';
+
+describe('sidebar/type-coercions', () => {
```
robertknight:
```suggestion
describe('shared/type-coercions', () => {
```

This needs updating to reflect the new location.

src/shared/type-coercions.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * Type conversion methods that gently coerce incoming configuration values to
```
robertknight:
```suggestion
 * Type conversion methods that gently coerce incoming configuration values to
```

By "gently" are you referring to the fact that these functions don't throw if the values are primitives or object/array literals?

LMS007: ya, might be better to change that word

PR #1965, opened 2020-03-27
Only create one highlight element for adjacent text nodes

robertknight:
When highlighting a run of adjacent text nodes, wrap the group of
adjacent text nodes in a single `<hypothesis-highlight>` element rather
than wrapping each text node separately. Adjacent text nodes may be
created if:

 - A sub-range of the text is highlighted by Hypothesis, and the
   highlight is later removed, for example when switching groups.
 - The text nodes were generated programatically by JavaScript rather
   than by the browser's HTML parser

For screen reader users, this reduces the number of "annotation
start"/"annotation end" comments they will hear in this scenario. It is
also likely to benefit other code which interacts with highlighted text.

### DIFF
--- a/src/annotator/highlighter.js
+++ b/src/annotator/highlighter.js
@@ -10,26 +10,43 @@ export function highlightRange(normedRange, cssClass = 'hypothesis-highlight') {
   const white = /^\s*$/;
 
   // Find text nodes within the range to highlight.
-  //
-  // We ignore text nodes that contain only whitespace to avoid inserting
-  // elements in places that can only contain a restricted subset of nodes such
-  // as table rows and lists. This does mean that there may be the odd abandoned
-  // whitespace node in a paragraph that is skipped but better than breaking
-  // table layouts.
-  const textNodes = normedRange
-    .textNodes()
-    .filter(node => !white.test(node.nodeValue));
+  const textNodes = normedRange.textNodes();
 
-  // Wrap each text node with a `<hypothesis-highlight>` element.
-  const highlights = [];
+  // Group text nodes into spans of adjacent nodes. If a group of text nodes are
+  // adjacent, we only need to create one highlight element for the group.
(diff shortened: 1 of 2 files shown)


robertknight:
> I should note that this is also the case prior to this diff, so is that expected?

Yes. This patch only makes a difference when the range to be highlighted contains of adjacent _text_ nodes only, no `<hypothesis-highlight>`s or other elements. For example:

```
// Initial state
<p>[One] [two] [three]</p>

// After highlighting "two three" on master:
<p>[One] <hypothesis-highlight>[two]</hypothesis-highlight><hypothesis-highlight>[three]</hypothesis-highlight></p>

// After highlighting "two three" on this branch:
<p>[One] <hypothesis-highlight>[two][three]</hypothesis-highlight></p>
```

The browser's HTML parser will never produce a series of adjacent text nodes like the above AFAIK, but they can get produced if JavaScript generates the DOM or using Hypothesis if you do the following:

1. Select part of the text in a DOM element
2. Highlight the text
3. Remove the highlight

Step 2 will currently split the text node at the points where the highlight starts and ends, and then leave it split after the highlight is removed:

```
[ ... ] - One text node

// Initial state
<p>[One two three]</p>

// State after inserting highlight around "two"
<p>[One] <hypothesis-highlight> [two] </hypothesis-highlight> [three]</p>

// State after removing highlight:
<p>[One] [two] [three]</p>
```

The fact that removing a highlight leaves adjacent text nodes behind is something we should probably also change, but since such sequences can be generated by other means, I decided to fix what happens at the insert.

> I did notice that when creating a highlight around another highlight such that one of the two highlights is a substring, then it seems to wrap them in a single hypothesis-highlight tag, but it does not remove the inner tags, is that expected?

Currently yes. I've been thinking that we might want to merge or combine highlights in certain situations to reduce visual noise and prevent overlapping highlights from obscuring the text underneath.

PR #1978, opened 2020-03-30
WIP - Implement standard keyboard navigation for menus

robertknight:
This is some rather rough and definitely incomplete code for implementing the [standard keyboard interaction](https://www.w3.org/TR/wai-aria-practices/#menubutton) pattern for menu buttons:
 
- Focus first item in menu when menu is opened
- Focus menu button when menu is closed unless user has closed the menu by performing some action which has already focused an element outside of the menu
- Support navigating menu items using arrow keys rather than tab stops. Submenus are opened by pressing the _right_ arrow key and closed by pressing the _left_ arrow key even though we don't actually display the menu to the right.

Some notes about the implementation:

- For the specific use case of a menu popup, a roving tabindex (where one item has a tabindex of 0 and all other items have a tabindex of -1) doesn't seem necessary since we have no need to preserve the focused item when closing and re-opening a menu. When a menu is opened, focus always goes to the first item
- To make the presentation of all menu items consistent in a screen reader, I removed the link elements for menu items that open an external link and added a call to `window.open(...)` instead. I'm frankly not certain about this
- I haven't implemented the affordance where you can press a letter to move focus instantly to the first menu item whose label begins with that key

### DIFF
--- a/src/sidebar/components/menu.js
+++ b/src/sidebar/components/menu.js
@@ -6,6 +6,17 @@ import propTypes from 'prop-types';
 import useElementShouldClose from './hooks/use-element-should-close';
 import SvgIcon from './svg-icon';
 
+let nextId = 0;
+
+function useId(prefix) {
+  const idRef = useRef(null);
+  if (!idRef.current) {
+    ++nextId;
+    idRef.current = prefix + '-' + nextId;
+  }
+  return idRef.current;
+}
+
 // The triangular indicator below the menu toggle button that visually links it
 // to the menu content.
 const menuArrow = className => (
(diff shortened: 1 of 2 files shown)


src/sidebar/components/menu.js
```diff
@@ -21,6 +32,80 @@ const menuArrow = className => (
  */
 let ignoreNextClick = false;
 
+function isElementVisible(element) {
+  return element.offsetParent !== null;
+}
+
+function MenuList({ align, menuButtonId, contentClass, closeMenu, children }) {
```
robertknight: I'm thinking here that if we split out a component that wraps the _content_ of a menu, then we can let it handle the keyboard navigation when an item in that menu is focused. We can then re-use this in `GroupListItem` to wrap the submenu.

src/sidebar/components/menu.js
```diff
@@ -21,6 +32,80 @@ const menuArrow = className => (
...
+  const menuRef = useRef(null);
+
+  const handleMenuKeyDown = event => {
+    const menuItems = Array.from(
+      menuRef.current.querySelectorAll('[role^="menuitem"]')
```
robertknight: The `[role^=menuitem]` selector here matches child roles of `menuitem` such as `menuitemradio`.

src/sidebar/components/menu.js
```diff
@@ -21,6 +32,80 @@ const menuArrow = className => (
  */
 let ignoreNextClick = false;
 
+function isElementVisible(element) {
+  return element.offsetParent !== null;
```
robertknight: This trick shamelessly taken from Stack Overflow, which I forgot to link here.

robertknight:
I'm closing this proof of concept PR. @LMS007 is working on this feature.

PR #1992, opened 2020-04-03
Menu key navigation

LMS007:
The goal of this PR is to fully finish accessibility requirements for the menu component. The primary goals for that is keyboard navigation of the menu system. This particular PR is not complete, but does move us towards that goal. This is roughtly modeleed after these examples
https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html
https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-2/menubar-2.html

_With the major change being that an item can both have its own action or toggle a submenu._

I have taken Rob's branch `p-menu-item-arrow-keys` and used that as a basis for this PR with some changes. In short, the component from that branch called `MenuList` has been refactored into a component called `MenuKeyboardNavigation` that can wrap a menu or a submenu and provide some common keyboard functionality.

Along with the accessibility goal I took the time to fix some minor CSS and coding issues I found with the menu (discussed below)

**Mouse Navigation**
All click navigations with the mouse should work exactly as before. There are some minor css differences that I will point out below.

**Keyboard Navigation**
Keyboard navigation works as follows: Click an element next to the menu button with the mouse such as the search bar. Then, using the keyboard, tab back to the menu button so you see the blue focus halo. Press `enter`. Once you do this, then the focus should be moved to the first item with the blue halo. From here, keyboard nav should work and there should always be a blue halo on the items. Every time you drill deeper into a menu, focus should be placed on the first item in that menu. When you back out of a submenu, then focus should be placed back on the parent controlling item.

**Operation:**
- `enter` or `space` or `right arrow` to drill down into a submenu, or click an item
- `up` or `down` to navigate within a menu level
- `home`, `end` to navigate to the first / last items in the menu level
- `left` to ascend up to the parent menu level
- `esc` to close menu

**Other issues or fixes worth pointing out**

- Previous it turns out there was no way to actually click on a link in a menu with the keyboard. I was surprised by that, but I suspect it has to do with preact killing off the a `<a>` before the event makes it to the bowser. I have demonstrated it here: https://jsfiddle.net/ft8z1wux/7/ and there is a hack fix in the code with a timeout. I have added a few hack fixes and commented on them.

- The toggle button is no longer focusable. You can still click on it to active it or when you're on a menu item that owns one, press `right`

- The blue halo gets a little covered up on the submenu. I did spend some time trying to figure out how to force it to render above its siblings but I was unsuccessful. It may mean we live with that for now or change its outline styling. 

- I have not fully tested this with the gauntlet of screen readers yet. I'm going to do that more next week. There may be some extra label(s) to add to inform the user when a toggle button is available on a menu item.

- Moving focus off the menu in my opinion should _close_ the menu, but that is tricky. A simple `onBlur` won't cut is as that only applied to items, not the menu wrapper. The w3 examples actually do this and w/o looking at the code it seems like there is a short delay before the menu hides after focus is removed. I'll look into how that was achieved next week, but I think we can merge this w/o that feature and add it later on.
https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html

**CSS changes**

- Toggle buttons get darker when their submenu is expanded and when the menu item they belong to has `:hover`. This helps make them more distinct.

- Links are now red on hover across the entire menu item. Previously, only link text was red, but now the left/right icons turn red too.

- Submenu bottom border is darker and I added a top border to help frame it from its parent. 

**TODO:**
- Minor css issues with sort and post menus. The min width is no longer the same and needs to be fixed.
- Add keyboard navigation ability to the sort and post menus.
- Tests
- A bit of css clean up and comments need to be added here and there

menu

### DIFF
--- a/src/sidebar/components/group-list-item.js
+++ b/src/sidebar/components/group-list-item.js
@@ -50,6 +50,11 @@ function GroupListItem({
     }
   };
 
+  /**
+   * Opens or closes the submenu.
+   *
+   * @param {MouseEvent|KeyboardEvent} event
+   */
   const toggleSubmenu = event => {
     event.stopPropagation();
 
@@ -85,9 +90,11 @@ function GroupListItem({
       onToggleSubmenu={toggleSubmenu}
       submenu={
         <Fragment>
-          <ul>
+          {/* Giving the role="menu" allows screen readers to have
+          correct number of subitems */}
+          <ul role="menu">
             {activityUrl && (
-              <li>
+              <li role="none">
                 <MenuItem
                   href={activityUrl}
                   icon="external"
@@ -97,7 +104,7 @@ function GroupListItem({
               </li>
             )}
             {activityUrl && (
-              <li>
+              <li role="none">
                 <MenuItem
                   onClick={copyLink}
                   icon="copy"
@@ -107,7 +114,7 @@ function GroupListItem({
               </li>
             )}
             {canLeaveGroup && (
-              <li>
+              <li role="none">
                 <MenuItem
                   icon="leave"
                   isSubmenuItem={true}
--- a/src/sidebar/components/group-list-item.js
+++ b/src/sidebar/components/group-list-item.js
@@ -90,11 +90,9 @@ function GroupListItem({
       onToggleSubmenu={toggleSubmenu}
       submenu={
         <Fragment>
-          {/* Giving the role="menu" allows screen readers to have
-          correct number of subitems */}
-          <ul role="menu">
+          <ul>
             {activityUrl && (
-              <li role="none">
+              <li>
                 <MenuItem
                   href={activityUrl}
                   icon="external"
@@ -104,7 +102,7 @@ function GroupListItem({
               </li>
             )}
             {activityUrl && (
-              <li role="none">
+              <li>
                 <MenuItem
                   onClick={copyLink}
                   icon="copy"
@@ -114,7 +112,7 @@ function GroupListItem({
               </li>
             )}
             {canLeaveGroup && (
-              <li role="none">
+              <li>
                 <MenuItem
                   icon="leave"
                   isSubmenuItem={true}
--- a/src/sidebar/components/menu-item.js
+++ b/src/sidebar/components/menu-item.js
@@ -21,6 +21,7 @@ import SvgIcon from '../../shared/components/svg-icon';
  *
  * For items that have submenus, the `MenuItem` will call the `renderSubmenu`
  * prop to render the content of the submenu, when the submenu is visible.
+ * Note that the `submenu` is not supported for link (`href`) items.
  */
 export default function MenuItem({
   href,
@@ -228,7 +229,7 @@ MenuItem.propTypes = {
   /**
    * If present, display a button to toggle the sub-menu associated with this
    * item and indicate the current state; `true` if the submenu is visible.
-   * Note. Omit this prop if there is no `submenu`.
+   * Note. Omit this prop, or set it to null, if there is no `submenu`.
    */
   isSubmenuVisible: propTypes.bool,
 
--- a/src/sidebar/components/menu-item.js
+++ b/src/sidebar/components/menu-item.js
@@ -155,6 +155,10 @@ export default function MenuItem({
 
         {hasSubmenuVisible && (
           <div
+            // We should not have a <button> inside of the menu item itself
+            // but we have a non-standard mechanism with the toggle control
+            // requiring an onClick event nested inside a "menuitemradio|menuitem".
+            // Therefore, a static element with a role="none" is necessary here.
             role="none"
             icon={isSubmenuVisible ? 'collapse-menu' : 'expand-menu'}
             className="menu-item__toggle"
--- a/src/sidebar/components/menu-item.js
+++ b/src/sidebar/components/menu-item.js
@@ -1,11 +1,13 @@
 import classnames from 'classnames';
 import { Fragment, createElement } from 'preact';
-import { useRef } from 'preact/hooks';
+import { useEffect, useRef } from 'preact/hooks';
 import propTypes from 'prop-types';
 
+import { normalizeKeyName } from '../../shared/browser-compatibility-utils';
+
+import SvgIcon from '../../shared/components/svg-icon';
 import MenuKeyboardNavigation from './menu-keyboard-navigation';
 import Slider from './slider';
-import SvgIcon from '../../shared/components/svg-icon';
 
 /**
  * An item in a dropdown menu.
@@ -44,6 +46,7 @@ export default function MenuItem({
   const hasRightIcon = icon && isSubmenuItem;
 
   const menuItemRef = useRef(null);
+  const focusTimer = useRef(null);
 
   let renderedIcon = null;
   if (icon !== 'blank') {
@@ -61,18 +64,25 @@ export default function MenuItem({
   const hasSubmenuVisible = typeof isSubmenuVisible === 'boolean';
   const isRadioButtonType = typeof isSelected === 'boolean';
 
+  useEffect(() => {
+    return () => {
+      // unmount
+      clearTimeout(focusTimer.current);
+    };
+  }, []);
+
   const onCloseSubmenu = event => {
     if (onToggleSubmenu) {
       onToggleSubmenu(event);
     }
     // The focus won't work without delaying rendering.
-    setTimeout(() => {
+    focusTimer.current = setTimeout(() => {
       menuItemRef.current.focus();
     });
   };
 
   const onKeyDown = event => {
-    switch (event.key) {
+    switch (normalizeKeyName(event.key)) {
       case 'ArrowRight':
         if (onToggleSubmenu) {
           event.stopPropagation();
@@ -117,9 +127,6 @@ export default function MenuItem({
     // The menu item is a clickable button or radio button.
     // In either case there may be an optional submenu.
 
-    const expandedProp = {
-      'aria-expanded': hasSubmenuVisible ? isSubmenuVisible : undefined,
-    };
     menuItem = (
       <div
         ref={menuItemRef}
@@ -135,7 +142,7 @@ export default function MenuItem({
         role={isRadioButtonType ? 'menuitemradio' : 'menuitem'}
         aria-checked={isRadioButtonType ? isSelected : undefined}
         aria-haspopup={hasSubmenuVisible}
-        {...expandedProp}
+        aria-expanded={hasSubmenuVisible ? isSubmenuVisible : undefined}
       >
         {hasLeftIcon && (
           <div className="menu-item__icon-container">{leftIcon}</div>
--- a/src/sidebar/components/menu-item.js
+++ b/src/sidebar/components/menu-item.js
@@ -1,9 +1,9 @@
 import classnames from 'classnames';
 import { Fragment, createElement } from 'preact';
+import { useRef } from 'preact/hooks';
 import propTypes from 'prop-types';
 
-import { onActivate } from '../util/on-activate';
-
+import MenuKeyboardNavigation from './menu-keyboard-navigation';
 import Slider from './slider';
 import SvgIcon from '../../shared/components/svg-icon';
 
@@ -38,12 +38,11 @@ export default function MenuItem({
 }) {
   const iconClass = 'menu-item__icon';
   const iconIsUrl = icon && icon.indexOf('/') !== -1;
-  const labelClass = classnames('menu-item__label', {
-    'menu-item__label--submenu': isSubmenuItem,
-  });
 
   const hasLeftIcon = icon || isSubmenuItem;
   const hasRightIcon = icon && isSubmenuItem;
+
+  const menuItemRef = useRef(null);
 
   let renderedIcon = null;
   if (icon !== 'blank') {
@@ -56,49 +55,102 @@ export default function MenuItem({
   const leftIcon = isSubmenuItem ? null : renderedIcon;
   const rightIcon = isSubmenuItem ? renderedIcon : null;
 
-  return (
-    <Fragment>
-      {/* FIXME-A11Y */}
-      {/* eslint-disable-next-line jsx-a11y/role-supports-aria-props */}
-      <div
-        aria-checked={isSelected}
+  // menuItem can be either a link or a button
+  let menuItem;
+  const hasSubmenuVisible = typeof isSubmenuVisible === 'boolean';
+  const isRadioButtonType = typeof isSelected === 'boolean';
+
+  const onCloseSubmenu = event => {
+    if (onToggleSubmenu) {
+      onToggleSubmenu(event);
+    }
+    // The focus won't work without delaying rendering.
+    setTimeout(() => {
+      menuItemRef.current.focus();
+    });
+  };
+
+  const onKeyDown = event => {
+    switch (event.key) {
+      case 'ArrowRight':
+        if (onToggleSubmenu) {
+          event.stopPropagation();
+          event.preventDefault();
+          onToggleSubmenu(event);
+        }
+        break;
+      case 'Enter':
+      case ' ':
+        if (onClick) {
+          // Let event propagate so the menu closes
+          onClick(event);
+        }
+    }
+  };
+  if (href) {
+    // The menu item is a link
+    menuItem = (
+      <a
+        ref={menuItemRef}
         className={classnames('menu-item', {
-          'menu-item--submenu': isSubmenuItem,
+          'is-submenu': isSubmenuItem,
+          'is-disabled': isDisabled,
+        })}
+        href={href}
+        target="_blank"
+        tabIndex="-1"
+        rel="noopener noreferrer"
+        role="menuitem"
+        onKeyDown={onKeyDown}
+      >
+        {hasLeftIcon && (
+          <div className="menu-item__icon-container">{leftIcon}</div>
+        )}
+        <span className="menu-item__label">{label}</span>
+        {hasRightIcon && (
+          <div className="menu-item__icon-container">{rightIcon}</div>
+        )}
+      </a>
+    );
+  } else {
+    // The menu item is a clickable button or radio button.
+    // In either case there may be an optional submenu.
+
+    const expandedProp = {
+      'aria-expanded': hasSubmenuVisible ? isSubmenuVisible : undefined,
+    };
+    menuItem = (
+      <div
+        ref={menuItemRef}
+        className={classnames('menu-item', {
+          'is-submenu': isSubmenuItem,
           'is-disabled': isDisabled,
           'is-expanded': isExpanded,
           'is-selected': isSelected,
         })}
-        role="menuitem"
-        {...(onClick && onActivate('menuitem', onClick))}
+        tabIndex="-1"
+        onKeyDown={onKeyDown}
+        onClick={onClick}
+        role={isRadioButtonType ? 'menuitemradio' : 'menuitem'}
+        aria-checked={isRadioButtonType ? isSelected : undefined}
+        aria-haspopup={hasSubmenuVisible}
+        {...expandedProp}
       >
-        <div className="menu-item__action">
-          {hasLeftIcon && (
-            <div className="menu-item__icon-container">{leftIcon}</div>
-          )}
-          {href && (
-            <a
-              className={labelClass}
-              href={href}
-              target="_blank"
-              rel="noopener noreferrer"
-            >
-              {label}
-            </a>
-          )}
-          {!href && <span className={labelClass}>{label}</span>}
-          {hasRightIcon && (
-            <div className="menu-item__icon-container">{rightIcon}</div>
-          )}
-        </div>
-        {typeof isSubmenuVisible === 'boolean' && (
+        {hasLeftIcon && (
+          <div className="menu-item__icon-container">{leftIcon}</div>
+        )}
+        <span className="menu-item__label">{label}</span>
+        {hasRightIcon && (
+          <div className="menu-item__icon-container">{rightIcon}</div>
+        )}
+
+        {hasSubmenuVisible && (
           <div
+            role="none"
+            icon={isSubmenuVisible ? 'collapse-menu' : 'expand-menu'}
             className="menu-item__toggle"
-            // We need to pass strings here rather than just the boolean attribute
-            // because otherwise the attribute will be omitted entirely when
-            // `isSubmenuVisible` is false.
-            aria-expanded={isSubmenuVisible ? 'true' : 'false'}
-            aria-label={`Show actions for ${label}`}
-            {...onActivate('button', onToggleSubmenu)}
+            onClick={onToggleSubmenu}
+            title={`Show actions for ${label}`}
           >
             <SvgIcon
               name={isSubmenuVisible ? 'collapse-menu' : 'expand-menu'}
@@ -107,9 +159,20 @@ export default function MenuItem({
           </div>
         )}
       </div>
-      {typeof isSubmenuVisible === 'boolean' && (
+    );
+  }
+  return (
+    <Fragment>
+      {menuItem}
+      {hasSubmenuVisible && (
         <Slider visible={isSubmenuVisible}>
-          <div className="menu-item__submenu">{submenu}</div>
+          <MenuKeyboardNavigation
+            closeMenu={onCloseSubmenu}
+            visible={isSubmenuVisible}
+            className="menu-item__submenu"
+          >
+            {submenu}
+          </MenuKeyboardNavigation>
         </Slider>
       )}
     </Fragment>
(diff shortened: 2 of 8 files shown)


src/sidebar/components/menu-item.js
```diff
@@ -56,49 +55,109 @@ export default function MenuItem({
...
+  const hasSubmenuVisible = typeof isSubmenuVisible === 'boolean';
+  const isRadioButtonType = typeof isSelected === 'boolean';
+
+  const onCloseSubmenu = event => {
+    if (onToggleSubmenu) {
+      onToggleSubmenu(event, false);
+      // TODO: figure out something better than this timeout hack if possible.
+      // The focus won't work without delaying rendering.
+      setTimeout(() => {
+        menuItemRef.current.focus();
```
robertknight, line 69: Can you explain what problem you ran into here?

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -0,0 +1,99 @@
...
+  children,
+  isVisible,
+}) {
+  const menuRef = useRef(null);
+
+  useEffect(() => {
+    if (isVisible) {
+      // TODO: figure out something better than this timeout hack if possible.
+      // The focus won't work without delaying rendering.
+      setTimeout(() => {
```
robertknight, line 28: Can you explain why this was needed. Effects (both `useEffect` and `useLayoutEffect`) run after the DOM is updated so you should be able to get a reference to the menu item either way.

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -0,0 +1,99 @@
...
+
+    let focusedIndex = menuItems.findIndex(el =>
+      el.contains(document.activeElement)
+    );
+
+    let handled = false;
+
+    switch (event.key) {
+      case 'ArrowLeft':
+      case 'Esc':
```
robertknight, line 50: The `event.key` value is "Escape" rather than "Esc" isn't it?

src/styles/sidebar/components/menu-item.scss
```diff
@@ -143,17 +134,26 @@ $menu-item-padding: 10px;
...
+    background-color: var.$grey-4;
+    color: var.$grey-6;
   }
 
   &-icon {
     width: 12px;
     height: 12px;
   }
+
+  svg {
```
robertknight, line 146: Is this not covered by the `&-icon` rule above?

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -0,0 +1,99 @@
...
+import { useEffect, useRef } from 'preact/hooks';
+import propTypes from 'prop-types';
+
+function isElementVisible(element) {
+  return element.offsetParent !== null;
+}
+
+/**
+ * Menu wrapper component to facilitate keyboard navigation of a list of
+ * <MenuItem> components.
```
robertknight, line 11: I think it would be useful to clarify that this is a helper component used by `Menu` and `MenuItem` to wrap a list of `MenuItem`s, and that other components shouldn't need to use it directly.

src/sidebar/components/menu-item.js
```diff
@@ -56,49 +55,109 @@ export default function MenuItem({
...
+          <div className="menu-item__icon-container">{rightIcon}</div>
+        )}
+      </a>
+    );
+  } else {
+    // The menu item is a clickable button or radio button.
+    // In either case there may be an optional submenu.
+    const buttonProps = {
+      role: isSelected ? 'menuitemradio' : 'menuitem',
+      'aria-checked': isRadioButtonType ? isSelected.toString() : false,
```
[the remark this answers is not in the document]
LMS007: Oh, I forgot about that, I think you filed that ticket for preact right?

src/sidebar/components/menu-keyboard-navigation.js
LMS007: indeed! this was never working

src/sidebar/components/group-list-item.js
```diff
@@ -85,9 +90,11 @@ function GroupListItem({
       onToggleSubmenu={toggleSubmenu}
       submenu={
         <Fragment>
-          <ul>
+          {/* Giving the role="menu" allows screen readers to have
+          correct number of subitems */}
+          <ul role="menu">
```
LMS007, line 95: I noticed chromevox does not read the correct # of subitems unless we have a second `role="menu"` here.

src/sidebar/components/menu-item.js
```diff
@@ -56,49 +55,102 @@ export default function MenuItem({
...
+  const hasSubmenuVisible = typeof isSubmenuVisible === 'boolean';
+  const isRadioButtonType = typeof isSelected === 'boolean';
+
+  const onCloseSubmenu = event => {
+    if (onToggleSubmenu) {
+      onToggleSubmenu(event);
+    }
+    // The focus won't work without delaying rendering.
+    setTimeout(() => {
+      menuItemRef.current.focus();
```
LMS007, line 69: I still don't have a fix for focus other than this delay

src/sidebar/components/menu.js
```diff
@@ -96,7 +106,11 @@ export default function Menu({
   // It should also close if the user presses a key which activates menu items.
   const handleMenuKeyDown = event => {
     if (event.key === 'Enter' || event.key === ' ') {
-      closeMenu();
+      // Don't close the menu right away, the link still needs to be present at least one
+      // more render cycle for them to be followed.
+      setTimeout(() => {
```
LMS007, line 111: Shy of a callback, setting the close state from its parent, and a new `open` prop, I can't remove this timeout.

Notice that closeMenu does work from useElementShouldClose, but I suspect that is become the event by then already bubbled to the body, where as in this case, it has not.

src/styles/sidebar/components/menu-item.scss
LMS007: Ya, this is still needed, but I think it can be refactored so its simpler and less confusing.

src/sidebar/components/menu-keyboard-navigation.js
LMS007: I think this has to do with `display: contentVisible ? '' : 'none',` in slider.js, but I don't have a more appropriate fix at this time

src/sidebar/components/test/menu-item-test.js
```diff
@@ -19,125 +25,225 @@ describe('MenuItem', () => {
...
-    assert.equal(wrapper.find('Slider').children().text(), 'Submenu content');
+    it('calls `onToggleSubmenu` when the right arrow key is pressed', () => {
+      const fakeOnToggleSubmenu = sinon.stub();
+      const wrapper = createMenuItem({
+        isSubmenuVisible: true,
+        submenu: <div>Submenu content</div>,
+        onToggleSubmenu: fakeOnToggleSubmenu,
+      });
+      wrapper.find('.menu-item').simulate('keydown', { key: 'ArrowRight' });
+      assert.isTrue(fakeOnToggleSubmenu.called);
```
robertknight, line 228: It looks to me like this test duplicates the "calls the `onToggleSubmenu` callback when pressing arrow right" test above?

src/sidebar/components/test/menu-item-test.js
```diff
@@ -19,125 +25,225 @@ describe('MenuItem', () => {
...
+    });
+
+    it('sets focus to div.menu-item when the submenu is closed via `closeMenu`', () => {
```
robertknight, line 231: I prefer test descriptions to be written from the user's perspective where possible, as that makes it easy to see what the user-facing relevance of a particular test is (and thus, makes it easier to determine whether a test is still relevant when refactoring). In this case, `div.menu-item` means the "submenu's parent item"?

src/sidebar/components/test/menu-keyboard-navigation-test.js
```diff
@@ -0,0 +1,119 @@
...
+    it(`calls \`closeMenu\` callback when the ${key} is pressed`, () => {
+      const wrapper = createMenuItem({ visible: true }).find('div');
+      wrapper.simulate('keydown', { key });
+      assert.called(fakeCloseMenu);
+    });
+  });
+
+  // useFakeTimers does not work with checkAccessibility
+  // so wrap these tests in their own describe block
+  describe('fake timer', () => {
```
robertknight, line 54:
```suggestion
  describe('keyboard navigation', () => {
```

"fake timer" is more of a detail of how these tests work rather than a description of what feature or scenario is being tested.

src/sidebar/components/test/menu-test.js
```diff
@@ -161,9 +164,13 @@ describe('Menu', () => {
     it(`${
       shouldClose ? 'closes' : "doesn't close"
     } when user performs a "${eventType}" (key: "${key}") on menu content`, () => {
+      const clock = sinon.useFakeTimers();
```
robertknight, line 167: There is a hazard with making global changes to the environment in a test and restoring them at the end. The problem is that if the test fails mid-way through, the clock won't be restored and this will cause a cascade of failures in other tests. I can suggest two patterns to avoid this:

1. Use a `try..finally` block:

```js
const clock = sinon.useFakeTimers();
try {
  ...
} finally {
  // Still runs if `try` block throws an exception, but without stopping the exception from
  // propagating
  clock.restore();
}
```

2. Put the restore process inside a top-level `afterEach` block which runs conditionally only if needed:

```js
describe('MyWidget', () => {
  let clock;

  // Some tests call `clock = sinon.useFakeTimers()` in
  // their setup code, others do not.

  afterEach(() => {
    if (clock) {
      clock.restore();
    }
  ));
});
```

robertknight:
I left some notes on the PR description here: https://hyp.is/go?url=https%3A%2F%2Fgithub.com%2Fhypothesis%2Fclient%2Fpull%2F1992&group=wqvn7Lar

I took a look at the issue with focus rings in submenu items and found that it was a problem with the `Slider` component, which is fixed in https://github.com/hypothesis/client/pull/2039.

src/sidebar/components/group-list-item.js
robertknight: I think it would be preferable to put this on the `.menu-item__submenu` div if possible, otherwise each time a dev authors a new component that renders submenus they will need to know to do this.

LMS007:
> There is one more interaction issue



> From a user-facing perspective this is looking really good. I had a read through the code and that is looking good too, though there is quite a lot here and I think it will take me a couple of passes to review fully.
> 
> I opened a [separate PR](https://github.com/hypothesis/client/pull/2039) to address an issue with the `Slider` component that is causing focus rings of submenu items to be clipped.
> 
> There is one more interaction issue which I think will be quite confusing for users, but I think we should address only after this PR is merged: Currently menu items that are links don't respond to the "Space" key whereas other menu items do. That behavior is normal for buttons and links, but I think it is surprising in this context since both types of menu item look the same visually, and native menus don't have different behavior depending on what kind of action is triggered.

Right, I'm assuming we could do something like window.location when we detect space on a link. I may need to look at how the examples were done in w3.org. I'll have to get into the debugger a bit for that

src/sidebar/components/menu-item.js
```diff
@@ -56,49 +55,102 @@ export default function MenuItem({
...
+        )}
+        <span className="menu-item__label">{label}</span>
+        {hasRightIcon && (
+          <div className="menu-item__icon-container">{rightIcon}</div>
+        )}
+      </a>
+    );
+  } else {
+    // The menu item is a clickable button or radio button.
+    // In either case there may be an optional submenu.
```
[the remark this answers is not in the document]
LMS007: Okay, that could also be changed so thats not a condition, but seems reasonable to leave for now

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -0,0 +1,105 @@
...
+
+function isElementVisible(element) {
+  return element.offsetParent !== null;
+}
+
+/**
+ * Helper component used by Menu and MenuItem to facilitate keyboard navigation of a
+ * list of <MenuItem> components. This component should not be used directly.
+ *
+ * Note that `ArrowRight` shall be handled by the <MenuItem> directly and
```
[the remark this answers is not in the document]
LMS007: indeed!

src/sidebar/components/menu.js
```diff
@@ -96,7 +106,11 @@ export default function Menu({
   // It should also close if the user presses a key which activates menu items.
   const handleMenuKeyDown = event => {
     if (event.key === 'Enter' || event.key === ' ') {
-      closeMenu();
+      // Don't close the menu right away, the link still needs to be present at least one
+      // more render cycle for them to be followed.
```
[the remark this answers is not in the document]
LMS007: Well, more specifically I think this works because the event bubbles up when you call setTimeout allowing it be handled before the anchor is deleted. I demonstrated this here 
https://jsfiddle.net/ft8z1wux/7
How or why that happens in a browser is not clear to me, so we should change that comment around

src/sidebar/components/test/menu-item-test.js
```diff
@@ -19,125 +25,225 @@ describe('MenuItem', () => {
...
-    // The actual icon for the submenu should be shown on the right.
-    assert.equal(iconSpaces.at(1).children().length, 1);
-  });
+    it('calls the `onToggleSubmenu` callback when pressing arrow right', () => {
+      const fakeOnToggleSubmenu = sinon.stub();
+      const wrapper = createMenuItem({
+        isSubmenuVisible: true,
+        onToggleSubmenu: fakeOnToggleSubmenu,
+      });
+      wrapper.find('.menu-item').simulate('keydown', { key: 'ArrowRight' });
```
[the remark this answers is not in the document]
LMS007: Or fail tests that don't call assert, but that could be tricky

src/sidebar/components/test/menu-item-test.js
LMS007: yes indeed, good catch

src/sidebar/components/test/menu-item-test.js
LMS007: correct yes that was name was bad!

src/sidebar/components/test/menu-test.js
LMS007: good idea

robertknight:
I cherry-picked b2887be5ab3b29cafc23c6eb531f682482748213 onto this branch to make testing the groups sub-menus easier.

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <div
+      role="presentation"
+      className={className}
+      ref={menuRef}
+      onKeyDown={onKeyDown}
+    >
+      {/* role="menu" helps screen readers separate submenu items from parent menu items */}
+      <div role="menu">{children}</div>
```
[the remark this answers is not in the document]
LMS007: Yes, it caused the linter to complain about having role="menu" along with `onKeyDown` This one one way to sidestep that, the other way would be a to set ignore on the linter.

src/sidebar/components/menu.js
```diff
@@ -76,6 +78,14 @@ export default function Menu({
       event.preventDefault();
       return;
     }
+    // State variable so we know to set focus() on the first item when opened
+    // via the keyboard.
+    if (!isOpen && event.type === 'click') {
+      setOpenedByKeyboard(true);
```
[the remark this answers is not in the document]
LMS007: I meant to circle back to this in more detail but forgot (diff is a bit large)
Turns out when you click on the menu with a mouse or a keyboard, both the events end up being "MouseEvent" for some strange reason (at least in Chrome), but the difference is the type when you click with a mouse is "mousedown" and the type for the keyboard is "click" which does feel totally wrong.

We probably need a comment here or maybe get to the bottom of the reason for this.

src/sidebar/components/menu.js
```diff
@@ -96,7 +106,11 @@ export default function Menu({
   // It should also close if the user presses a key which activates menu items.
   const handleMenuKeyDown = event => {
     if (event.key === 'Enter' || event.key === ' ') {
-      closeMenu();
+      // The link will not follow if its removed directly from an keypress event.
```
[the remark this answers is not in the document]
LMS007: Yes, I'm not sure if you saw my JSfiddle on this, but it can be demonstrated with just a simple few liens of js.

src/sidebar/components/test/menu-item-test.js
```diff
@@ -8,8 +9,13 @@ import { checkAccessibility } from '../../../test-util/accessibility';
 import mockImportedComponents from '../../../test-util/mock-imported-components';
 
 describe('MenuItem', () => {
-  const createMenuItem = props =>
-    mount(<MenuItem label="Test item" {...props} />);
+  const createMenuItem = props => {
+    const container = document.createElement('div');
+    document.body.appendChild(container);
+    return mount(<MenuItem label="Test item" {...props} />, {
+      attachTo: container,
```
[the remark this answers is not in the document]
LMS007: yes, we use that pattern in tag-editor-test, I'll use it here too.

src/sidebar/components/test/menu-item-test.js
```diff
@@ -113,9 +119,12 @@ describe('MenuItem', () => {
   });
 
   describe('submenu', () => {
+    afterEach(() => {});
+
     it('shows the submenu indicator if `isSubmenuVisible` is a boolean', () => {
       const wrapper = createMenuItem({
         isSubmenuVisible: true,
+        submenu: <div role="menuitem">Submenu content</div>,
```
LMS007, line 127: Some of these tests were causing prop error warnings because `children` is required for the `menu-keyboard-navigation` component

src/sidebar/components/test/menu-keyboard-navigation-test.js
```diff
@@ -107,6 +113,13 @@ describe('MenuKeyboardNavigation', () => {
       wrapper.simulate('keydown', { key: 'Home' });
       assert.equal(document.activeElement.innerText, 'Item 1');
     });
+
+    it('does not throw an error when unmounting the component before the focus timeout finishes', () => {
+      const wrapper = createMenuItem({ visible: true });
+      wrapper.unmount();
+      clock.tick(1);
```
LMS007, line 120: I confirmed this test fails when commenting out the respective clearTimeout() call in the component

src/sidebar/components/test/menu-item-test.js
```diff
@@ -220,19 +233,43 @@ describe('MenuItem', () => {
...
+            .closeMenu({ key: 'Enter' });
+        });
+        clock.tick(1);
+        assert.equal(document.activeElement.className, 'menu-item');
+      } finally {
+        clock.restore();
+      }
+    });
+
+    it('does not throw an error when unmounting the component before the focus timeout finishes', () => {
```
LMS007, line 254: I confirmed this test fails when commenting out the respective clearTimeout() call in the component

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -20,17 +22,22 @@ export default function MenuKeyboardNavigation({
   visible,
 }) {
   const menuRef = useRef(null);
+  const focusTimer = useRef(null);
```
robertknight, line 25: You don't need to use a ref here since you're only referencing the value in the `useEffect` callback. You can just declare a `let focusTimer` variable at the top of that function.

src/sidebar/components/menu-keyboard-navigation.js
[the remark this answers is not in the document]
robertknight: I would prefer to suppress the lint warning rather than do something unnecessary with the DOM structure.

src/sidebar/components/menu.js
```diff
@@ -73,13 +75,17 @@ export default function Menu({
     if (event.type === 'mousedown') {
       ignoreNextClick = true;
     } else if (event.type === 'click' && ignoreNextClick) {
+      // ignore mouseup event
```
robertknight, line 78: I presume this comment means "ignore `click` event _caused_ by a mouse release", rather than an actual `mouseup` event which is a distinct DOM event?

src/sidebar/components/menu.js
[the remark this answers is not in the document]
robertknight: OK, can you change the comment to something like "The browser will not open the link if the link element is removed from within the keypress event that triggers it."? That's a bit verbose but hopefully more obvious for future readers.

src/sidebar/components/test/menu-item-test.js
```diff
@@ -113,9 +119,12 @@ describe('MenuItem', () => {
   });
 
   describe('submenu', () => {
+    afterEach(() => {});
```
robertknight, line 122: This `afterEach` does nothing and can be removed.

src/sidebar/components/menu-item.js
```diff
@@ -60,18 +64,25 @@ export default function MenuItem({
   const hasSubmenuVisible = typeof isSubmenuVisible === 'boolean';
   const isRadioButtonType = typeof isSelected === 'boolean';
 
+  useEffect(() => {
+    return () => {
+      // unmount
+      clearTimeout(focusTimer.current);
```
robertknight, line 70: I assume this does not throw an exception or display a warning if `focusTimer.current` is `null`?

src/sidebar/components/menu.js
LMS007: yep, I'll make that more clear

src/sidebar/components/menu-item.js
LMS007: correct

src/sidebar/components/menu-keyboard-navigation.js
LMS007: I was trying to avoid the `exhaustive-deps` warning about that, but I'll change it, then add that comment.

src/sidebar/components/menu-item.js
robertknight: I see what you mean. I can't immediately see why this is happening. As long as the reason for having this timeout is documented in the code, I think we can live with it for now until one of us has time to debug this more thoroughly.

src/sidebar/components/menu-item.js
```diff
@@ -56,49 +59,107 @@ export default function MenuItem({
...
+          <div className="menu-item__icon-container">{leftIcon}</div>
+        )}
+        <span className="menu-item__label">{label}</span>
+        {hasRightIcon && (
+          <div className="menu-item__icon-container">{rightIcon}</div>
+        )}
+
+        {hasSubmenuVisible && (
           <div
+            role="none"
```
robertknight, line 158: Maybe add a brief comment here about why the use of `role="none`.

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -0,0 +1,118 @@
...
+function isElementVisible(element) {
+  return element.offsetParent !== null;
+}
+
+/**
+ * Helper component used by Menu and MenuItem to facilitate keyboard navigation of a
+ * list of <MenuItem> components. This component should not be used directly.
+ *
+ * Note that `ArrowRight` shall be handled by the parent <MenuItem> directly and
+ * all other focus() related  navigation is handled here.
```
robertknight, line 16:
```suggestion
 * all other focus() related navigation is handled here.
```

Extra space removed.

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -0,0 +1,118 @@
...
+      event.preventDefault();
+      menuItems[focusedIndex].focus();
+    }
+  };
+
+  return (
+    // This element needs to have role="menu" to facilitate readers
+    // correctly enumerating discrete submenu items, but it also needs
+    // to facilitate keydown events for navigation. Disable the linter
+    // error so it can do both.
```
robertknight, line 97: Great comment 👍

src/sidebar/components/test/menu-test.js
```diff
@@ -161,8 +169,11 @@ describe('Menu', () => {
     it(`${
       shouldClose ? 'closes' : "doesn't close"
     } when user performs a "${eventType}" (key: "${key}") on menu content`, () => {
+      clock = sinon.useFakeTimers();
```
robertknight, line 172: Can you add a short note about why this is needed?

src/styles/sidebar/components/menu-item.scss
```diff
@@ -3,96 +3,91 @@
...
 
 $menu-item-padding: 10px;
+$item-height: 40px;
 
-.menu-item {
-  @include focus.outline-on-keyboard-focus;
+// The container for the clickable area of the menu item which triggers an
+// action or link. For menu items without submenus, this covers the full area of
+// the menu item. For menu items with submenus, this covers the full area
+// except for the toggle button that opens the submenu.
```
robertknight, line 11: I wasn't initially completely sure which CSS this comment referred to. I think it would have been more obvious if the `a.menu-item` rule was nested under `.menu-item`, but AFAIK it isn't possible to do this with SASS (`a&:hover` isn't allowed for example as  the `&` parent selector must appear at the start of the selector).

Let's leave as-is for the moment.

PR #1993, opened 2020-04-06
Remove annotation-mapper service and associated Angular events

robertknight:
This PR removes the legacy annotation-mapper service and the Angular events that it used to trigger. It is part of https://github.com/hypothesis/client/issues/1994.

This service was used to load annotations into the sidebar after they were fetched from the API as well as to delete/flag annotations and update the store afterwards. Previously the methods in this service would make an API call or otherwise process the input data, and then emit an Angular event which was picked up elsewhere in the app (`rootThread` or `<hypothesis-app>`) which would in turn dispatch a store action.

In line with other recent changes, this has been reworked so that:

 - Angular events which were only used to trigger store actions are replaced with direct calls to the store
 - Methods that update or delete annotations via the API and then update the store afterwards are consolidated in the annotations service (`src/sidebar/services/annotations.js`).

In detail, the annotation-mapper service had 4 methods which have been replaced as follows:

- `loadAnnotations` - This method was used to load annotations fetched from the API into the store. This has been replaced with a direct call to `store.addAnnotations`.
- `createAnnotation` - This method was unused as it has already been replaced by the `create` method in the `annotations` service
- `flagAnnotation` - This was called to flag an annotation in the API and update the store. This has been replaced with a `flag` method in the `annotations` service.
- `deleteAnnotation` - This was called to delete an annotation in the API and update the store. This has been replaced with a `delete` method in the `annotations` service

There is one other related change in the `streamer` service which removes use of `$rootScope.$apply` that is no longer needed.


src/sidebar/components/test/annotation-action-bar-test.js
```diff
@@ -276,26 +275,12 @@ describe('AnnotationActionBar', () => {
         button.props().onClick();
       });
 
-      assert.calledWith(fakeAnnotationMapper.flagAnnotation, fakeAnnotation);
-    });
-
-    it('updates flag state in store after flagging on service is successful', async () => {
```
robertknight, line 282: Updating the flag state in the store after calling the API is now done by `annotationsService.flag`.

src/sidebar/components/test/stream-content-test.js
```diff
@@ -44,9 +40,7 @@ describe('StreamContentController', function () {
     };
 
     fakeApi = {
-      search: sinon.spy(function () {
-        return Promise.resolve({ rows: [], total: 0 });
-      }),
+      search: sinon.stub().resolves({ rows: [], replies: [], total: 0 }),
```
robertknight, line 43: This updates the fake API response to include the `replies` field, which was missing before.

src/sidebar/services/annotations.js
```diff
@@ -116,6 +116,22 @@ export default function annotationsService(api, store) {
     });
   }
 
+  /**
+   * Delete an annotation via the API and update the store.
+   */
+  async function delete_(annotation) {
```
robertknight, line 122: The function is named `delete_` because you can't call a function or binding `delete`, although you _can_ use that name for a method.

src/sidebar/services/streamer.js
```diff
@@ -81,32 +65,27 @@ export default function Streamer(
   }
 
   function handleSocketOnMessage(event) {
-    // Wrap message dispatches in $rootScope.$apply() so that
-    // scope watches on app state affected by the received message
-    // are updated
-    $rootScope.$apply(function () {
```
robertknight, line 87: The change here just removes a `$rootScope.$apply(...)` "wrapper" around the code. Historically this was needed to update the UI after WebSocket messages were processed, but this is obsolete now as Redux store subscribers handle updating the UI after processing actions.

PR #2014, opened 2020-04-08
Update Preact

robertknight:
This fixes an issue where an uncaught error thrown inside an `act` callback would
stop any `useEffect` effects from running for the rest of the test
session, which was pretty confusing to debug.

See https://github.com/preactjs/preact/releases/tag/10.4.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "npm-packlist": "^2.0.1",
     "postcss": "^7.0.13",
     "postcss-url": "^8.0.0",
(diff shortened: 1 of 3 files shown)


robertknight:
CI run looks good. I'll deploy this tomorrow.

robertknight:
CC @lyzadanger as I think you originally reported the issue internally where one test failure could surprisingly cause a cascade of failures in other tests.

PR #2015, opened 2020-04-08
Add assert.rejects()

LMS007:
I attempted to remove `assertPromiseIsRejected` a while back. https://github.com/hypothesis/client/pull/1876/

I did some research on Chai's helpers https://www.chaijs.com/guide/helpers/ 
From the docs, it appears that helpers, both methods or properties, are only available through the `expect` / `should` interfaces rather than the `assert` interface (to the best of what I could tell)

I decided to just polyfill the `rejects` property directly on the global assert object in bootstrap.js. The new file `assert-polyfills` will serve as new methods that can attach to the assert object via its `patch` method. (see bootstrap.js)


src/sidebar/test/bootstrap.js
```diff
@@ -1,6 +1,22 @@
...
+    const error = new Error();
+    error.name = 'ErrorNotCaught';
+    throw error;
+  } catch (e) {
+    if (e.name === 'ErrorNotCaught') {
+      throw new Error(
+        'Expected to catch the rejected promise but it was not thrown'
+      );
+    }
+    assert.equal(error, e.message);
```
robertknight, line 16: From experience with using `assert.throws`, I think it would be useful to allow the use of partial matching by passing a regex as the value of `error`.

src/sidebar/test/bootstrap.js
```diff
@@ -1,6 +1,22 @@
 // Expose the sinon assertions.
 sinon.assert.expose(assert, { prefix: null });
 
+assert.rejects = async (asyncFn, error) => {
```
robertknight, line 4: I was slightly confused initially as here `asyncFn` is actually a _promise_ rather than a function that returns a Promise, which is what the name `asyncFn` implies to me.

robertknight, line 4: It might be useful to create a module that exports a function which can be used to patch `assert`, like we do with `sinon.assert.expose` above. This would live in `src/test-util`.

This way we can package it up for re-use in other applications.

src/sidebar/test/bootstrap.js
LMS007: good point, I think I named that before I sorted out how I wanted the param to look.

src/sidebar/test/bootstrap.js
```diff
@@ -1,6 +1,10 @@
 // Expose the sinon assertions.
 sinon.assert.expose(assert, { prefix: null });
 
+// Patch any extra assert methods defined in assert-polyfills
+import patch from '../../test-util/assert-polyfills';
+patch(assert);
```
LMS007, line 6: @robertknight I'm just not sure about the names or if this was what you were expecting.

robertknight: Hi @LMS007 - This looks good to me except that the term _polyfill_ specifically refers to an implementation of a feature that is part of the documented behaviour or API of a system, but which is missing from the current implementation. Here we are adding extra methods to `assert` that are not part of its normal API.

Perhaps `assert-extensions` or `assert-methods`, with corresponding changes to comments in the assert module.

src/test-util/assert-methods.js
```diff
@@ -0,0 +1,45 @@
...
+ * The `promiseResult` is awaited, and then upon any exception, the resulting
+ * error message is matched against the `errorMessage` string or regex. This
+ * method throws an error if `promiseResult` does not throw an error or will
+ * throw an assertion error if the caught error message differs from the
+ * `errorMessage` parameter.
+ *
+ * e.g. await rejects(someAsyncFunction(), /expected error/g);
+ *
+ * @param {Promise} promiseResult - The returned promise a function to test
+ * @param errorMessage {RegEx|String} - A string or regex that matches the error
```
robertknight, line 13:
```suggestion
 * @param {RegEx|string} errorMessage - A string or regex that matches the error
```

src/test-util/assert-methods.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+};
+
+/**
+ * Patches the assert object with additional custom helper methods
+ * defined in this module.
+ *
+ * @param {Object} assert - global assertion object.
+ */
+export default function patch(assert) {
```
robertknight, line 43: I would be inclined to export `patch` as a named rather than default export, since it doesn't seem obvious to me if you read `import something from '.../assert-methods` what "something" is likely to be.

Issue #1995, opened 2020-04-06
Highlights on PDFs reduces text contrast


robertknight:
On web pages, highlights are drawn _behind_ the highlighted text using CSS `background-color`. On PDFs however, the selectable text layer is actually a transparent layer on top of a canvas which is where the text that the user sees is rendered. As a result, the highlights are rendered _on top_ of the text and they end up reducing its visibility:

<img width="590" alt="Screenshot 2020-04-06 10 50 41" src="https://user-images.githubusercontent.com/2458/78546154-dbb4b480-77f4-11ea-9511-a1f962732e33.png">

This problem is exacerbated by the issue that some PDFs have text that is not super sharp / contrasting in the first place.

Closely related issue: https://github.com/hypothesis/client/issues/408. The difference from that issue is that this one applies even when there are no overlapping annotations or one layer of overlap.

Issue #408, opened 2017-05-29
Overlapping annotations on PDFs cause text to be unreadable


dwhly:
### Steps to reproduce
1. Open this link: https://hyp.is/1HFqPDR4EeeDM2deDdF8Rw/www.heartland.org/_template-assets/documents/Books/Why%20Scientists%20Disagree%20Second%20Edition%20with%20covers.pdf

### Expected behaviour
The selected text on the PDF should be clear and visible on the page.

### Actual behaviour
There are so many overlapping highlights that the text is nearly completely obscured.

### Browser/system information
Chrome / extension

### Additional details
Seems like the way we use layers to communicate highlights results in diminished visibility of the text with each layer for PDFs.  May need to fundamentally change approach, or use lighter gradients for the 2nd, 3rd and 4th overlaps.  Perhaps after 3-4 overlaps we shouldn't darken further.

robertknight:
https://github.com/hypothesis/client/issues/1995 explains the underlying issue and why it happens with PDFs but not web pages. In short, highlights in PDFs are drawn _on top of_ the text whereas in web pages they are drawn _underneath_ the text.

robertknight:
The situation has been significantly improved with https://github.com/hypothesis/client/pull/1996, although there is still some loss of readability as the highlights are still drawn _on top_ of the text in PDFs.

PR #2017, opened 2020-04-09
Improve readability of PDF highlights

robertknight:
Improve the readability of highlights on PDFs by creating the highlights
using SVG elements positioned just above the page's canvas instead of
using the CSS `background-color` property on the
`<hypothesis-highlight>` elements in the page's text layer.

Using SVG elements placed in the DOM like this allows us to control how
the highlight is blended with the content underneath using CSS
`mix-blend-mode`. Using the `multiply` blend mode [2] means that highlights
will darken the content below rather than making dark text in the
canvas appear lighter and muddier. Additionally this approach gives us
more control over the appearance of overlapping highlights. Note that
for the custom blending to work, it is important that the SVG is in the
same stacking context as the canvas [1]

We still need to keep the `<hypothesis-highlight>` elements in the text layer
for interactive functionality (eg. interacting with highlights using the
keyboard or pointer). The SVG highlight is associated with the
`<hypothesis-highlight>` via an `svgHighlight` property so that the SVG
can be removed when the highlight itself is removed.

An additional benefit of this approach is that it will allow us to bump up the contrast used
for highlights on web pages, as some users have reported they are too feint (eg. https://github.com/hypothesis/product-backlog/issues/1099),
without negatively affecting readability of highlighted text in PDFs.

[1] https://drafts.fxtf.org/compositing-1/#csscompositingrules_CSS
[2] https://drafts.fxtf.org/compositing-1/#valdef-blend-mode-multiply

Fixes https://github.com/hypothesis/client/issues/408
Fixes https://github.com/hypothesis/client/issues/1995

----

Example: https://docdrop.org/static/drop-pdf/The-Pedagogy-of-Listening_Retyped-WZ1Tv.pdf

(Note that the kerning issues in this PDF are an issue with either PDF.js or the PDF itself) 

**Master:**

pdf-highlighting-master

**This branch:**

pdf-highlighting-with-svgs

----

- [x] Investigate highlight creation and scrolling performance compared with master
- [x] Test with more PDFs
- [x] Write tests
- [x] Test behavior in browsers that don't support `mix-blend-mode` (IE 11, Edge < 79) (Can still use SVG, but may need to use an `feBlend` or fall back to "normal" blending with lower opacity)
- [x] Make highlight visibility toggle button work with SVG-based highlights

### DIFF
--- a/src/annotator/highlighter.js
+++ b/src/annotator/highlighter.js
@@ -1,3 +1,122 @@
+const SVG_NAMESPACE = 'http://www.w3.org/2000/svg';
+
+/**
+ * Polyfill for `element.closest(selector)`, only needed for IE 11.
+ */
+function closest(element, selector) {
+  while (element) {
+    if (element.matches(selector)) {
+      return element;
+    }
+    element = element.parentElement;
+  }
+  return null;
+}
+
+/**
+ * Return the canvas element underneath a highlight element in a PDF page's
+ * text layer.
+ *
+ * Returns `null` if the highlight is not above a PDF canvas.
+ *
+ * @param {HTMLElement} highlightEl -
+ *   A `<hypothesis-highlight>` element in the page's text layer
+ * @return {HTMLCanvasElement|null}
+ */
+function getPdfCanvas(highlightEl) {
+  // This code assumes that PDF.js renders pages with a structure like:
+  //
+  // <div class="page">
+  //   <div class="canvasWrapper">
+  //     <canvas></canvas> <!-- The rendered PDF page -->
+  //   </div>
+  //   <div class="textLayer">
+  //      <!-- Transparent text layer with text spans used to enable text selection -->
+  //   </div>
+  // </div>
+  //
+  // It also assumes that the `highlightEl` element is somewhere under
+  // the `.textLayer` div.
+
+  const pageEl = closest(highlightEl, '.page');
+  if (!pageEl) {
+    return null;
+  }
+
+  const canvasEl = pageEl.querySelector('.canvasWrapper > canvas');
+  if (!canvasEl) {
+    return null;
+  }
+
+  return canvasEl;
+}
+
+/**
+ * Draw highlights in an SVG layer overlaid on top of a PDF.js canvas.
+ *
+ * Returns `null` if `highlightEl` is not above a PDF.js page canvas.
+ *
+ * @param {HTMLElement} highlightEl -
+ *   An element that wraps the highlighted text in the transparent text layer
+ *   above the PDF.
+ * @return {SVGElement|null} -
+ *   The SVG graphic element that corresponds to the highlight or `null` if
+ *   no PDF page was found below the highlight.
+ */
+function drawHighlightsAbovePdfCanvas(highlightEl) {
+  const canvasEl = getPdfCanvas(highlightEl);
+  if (!canvasEl) {
+    return null;
+  }
+
+  let svgHighlightLayer = canvasEl.parentElement.querySelector(
+    '.hypothesis-highlight-layer'
+  );
+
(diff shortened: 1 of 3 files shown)


robertknight:
I've switched from one `<svg>` per highlight to one `<svg>` per page. This seems to have improved performance somewhat. I have also added tests and improved documentation.

Some more screenshots:

**Master:**

Pedagogy of Learning highlight - CSS background

**This branch:**

Pedagogy of Learning PDF highlight - SVG

robertknight:
The visual effects shown above rely on [support for the CSS `mix-blend-mode` property](https://caniuse.com/#search=mix-blend) which is not available in IE 11 and Edge < 79. I have implemented a fallback which visually looks the same as on master, except that non-overlapping and overlapping highlights are all rendered with a uniform shade, limiting the impact to readability:

**Modern browsers:**

Pedagogy of Learning - modern

**IE 11 / Edge < 79:**

Pedagogy of Learning - SVG no-blend fallback

In non-IE browsers you can trigger this fallback by locally modifying the `isCSSPropertySupported` function in `highlighter.js` to return `false`.

src/annotator/highlighter.js
```diff
@@ -78,6 +233,10 @@ export function removeHighlights(highlights) {
       const children = Array.from(h.childNodes);
       replaceWith(h, children);
     }
+
+    if (h.svgHighlight) {
```
LMS007, line 237: Is there any case where svgHighlight prop would be missing and would that be a problem?

src/annotator/test/highlighter-test.js
```diff
@@ -6,6 +8,60 @@ import {
...
+}
+
+/**
+ * Highlight the text in a fake PDF page.
+ *
+ * @param {HTMLElement} - HTML element into which `PdfPage` component has been
+ *   rendered
+ * @return {HTMLElement} - `<hypothesis-highlight>` element
+ */
+function highlightPdfRange(pdfPage) {
```
LMS007, line 41: pdfPage is also the name of the function above, I would suggest a discrete name to avoid confusion.

src/annotator/highlighter.js
robertknight: This property won't be set for highlights on HTML content (ie. not a PDF), in which case it will just evaluate to `undefined` here and this branch will be skipped.

src/annotator/highlighter.js
```diff
@@ -1,3 +1,146 @@
...
+function isCSSPropertySupported(property, value) {
+  if (typeof CSS !== 'function' || typeof CSS.supports !== 'function') {
+    /* istanbul ignore next */
+    return false;
+  }
+  return CSS.supports(property, value);
+}
+
+/**
+ * Polyfill for `element.closest(selector)`, only needed for IE 11.
```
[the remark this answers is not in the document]
robertknight: Indeed. I've updated this comment. In some circles this would be described as a [ponyfill](https://kikobeats.com/polyfill-ponyfill-and-prollyfill/).

robertknight:
> I did notice some strange artifact on this example pdf, but its not clear where in the DOM is appearing from

Ah, good catch. On my system at least it turned out to be quite reproducible on many PDFs when zooming in and out. These highlights turned out to be "placeholder highlights" which are connected to how we anchor to content on PDF pages which have not yet finished rendering. They are supposed to be visually hidden, so we shouldn't create SVGs for them. See https://github.com/hypothesis/client/pull/2017/commits/d2fa03c596dce711804b1068c90730bdf6d66774 for the fix and a test case.

src/annotator/highlighter.js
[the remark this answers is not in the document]
LMS007: Looking at that document..."Prollyfill" - lol

PR #2021, opened 2020-04-10
Various fixes for IE11

LMS007:
- Add es2018 Polyfill check for Promise existence
- Add isIE11() and isMacOS() helper methods
- Fix tag-editor to allow for IE11 specific arrow key events
- Fix markdown-editor to allow for IE11 specific arrow key events
- Add conditional `dir` attribute when IE is detected

There is still one console error remaining and a few self-closing tag warnings, but the sidebar does appear to be working again

Image4

### DIFF
--- a/src/shared/browser-compatibility-utils.js
+++ b/src/shared/browser-compatibility-utils.js
@@ -1,0 +1,22 @@
+/**
+ * Normalize a keyboard event key name.
+ *
+ * Several old browsers, such as IE11, use alternate key
+ * names for keyboard events. If any abnormal keys are used,
+ * this method returns the normalized name so our UI
+ * components don't require a special case.
+ *
+ * @param {string} key - The keyboard event `key` name
+ * @return {string} - Normalized `key` name
+ */
+export function normalizeKeyName(key) {
+  const mappings = {
+    Left: 'ArrowLeft',
+    Up: 'ArrowUp',
+    Down: 'ArrowDown',
+    Right: 'ArrowRight',
+    Spacebar: ' ',
+    Del: 'Delete',
+  };
+  return mappings[key] ? mappings[key] : key;
+}
--- a/src/shared/test/user-agent-test.js
+++ b/src/shared/test/user-agent-test.js
@@ -1,6 +1,6 @@
 import { isIE11, isMacOS } from '../user-agent';
 
-describe('sidebar/util/user-agent', () => {
+describe('shared/user-agent', () => {
   describe('isIE11', () => {
     it('returns true when the user agent is IE 11', () => {
       assert.isTrue(
--- a/src/shared/test/user-agent-test.js
+++ b/src/shared/test/user-agent-test.js
@@ -1,0 +1,37 @@
+import { isIE11, isMacOS } from '../user-agent';
+
+describe('sidebar/util/user-agent', () => {
+  describe('isIE11', () => {
+    it('returns true when the user agent is IE 11', () => {
+      assert.isTrue(
+        isIE11('Mozilla/5.0 (Windows NT 10.0; Trident/7.0; rv:11.0) like Gecko')
+      );
+    });
+
+    it('returns false when the user agent is not IE 11', () => {
+      assert.isFalse(
+        isIE11(
+          'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.92 Safari/537.36 Edg/80.0.361.109'
+        )
+      );
+    });
+  });
+
+  describe('isMacOS', () => {
+    it('returns true when the user agent is a Mac', () => {
+      assert.isTrue(
+        isMacOS(
+          'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.163 Safari/537.36'
+        )
+      );
+    });
+
+    it('returns false when the user agent is not a Mac', () => {
+      assert.isFalse(
+        isMacOS(
+          'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.92 Safari/537.36 Edg/80.0.361.109'
+        )
+      );
+    });
+  });
+});
--- a/src/sidebar/components/markdown-editor.js
+++ b/src/sidebar/components/markdown-editor.js
@@ -9,6 +9,7 @@ import {
   toggleBlockStyle,
   toggleSpanStyle,
 } from '../markdown-commands';
+import { normalizeKeyName } from '../../shared/browser-compatibility-utils';
 import { isMacOS } from '../../shared/user-agent';
 
 import MarkdownView from './markdown-view';
@@ -212,9 +213,8 @@ function Toolbar({ isPreviewing, onCommand, onTogglePreview }) {
       lowerLimit = buttonIds.help;
     }
     let newFocusedElement = null;
-    switch (e.key) {
+    switch (normalizeKeyName(e.key)) {
       case 'ArrowLeft':
-      case 'Left': // IE11
         if (rovingElement <= lowerLimit) {
           newFocusedElement = upperLimit;
         } else {
@@ -222,7 +222,6 @@ function Toolbar({ isPreviewing, onCommand, onTogglePreview }) {
         }
         break;
       case 'ArrowRight':
-      case 'Right': // IE11
         if (rovingElement >= upperLimit) {
           newFocusedElement = lowerLimit;
         } else {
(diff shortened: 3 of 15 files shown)


src/sidebar/components/markdown-editor.js
```diff
@@ -213,13 +214,15 @@ function Toolbar({ isPreviewing, onCommand, onTogglePreview }) {
     let newFocusedElement = null;
     switch (e.key) {
       case 'ArrowLeft':
+      case 'Left': // IE11
```
robertknight, line 217: Rather than needing to do this in each component which handles this event, there is a way we can define a global hook that pre-processes events before they are passed to components.

On application startup (so, from `src/sidebar/index.js`) you would need something like the following:

```js
import { options } from 'preact/hooks';

const prevEventHook = options.event;
options.event = event => {
  if (event.type === 'keydown' || event.type === 'keyup') {
    // Normalize `event.key`
  }
  prevEventHook(event);
};
```

The advantage is that we can re-use this in other applications more easily.

src/sidebar/util/user-agent.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Helper methods to identify browser versions and os types
+ */
+
+export const isIE11 = (_window = window) => {
+  return _window.navigator.userAgent.indexOf('rv:11') >= 0;
```
robertknight, line 6: This test might be overly generic. If we have test for IE specifically, I would probably rather test for the `Trident/7.0` part, since "Trident" is the codename for IE's browsing engine.

src/sidebar/util/user-agent.js
```diff
@@ -0,0 +1,11 @@
...
+};
+
+export const isMacOS = (_window = window) => {
+  return _window.navigator.userAgent.indexOf('Mac OS') >= 0;
```
robertknight, line 10: These functions could accept the user agent string as an argument rather than `window`, with a default value of `navigator.userAgent`, since `navigator.userAgent` is a global that exists in all browsers we support (and many much older ones we do not).

src/sidebar/components/markdown-editor.js
robertknight: For reference, the full list of available hooks [is defined here](https://github.com/preactjs/preact/blob/3426ec572ca465fbe959692625dacdffab75e2a5/src/index.d.ts#L255).

LMS007: I'm only seeing `unmount`, `diffed` and `vnode`. `event` is undefined so when I override options.event I wind up with all events broken.

src/sidebar/util/user-agent.js
LMS007: I agree

src/sidebar/components/annotation-quote.js
```diff
@@ -22,6 +24,9 @@ function AnnotationQuote({ annotation, settings = {} }) {
   // as part of the annotation anchoring process.
   const documentLanguage = '';
 
+  // IE11 does not support `dir=` attribute.
+  const dirProp = useMemo(() => (isIE11() ? {} : { dir: 'auto' }), []);
```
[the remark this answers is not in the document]
LMS007: https://www.w3.org/International/tests/repo/results/the-dir-attribute-auto

 IE11 is not the easiest to debug or even understand but when adding the dir= attribute to an `<input>`, for example, it no longer renders in a predictable way.  The layout will be severely broken and more or less unusable (or at least the element with that attribute is unusable). I see no way of getting IE11 to work when adding this attribute to an input or textarea so I think its best left to a conditional while we support the browser.

src/sidebar/components/markdown-editor.js
robertknight: > I'm only seeing unmount, diffed and vnode. event is undefined so when I override options.event I wind up with all events broken.

Ah, my choice of terminology confused things. What I talked about as "hooks" above are called `options` in the Preact docs. The `hooks/src/index.js` file refers to the _other_ kind of hook which is React hooks (`useState`, `useEffect` etc.)

My code example above was also wrong. [Per the docs](https://qa.hyp.is/2cCBVn4ZEeqNpI8AiXx0Jw/preactjs.com/guide/v10/options), the `event` hook/option needs to _return_ the event, rather than just pass it on. Here is a working example:

```js
// Quake key navigation support.
import { options } from 'preact';

const bindings = {
  a: 'ArrowLeft',
  w: 'ArrowUp',
  s: 'ArrowDown',
  d: 'ArrowRight',
};
const prevEventHook = options.event;
options.event = event => {
  if (event.key in bindings) {
    const key = bindings[event.key];
    Object.defineProperty(event, 'key', {
      configurable: true,
      enumerable: true,
      value: key,
    });
  }
  return prevEventHook ? prevEventHook(event) : event;
};
```

src/sidebar/components/annotation-quote.js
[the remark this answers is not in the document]
robertknight: > IE11 is not the easiest to debug or even understand but when adding the dir= attribute to an <input>, for example, it no longer renders in a predictable way. 

I found https://stackoverflow.com/questions/38638718 this morning and was able to reproduce. Attempting to do `someInputElement.dir = 'auto'` causes an exception in IE. You can set it to "ltr", "rtl" or an empty string though.

I'm guessing what you saw was the result of this exception in the middle of an update causing all further UI processing to stop and leaving the UI in an arbitrarily broken state, which admittedly is super confusing if you haven't seen that there is an error in the browser console.

On this basis though, you're right that we need to not set `dir` to `auto` in IE. It looks to me like setting it to an empty string should behave the same way as not setting it. In that case you could do something like:

```js
// `dirAuto` is a constant that is defined as the string "auto" in modern browsers
// and the empty string in IE.
import { dirAuto } from '../util/dom'

function SomeWidget() {
  return <input dir={dirAuto} />
}
```

robertknight:
@LMS007 - Unrelated to the content of this PR: When pushing fixup commits to a branch that need to be squashed before merge, put "WIP" or "fixup" in the commit title and the "WIP bot" check will set a commit status to warn devs before merging.

src/sidebar/components/markdown-editor.js
LMS007: awe, that makes more sense, This will be a lot better than trying to maintain special events everywhere

LMS007:
> WIP

good to know, thanks Rob

src/sidebar/components/annotation-quote.js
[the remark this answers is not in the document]
LMS007: I went with Method C, mostly because I was attempted the renderer-option with the key remapping (below) and we can limit it to only taking effect for IE11 so it won't harm performance for anything else.

src/sidebar/components/markdown-editor.js
LMS007: Turns out that `defineProperty ` this will not work in IE11 for the event object. We should discuss more options tomorrow.

src/shared/test/user-agent-test.js
```diff
@@ -0,0 +1,37 @@
+import { isIE11, isMacOS } from '../user-agent';
+
+describe('sidebar/util/user-agent', () => {
```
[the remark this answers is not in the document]
LMS007: oops!

src/sidebar/components/tag-editor.js
```diff
@@ -34,6 +36,9 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
     setSuggestionsListOpen(false);
   });
 
+  // Convenient boolean  for ie detection.
```
robertknight, line 39: I don't think this comment is needed as the purpose of the code seems pretty obvious.

src/sidebar/index.js
```diff
@@ -56,6 +57,9 @@ const isSidebar = !(
   window.location.pathname.startsWith('/a/')
 );
 
+// Install prect renderer options for IE11 quirks
```
robertknight, line 60:
```suggestion
// Install Preact renderer options to work around IE11 quirks
```

src/shared/browser-compatibility-utils.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * IE11 uses alternate key names for its keyboard events.
+ * If any abnormal keys are used (via IE11), this method
+ * returns the normalized name so our UI components
+ * don't require a special case.
+ *
+ * @param {string} key - The keyboard event `key` name
+ * @return {string} - Normalized `key` name
+ */
+export function getNormalKey(key) {
```
robertknight, line 12: Can I suggest `normalizeKeyName` as a name for this? I think it may not be obvious to a reader what a "normal" key is (as opposed to _normalizing_ the key name).

src/shared/browser-compatibility-utils.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Normalize a keyboard event key name.
+ *
+ * IE11 uses alternate key names for its keyboard events.
+ * If any abnormal keys are used (via IE11), this method
```
robertknight, line 5: Strictly speaking is is not just IE 11 that uses non-standard names for keys. There is an [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) has details. Firefox <= 36 is the main browser that uses non-standard names for common keys, although I would imagine that the usage of Firefox versions that old is tiny to non-existent.

LMS007: I'll rework that doc so its correct

PR #2077, opened 2020-04-23
Move margin between `ThreadCard`s to `ThreadList`

robertknight:
Move the bottom margin on thread cards from the root `<div>` rendered by
`ThreadCard` to the wrapper div that contains it rendered by
`ThreadList`.

Aside from making conceptual sense that the list component controls the
spacing of list items, this also fixes an issue where the thread list
would sometimes jump when scrolling. This happened because the height of
individual threads was computed by calling `getElementHeightWithMargins`
on the wrapper div, which only accounted for margins on the wrapper div,
not margins on descendants. Note that due to margin collapsing, the descendant
margins did not contribute to the `getBoundingClientRect()` of the
wrapper div.

### DIFF
--- a/src/sidebar/components/thread-list.js
+++ b/src/sidebar/components/thread-list.js
@@ -25,8 +25,13 @@ function getScrollContainer() {
 }
 
 /**
- * Render a list of threads, but only render those that are in or near the
- * current browser viewport.
+ * Render a list of threads.
+ *
+ * The thread list is "virtualized", meaning that only threads in or near the
+ * viewport are rendered. This is critical for performance and memory usage as
+ * annotations (and replies) are complex interactive components whose
+ * user-defined content may include rich media such as images, audio clips,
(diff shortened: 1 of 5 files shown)


src/sidebar/components/thread-list.js
```diff
@@ -25,8 +25,13 @@ function getScrollContainer() {
...
 /**
- * Render a list of threads, but only render those that are in or near the
- * current browser viewport.
+ * Render a list of threads.
+ *
+ * The thread list is "virtualized", meaning that only threads in or near the
+ * viewport are rendered. This is critical for performance and memory usage as
+ * annotations (and replies) are complex interactive components whose
+ * user-defined content may include rich media such as images, audio clips,
+ * embedded YouTube videos, rendered math and more.
```
robertknight, line 34: I took the liberty of adding this note while I was in this code because several developers have asked in the past why we bother with this.

PR #2078, opened 2020-04-23
Convert stream content to Preact

robertknight:
This is a fairly direct conversion of the `<stream-content>` component, for the `/stream` route,
from AngularJS to Preact. This is a countepart to https://github.com/hypothesis/client/pull/2074.

There is one minor functional improvement which is to display an error notification
if fetching annotations from the API fails.


robertknight:
> I think it would be optimal to pull out some of the query logic into a service, but for now, is fine.

I agree. I think it will make sense to centralize the logic for fetching annotations for the stream/single annotation/sidebar in the existing `loadAnnotationsService`.

Issue #2083, opened 2020-04-24
openSidebar is not working from via


LMS007:
### Steps to reproduce
Open up a pdf (or html) in via as follows

https://via.hypothes.is/https://arxiv.org/pdf/1908.04683v3.pdf?via.open_sidebar=false

or (new way)

https://via.hypothes.is/https://arxiv.org/pdf/1908.04683v3.pdf?via.client.openSidebar=false

### Expected behaviour
The sidebar should be closed

### Actual behaviour
Sidebar remains open.

### Additional details
It looks like the client is casting that value from the hypothesisConfig to a boolean in host-config, but the annotator gets its value in a different way than the sidebar itself, so it never sees the converted value from `host-config.js`. We may need to implement something similar what was done in host-config on the annotator itself.

jon-betts:
Probably best to accept a few things as falsy:

* "False"
* "false"
* "0"
* ""

Issue #2058, opened 2020-04-21
Confusing experience for assistive tech & keyboard users when sidebar is closed


klemay:
Part of WCAG 2.0 Criterion 4.1.2 Name, Role, Value 

When the Hypothesis panel is hidden visually (collapsed), it is still perceptible to assistive technologies, and accessible with a keyboard. This could be confusing to: 

- Non-visual users when the content of the panel is communicated as collapsed. 
- Low vision users who rely on a screen reader and so will hear screen reader output while not seeing the panel on screen. 
- Keyboard users who will lose their place as the content they are interacting with is off screen. 

Ensure that any content that is hidden visually is not in the focus order and is hidden from assistive technologies.

PR #2095, opened 2020-04-27
Add visibility hidden to iframe when closed

LMS007:
- Add visibility hidden to the sidebar iframe when its closed (and after it finishes animation). This prevents all interaction from the components of the sidebar.

- Add an optional coerce method to the hostPageSetting so that any incoming values can be coerced if needed. This may be required for values from from via that may only be a string type.

- `openSidebar` config value gets coerced to a boolean in the annotator settings (just like it does in the host-config).

---------

Test in 2 ways:

1. change dev-server.js to have openSidebar false (or true) & restart the client server.
`is-closed` should be added when its `false` to the `<iframe>`  but not added otherwise.

2. Open and close the sidebar and the `is-closed` class should also be added or removed corresponding to the state of the sidebar after show/hide events are fired.

Fixes https://github.com/hypothesis/client/issues/2058
Fixes https://github.com/hypothesis/client/issues/2083

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -1,4 +1,5 @@
 import settingsFrom from './settings';
+import { toBoolean } from '../../shared/type-coercions';
 
 /**
  * Reads the Hypothesis configuration from the environment.
@@ -28,6 +29,7 @@ export default function configFrom(window_) {
     onLayoutChange: settings.hostPageSetting('onLayoutChange'),
     openSidebar: settings.hostPageSetting('openSidebar', {
       allowInBrowserExt: true,
+      coerce: toBoolean,
     }),
     query: settings.query,
     requestConfigFromFrame: settings.hostPageSetting('requestConfigFromFrame'),
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -29,6 +29,7 @@ export default function configFrom(window_) {
     onLayoutChange: settings.hostPageSetting('onLayoutChange'),
     openSidebar: settings.hostPageSetting('openSidebar', {
       allowInBrowserExt: true,
+      // Coerce value to a boolean because it may come from via as a string
       coerce: toBoolean,
     }),
     query: settings.query,
--- a/src/annotator/host.coffee
+++ b/src/annotator/host.coffee
@@ -40,6 +40,9 @@ module.exports = class Host extends Guest
     .attr('title', 'Hypothesis annotation viewer')
     .addClass('h-sidebar-iframe')
 
+    if config.openSidebar == false
+      app.addClass('is-closed')
+
     externalContainer = null
 
     if config.externalContainerSelector
--- a/src/annotator/host.coffee
+++ b/src/annotator/host.coffee
@@ -40,9 +40,6 @@ module.exports = class Host extends Guest
     .attr('title', 'Hypothesis annotation viewer')
     .addClass('h-sidebar-iframe')
 
-    if config.openSidebar == false
-      app.addClass('is-closed')
-
     externalContainer = null
 
     if config.externalContainerSelector
--- a/src/styles/annotator/annotator.scss
+++ b/src/styles/annotator/annotator.scss
@@ -56,6 +56,10 @@ var.$base-font-size: 14px;
     position: relative;
   }
 
+  .is-closed {
+    visibility: hidden;
+  }
+
   @include icons.icons {
     font-family: 'h';
   }
--- a/src/styles/annotator/annotator.scss
+++ b/src/styles/annotator/annotator.scss
@@ -10,6 +10,7 @@
 @use './highlights';
 
 var.$base-font-size: 14px;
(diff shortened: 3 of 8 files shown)


src/annotator/config/index.js
```diff
@@ -28,6 +29,7 @@ export default function configFrom(window_) {
     onLayoutChange: settings.hostPageSetting('onLayoutChange'),
     openSidebar: settings.hostPageSetting('openSidebar', {
```
LMS007, line 30: There may be other values we want to force as a boolean, but only this one was needed for this ticket... and mostly because we only assume things coming from via are a string.

src/annotator/config/index.js
```diff
@@ -28,6 +29,7 @@ export default function configFrom(window_) {
...
       allowInBrowserExt: true,
+      coerce: toBoolean,
```
LMS007, line 32: There may be other values we want to force as a boolean, but only this one was needed for this ticket... and only because we only assume things coming from via are a string. (not all these settings from from via)

src/styles/annotator/annotator.scss
```diff
@@ -56,6 +56,10 @@ var.$base-font-size: 14px;
     position: relative;
   }
 
+  .is-closed {
```
robertknight, line 59: Hmm, rather than adding a new CSS class and logic to add/remove it, perhaps it would be easier in this scenario to leverage the existing `.annotator-collapsed` class that is set on the parent `.annotator-frame` with a descendant selector?

src/styles/annotator/annotator.scss
```diff
@@ -56,6 +56,10 @@ var.$base-font-size: 14px;
...
+    visibility: hidden;
```
robertknight, line 60: There is an issue at the moment that the visual transition to hide the sidebar is animated, but the change in iframe visibility happens immediately. This creates a flicker when closing the sidebar where clicking the "Close" button removes all the sidebar content immediately but the container gradually slides out of view.

The default sidebar open/close transition is only 0.15s so it may be hard to see what is going on, if you slow the transition down, the effect is more obvious.

src/annotator/host.coffee
```diff
@@ -40,6 +40,9 @@ module.exports = class Host extends Guest
     .attr('title', 'Hypothesis annotation viewer')
     .addClass('h-sidebar-iframe')
 
+    if config.openSidebar == false
+      app.addClass('is-closed')
```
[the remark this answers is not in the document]
robertknight: Update - after thinking a bit more on this I have another idea. I suggest that perhaps we could leverage the existing `annotator-collapsed` class which is used on the iframe's parent `annotator-frame` element.

src/styles/annotator/annotator.scss
robertknight: After some digging I learned that you can specify a transition on `visibility`, but since it is a boolean it doesn't animate gradually between the two states. Instead it just [defers the property change](https://stackoverflow.com/questions/27900053/css-transition-with-visibility-not-working) until the end of the transition, which is what we want here.

Interestingly, with this diff:

```diff
diff --git a/src/styles/annotator/annotator.scss b/src/styles/annotator/annotator.scss
index 4cca9199a..a9ddb01f8 100644
--- a/src/styles/annotator/annotator.scss
+++ b/src/styles/annotator/annotator.scss
@@ -11,6 +11,8 @@
 
 var.$base-font-size: 14px;
 
+$sidebar-collapse-transition-time: 2s;
+
 // Sidebar
 .annotator-frame {
   // a CSS reset which attempts to prevent the host page's styles
@@ -54,10 +56,12 @@ var.$base-font-size: 14px;
     width: 100%;
     z-index: 3;
     position: relative;
-  }
 
-  .is-closed {
-    visibility: hidden;
+    transition: visibility $sidebar-collapse-transition-time;
+
+    &.is-closed {
+      visibility: hidden;
+    }
   }
 
   @include icons.icons {
@@ -193,7 +197,7 @@ var.$base-font-size: 14px;
 
 @media screen and (min-width: 37.5em) {
   .annotator-frame {
-    transition: margin-left 0.15s cubic-bezier(0.55, 0, 0.2, 0.8);
+    transition: margin-left $sidebar-collapse-transition-time cubic-bezier(0.55, 0, 0.2, 0.8);
     width: 428px;
     margin-left: -428px;
   }
```

When closing the sidebar, the iframe is not hidden until the end of the transition, which is what we want. When opening the sidebar it is shown immediately however, which is also what we want, but not what I thought would happen (I thought it would wait until the end of the transition before un-hiding the iframe).

LMS007: I'm surprised I missed this, I think I got caught up in tests, but good catch never the less.  

I see, so you're only applying the transition when its collapsing and that will delay the effect until it reaches the end of the delay. Then when it opens, no delay, no transition.

LMS007:
Added ticket
https://github.com/hypothesis/client/issues/2098

PR #2107, opened 2020-05-03
Use Identity API for login with Firefox

diegodlh:
This uses [Identity API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity) for login with the Firefox extension, as discussed [here](https://github.com/hypothesis/browser-extension/issues/316#issuecomment-602286222). It makes use of Identity API support proposed in a separate [pull request](https://github.com/hypothesis/browser-extension/pull/341) in the browser-extension repo.

I tried to have the new method as parallalel as possible to the old (web message) one in src/sidebar/util/oauth-client.js.

There is a bug in Chrome: whereas in Firefox the 'auth' cookie set at login is shared between the auth popup window launched by launchWebAuthFlow() and regular windows, this is not the case in Chrome. As a result:
1) If I log in from the sidebar, I am not logged in automatically in the website.
2) If I log out from the website, I am not logged out from the sidebar. Because [the sidebar logout doesn't actually log out](https://github.com/hypothesis/client/issues/2108) (i.e., unset the 'auth' cookie) but rather it just revokes the auth token, there is no way to log out from the sidebar.

For this reason, and because the Chrome extension already works with the old login method, I decided to enable this Identity-API method for appType === 'firefox-extension' only.

The Lint test fails with 'chrome is not defined'. Can you help me fix this?


diegodlh:
Just so it's easier to understand what I did, the code inside the if block (i.e., appType !== 'firefox-extension') is the same as it was before, only it's indented now. The new code is inside the else block (i.e., appType === 'firefox-extension').

diegodlh:
Just for the record, this has been implemented in an [unofficial extension](https://github.com/diegodlh/browser-extension/releases/latest). See discussion [here](https://github.com/hypothesis/browser-extension/issues/310#issuecomment-637715653).

PR #2129, opened 2020-05-06
Fix "New annotation"/"New page note" label

robertknight:
The button label should be "New annotation" if text is selected or "New
page note otherwise", but it was the wrong way round.

Fixes a regression introduced in #2102


robertknight:
I'm going to go ahead and merge this as the code change is trivial and there is currently a regression master.

PR #2135, opened 2020-05-07
Focus first item in menu when opened by any input method

robertknight:
Previously the `Menu` component would focus the first item in the list if it was
opened by pressing Enter or Space but not if it was clicked. This meant that it
was not possible to mix input modalities (open menu with mouse, navigate using
keyboard) easily when using the menu.  More importantly though, if the menu was opened
using a screen reader shortcut (eg. VoiceOver Key + Space on macOS) per the
screen reader's verbal instructions, the focus didn't change. My understanding
is that SR shortcuts generally simulate mouse rather than keyboard input.

In my testing with various desktop and mobile browsers I didn't notice any problems introduced by this change, but can you think of anything I have overlooked @LMS007?

### DIFF
--- a/src/sidebar/components/menu.js
+++ b/src/sidebar/components/menu.js
@@ -55,7 +55,6 @@ export default function Menu({
   title,
 }) {
   const [isOpen, setOpen] = useState(defaultOpen);
-  const [openedByKeyboard, setOpenedByKeyboard] = useState(false);
 
   // Notify parent when menu is opened or closed.
   const wasOpen = useRef(isOpen);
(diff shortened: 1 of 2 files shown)


src/sidebar/components/menu.js
```diff
@@ -179,7 +167,7 @@ export default function Menu({
             onClick={closeMenu}
             onKeyDown={handleMenuKeyDown}
           >
-            <MenuKeyboardNavigation visible={openedByKeyboard}>
+            <MenuKeyboardNavigation visible={true}>
```
robertknight, line 170: nb. Although `visible` is always `true` in this component, `MenuItem` still uses it and sets it depending on whether the submenu is visible or not. We still want that behavior.

robertknight:
> Thanks Rob, but was there a ticket for this?

No, not that I'm aware of.

Issue #2053, opened 2020-04-20
Remove unused SVG icons


LMS007:
It looks like there are some unused icons in our code base. We should examine all icons and remove any svg objects that are no longer used. It also may be worth inventing a new name convention for svg files so they can be searched for more easily in the future so its very clear if they are not used. A search for something like "filter" is not terrible obvious.

**Possibly deprecated** 

- [ ] account.svg
- [ ] filter.svg

robertknight:
I updated the description to remove a discussion about unifying icons in the LMS app and client. This has been done by making both use the `SvgIcon` component from the frontend-shared package.

robertknight:
These icons were removed in 84cb056fd054ac37f67324e92def4652767f311a.

PR #2140, opened 2020-05-08
SvgIcon changes

LMS007:
- Add ability for SvgIcon to use a trusted icon directly. This serves 2 purposes along with being aligned with LMS.

- 1. It is more readable to have the icon path used in place rather than a level of indirection from a registered name in another file.
- 2. Keeping track of abandoned icons should be easier because searching for a path and ".svg" suffix is a lot easier than a name that could be very common such as "groups" -- which is a real icon.

See https://github.com/hypothesis/lms/blob/master/lms/static/scripts/frontend_apps/components/SvgIcon.js

To the best of my knowledge, `require` will cache the file so even through we may call it more than once on the same SVG, there should not be a significate performance hit for doing that relative to a single registered icon file.

If we like this, then going forward, we can convert to using trustMarkup and remove `icon.js` when finished. 

**current**

icons.js
```
export default {
...
spinner: require('../images/icons/spinner.svg'),
}
```
component
```
<SvgIcon
    name="spinner"
    />
```

**changed**

component
```
import { trustMarkup } from '../utils/trusted';

<SvgIcon
      src={trustMarkup(require('../../../images/spinner.svg'))}
    />
```

fixes https://github.com/hypothesis/client/issues/2053

### DIFF
--- a/src/shared/components/svg-icon.js
+++ b/src/shared/components/svg-icon.js
@@ -20,6 +20,10 @@ let iconRegistry = {};
  * Component that renders icons using inline `<svg>` elements.
  * This enables their appearance to be customized via CSS.
  *
+ * Supply either a registerer icon `name` or a trusted SVG object
+ * to the `src` prop.
+ * a
+ *
  * This matches the way we do icons on the website, see
  * https://github.com/hypothesis/h/pull/3675
  */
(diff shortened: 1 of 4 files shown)


LMS007:
closed by mistake, need to re-open

Issue #2142, opened 2020-05-11
Don't show groups menu arrows in LMS app


klemay:
We've removed the "Leave group" option for LMS users, but the groups dropdown still shows the arrows that allow users to expand/collapse the options for each group:

![Peek 2020-05-11 15-36](https://user-images.githubusercontent.com/12087705/81575533-2639cb80-9375-11ea-8e95-4ce412710efb.gif)

We should remove these arrows to avoid confusing users.

PR #2155, opened 2020-05-11
Hide menu item toggle button if there are no items

LMS007:
Pass undefined to MenuItem to hide the submenu and its toggle button.

--------

This has to be tested in canvas with `feature_flags.section_groups = true`
and  `allowLeavingGroups` to be set to false in the service object from in .js-config -- I don't think that is being passed in yet from LMS (I could be wrong). In any case, it may be hard to test in practice, but the `<MenuItem>` component was already set up to hide the toggle button, but we were not correctly passing `undefined` to the `isSubmenuVisible` prop. This fixes that! 

Fixes https://github.com/hypothesis/client/issues/2142

### DIFF
--- a/src/sidebar/components/group-list-item.js
+++ b/src/sidebar/components/group-list-item.js
@@ -83,7 +83,7 @@ function GroupListItem({
       isDisabled={!isSelectable}
       isExpanded={hasActionMenu ? isExpanded : false}
       isSelected={isSelected}
-      isSubmenuVisible={isExpanded}
+      isSubmenuVisible={hasActionMenu ? isExpanded : undefined}
       label={group.name}
       onClick={isSelectable ? focusGroup : toggleSubmenu}
       onToggleSubmenu={toggleSubmenu}
--- a/src/sidebar/components/test/group-list-item-test.js
+++ b/src/sidebar/components/test/group-list-item-test.js
@@ -169,11 +169,13 @@ describe('GroupListItem', () => {
 
   it('expands submenu if `isExpanded` is `true`', () => {
     const wrapper = createGroupListItem(fakeGroup, { isExpanded: true });
+    assert.isTrue(wrapper.find('MenuItem').prop('isSubmenuVisible'));
     assert.isTrue(wrapper.find('MenuItem').first().prop('isExpanded'));
   });
 
   it('collapses submenu if `isExpanded` is `false`', () => {
     const wrapper = createGroupListItem(fakeGroup, { isExpanded: false });
+    assert.isFalse(wrapper.find('MenuItem').prop('isSubmenuVisible'));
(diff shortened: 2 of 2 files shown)


src/sidebar/components/test/group-list-item-test.js
```diff
@@ -169,11 +169,13 @@ describe('GroupListItem', () => {
 
   it('expands submenu if `isExpanded` is `true`', () => {
     const wrapper = createGroupListItem(fakeGroup, { isExpanded: true });
+    assert.isTrue(wrapper.find('MenuItem').prop('isSubmenuVisible'));
```
LMS007, line 172: There were no tests for this value as a boolean, so I added these

src/sidebar/components/group-list-item.js
```diff
@@ -25,6 +25,7 @@ function GroupListItem({
   const activityUrl = group.links.html;
   const hasActionMenu = activityUrl || group.canLeave;
   const isSelectable = !group.scopes.enforced || group.isScopedToUri;
+  const isSubmenuEmpty = !activityUrl && !group.canLeave && isSelectable;
```
robertknight, line 28: There is already a `hasActionMenu` variable here which can be used. Note that whether or not a group is selectable doesn't affect whether it has menu items. A group might be unselectable because it is restricted to certain URLs, but the menu item to go to its activity page is still available. The tests on this PR pass with the following diff:

```diff
diff --git a/src/sidebar/components/group-list-item.js b/src/sidebar/components/group-list-item.js
index 63c32d621..4744a7644 100644
--- a/src/sidebar/components/group-list-item.js
+++ b/src/sidebar/components/group-list-item.js
@@ -25,7 +25,6 @@ function GroupListItem({
   const activityUrl = group.links.html;
   const hasActionMenu = activityUrl || group.canLeave;
   const isSelectable = !group.scopes.enforced || group.isScopedToUri;
-  const isSubmenuEmpty = !activityUrl && !group.canLeave && isSelectable;
 
   const focusedGroupId = useStore(store => store.focusedGroupId());
   const isSelected = group.id === focusedGroupId;
@@ -84,7 +83,7 @@ function GroupListItem({
       isDisabled={!isSelectable}
       isExpanded={hasActionMenu ? isExpanded : false}
       isSelected={isSelected}
-      isSubmenuVisible={isSubmenuEmpty ? undefined : isExpanded}
+      isSubmenuVisible={hasActionMenu ? isExpanded : undefined}
       label={group.name}
       onClick={isSelectable ? focusGroup : toggleSubmenu}
       onToggleSubmenu={toggleSubmenu}
```

Re-reading this code though has be asking - why do we have two separate props for submenu visibility? It looks to me like the `isExpanded` prop is only used to add the `is-expanded` class if the menu item has an open submenu. That logic could be changed to add the `is-expanded` class if `isSubmenuVisible === true` instead.

Issue #2035, opened 2020-04-14
Deleting an annotation causes the page to scroll to top


kevinjalbert:
### Steps to reproduce

1. Create an annotation
2. Scroll down the page
3. Delete annotation

### Expected behaviour

Page's scroll position shouldn't change.

### Actual behaviour

Page's scrolls to the top of the page.

### Browser/system information

- macOS 10.15.4
- Brave (v 1.5.123 && v 1.7.92)

### Additional details

[GIF showing issue](https://monosnap.com/direct/rNANt1vJEat670oEF9diIwwEapOEFD)

This is pretty frustrating (especially on mobile), if I'm trying to correct a highlight (expanding/contracting the highlight itself).

I saw https://github.com/hypothesis/client/issues/566, although that seems to be before the delete of the annotation actually occurs.

mkdir-washington-edu:
Also heard about this here: https://hypothesis.zendesk.com/agent/tickets/10064

I tested on MacOS 10.14, Chrome 81, FF 76. This occurs with the extension, the Bookmarklet, and in Via. It occurs with a single annotation on the page and multiple annotations on the page.

PR #2225, opened 2020-06-01
Prevent scrolling to annotations from links or buttons

LMS007:
If an event is triggered from a link or button of decent of a link or button, then prevent scrollToAnnotation from firing. It is assumed in these cases that a more specific action was intended and the page shall not scroll to the underlying annotation.

------

To see the effects of this bug, click on the annotation column and then click any button or hyperlink. In almost all cases, you'll see the page scroll to that annotation if was was not already scrolled to that position.  

I initially I attempted to fix all the buttons that were affected, but then I realized the scope of this issue was pretty wide and touched every component inside thread-card . Everything that is clicked on in the thread-card basically allows `scrollToAnnotation` to be triggered unless any specific click handler calls stopPropagation. --- few do. Rather than trying to shore up all the leafs, this approach just prevents the scrolling at the top level  if any `button` or `a` was clicked or any sub-element of a `button` or `a` is clicked.  

fixes https://github.com/hypothesis/client/issues/2035


src/sidebar/components/thread-card.js
```diff
@@ -33,10 +33,24 @@ function ThreadCard({ frameSync, settings = {}, thread }) {
     [frameSync]
   );
 
+  /**
+   * Is the target's event an <a> or <button> element, or does it have
+   * either as an ancestor?
+   */
+  const isFromButtonOrLink = target => {
+    return !!target.closest('button') || !!target.closest('a');
```
robertknight, line 41: We'll need to use the [`closest` polyfill](https://github.com/hypothesis/client/blob/0d15ec5dfcdec56b6394ebf1e1daa8bf9572edbf/src/annotator/util/dom-element.js#L11) for IE 11 support since we won't be dropping support for that browser until July 1st. You'll need to move that file to `src/shared/` since files in `src/sidebar` are not supposed to depend on modules in `src/annotator` and vice-versa.

src/sidebar/components/test/thread-card-test.js
```diff
@@ -103,12 +103,21 @@ describe('ThreadCard', () => {
       assert.calledWith(fakeFrameSync.focusAnnotations, sinon.match(['myTag']));
     });
 
-    it('unfocuses the annotation thread when mouse exits', () => {
```
robertknight, line 106: Was this test removed by mistake?

LMS007: yes, apparently it was. thanks

src/sidebar/components/thread-card.js
LMS007: okay, I moved this to /shared for the time being so it follows our standard between sidebar and annotator

PR #2250, opened 2020-06-22
Remove Node v14 workaround for Karma

robertknight:
Run tests under Puppeteer again as the issue with Node v14 has been
resolved.


robertknight:
I've tested this with Node v14.3.0 locally.

Issue #2249, opened 2020-06-22
CI builds sometimes only run a subset of tests


robertknight:
We've seen an issue in CI, and Lyza and myself have also experienced it locally too, where builds occasionally "succeed" but only run a fraction (typically 300-500 it seems) of the tests.

When this happens, codecov will typically report a sudden unexpected drop in code coverage.

Original report: https://github.com/hypothesis/client/pull/2244#issuecomment-646852343

robertknight:
I noticed this in the CI logs for a recent master build. Probably unrelated, but I'll note it anyway:

```
[0622/134713.594272:INFO:CONSOLE(342905)] "Uncaught (in promise) TypeError: Cannot read property 'references' of undefined", source: http://localhost:9876/absolute/tmp/92430908e258191cffbe541e795258ee.browserify.js?352a30a79692f7b6bedaf8759b1b434258a231ce (342905)
```

robertknight:
Example of a build of the `master` branch where this happened: https://jenkins.hypothes.is/job/client/job/master/1986/console

robertknight:
We updated the version of Alpine that is used on Jenkins CI and I don't think we've seen the problem since there, although Lyza did report that it happened on her system.

https://github.com/hypothesis/client/pull/2263 has modified the CI build so that it logs which tests have executed, which should help debugging if the issue happens again.

Reproducing the issue on my system may be more difficult now because I updated my laptop and the new one is faster.

One thing I probably will try before closing this as "maybe fixed - wait until it happens again so I can continue debugging" is to write a script that just runs the test suite in a loop many times and checking whether the issue occurred. It may also be worth checking to see if it happens in an environment where the system is resource constrained (eg. in a Docker container with reduced CPU limits).

robertknight:
I was able to reproduce this issue locally with the following script:

```py
import re
import subprocess
import sys

max_iterations = 100
for i in range(0, max_iterations):
    print(f"Running test iteration {i+1} of {max_iterations}")

    proc = subprocess.run(
        "node_modules/.bin/gulp test",
        shell=True,
        stdout=subprocess.PIPE,
        check=True,
    )
    stdout = proc.stdout.decode()

    tests_run_match = re.search("TOTAL: ([0-9]+) SUCCESS", stdout)

    if not tests_run_match:
        print(f"Tests failed on run {i}")
        print(stdout)
        sys.exit(1)

    expected_tests = 2439
    tests_run = int(tests_run_match.group(1))
[...]

PR #2263, opened 2020-06-23
Log which tests have executed during CI runs

robertknight:
During CI runs, but not local development, log which individual tests
have executed. This may help with debugging https://github.com/hypothesis/client/issues/2249.


robertknight:
I'm going to merge this and then watch out for the problem described in #2249 occurring in future PRs.

PR #2268, opened 2020-06-24
Begin to add JSDoc types for client configuration objects

robertknight:
The client's "sidebar" and "annotator" applications query configuration
from various sources (the host page, the H backend / browser extension
build, added dynamically by the app).

This commit begins the process of adding a centralized place to document
the types of these configuration objects and making typechecked modules
in `src/sidebar/util` use them.

~~Something that needs to be sorted before merging this is to decide on usage of "config" vs
"settings" as we're using both quite freely at the moment, and I realize as I write this that I've created
a file called `config.js` with types that mostly have a `Settings` suffix (!)~~ - I went with `Config` in the type names as it is more common for us to use "configuration" or "config" rather than "settings" overall for application configuration.

### DIFF
--- a/src/sidebar/util/groups.js
+++ b/src/sidebar/util/groups.js
@@ -1,4 +1,7 @@
-/** @typedef {import('../../types/api').Group} Group */
+/**
+ * @typedef {import('../../types/config').HostConfig} HostConfig
+ * @typedef {import('../../types/api').Group} Group
+ */
 
 import escapeStringRegexp from 'escape-string-regexp';
 import serviceConfig from '../service-config';
@@ -9,14 +12,15 @@ import serviceConfig from '../service-config';
  * explicitly disallowed in the service configuration of the
  * `settings` object.
  *
- * @param {object} settings
+ * @param {HostConfig} settings
  * @return {boolean}
  */
 function allowLeavingGroups(settings) {
-  const config = serviceConfig(settings) || {};
-  return typeof config.allowLeavingGroups === 'boolean'
-    ? config.allowLeavingGroups
-    : true;
+  const config = serviceConfig(settings);
+  if (!config) {
+    return true;
+  }
+  return !!config.allowLeavingGroups;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


src/types/config.js
```diff
@@ -0,0 +1,95 @@
...
+ * @prop {string} [usernameUrl]
+ *   URL template for username links
+ */
+
+/**
+ * The `settings` object used in the sidebar app that is a result of merging
+ * (filtered and validated) configuration from the host page with configuration
+ * from h / the browser extension.
+ *
+ * @typedef {HostSettings & SidebarSettings} MergedSettings
```
robertknight, line 91: This is an example of creating an "intersection" type which combines all the properties from two existing types.

src/sidebar/util/session.js
```diff
@@ -28,14 +31,15 @@ export function shouldShowSidebarTutorial(sessionState) {
...
  */
 export function shouldAutoDisplayTutorial(isSidebar, profile, settings) {
   const shouldShowBasedOnProfile =
     typeof profile.preferences === 'object' &&
     !!profile.preferences.show_sidebar_tutorial;
-  const service = serviceConfig(settings) || {};
+
+  const service = serviceConfig(settings) || { onHelpRequestProvided: false };
   return (
     isSidebar && !service.onHelpRequestProvided && shouldShowBasedOnProfile
```
robertknight, line 44: Without the change to the object literal here, TS complains that `service` might not have an `onHelpRequestProvided` property as it may have been initialized to `{}`, which is inferred to be an object type with no properties.

src/sidebar/util/groups.js
```diff
@@ -9,14 +12,15 @@ import serviceConfig from '../service-config';
...
 function allowLeavingGroups(settings) {
-  const config = serviceConfig(settings) || {};
-  return typeof config.allowLeavingGroups === 'boolean'
-    ? config.allowLeavingGroups
-    : true;
+  const config = serviceConfig(settings);
+  if (!config) {
+    return true;
+  }
+  return !!config.allowLeavingGroups;
```
robertknight, line 23: This refactor avoids TS complaining that `config` might not have an `allowLeavingGroups` property, if it was initialized to `{}` (which is inferred to be an object with no properties).

PR #2269, opened 2020-06-24
Enable typechecking for `src/shared/*.js`

robertknight:
Enable typechecking of annotated functions in src/shared/, excluding
subdirectories for the moment.

This found a few minor mistakes with existing JSDoc usage and a missing null check.

### DIFF
--- a/src/shared/discovery.js
+++ b/src/shared/discovery.js
@@ -2,7 +2,7 @@
  * Callback invoked when another frame is discovered in this window which runs
  * the Hypothesis sidebar or annotation layer code.
  *
- * @type {Function} DiscoveryCallback
+ * @callback DiscoveryCallback
  * @param {Window} source - The frame that was discovered.
  * @param {string} origin - The origin to use when posting messages to this frame.
  * @param {string} token - A random identifier used by this frame.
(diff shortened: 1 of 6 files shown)


src/shared/discovery.js
```diff
@@ -2,7 +2,7 @@
  * Callback invoked when another frame is discovered in this window which runs
  * the Hypothesis sidebar or annotation layer code.
  *
- * @type {Function} DiscoveryCallback
+ * @callback DiscoveryCallback
```
LMS007, line 5: could this be `@class` too? I'm just trying to understand if this is simply a better vs best or right vs wrong?

src/shared/discovery.js
```diff
@@ -115,7 +115,7 @@ export default class Discovery {
     // window and send messages to them.
     const queue = [this.target.top];
     while (queue.length > 0) {
-      const parent = queue.shift();
+      const parent = /** @type {Window} */ (queue.shift());
```
LMS007, line 118: is this casting?

src/shared/discovery.js
robertknight: `@type {Function}` isn't legal syntax for declaring a function/callback type with its parameters and return value. I think `@typedef {Function} DiscoveryCallback` would also have worked here, but `@callback` is shorter.

src/shared/discovery.js
robertknight: Yes, exactly. The general syntax is `/** @type {<Type>} */ (<JS expression>)`. The braces around the expression are important.

PR #2275, opened 2020-06-25
Typecheck src/sidebar/services

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/2273**~~

This PR enables initial typechecking of modules in src/sidebar/services and fixes up many issues/incompleteness with the existing JSDoc annotations.

Aside from finding some over-complex and error-prone code in [filter query parsing](https://github.com/hypothesis/client/pull/2273) this also turned up a potential issue in `src/sidebar/services/stream-filter.js` where we set a `case_sensitive` flag on the client for certain fields but nothing explicitly uses it on the backend. I haven't altered the code in this commit but have just flagged the issue. We also continue to set some fields which are still parsed on the backend but [comments in the source](https://github.com/hypothesis/h/blob/master/h/streamer/filter.py) indicate are ignored.

### DIFF
--- a/src/sidebar/services/groups.js
+++ b/src/sidebar/services/groups.js
@@ -4,6 +4,8 @@ import { combineGroups } from '../util/groups';
 import { awaitStateChange } from '../util/state';
 import { watch } from '../util/watch';
 
+/** @typedef {import('../../types/api').Group} Group */
+
 const DEFAULT_ORG_ID = '__default__';
 
 /**
@@ -11,8 +13,10 @@ const DEFAULT_ORG_ID = '__default__';
  */
 const DEFAULT_ORGANIZATION = {
   id: DEFAULT_ORG_ID,
+  name: '__DEFAULT__',
   logo:
     'data:image/svg+xml;utf8,' +
+    // @ts-ignore - TS doesn't know about .svg files.
     encodeURIComponent(require('../../images/icons/logo.svg')),
 };
 
@@ -93,7 +97,7 @@ export default function groups(
    * @param {boolean} isLoggedIn
    * @param {string|null} directLinkedAnnotationGroupId
    * @param {string|null} directLinkedGroupId
-   * @return {Group[]}
+   * @return {Promise<Group[]>}
    */
   async function filterGroups(
     groups,
@@ -107,6 +111,7 @@ export default function groups(
       if (
         directLinkedGroup &&
         !directLinkedGroup.isScopedToUri &&
+        directLinkedGroup.scopes &&
         directLinkedGroup.scopes.enforced
       ) {
         groups = groups.filter(g => g.id !== directLinkedGroupId);
@@ -365,6 +370,7 @@ export default function groups(
     addGroupsToStore(groups);
 
     if (error) {
+      // @ts-ignore - TS can't track the type of `error` here.
       toastMessenger.error(`Unable to fetch groups: ${error.message}`, {
         autoDismiss: false,
       });
--- a/src/sidebar/services/oauth-auth.js
+++ b/src/sidebar/services/oauth-auth.js
@@ -1,10 +1,12 @@
-import EventEmitter from 'tiny-emitter';
+import { TinyEmitter } from 'tiny-emitter';
 
 import serviceConfig from '../service-config';
 import OAuthClient from '../util/oauth-client';
 import { resolve } from '../util/url';
 
 /**
+ * @typedef {import('../util/oauth-client').TokenInfo} TokenInfo
+ *
  * @typedef RefreshOptions
  * @property {boolean} persist - True if access tokens should be persisted for
  *   use in future sessions.
@@ -32,7 +34,7 @@ export default function auth(
 ) {
   /**
    * Authorization code from auth popup window.
-   * @type {string}
+   * @type {string|null}
    */
   let authCode;
 
@@ -41,7 +43,7 @@ export default function auth(
    *
    * Resolves to `null` if the user is not logged in.
    *
-   * @type {Promise<TokenInfo|null>}
+   * @type {Promise<TokenInfo|null>|null}
    */
   let tokenInfoPromise;
 
@@ -127,7 +129,7 @@ export default function auth(
       });
   }
 
-  const emitter = new EventEmitter();
+  const emitter = new TinyEmitter();
 
   /**
    * Listen for updated access & refresh tokens saved by other instances of the
@@ -162,7 +164,7 @@ export default function auth(
   /**
    * Retrieve an access token for the API.
    *
-   * @return {Promise<string>} The API access token.
+   * @return {Promise<string|null>} The API access token or `null` if not logged in.
    */
   function tokenGetter() {
     if (!tokenInfoPromise) {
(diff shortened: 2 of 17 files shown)


src/sidebar/services/groups.js
```diff
@@ -4,15 +4,19 @@ import { combineGroups } from '../util/groups';
...
+/** @typedef {import('../../types/api').Group} Group */
+
 const DEFAULT_ORG_ID = '__default__';
 
 /**
  * FIXME: There is almost assuredly a better way to handle a fallback, default logo
  */
 const DEFAULT_ORGANIZATION = {
   id: DEFAULT_ORG_ID,
+  name: '__DEFAULT__',
```
robertknight, line 16: `name` is a mandatory parameter of organizations returned from the server, so I added an unused value here.

src/sidebar/services/groups.js
```diff
@@ -107,6 +111,7 @@ export default function groups(
       if (
         directLinkedGroup &&
         !directLinkedGroup.isScopedToUri &&
+        directLinkedGroup.scopes &&
```
robertknight, line 114: It shouldn't technically be possible for `isScopedToUri` to be `true` and `scopes` to be missing, but this check makes it obvious to TS and the reader that we are accounting for this.

src/sidebar/services/groups.js
```diff
@@ -365,6 +370,7 @@ export default function groups(
     addGroupsToStore(groups);
 
     if (error) {
+      // @ts-ignore - TS can't track the type of `error` here.
```
robertknight, line 373: I didn't get to the bottom of this, but I didn't feel it was too critical.

src/sidebar/services/oauth-auth.js
```diff
@@ -1,10 +1,12 @@
-import EventEmitter from 'tiny-emitter';
+import { TinyEmitter } from 'tiny-emitter';
```
robertknight, line 1: As in another recent PR, `TinyEmitter` is exactly the same as `EventEmitter`, but only `TinyEmitter` has type information available in the `tiny-emitter` package (without us having to write our own type info for that package).

src/sidebar/services/oauth-auth.js
```diff
@@ -205,50 +207,52 @@ export default function auth(
 
     const origToken = tokenInfoPromise;
 
-    return tokenInfoPromise.then(token => {
-      if (!token) {
-        // No token available. User will need to log in.
-        return null;
-      }
+    return /** @type {Promise<TokenInfo|null>} */ (tokenInfoPromise).then(
```
robertknight, line 210: There are no changes to the body of the `then` callback. It was just necessary in this context to give TS a bit of help to understand the type of `tokenInfoPromise` in this slightly complex method. It is also useful for the human reader IMO.

src/sidebar/services/oauth-auth.js
```diff
@@ -282,7 +286,10 @@ export default function auth(
       tokenInfoPromise,
       oauthClient(),
     ]);
-    await client.revokeToken(token.accessToken);
+
+    if (token) {
+      await client.revokeToken(token.accessToken);
+    }
```
robertknight, line 292: I didn't find a code path where `token` was definitely going to be `null`, but it also wasn't obvious that it _couldn't_ be null and so TS complained about that.

src/sidebar/services/root-thread.js
```diff
@@ -85,15 +92,7 @@ export default function RootThread(
...
-    return buildThread(state.annotations.annotations, {
-      forceVisible: truthyKeys(state.selection.forceVisible),
-      expanded: store.expandedThreads(),
-      highlighted: state.selection.highlighted,
-      selected: truthyKeys(store.getSelectedAnnotationMap() || {}),
-      sortCompareFn: sortFn,
-      filterFn: filterFn,
-      threadFilterFn: threadFilterFn,
-    });
+    return buildThread(state.annotations.annotations, options);
```
robertknight, line 95: Previously `filterFn` and `threadFilterFn` could be undefined, which doesn't match the JSDoc for `options` which says that `filterFn` is an optional property that, if set, should be a function. This change restructures this code so we don't set these optional properties at all if there is no filter.

src/sidebar/services/stream-filter.js
```diff
@@ -7,16 +50,16 @@
  */
 export default class StreamFilter {
   constructor() {
-    this.resetFilter();
+    this.filter = defaultFilter();
```
robertknight, line 53: This change makes it more obvious to the reader and to TS that `this.filter` always gets initialised before it is used.

src/sidebar/services/service-url.js
```diff
@@ -20,6 +20,7 @@ import * as urlUtil from '../util/url';
  * always returns empty strings as the URLs. After the links object has been
  * received from the API this function starts returning the real URLs.
  *
+ * @callback ServiceUrlGetter
  * @param {string} linkName - The name of the link to expand
  * @param {object} params - The params with which to expand the link
  * @returns {string} The expanded absolute URL, or an empty string if the
```
LMS007, line 26: Should we bother to clean up these in another pass? return vs returns?

src/sidebar/services/stream-filter.js
```diff
@@ -1,3 +1,46 @@
...
+ * @prop {any} value
+ * @prop {boolean} case_sensitive - TODO: Backend doesn't use this at present,
+ *   but it seems important for certain fields (eg. ID).
+ */
+
+/**
+ * @typedef Filter
+ * @prop {string} match_policy - TODO: Remove this, the backend doesn't use it any more.
+ * @prop {FilterClause[]} clauses
+ * @prop {Object} actions - TODO: Remove this, the backend doesn't use it any more.
```
LMS007, line 21: This is a bit confusing to me. Maybe I just need to understand TS a bit more. But its not clear when we can get away with {Object} vs something custom ...

```
@typedef Actions
@prop {boolean} [create]
@prop {boolean} [update]
@prop {boolean} [delete]
```

src/sidebar/services/stream-filter.js
```diff
@@ -1,3 +1,46 @@
...
+ * @prop {boolean} [actions.create]
```
LMS007, line 22: I realize this is `todo remove`, but in general If these are props of actions, I would prefer (if possible) 2 spaces to help visually see they belong to actions rather than than FIlter. that would help clarify nested structures for me.

```
 * @prop {Object} actions - TODO: Remove this, the backend doesn't use it any more.
 *   @prop {boolean} [actions.create]
 *   @prop {boolean} [actions.update]
 *   @prop {boolean} [actions.delete]
```

src/sidebar/services/tags.js
```diff
@@ -21,7 +21,7 @@ export default function tags(localStorage) {
    * Return a list of tag suggestions matching `query`.
    *
    * @param {string} query
-   * @param {number} limit - Optional limit of the results.
+   * @param {number|null} limit - Optional limit of the results.
```
LMS007, line 24: `@param {number|null} [limit=null]`

src/sidebar/services/test/groups-test.js
```diff
@@ -507,13 +507,13 @@ describe('groups', function () {
       });
     });
 
-    it('injects a defalt organization if group is missing an organization', function () {
+    it('injects a default organization if group is missing an organization', function () {
```
LMS007, line 510: `defalt` looks like a kyle word :)

src/sidebar/services/test/groups-test.js
```diff
@@ -507,13 +507,13 @@ describe('groups', function () {
...
       const svc = service();
       const groups = [{ id: '39r39f', name: 'Ding Dong!' }];
       fakeApi.groups.list.returns(Promise.resolve(groups));
       return svc.load().then(groups => {
         assert.isObject(groups[0].organization);
-        assert.hasAllKeys(groups[0].organization, ['id', 'logo']);
+        assert.hasAllKeys(groups[0].organization, ['id', 'name', 'logo']);
```
LMS007, line 516: 👍

src/sidebar/util/groups.js
```diff
@@ -65,6 +65,10 @@ export function combineGroups(userGroups, featuredGroups, uri, settings) {
   return groups;
 }
 
+/**
+ * @param {Group} group
+ * @param {string} uri
+ */
```
LMS007, line 71: I'm wondering how many of the `@return` docs we need to flush out or just not worry if TS does not complain?

src/types/api.js
```diff
@@ -12,12 +12,18 @@
...
  * @prop {string} created
+ * @prop {string} updated
+ * @prop {string[]} tags
+ * @prop {string} text
+ * @prop {string} uri
+ * @prop {string} user
  */
 
 /**
  * TODO - Fill out remaining properties
```
LMS007, line 23: is this TODO done?

src/types/api.js
```diff
@@ -30,6 +36,12 @@
...
  */
 
+/**
+ * @typedef GroupScopes
+ * @prop {boolean} enforced
+ * @prop {string[]} uri_patterns;
+ */
+
 /**
  * TODO - Fill out remaining properties
```
LMS007, line 46: same here

src/sidebar/services/stream-filter.js
robertknight: There are two different uses of `Object` to be aware of:

1. The first is to define a property or parameter type as "some kind of value that is not null/undefined" but you don't have enough information to be more specific. `Function` is similar, but for functions specifically.
2. The second use case, which is what we're doing here, is defining an object type "inline" as part of a parameter definition. The first line has `@param {Object} <parameter name>` and that is followed by several `@param {<property type>} <parameter name>.<property name>` lines.

src/sidebar/services/test/groups-test.js
robertknight: I did not check the git history here 😜

src/sidebar/util/groups.js
robertknight: Providing you specify the types of the parameters, TS can infer the return type automatically. However there are reasons why you may want to specify it explicitly:

1. To help human readers
2. To get TS to check that the returned type is what it should be (for example, if a function is documented with `@return {Annotation}` and you have a code path that returns `null` or some other value that is not compatible with this, TS will complain, which is good because other code using that function needs to handle these other possibilities!)
3. In some cases, this can enable more readable/succinct error messages if two types don't match up (for example, if you call a function that returns an object and then try to access a property that doesn't exist on that object, TS will report "Cannot access property X on type <Name>" instead of "Cannot access property X on type { < list of all the properties that the function returns > }")

We'll need to figure out the right balance here, but my approach so far has been to omit the return type if the function is short and the return type is "obvious", but specify it otherwise.

src/sidebar/services/service-url.js
robertknight: According to [the reference](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) either `@returns` or `@return` works, so we should just pick one of the two as a convention and stick to it.

src/sidebar/services/stream-filter.js
robertknight: I like this. Will do.

src/sidebar/services/tags.js
robertknight: I didn't specify a default here because TS infers it automatically from the `= null` in the parameter list below.

src/types/api.js
robertknight: Not yet. Going over the TODOs in this file is still to be done.

src/types/api.js
robertknight: Not yet. Going over the TODOs in this file is still to be done.

robertknight:
I made a change to intent nested `@prop`, as in:

```js
/**
 * ...
 * @prop {Object} foo
 *  @prop {string} foo.bar
 *  @prop {string} foo.baz
 */
```

We'll need to remember to do this elsewhere.

src/sidebar/util/groups.js
LMS007: I have been noticing that error today and its more helpful to see the list of types when possible. I suppose we could do multiple passes and fill in return values.

PR #2277, opened 2020-06-25
Upgrade to ESLint v7.x

robertknight:
Update eslint-config-hypothesis which was in turn holding back various
other ESLint packages.

The newest update to eslint-plugin-react-hooks found an incorrect usage
of `useCallback` in `ThreadCard` where `debounce` was being called on
every render instead of only when certain dependencies changed.


robertknight:
The tests passed so I'm going to go ahead and merge this.

Issue #2270, opened 2020-06-24
Search and selecting annotations should not be additive


LMS007:
### Steps to reproduce
1. Search for some test using the search bar that will return 0 results.
2. Click an highlighted annotation on the host page

### Expected behaviour
That annotation should show up in the sidebar

### Actual behaviour
The annotation is still hidden because both filters are applied.

![search2](https://user-images.githubusercontent.com/3939074/85626782-65b93000-b622-11ea-82d9-3470e9307ce2.gif)

### Steps to reproduce
1. Click an highlighted annotation on the host page
2. Search for some text not found in that single annotation, but in another.

### Expected behaviour
The search does not produce any results.

### Actual behaviour
The search should clear out the selected annotation and show the search results among the entire set of annotations on the page

![select](https://user-images.githubusercontent.com/3939074/85626645-31457400-b622-11ea-80c8-c944b66380f4.gif)

PR #2290, opened 2020-06-29
Search and Select annotation clear one another

LMS007:
Fix issue where searching and selecting an annotation will be additive and create an unexpected result.

1. When searching for an annotation, it clears the selection.
2. When selecting an annotation, the search is cleared.

fixes https://github.com/hypothesis/client/issues/2270

Testing with LMS: There are a lot of paths to testing this,I tried to mix up a few varieties, but you can see the filters stacking on master but clear eachother out with the fix. The focused user seems to work as intended.

**Master**
old_filter

-----------

**This branch**
new_filter

### DIFF
--- a/src/sidebar/components/search-input.js
+++ b/src/sidebar/components/search-input.js
@@ -29,7 +29,10 @@ export default function SearchInput({ alwaysExpanded, query, onSearch }) {
 
   const onSubmit = e => {
     e.preventDefault();
-    onSearch(input.current.value);
+    // Don't submit an empty query
+    if (input.current.value.trim()) {
+      onSearch(input.current.value);
+    }
   };
 
   // When the active query changes outside of this component, update the input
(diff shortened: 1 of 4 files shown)


src/sidebar/components/search-input.js
```diff
@@ -29,7 +29,10 @@ export default function SearchInput({ alwaysExpanded, query, onSearch }) {
 
   const onSubmit = e => {
     e.preventDefault();
-    onSearch(input.current.value);
+    // Don't submit an empty query
+    if (input.current.value.trim()) {
```
LMS007, line 33: This now became an issue because just clicking the search icon would normally invoke the search. Prior to this change, it didn't cause any side effects because the selection was not being cleared. But now that simple action will clear out any selected annotation -- which a user wouldn't want unless they were actually making a search. The simple fix is to confirm there is something in the input before firing off the search event.

PR #2297, opened 2020-07-02
Send notifications to Slack about client deployment events

robertknight:
This PR adds notifications about QA and production deployments to the client's Jenkinsfile. The motivation came from a [Slack discussion](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1593696733351500?thread_ts=1593696631.350300&cid=C4K6M7P5E)

When reviewing this, I recommend looking at the whitespace only diff, since I had to wrap several large blocks in a `try..catch` expression.

To understand the syntax here, note that the client uses Jenkins "[Scripted Pipeline](https://www.jenkins.io/doc/book/pipeline/syntax/#flow-control)" as opposed to a "declarative pipeline". Hence `try...catch` is used to catch failures rather than the `post { ... }` blocks you'll find in the Jenkinsfile in the "deployment" repo.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,11 @@
 #!groovy
 
+// `color` can be a color code or one of "good", "warning" or "danger".
+// See https://www.jenkins.io/doc/pipeline/steps/slack/.
+def postSlack(color, message) {
+    slackSend channel: "#eng-frontend", color: color, message: message
+}
+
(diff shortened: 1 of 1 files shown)


robertknight:
I'm going to merge this and do and end-to-end test. We can change details (eg. which Slack channel to send to) later. @lyzadanger - would you mind doing a post-hoc review when you're about later?

PR #2298, opened 2020-07-02
Improve typechecking on annotation-metatdata

LMS007:
Further fill in the {Annotation} typedef

----

pretty straightforward PR, I made a choice to nest objects all under the `Annotation` type rather than make custom `@typdefs` for the sub-objects that really wouldn't stand on their own or have much meaning by themselves.

### DIFF
--- a/src/sidebar/util/annotation-metadata.js
+++ b/src/sidebar/util/annotation-metadata.js
@@ -8,9 +8,7 @@
 
 /** Extract a URI, domain and title from the given domain model object.
  *
- * @param {object} annotation
- * @returns {object} An object with three properties extracted from the model:
- *   uri, domain and title.
+ * @param {Annotation} annotation
  *
  */
 export function documentMetadata(annotation) {
@@ -267,9 +265,9 @@ export function location(annotation) {
     const targets = annotation.target || [];
     for (let i = 0; i < targets.length; i++) {
       const selectors = targets[i].selector || [];
-      for (let k = 0; k < selectors.length; k++) {
-        if (selectors[k].type === 'TextPositionSelector') {
-          return /** @type {TextPositionSelector} */ (selectors[k]).start;
+      for (const selector of selectors) {
+        if (selector.type === 'TextPositionSelector') {
+          return selector.start;
         }
       }
     }
--- a/src/sidebar/util/annotation-metadata.js
+++ b/src/sidebar/util/annotation-metadata.js
@@ -3,11 +3,12 @@
  */
 
 /** @typedef {import('../../types/api').Annotation} Annotation */
+/** @typedef {import('../../types/api').TextPositionSelector} TextPositionSelector */
+/** @typedef {import('../../types/api').TextQuoteSelector} TextQuoteSelector */
 
 /** Extract a URI, domain and title from the given domain model object.
  *
- * @param {object} annotation An annotation domain model object as received
- *   from the server-side API.
+ * @param {object} annotation
  * @returns {object} An object with three properties extracted from the model:
  *   uri, domain and title.
  *
@@ -36,7 +37,7 @@ export function documentMetadata(annotation) {
 /**
  * Return the domain and title of an annotation for display on an annotation
  * card.
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  */
 export function domainAndTitle(annotation) {
   return {
@@ -47,10 +48,7 @@ export function domainAndTitle(annotation) {
 }
 
 /**
- * Returns the uri or the incontext link from an annotation,
- * or null if the protocol is not http(s).
- *
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  */
 function titleLinkFromAnnotation(annotation) {
   let titleLink = /** @type {string|null} */ (annotation.uri);
@@ -72,7 +70,7 @@ function titleLinkFromAnnotation(annotation) {
 /**
  * Returns the domain text from an annotation.
  *
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  */
 function domainTextFromAnnotation(annotation) {
   const document = documentMetadata(annotation);
@@ -95,7 +93,7 @@ function domainTextFromAnnotation(annotation) {
  * Returns the title text from an annotation and crops it to 30 chars
  * if needed.
  *
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  */
 function titleTextFromAnnotation(annotation) {
   const document = documentMetadata(annotation);
@@ -122,7 +120,7 @@ export function isReply(annotation) {
  * "New" means this annotation has been newly created client-side and not
  * saved to the server yet.
  *
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  */
 export function isNew(annotation) {
   return !annotation.id;
@@ -131,7 +129,7 @@ export function isNew(annotation) {
 /**
  * Return `true` if the given annotation is public, `false` otherwise.
  *
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  */
 export function isPublic(annotation) {
   let isPublic = false;
@@ -157,7 +155,7 @@ export function isPublic(annotation) {
  * as opposed to a Page Note which refers to the whole document or a reply,
  * which refers to another annotation.
  *
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  */
 function hasSelector(annotation) {
   return !!(
@@ -173,7 +171,7 @@ function hasSelector(annotation) {
  * Returns false if anchoring is still in process but the flag indicating that
  * the initial timeout allowed for anchoring has expired.
  *
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  */
 export function isWaitingToAnchor(annotation) {
   return (
@@ -186,7 +184,7 @@ export function isWaitingToAnchor(annotation) {
 /**
  * Has this annotation hidden by moderators?
  *
- * @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
  * @return {boolean}
  */
 export function isHidden(annotation) {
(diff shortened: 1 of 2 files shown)


src/types/api.js
```diff
@@ -17,6 +17,32 @@
...
+ *   @prop {string[]} permissions.read
+ *   @prop {string[]} permissions.update
+ *   @prop {string[]} permissions.delete
+ *
+ * @prop {Object[]} target
+ *   @prop {string} target.source
+ *   @prop {Object[]} target.selector
+ *     @prop {string} target.selector.type
+ *     @prop {string} target.selector.exact
+ *     @prop {number} target.selector.start
```
robertknight, line 38: The properties that a selector has depends on the type of the selector. Creating a type with all the properties and marking them optional is one way of representing this. A better way though is to use a _union_:

```js
/**
 * @typedef TextQuoteSelector
 * @prop {'TextQuoteSelector'} type
 * @prop {string} exact
 * @prop {string} [prefix]
 * @prop {string} [suffix]
 */

/**
 * @typedef TextPositionSelector
 * @prop {'TextPositionSelector'} type
 * @prop {number} start
 * @prop {number} end
 */

/**
 * @typedef {TextPositionSelector|TextQuoteSelector} Selector
 */
```

The advantage of this is that it provides better documentation to the reader, and it also enables TS to do
more accurate checking that you're accessing props which exist:

```js
const selector = ...
if (selector.type === 'TextQuoteSelector') {
  // TS knows that `selector` is a `TextQuoteSelector` here and
  // has the properties "exact", "suffix" and "prefix".
}
```

src/types/api.js
```diff
@@ -17,6 +17,32 @@
...
+ *
+ * @prop {Object} moderation
+ *   @prop {number} moderation.flagCount
+ *
+ * @prop {Object} links
+ *   @prop {string} links.incontext
+ *   @prop {string} links.html
```
robertknight, line 45:
```suggestion
 * @prop {Object} links
 *   @prop {string} [links.incontext]
 *   @prop {string} [links.html]
```

All of the link types are optional, so the client needs to check before it uses them.

src/types/api.js
```diff
@@ -17,6 +17,32 @@
...
+ *   @prop {string} document.title
+ *
+ * @prop {Object} permissions
+ *   @prop {string[]} permissions.read
+ *   @prop {string[]} permissions.update
+ *   @prop {string[]} permissions.delete
+ *
+ * @prop {Object[]} target
+ *   @prop {string} target.source
+ *   @prop {Object[]} target.selector
```
robertknight, line 35:
```suggestion
 *   @prop {Object[]} [target.selector]
```

The `selector` property is optional. Page notes do not have it.

src/types/api.js
```diff
@@ -17,6 +17,32 @@
...
+ *     @prop {string} target.selector.type
+ *     @prop {string} target.selector.exact
+ *     @prop {number} target.selector.start
+ *
+ * @prop {Object} moderation
```
robertknight, line 40:
```suggestion
 * @prop {Object} [moderation]
```

The `moderation` parameter may be missing.

robertknight: Related to this, I expected that if a property is marked as optional then you'd get an error if you try to use it without checking whether it exists or not. That doesn't appear to be the case though. I'm unclear if that is [intended behaviour or a bug](https://github.com/microsoft/TypeScript/issues/39111), as this is different to how it works in `.ts` files.

In any case, it is useful for human readers to convey whether a property is optional or not.

src/types/api.js
LMS007: That makes a lot more sense, I was just bliding filling props as I saw them used here and there. One advantage of doing this TS documentation is we now have a concrete.definition for the shit we're using!

src/types/api.js
LMS007: Sure. 
My method so far has been to make things not-optional until TS complains. I'm guessing as we fill things in more, these errors will popup and then we can go back and make things optional as they are needed.  Because you have knowledge of this, we can do it now, but if I'm to liberal here, and for example, I make a prop optional that really should not be, then we'll never get an error thats its incorrect unless we just visually catch it.

src/types/api.js
```diff
@@ -17,6 +17,32 @@
  * @prop {string} text
  * @prop {string} uri
  * @prop {string} user
+ * @prop {boolean} hidden
+ * @prop {boolean} [$highlight]
+ * @prop {boolean} [$orphan]
+ * @prop {boolean} [$anchorTimeout]
```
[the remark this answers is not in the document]
LMS007: In that case, how about another type with a union?

src/sidebar/util/annotation-metadata.js
```diff
@@ -42,8 +46,14 @@ export function domainAndTitle(annotation) {
   };
 }
 
+/**
+ * Returns the uri or the incontext link from an annotation,
+ * or null if the protocol is not http(s).
```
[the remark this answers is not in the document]
LMS007: roger

src/sidebar/util/annotation-metadata.js
```diff
@@ -224,7 +269,7 @@ export function location(annotation) {
       const selectors = targets[i].selector || [];
       for (let k = 0; k < selectors.length; k++) {
         if (selectors[k].type === 'TextPositionSelector') {
-          return selectors[k].start;
+          return /** @type {TextPositionSelector} */ (selectors[k]).start;
```
LMS007, line 272: I was a bit surprised by this. Without the casting, I saw this error

```
src/sidebar/util/annotation-metadata.js:271:31 - error TS2339: Property 'start' does not exist on type 'TextQuoteSelector | TextPositionSelector'.
  Property 'start' does not exist on type 'TextQuoteSelector'.

271           return selectors[k].start;
```
It does not appear to be the case that typescript looks at the condition and can infer that the type must be TextPositionSelector.

src/sidebar/util/annotation-metadata.js
```diff
@@ -2,16 +2,20 @@
...
 
+/** @typedef {import('../../types/api').Annotation} Annotation */
+/** @typedef {import('../../types/api').TextPositionSelector} TextPositionSelector */
+/** @typedef {import('../../types/api').TextQuoteSelector} TextQuoteSelector */
+
 /** Extract a URI, domain and title from the given domain model object.
  *
- * @param {object} annotation An annotation domain model object as received
- *   from the server-side API.
+ * @param {object} annotation
```
robertknight, line 11:
```suggestion
 * @param {Annotation} annotation
```

src/sidebar/util/annotation-metadata.js
```diff
@@ -2,16 +2,20 @@
...
  * @returns {object} An object with three properties extracted from the model:
```
robertknight, line 12: It wasn't part of your changes, but I would recommend to remove the `@returns` line entirely for `documentMetadata`. It doesn't provide useful information to the reader beyond what is obvious from the code and the `{object}` type overrides the inferred return type with something less useful (`object` or `Object` to TS means "could be anything").

src/sidebar/util/annotation-metadata.js
robertknight: It looks like the inference only works when the expression that returns the object you are checking is "simple". If I change the code to replace `selectors[k]` with a variable, then it works:

```diff
diff --git a/src/sidebar/util/annotation-metadata.js b/src/sidebar/util/annotation-metadata.js
index a69ad2044..58845a8c7 100644
--- a/src/sidebar/util/annotation-metadata.js
+++ b/src/sidebar/util/annotation-metadata.js
@@ -268,8 +268,9 @@ export function location(annotation) {
     for (let i = 0; i < targets.length; i++) {
       const selectors = targets[i].selector || [];
       for (let k = 0; k < selectors.length; k++) {
-        if (selectors[k].type === 'TextPositionSelector') {
-          return /** @type {TextPositionSelector} */ (selectors[k]).start;
+        const selector = selectors[k];
+        if (selector.type === 'TextPositionSelector') {
+          return selector.start;
         }
       }
     }
```

I've seen this happen in other contexts too. I would guess it is because figuring out if two different uses of `selectors[k]` will return the same object requires much smarter analysis (for example, making sure that the array wasn't modified in any way in between).

robertknight: If making the above change, you might as well replace the `for` loop with a `for ... of` loop.

src/sidebar/util/annotation-metadata.js
```diff
@@ -236,28 +281,30 @@ export function location(annotation) {
...
     return null;
   }
-  const target = ann.target[0];
+  const target = annotation.target[0];
   if (!target.selector) {
     return null;
   }
   const quoteSel = target.selector.find(s => s.type === 'TextQuoteSelector');
-  return quoteSel ? quoteSel.exact : null;
+  return quoteSel ? /** @type {TextQuoteSelector}*/ (quoteSel).exact : null;
```
robertknight, line 309: 👍

src/types/api.js
```diff
@@ -5,6 +5,25 @@
...
+
+/**
+ * @typedef TextPositionSelector
+ * @prop {'TextPositionSelector'} type
+ * @prop {number} start
+ * @prop {number} end
+ */
+
+/**
+ * @typedef {TextQuoteSelector | TextPositionSelector} Selector
```
robertknight, line 24: There is one other type of selector that is actively used by the client, an XPath "range" selector. We can add that later.

src/types/api.js
```diff
@@ -17,6 +36,31 @@
  * @prop {string} text
  * @prop {string} uri
  * @prop {string} user
+ * @prop {boolean} hidden
+ *
+ * @prop {Object} document
+ *   @prop {string} document.title
```
robertknight, line 42: I'm not sure if the `document` and `document.title` properties will always be present. You'd have to check the backend code to find out.

src/types/api.js
LMS007: yep, I saw that! adding that in a follow up PR

src/sidebar/util/annotation-metadata.js
LMS007: okay, thanks for sorting that out. Good to know. I feel like we may need to compile a list of "gotchas" for this TS stuff :)

src/sidebar/util/annotation-metadata.js
LMS007: I guess its good to point out that unless any return doc is very specific and asserts correctness, then we shouldn't add it in the first place.

src/types/api.js
LMS007: I'll check on this and change if need.

PR #2315, opened 2020-07-06
Typecheck utils

LMS007:
- Adding a few more params and defs for various modules inside of the util/ and services/ folder.
- Still lots more to go, but this incrementally moves ot closer to having param completeness.
- There are still missing params, but they entail larger changes that would balloon the PR so I want to keep the changes small.

### DIFF
--- a/src/sidebar/services/analytics.js
+++ b/src/sidebar/services/analytics.js
@@ -34,7 +34,6 @@ export const events = {
  * behavior across different environments in which the client is used.
  *
  * @param {Window} win
- * @param {Object} settings - Settings rendered into sidebar HTML
  * @return {string}
  */
 function clientType(win, settings = {}) {
@@ -66,8 +65,17 @@ function clientType(win, settings = {}) {
 }
 
 /**
+ * @typedef Analytics
+ * @prop {() => any} sendPageView
+ * @prop {(action: string, label?: string, value?: number) => void} track
+ * @prop {Object.<string, string>} events
+ */
+
+/**
  * Analytics service for tracking page views and user interactions with the
  * application.
+ * @param {Window} $window - Test seam
+ * @return {Analytics}
  */
 // @ngInject
 export default function analytics($window, settings) {
@@ -78,6 +86,7 @@ export default function analytics($window, settings) {
   // is replaced when analytics.js fully loads.
   //
   // See https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference
+  // @ts-ignore The window interface needs to be expanded to include this property
   const commandQueue = () => $window.ga || noop;
 
   return {
--- a/src/sidebar/services/annotations.js
+++ b/src/sidebar/services/annotations.js
@@ -1,3 +1,6 @@
+/** @typedef {import('../../types/api').Annotation} Annotation */
+/** @typedef {import('../../types/annotator').AnnotationData} AnnotationData */
+
 /**
  * A service for creating, manipulating and persisting annotations and their
  * application-store representations. Interacts with API services as needed.
@@ -17,6 +20,8 @@ export default function annotationsService(api, store) {
   /**
    * Apply changes for the given `annotation` from its draft in the store (if
    * any) and return a new object with those changes integrated.
+   *
+   * @param {Annotation} annotation
    */
   function applyDraftChanges(annotation) {
     const changes = {};
@@ -36,6 +41,10 @@ export default function annotationsService(api, store) {
 
   /**
    * Extend new annotation objects with defaults and permissions.
+   *
+   * @param {AnnotationData} annotationData
+   * @param {Date} now
+   * @return {Annotation}
    */
   function initialize(annotationData, now = new Date()) {
     const defaultPrivacy = store.getDefault('annotationPrivacy');
@@ -50,27 +59,29 @@ export default function annotationsService(api, store) {
     // as it has not been persisted to the service.
     const $tag = generateHexString(8);
 
-    let permissions = defaultPermissions(userid, groupid, defaultPrivacy);
-
-    // Highlights are peculiar in that they always have private permissions
-    if (metadata.isHighlight(annotationData)) {
-      permissions = privatePermissions(userid);
-    }
-
-    return Object.assign(
+    /** @type {Annotation} */
+    const annotation = Object.assign(
       {
         created: now.toISOString(),
         group: groupid,
-        permissions,
+        permissions: defaultPermissions(userid, groupid, defaultPrivacy),
         tags: [],
         text: '',
         updated: now.toISOString(),
         user: userid,
         user_info: userInfo,
         $tag: $tag,
+        hidden: false,
+        links: {},
       },
       annotationData
     );
+
+    // Highlights are peculiar in that they always have private permissions
+    if (metadata.isHighlight(annotation)) {
+      annotation.permissions = privatePermissions(userid);
+    }
+    return annotation;
   }
 
   /**
--- a/src/sidebar/util/group-list-item-common.js
+++ b/src/sidebar/util/group-list-item-common.js
@@ -9,7 +9,3 @@
 export function orgName(group) {
   return group.organization && group.organization.name;
 }
-
-export function trackViewGroupActivity(analytics) {
-  analytics.track(analytics.events.GROUP_VIEW_ACTIVITY);
-}
--- a/src/types/annotator.js
+++ b/src/types/annotator.js
@@ -1,0 +1,34 @@
+/**
+ * Type definitions for objects passed between the annotator and sidebar.
+ */
+
+/** @typedef {import("./api").Target} Target */
+
+/**
+ * @typedef AnnotationData
+ * @prop {string} uri
+ * @prop {Target[]} target
+ * @prop {string} $tag
+ * @prop {boolean} [$highlight]
+ * @prop {DocumentMetadata} document
+ */
+
+/**
+ * @typedef DocumentMetadata
+ * @prop {string} title
+ * @prop {Object[]} link
+ *   @prop {string} link.rel
+ *   @prop {string} link.type
+ * // html pages
+ * @prop {Object.<string, string[]>} [dc]
+ * @prop {Object.<string, string[]>} [eprints]
+ * @prop {Object.<string, string[]>} [facebook]
+ * @prop {Object.<string, string[]>} [highwire]
+ * @prop {Object.<string, string[]>} [prism]
+ * @prop {Object.<string, string[]>} [twitter]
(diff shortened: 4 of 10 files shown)


src/sidebar/services/analytics.js
```diff
@@ -34,7 +36,6 @@ export const events = {
  * behavior across different environments in which the client is used.
  *
  * @param {Window} win
- * @param {Object} settings - Settings rendered into sidebar HTML
```
LMS007, line 37: Seems too general, so remove for now, fill it correctly later

src/types/api.js
```diff
@@ -73,6 +73,34 @@
  * @prop {boolean} [$anchorTimeout]
  */
 
+/**
+ * @typedef AnnotationData
```
robertknight, line 77: I assume this is meant to be a typedef for the object that is exchanged between the sidebar and annotator to represent an annotation?

I would suggest to put that in a separate file (`types/annotator.js`?) since it isn't something that comes from/is sent to the hypothesis API directly.

src/types/api.js
```diff
@@ -73,6 +73,34 @@
...
+ * @prop {string} uri
+ * @prop {string} $tag
+ * @prop {string} [$highlight]
+ * @prop {SelectSource[]} target
```
robertknight, line 81: This is the same as the `target` property of the `Annotation` type above isn't it. Perhaps unify that and `SelectSource` into a single `Target` definition? The `target` property would then be a `Target[]`.

src/types/api.js
```diff
@@ -73,6 +73,34 @@
...
+/**
+ * @typedef SelectSource
+ * @prop {string} source
+ * @prop {Selector[]} selector
+ */
+
+/**
+ * @typedef Document
+ * @prop {string} title
+ * @prop {string[]} link
```
robertknight, line 94: `link` is an array of objects with `href` and `ref` properties rather than strings. See https://github.com/hypothesis/client/blob/c2ca6c8f6c3afbfddfacc0ebcf158ae610a1ea2f/src/annotator/plugin/document.js#L189. Note that the `ref` and `type` properties on that line are only currently written/read on the "annotator" side of the application.

src/types/api.js
LMS007: Is the same true for `document` between AnnotationData.document vs Annotation.document.

src/types/api.js
LMS007: good catch! I'm not sur eif they are required or not, but I lean on making everything required until I see errors.

src/types/api.js
LMS007: bah! yes, missed that.

src/sidebar/util/annotation-metadata.js
```diff
@@ -206,8 +207,10 @@ export function isHighlight(annotation) {
   if (annotation.$highlight) {
     return true;
   }
+  // treat like @type Annotation from here
+  const _annotation = /**@type {Annotation}*/ (annotation);
```
robertknight, line 211: This cast hides a possible error if the parameter is an `AnnotationData` rather than an `Annotation`. If any of the code below tried to access a property of the object that only existed on an `annotation`, that could have unexpected behaviour.

One way to resolve this problem would be to modify the code in `src/sidebar/services/annotations.js` so that it only calls `isHighlight` with an `Annotation`:

```diff
diff --git a/src/sidebar/services/annotations.js b/src/sidebar/services/annotations.js
index ef7f07d7d..e55dc96f9 100644
--- a/src/sidebar/services/annotations.js
+++ b/src/sidebar/services/annotations.js
@@ -60,18 +60,12 @@ export default function annotationsService(api, store) {
     // as it has not been persisted to the service.
     const $tag = generateHexString(8);
 
-    let permissions = defaultPermissions(userid, groupid, defaultPrivacy);
-
-    // Highlights are peculiar in that they always have private permissions
-    if (metadata.isHighlight(annotationData)) {
-      permissions = privatePermissions(userid);
-    }
-
-    return Object.assign(
+    /** @type {Annotation} */
+    const annotation = Object.assign(
       {
         created: now.toISOString(),
         group: groupid,
-        permissions,
+        permissions: defaultPermissions(userid, groupid, defaultPrivacy),
         tags: [],
         text: '',
         updated: now.toISOString(),
@@ -79,8 +73,15 @@ export default function annotationsService(api, store) {
         user_info: userInfo,
         $tag: $tag,
       },
-      /**@type {Object}*/ (annotationData)
+      annotationData
     );
+
+    // Highlights are peculiar in that they always have private permissions
+    if (metadata.isHighlight(annotation)) {
+      annotation.permissions = privatePermissions(userid);
+    }
+
+    return annotation;
   }
 
   /**
```

With this diff, no changes to `isHighlight` are needed. Note that in that diff I have assumed that the issue mentioned above with the cast from `AnnotationData` to `Annotation` has been fixed.

src/sidebar/util/group-list-item-common.js
```diff
@@ -10,6 +10,9 @@ export function orgName(group) {
   return group.organization && group.organization.name;
 }
 
+/**
+ * @param {import("../services/analytics").Analytics} analytics
+ */
 export function trackViewGroupActivity(analytics) {
```
robertknight, line 16: This function isn't actually used anywhere in the codebase any more. We could just remove it and associated tests.

src/types/api.js
```diff
@@ -33,6 +33,12 @@
  * @typedef {TextQuoteSelector | TextPositionSelector | RangeSelector} Selector
  */
 
+/**
+ * @typedef Target
+ * @prop {string} target.source
+ * @prop {Selector[]} [target.selector]
```
robertknight, line 39:
```suggestion
 * @prop {string} source
 * @prop {Selector[]} [selector]
```

I'm surprised this didn't generate an error but in any case, the parent property name is only needed for "inline" typedefs inside a parameter list or other typedef.

src/types/api.js
```diff
@@ -113,5 +117,12 @@
  * @prop {boolean} canLeave
  */
 
+/**
+ * @typedef Analytics
+ * @prop {() => any} sendPageView
+ * @prop {(action: string, label?: string, value?: number) => void} track
+ * @prop {Object} events
```
robertknight, line 124: This service interface definition isn't part of the Hypothesis API. Can you move it to `src/sidebar/services/analytics.js`?

src/types/annotator.js
```diff
@@ -0,0 +1,31 @@
...
+ * @typedef AnnotationData
+ * @prop {string} uri
+ * @prop {string} $tag
+ * @prop {string} [$highlight]
+ * @prop {Target[]} target
+ * @prop {Document} document
+ */
+
+/**
+ * @typedef Document
```
robertknight, line 17: Can I suggest renaming this to say, DocumentMetadata` to avoid confusion with the DOM `Document` interface.

src/types/annotator.js
```diff
@@ -0,0 +1,31 @@
...
+ * @prop {string} title
+ * @prop {Object[]} link
+ *   @prop {string} link.rel
+ *   @prop {string} link.type
+ * @prop {Object} dc
+ * @prop {Object} eprints
+ * @prop {Object} facebook
+ * @prop {Object} highwire
+ * @prop {Object} prism
+ * @prop {Object} twitter
```
robertknight, line 27: On the annotator side of the application this metadata can be generated in two places:

- `src/annotator/plugin/pdf-metadata.js`
- `src/annotator/plugin/document.js`

It looks to me like only the `title` and `link` top-level properties are common between them and therefore the other fields may not be present. From a quick look over those two files it looks like in web pages there will also be `dc`, `eprints`, `facebook`, `highwire`, `prism` and `twitter`. In PDFs there will also be `documentFingerprint`.

Looking at the tests in `src/annotator/plugin/test/document-test.js`, the types of the dc, eprints etc. fields is a map between metadata field names and arrays of values, which can be expressed here as `Object.<string, string[]>`.

src/sidebar/services/annotations.js
```diff
@@ -69,7 +79,7 @@ export default function annotationsService(api, store) {
         user_info: userInfo,
         $tag: $tag,
       },
-      annotationData
+      /**@type {Object}*/ (annotationData)
```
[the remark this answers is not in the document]
LMS007: I totally missed that, good fix.

src/sidebar/util/annotation-metadata.js
LMS007: 👍

src/types/api.js
LMS007: Must have been a refactor copy paste error, my mistake. And yes, I agree, the error would have been nice.

src/types/annotator.js
LMS007: indeed, better name.

src/sidebar/util/annotation-metadata.js
```diff
@@ -3,6 +3,7 @@
  */
 
 /** @typedef {import('../../types/api').Annotation} Annotation */
+/** @typedef {import('../../types/annotator').AnnotationData} AnnotationData */
```
robertknight, line 6: The changes in this file are not needed any more because of the changes you made to the place where `isHighlight` is called.

src/types/annotator.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Type definitions for objects passes between the annotator and sidebar
```
robertknight, line 2:
```suggestion
 * Type definitions for objects passed between the annotator and sidebar
```

src/types/api.js
```diff
@@ -113,5 +117,12 @@
  * @prop {boolean} canLeave
  */
 
+/**
+ * @typedef Analytics
```
robertknight, line 121: This `@typedef` was moved to `services/analytics.js` so the one here should be removed.

src/sidebar/util/annotation-metadata.js
LMS007: yes, it only needs Annotation now, good catch

src/sidebar/services/analytics.js
```diff
@@ -65,9 +64,18 @@ function clientType(win, settings = {}) {
   return type;
 }
 
+/**
+ * @typedef Analytics
+ * @prop {() => any} sendPageView
+ * @prop {(action: string, label?: string, value?: number) => void} track
+ * @prop {Object} events
```
robertknight, line 71:
```suggestion
 * @prop {Object.<string, string>} events
```

src/sidebar/util/annotation-metadata.js
robertknight: There are still unnecessary changes here.

PR #2318, opened 2020-07-07
Improve typechecking for hypothesis stores

LMS007:
**Updated**

I stubbed out most of the actions and selectors for the stores. I have left some of the more complex stores which mean require some additional typedefs here and there for a future PR. This is intended to be a first pass at getting the store actions / selectors typechecking in reasonable shape.

### DIFF
--- a/src/sidebar/store/create-store.js
+++ b/src/sidebar/store/create-store.js
@@ -42,7 +42,7 @@ import { createReducer, bindSelectors } from './util';
  *  SessionStore &
  *  SidebarPanelsStore &
  *  ToastMessagesStore &
- *  ViewerStore} HypothesisStore
+ *  ViewerStore} SidebarStore
  */
 
 /**
@@ -63,7 +63,7 @@ import { createReducer, bindSelectors } from './util';
  * @param {Object[]} modules
  * @param {any[]} [initArgs] - Arguments to pass to each state module's `init` function
  * @param {any[]} [middleware] - List of additional Redux middlewares to use.
- * @return {HypothesisStore}
+ * @return {SidebarStore}
  */
 export default function createStore(modules, initArgs = [], middleware = []) {
   // Create the initial state and state update function.
@@ -117,7 +117,7 @@ export default function createStore(modules, initArgs = [], middleware = []) {
   }
 
   // Create the store.
-  const store = /** @type {HypothesisStore} */ (redux.createStore(
+  const store = /** @type {SidebarStore} */ (redux.createStore(
     reducer,
     initialState,
     enhancer
--- a/src/sidebar/store/create-store.js
+++ b/src/sidebar/store/create-store.js
@@ -8,6 +8,23 @@ import immutable from '../util/immutable';
 import { createReducer, bindSelectors } from './util';
 
 /**
+ * @typedef {import("../../types/api").State} State
+ * @typedef {import("../../types/api").Action} Action
+ *
+ * // Base redux store
+ * @typedef {import("redux").Store} ReduxStore<State, Action>
+ *
+ * // Custom stores
+ * @typedef {import("./modules/activity").ActivityStore<State, Action>} ActivityStore
+ * @typedef {import("./modules/annotations").AnnotationsStore<State, Action>} AnnotationsStore
+ * @typedef {import("./modules/defaults").DefaultsStore<State, Action>} Defaults
+ * TODO: add the rest of the stores...
+ *
+ * // Combine all stores
+ * @typedef {ReduxStore & ActivityStore & AnnotationsStore & Defaults} HypothesisStore
+ */
+
+/**
  * Create a Redux store from a set of _modules_.
  *
  * Each module defines the logic related to a particular piece of the application
@@ -25,6 +42,7 @@ import { createReducer, bindSelectors } from './util';
  * @param {Object[]} modules
  * @param {any[]} [initArgs] - Arguments to pass to each state module's `init` function
  * @param {any[]} [middleware] - List of additional Redux middlewares to use.
+ * @return {HypothesisStore}
  */
 export default function createStore(modules, initArgs = [], middleware = []) {
   // Create the initial state and state update function.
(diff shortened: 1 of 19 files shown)


src/sidebar/store/modules/annotations.js
```diff
@@ -441,6 +441,20 @@ const isWaitingToAnchorAnnotations = createSelector(
...
+ * @template S The type of state held by this store.
+ * @template A the type of actions which may be dispatched by this store.
+ * @typedef AnnotationsStore<S, A>
+ *
+ * // actions
+ * @prop {()=>A} addAnnotations
+ * // TODO: add the rest
+ *
+ * // selectors
+ * @prop {(a: S) => S} annotationCount
```
LMS007, line 454: I think this makes sense... takes a "state" and returns a piece of it. Same type for the param and return value.

src/sidebar/store/modules/annotations.js
```diff
@@ -441,6 +441,20 @@ const isWaitingToAnchorAnnotations = createSelector(
   annotations => annotations.some(metadata.isWaitingToAnchor)
 );
 
+/**
+ * @template S The type of state held by this store.
+ * @template A the type of actions which may be dispatched by this store.
+ * @typedef AnnotationsStore<S, A>
+ *
+ * // actions
+ * @prop {()=>A} addAnnotations
```
LMS007, line 450: @robertknight , your suggestion in slack was to just use {`typeof`}, but it feels like making this a function that returns a generic `Action` is a bit more accurate. Does that seem reasonable or do you foresee any issues with this approach?

src/sidebar/store/create-store.js
```diff
@@ -7,6 +7,23 @@ import immutable from '../util/immutable';
 
 import { createReducer, bindSelectors } from './util';
 
+/**
+ * @typedef {import("../../types/api").State} State
```
LMS007, line 11: TODO: I should move these to this file, they don't belong in the api

src/sidebar/store/modules/drafts.js
```diff
@@ -178,13 +180,29 @@ function getDraftIfNotEmpty(state, annotation) {
 /**
  * Returns a list of draft annotations which have no id.
  *
- * @return {Object[]}
+ * @return {Annotation[]}
```
LMS007, line 183: Seemed like a reasonable change to make

src/sidebar/store/modules/links.js
```diff
@@ -22,6 +22,13 @@ function updateLinksAction(newLinks) {
   return { type: 'UPDATE_LINKS', newLinks: newLinks };
 }
 
+/**
+ * @typedef LinksStore
+ *
+ * // actions
+ * @prop {typeof updateLinks} updateLinksAction
+ */
+
```
LMS007, line 31: Line 35: I'm a bit confused what `update: { UPDATE_LINKS: updateLinks }` is supposed to be
I'll take a fresh look at that later on, but something does not seem right to me here...

src/sidebar/store/modules/real-time-updates.js
```diff
@@ -140,7 +140,7 @@ function clearPendingUpdates() {
  * Return added or updated annotations received via the WebSocket
  * which have not been applied to the local state.
  *
- * @return {{[id: string]: Annotation}}
+ * @return {Object<string, boolean>}
```
LMS007, line 143: I took the liberty of making this change because I want to stick with one standard or another here. I was not sure if the existing type was correctly formatted or not. I could use some input.

src/sidebar/store/modules/real-time-updates.js
```diff
@@ -150,7 +150,7 @@ function pendingUpdates(state) {
  * Return IDs of annotations which have been deleted on the server but not
  * yet removed from the local state.
  *
- * @return {{[id: string]: boolean}}
+ * @return {Object<string, boolean>}
```
LMS007, line 153: same here

src/sidebar/store/modules/real-time-updates.js
robertknight: I'm fine with this change. Both mean exactly the same but we've been favouring the second style. I find the `Object.<KeyType, ValueType>` syntax slightly more readable in JSDoc `{...}` parens. A small downside is that you can't name the key parameter.

src/sidebar/store/create-store.js
```diff
@@ -7,6 +7,44 @@ import immutable from '../util/immutable';
...
+ *  DraftsStore &
+ *  FramesStore &
+ *  GroupsStore &
+ *  LinksStore &
+ *  RealTimeUpdatesStore &
+ *  SelectionStore &
+ *  SessionStore &
+ *  SidebarPanelsStore &
+ *  ToastMessagesStore &
+ *  ViewerStore} HypothesisStore
```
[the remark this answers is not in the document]
LMS007: indeed

src/sidebar/store/modules/annotations.js
```diff
@@ -441,6 +439,31 @@ const isWaitingToAnchorAnnotations = createSelector(
...
+ * @prop {typeof removeAnnotations} removeAnnotations
+ * @prop {typeof updateAnchorStatus} updateAnchorStatus
+ * @prop {typeof updateFlagStatus} updateFlagStatus
+ * @prop {typeof unhideAnnotation} unhideAnnotation
+ *
+ * // selectors
+ * @prop {() => number} annotationCount
+ * @prop {(a: string) => boolean} annotationExists
+ * @prop {(a: string) => Annotation} findAnnotationByID
+ * @prop {(a: string[]) => string[]} findIDsForTags
```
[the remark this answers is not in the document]
LMS007: I guess my question is are these function param types used for readability or just type compliance. Because one can always (and imo) should look up the proper function definition. However I could see this going both ways... the actual function definition is difference because it takes a `state` prop which the user should ignore.

hmmm... maybe I'm leaning towards using non-generic names. My fear is an unnecessary upkeep /parity down the road.

PR #2323, opened 2020-07-09
Improve typechecking in several shared modules

LMS007:
I'm not sure if this is our standard or not, but I'm ignoring all private methods in a class or (class-like) thing. Seems like overkill to add typechecking to those; however, the effort could still catch bugs. We could discuss further in slack.

### DIFF
--- a/src/shared/frame-rpc.js
+++ b/src/shared/frame-rpc.js
@@ -29,6 +29,13 @@
 
 const VERSION = '1.0.0';
 
+/**
+ *  @constructor
+ *  @param {Window} src
+ *  @param {Window} dst
+ *  @param {string} origin
+ *  @param {(Object<string, ((any) => any)>) | ((any) => any)} methods
+ */
 export default function RPC(src, dst, origin, methods) {
   if (!(this instanceof RPC)) return new RPC(src, dst, origin, methods);
   const self = this;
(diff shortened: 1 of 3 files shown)


robertknight:
> I'm not sure if this is our standard or not, but I'm ignoring all private methods in a class or (class-like) thing. Seems like overkill to add typechecking to those; however, the effort could still catch bugs. We could discuss further in slack.

I agree it is more valuable to focus on the exported functions and types from a module, since catching integration bugs and improving documentation for consumers are two of the main goals. As you said though, there are still opportunities to catch bugs from annotating internal functions.

My overall approach for existing code has been to start by focusing on coverage of the public interface of modules, and then improving the internal interfaces later when working on that code for other reasons. @lyzadanger's work on the threading code is a good example of this approach.

src/shared/frame-rpc.js
```diff
@@ -29,6 +29,13 @@
...
 const VERSION = '1.0.0';
 
+/**
+ *  @constructor
+ *  @param {Window} src
+ *  @param {Window} dst
+ *  @param {string} origin
+ *  @param {(Object<string, ((any) => any)>) | ((any) => any)} methods
+ */
 export default function RPC(src, dst, origin, methods) {
```
robertknight, line 39: This is fine, although we don't actually make use of the non-object version of `methods` anywhere in the code, so it would be equally OK to just strip out that support.

LMS007: I'll take it out

PR #2337, opened 2020-07-14
Typecheck various components

LMS007:
Intended to convert only `annotation-body`, but the mistake may have been picking a branch near the root rather than more of a leaf component with less dependances. I assume this will go smoother as we convert more components over.

- TagList
- TagEditor
- MarkdownEditor
- Excerpt
- AutocompleteList
- AnnotationBody

Generally, I used `jsdoc-from-proptypes.js` to create the props list, but in several cases I added further specificity where appropriate for function param types.


src/sidebar/components/tag-editor.js
```diff
@@ -61,7 +91,7 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
    * reset the activeItem and open the AutocompleteList
    */
   const updateSuggestions = () => {
-    const value = inputEl.current.value.trim();
+    const value = inputEl.current ? inputEl.current.value.trim() : '';
```
LMS007, line 94: In a number of cases, I got complaints about `current` being undefined. Sure, before and after a component mounts, its possible that this value is "undefined," however in practice I don't see this actually happening unless we have callbacks firing after a component unmounts. Therefore, I don't feel it's worthwhile adding a test for this condition. Frankly, I'm not even sure how to go about writing such a test.

robertknight: There will be scenarios where we know something cannot be null/undefined for reasons that require specific domain knowledge. In TypeScript there is an "I know what I'm doing" operator (`!`) for this: `inputEl.current!.value.trim()`. In JS unfortunately you have to be more verbose, eg. `/** @type {HTMLInputElement} */ (input.current).value.trim()`, or you could create a variable if you were going to reference the object multiple times. [Related upstream issue](https://github.com/microsoft/TypeScript/issues/23405).

src/sidebar/components/annotation-body.js
```diff
@@ -11,8 +11,24 @@ import MarkdownView from './markdown-view';
 import TagEditor from './tag-editor';
 import TagList from './tag-list';
 
+/** @typedef {import("../../types/api").Annotation} Annotation */
+
+/**
+ * @typedef AnnotationBodyProps
+ * @prop {Annotation} props.annotation - The annotation in question
```
robertknight, line 18: The `props.` prefixes are not needed here since you're defining a type explicitly rather than creating an "inline" type.

src/sidebar/components/annotation-body.js
```diff
@@ -11,8 +11,24 @@ import MarkdownView from './markdown-view';
...
+ * @prop {boolean} [props.isEditing] - Whether to display the body in edit mode (if true) or view mode.
+ * @prop {(a: Object<'tags', string[]>) => void} [props.onEditTags] - Callback invoked when the user edits tags.
+ * @prop {(a?: Object<'text', string>) => void} [props.onEditText] - Callback invoked when the user edits the content of the annotation body.
```
robertknight, line 21: The types for `onEditTags` and `onEditText` look weird to me. You'd expect the callback arguments to be just a list of strings and a string respectively. Then I noticed a comment in one of the other modules that said "simplify this after we finish the Preact conversion", so this is just documenting the weirdness accurately for now.

src/sidebar/components/autocomplete-list.js
```diff
@@ -4,6 +4,23 @@ import { useMemo } from 'preact/hooks';
...
+ * @typedef AutocompleteListProps
+ * @prop {number} [activeItem] - The index of the highlighted item.
+ * @prop {string} [id] - Optional unique HTML attribute id. This can be used
+ *   for parent `aria-controls` coupling.
+ * @prop {string} [itemPrefixId] - Optional unique HTML attribute id prefix
+ *   for each item in the list. The final value of each items' id is
+ *   `{itemPrefixId}{activeItem}`
+ * @prop {(Object|string)[]} list - The list of items to render. This can be a simple
+ *   list of strings or a list of objects when used with listFormatter.
+ * @prop {(a: Object|string, b?: number) => any} [listFormatter] - An optional formatter
```
robertknight, line 17: The meaning of `a` and `b` wasn't immediately obvious here. Can you use more specific names for them.

src/sidebar/components/autocomplete-list.js
```diff
@@ -4,6 +4,23 @@ import { useMemo } from 'preact/hooks';
...
 const defaultListFormatter = item => item;
+/**
+ * @typedef AutocompleteListProps
+ * @prop {number} [activeItem] - The index of the highlighted item.
+ * @prop {string} [id] - Optional unique HTML attribute id. This can be used
+ *   for parent `aria-controls` coupling.
+ * @prop {string} [itemPrefixId] - Optional unique HTML attribute id prefix
+ *   for each item in the list. The final value of each items' id is
+ *   `{itemPrefixId}{activeItem}`
+ * @prop {(Object|string)[]} list - The list of items to render. This can be a simple
```
robertknight, line 15: If you wanted to make the connection between `list` and `listFormatter` more obvious you could use a template parameter for the object type rather than `Object`. I'd only do that if it is easy and doesn't surface a bunch of extra complications though.

src/sidebar/components/autocomplete-list.js
```diff
@@ -46,6 +64,7 @@ export default function AutocompleteList({
           }}
           {...props}
         >
+          {/* @ts-ignore TODO: fix the type for this callback */}
```
robertknight, line 67: What was the issue here?

src/sidebar/components/excerpt.js
```diff
@@ -39,12 +52,35 @@ InlineControls.propTypes = {
...
+ * @typedef ExcerptProps
+ * @prop {Object} [children]
+ * @prop {boolean} [inlineControls] - If `true`, the excerpt provides internal
+ *   controls to expand and collapse the content. If `false`, the caller sets
+ *   the collapsed state via the `collapse` prop.  When using inline controls,
+ *   the excerpt is initially collapsed.
+ * @prop {boolean} [collapse] - If the content should be truncated if its height
+ *   exceeds `collapsedHeight + overflowThreshold`.  This prop is only used if
+ *   `inlineControls` is false.
+ * @prop {number} [collapsedHeight] - Maximum height of the container, in pixels,
```
robertknight, line 65: I don't think `collapsedHeight` is really optional from a look through the code. This was a mistake in the old `propTypes` documentation. It looks to me like `children` and `collapsedHeight` should both be required props, but everything else is optional.

src/sidebar/components/hooks/use-element-should-close.js
```diff
@@ -4,14 +4,28 @@ import { normalizeKeyName } from '../../../shared/browser-compatibility-utils';
...
+ * @prop {Node} base
+ */
+
+/**
+ * @typedef {Ref<HTMLButtonElement> & Ref<PreactElement>} PreactRef
+ */
+
+/**
+ * @typedef Ref2
  * @prop current {Node} - HTML node
```
robertknight, line 22: It would be good to find a way to reduce the number of different "ref" types here.

src/sidebar/components/markdown-editor.js
```diff
@@ -147,9 +165,14 @@ ToolbarButton.propTypes = {
...
 /**
- * @typedef {string} ButtonID
- * A unique string that can be one of one of:
- * [bold, italic, quote, link, image, math, numlist, list, preview, help]
+ * @typedef {'bold'|'italic'|'quote'|'link'|'image'|'math'|'numlist'|'list'|'preview'|'help'} ButtonID
+ */
+
+/**
+ * @typedef ToolbarProps
+ *  @prop {boolean} [isPreviewing] - `true` if the editor's "Preview" mode is active.
```
robertknight, line 173: The `@prop` lines don't need to be intended here.

src/sidebar/components/markdown-editor.js
```diff
@@ -147,9 +165,14 @@ ToolbarButton.propTypes = {
...
+ *  @prop {(a: ButtonID) => any} [onCommand] - Callback invoked with the selected command when a toolbar button is clicked.
+ *  @prop {() => any} [onTogglePreview] - Callback invoked when the "Preview" toggle button is clicked.
```
robertknight, line 175: I expected to see a boolean argument to `onTogglePreview` here, but looking at the code this type is correct. It is the caller's job to figure out whether the preview is being turned on or off.

src/sidebar/components/tag-editor.js
```diff
@@ -10,19 +10,46 @@ import { normalizeKeyName } from '../../shared/browser-compatibility-utils';
...
+ */
+
+/**
+ * @template T
+ * @typedef {import("preact/hooks").Ref<T>} Ref
+ */
+
+/**
+ * @template T
+ * @typedef {import("preact/hooks").PropRef<T>} PropRef
```
robertknight, line 25: The `PropRef` name that "preact/hooks" uses is a bit confusing. From looking at the `index.d.ts` file in the "preact/hooks" package it appears that the difference is that `PropRef` is nullable whereas `Ref` is not.

src/sidebar/components/tag-editor.js
```diff
@@ -115,12 +150,16 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
     setSuggestionsListOpen(false);
     setActiveItem(-1);
 
-    inputEl.current.value = '';
-    inputEl.current.focus();
+    if (inputEl.current) {
```
robertknight, line 153: Per my comment above, I think it is OK to skip the `if` check and override the null-ness if we know the value can't actually be null.

src/sidebar/components/tag-editor.js
```diff
@@ -146,7 +187,7 @@ function TagEditor({ onEditTags, tags: tagsService, tagList }) {
    * Opens the AutocompleteList on focus if there is a value in the input
    */
   const handleFocus = () => {
-    if (inputEl.current.value.trim().length) {
+    if (inputEl.current && inputEl.current.value.trim().length) {
```
robertknight, line 190: Also here, we can override the null-ness if we know the value can't be null.

src/sidebar/markdown-commands.js
```diff
@@ -83,7 +83,7 @@ function replaceText(state, pos, length, text) {
  * Convert the selected text into a Markdown link.
  *
  * @param {EditorState} state - The current state of the input field.
- * @param {LinkType} linkType - The type of link to insert.
+ * @param {LinkType} [linkType] - The type of link to insert.
```
robertknight, line 86: Side note - There is an ES5-style default value initialization inside `convertSelectionToLink` that we could modernize.

src/sidebar/components/tag-editor.js
LMS007: roger, casting seems appropriate here

src/sidebar/components/annotation-body.js
LMS007: oops, missed that!

src/sidebar/components/annotation-body.js
LMS007: so its not a simple string or list of strings, for some reason or another. I didn't dig into this, but perhaps there is a opportunity to simplifying this down

If you look at tag-editor.js, the current code is 

`onEditTags({ tags: tagList });`

but perhaps it could just be 
`onEditTags(tagList);`

with a few simple edits.

src/sidebar/components/autocomplete-list.js
LMS007: indeed!

src/sidebar/components/autocomplete-list.js
LMS007: I think I know what you mean, and that seems reasonable. I'll give it a go

src/sidebar/components/autocomplete-list.js
LMS007: I think I fixed it by the time I finished the draft of this PR... I rather forgot. Best to remove now :)

src/sidebar/components/excerpt.js
LMS007: good catch

src/sidebar/components/hooks/use-element-should-close.js
LMS007: Sorry, this is leftover cruft from my draft, I will clean this up.

src/sidebar/components/tag-editor.js
LMS007: indeed, 

this works
```
/** @type {Ref<HTMLInputElement>} */
  const inputEl = useRef(new HTMLInputElement());
```

```
this does not work
/** @type {Ref<HTMLInputElement>} */
  const inputEl = useRef();
```

which is why I switch it to a PropRef, but the PropRef name is very poor and not obvious at all. "NullableRef" would be better.

src/sidebar/components/tag-editor.js
LMS007: yep, i agree

src/sidebar/markdown-commands.js
LMS007: done

LMS007:
@robertknight, made most of the fixups you suggested. I also still feel a need to address the Ref types in `use-element-should-close`. I'm not convinced my hack is the right approach -- although, its an approach :)
0

src/sidebar/components/tag-editor.js
LMS007: Thinking more on this
```
/** @type {Ref<HTMLInputElement|null>} */
  const inputEl = useRef(null);

```
also could work, I'm just not sure what direction we want to proceed with.

robertknight: Can I request that we use this convention for non-nullable refs in this PR:

```js
const inputEl = useRef(/** @type {HTMLInputElement|null} */ (null));
```

With this, there is no need to import `Ref`.

LMS007:
@robertknight. I was not 100% clear on the standard for the refs and state hooks before throwing up this PR last week. I have made those adjustments now. Thanks for being patient.

src/sidebar/components/tag-list.js
```diff
@@ -5,8 +5,19 @@ import propTypes from 'prop-types';
 import { isThirdPartyUser } from '../util/account-id';
 import { withServices } from '../util/service-context';
 
+/** @typedef {import('../../types/api').Annotation} Annotation */
+
+/**
+ * @typedef TagListProps
+ * @prop {Annotation} annotation - Annotation that owns the tags.
```
robertknight, line 12: Seeing this dependency is making me think we should pass in a prop indicating whether or not to render links rather than determining it locally. That's an exercise for another time.

src/sidebar/components/autocomplete-list.js
```diff
@@ -4,6 +4,24 @@ import { useMemo } from 'preact/hooks';
 import propTypes from 'prop-types';
 
 const defaultListFormatter = item => item;
+/**
```
robertknight, line 7: Style nit - Can you add a new line after `defaultListFormatter`.

src/sidebar/components/autocomplete-list.js
```diff
@@ -12,6 +30,8 @@ const defaultListFormatter = item => item;
  * used by itself.
  *
  * Modeled after the "ARIA 1.1 Combobox with Listbox Popup"
+ * @template T
```
robertknight, line 33: Style nit, can you leave a new line between the description of a function and the `@template`/`@param`  comments

src/sidebar/components/autocomplete-list.js
```diff
@@ -12,6 +30,8 @@ const defaultListFormatter = item => item;
...
+ * @param {AutocompleteListProps<T>} props
  */
 
 export default function AutocompleteList({
```
robertknight, line 37: Style nit: I don't think TS cares, but for JSDoc that refers to a specific function/variable etc. there shouldn't be a new line between the JSDoc comment and the thing it refers to.

src/sidebar/components/excerpt.js
```diff
@@ -7,9 +7,17 @@ import observeElementSize from '../util/observe-element-size';
...
+ * @typedef InlineControlsProps
+ * @prop {boolean} [isCollapsed]
+ * @prop {(collapsed: boolean) => any} [setCollapsed]
+ * @prop {Object} [linkStyle]
+ */
+
 /**
  * An optional toggle link at the bottom of an excerpt which controls whether
  * it is expanded or collapsed.
+ * @param {InlineControlsProps} props
```
robertknight, line 20: Can you add a new line above `@param` here please.

src/sidebar/components/autocomplete-list.js
LMS007: yep, thanks for pointing these out

src/sidebar/components/autocomplete-list.js
LMS007: good to know, thanks

PR #2345, opened 2020-07-16
Improve typecheck for MenuKeyboardNavigationProps

LMS007:


src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -8,27 +8,45 @@ function isElementVisible(element) {
...
+ * @param {MenuKeyboardNavigationProps} props
  */
 export default function MenuKeyboardNavigation({
   className,
   closeMenu,
   children,
   visible,
 }) {
-  const menuRef = useRef(null);
+  const menuRef = /** @type {Ref<HTMLDivElement|null>} */ (useRef(null));
```
LMS007, line 39: @robertknight , the type needs to be div since its being added to a div element. I was not sure if you're example (in slack) was just using the HTMLFormElement as an example or literally everywhere? I'm assuming it was just an example.

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -8,27 +8,45 @@ function isElementVisible(element) {
...
 
   useEffect(() => {
     let focusTimer = null;
     if (visible) {
       focusTimer = setTimeout(() => {
         // The focus won't work without delaying rendering.
-        const firstItem = menuRef.current.querySelector('[role^="menuitem"]');
+        /** @type {HTMLFormElement|null} */
+        const firstItem = /** @type {Ref<HTMLDivElement>} */ (menuRef).current.querySelector(
```
LMS007, line 47: Still needing to cast everywhere since it does not know this is not null :\ maybe there is a better way to do this?

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -8,27 +8,45 @@ function isElementVisible(element) {
...
+ * @template T
+ * @typedef {import("preact/hooks").Ref<T>} Ref
+ */
+
+/**
+ * @typedef MenuKeyboardNavigationProps
+ * @prop {string} [className]
+ * @prop {(e: KeyboardEvent) => any} [closeMenu] - Callback when the menu is closed via keyboard input
+ * @prop {boolean} [visible] - When  true`, sets focus on the first item in the list
+ * @prop {Object} children - Array of nodes which may contain <MenuItems> or any nodes
```
LMS007, line 21: scripts/jsdoc-from-proptypes.js has this going to an `Object`, but I do wonder if there is a better type for children.

`{Element[]|Element}` seems a bit closer, but then again, maybe this is all superficial and irrelevant :)

src/sidebar/components/menu-keyboard-navigation.js
robertknight: Hi Kyle, the element type depends on the use. Depending on the use you may just be able to use `HTMLElement`.

Note in my Slack example that the cast is to the `useRef` _argument_ (ie. `useRef(/** @type {HTMLDivElement|null} / (null))`). Then everything else is inferred and you don’t need to import the Ref type. The need for null checks elsewhere is also currently avoided if you do this.

src/sidebar/components/menu-keyboard-navigation.js
robertknight: See my note above about specifying the type by casting the `useRef` argument.

src/sidebar/components/menu-keyboard-navigation.js
robertknight: Using `Object` for children is fine here I think. You can do something more specific but it would be a bit verbose. Also, it is important not to mix up the DOM Element type with Preact elements (`import(“preact”).JSX.Element`)

src/sidebar/components/menu-keyboard-navigation.js
LMS007: @robertknight , I missed that in your example, sorry. That makes more sense. Don't bother with refs at all, just assert the type of thing in the ref and the rest is easy. got it!

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -42,6 +57,7 @@ export default function MenuKeyboardNavigation({
 
   const onKeyDown = event => {
     const menuItems = Array.from(
+      /** @ts-ignore - assume current is not null */
```
LMS007, line 60: It would not allow a cast here... no "overlap" between null and HTMLElement prevented that. I think what it wants is a simple condition on `current`, but I think we can safely assert `current` is not null here so I added the `ignore`

robertknight: Ah - subtle problem here. You're missing the parens around the `null`  in the `useRef` call. If you add those, you don't need the `@ts-ignore` here. You can also remove the `Ref` import at the top of this file now.

In other words:

```diff
- const menuRef = useRef(/** @type {HTMLElement|null} */ null);
+ const menuRef = useRef(/** @type {HTMLElement|null} */ (null));
```

I appreciate it is non-obvious here that the "null-ness" is ignored whenever you access `menuRef.current`, but if we can accept this quirk for now and use it consistently, it will be easy to update everywhere in future if we find something better.

As an aside, in TS syntax (ie. not constrained to JS syntax) it is somewhat easier as you can do something like:

```ts
const aRef = useRef<HTMLElement>();
aRef.current!.querySelector(...) // `!` means "trust me, I know what I'm doing"
```

LMS007: ya, TS can handle that a bit more gracefully, 

**facepalm** on the parens

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -42,7 +57,8 @@ export default function MenuKeyboardNavigation({
 
   const onKeyDown = event => {
     const menuItems = Array.from(
-      menuRef.current.querySelectorAll('[role^="menuitem"]')
+      /** @type {NodeListOf<HTMLElement>} */
```
LMS007, line 60: I was seeing an error down below ...
`src/sidebar/components/menu-keyboard-navigation.js:105:33 - error TS2339: Property 'focus' does not exist on type 'Element'.`

I could either cast here, or there, but I believe that  querySelectorAll returns an Element array, but it needs to be an HTMLElement (or similar).

src/sidebar/components/menu-keyboard-navigation.js
```diff
@@ -8,27 +8,42 @@ function isElementVisible(element) {
   return element.offsetParent !== null;
 }
 
+/**
+ * @template T
+ * @typedef {import("preact/hooks").Ref<T>} Ref
```
robertknight, line 13: This `@typedef` is not needed any more due to the changed way that `useRef` is annotated.

PR #2347, opened 2020-07-16
Improve typechecking for Stores

LMS007:
Add SidebarStore type. This is an aggregation of the base ReduxStore + each one of our stores baked in. The fist store added here is simply the Activity store (as a first example)

There are some assumptions when making store types:

1. It is assumed they are transformed in `createStore` where the selectors and actions are flattened.

2. Some meta data (namely, the init method, namespace and update method) are refactored or removed.

These transformations are implied through the typing which explicitly ignores these methods and structure of the raw store before the transformation in create-store.

--------------

This is just a subset of the larger PR (no changes)
https://github.com/hypothesis/client/pull/2318

### DIFF
--- a/src/sidebar/store/create-store.js
+++ b/src/sidebar/store/create-store.js
@@ -8,19 +8,6 @@ import immutable from '../util/immutable';
 import { createReducer, bindSelectors } from './util';
 
 /**
- * // Base redux store
- * @typedef {import("redux").Store} ReduxStore
- *
- * // Custom stores
- * @typedef {import("./modules/activity").ActivityStore} ActivityStore
- * // TODO: add more stores
- *
- * // Combine all stores
- * @typedef {ReduxStore &
- *  ActivityStore} SidebarStore
- */
-
-/**
  * Create a Redux store from a set of _modules_.
  *
  * Each module defines the logic related to a particular piece of the application
@@ -37,8 +24,7 @@ import { createReducer, bindSelectors } from './util';
  *
  * @param {Object[]} modules
  * @param {any[]} [initArgs] - Arguments to pass to each state module's `init` function
- * @param {any[]} [middleware] - List of additional Redux middlewares to use.
- * @return {SidebarStore}
+ * @param {any[]} [middleware] - List of additional Redux middlewares to use
  */
 export default function createStore(modules, initArgs = [], middleware = []) {
   // Create the initial state and state update function.
@@ -102,5 +88,5 @@ export default function createStore(modules, initArgs = [], middleware = []) {
 
   Object.assign(store, boundActions, boundSelectors);
 
-  return /** @type {SidebarStore} */ (store);
+  return store;
 }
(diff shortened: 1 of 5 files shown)


src/sidebar/store/create-store.js
```diff
@@ -78,7 +92,11 @@ export default function createStore(modules, initArgs = [], middleware = []) {
   }
 
   // Create the store.
-  const store = redux.createStore(reducer, initialState, enhancer);
+  const store = /** @type {SidebarStore} */ (redux.createStore(
```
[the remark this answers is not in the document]
LMS007: agreed.

src/sidebar/store/index.js
```diff
@@ -54,6 +54,9 @@ import viewer from './modules/viewer';
  * the individual state modules. ie. `store.actionName(args)` dispatches an
  * action through the store and `store.selectorName(args)` invokes a selector
  * passing the current state of the store.
+ *
+ * @param {import('../../types/config').SidebarConfig} settings
+ * @return {import('.create-store').HypothesisStore}
```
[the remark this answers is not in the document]
LMS007: good catch! thanks

src/sidebar/store/create-store.js
```diff
@@ -7,6 +7,19 @@ import immutable from '../util/immutable';
...
+ * // Base redux store
+ * @typedef {import("redux").Store} ReduxStore
+ *
+ * // Custom stores
+ * @typedef {import("./modules/activity").ActivityStore} ActivityStore
+ * // TODO: add more stores
+ *
+ * // Combine all stores
+ * @typedef {ReduxStore &
+ *  ActivityStore} SidebarStore
```
robertknight, line 20: Ah - I've realized that these types should be moved to `store/index.js` because the set of modules that the store has is determined by the `store` function in `store/index.js` whereas `create-store.js` is a generic utility.

src/sidebar/store/create-store.js
```diff
@@ -87,5 +101,5 @@ export default function createStore(modules, initArgs = [], middleware = []) {
 
   Object.assign(store, boundActions, boundSelectors);
 
-  return store;
+  return /** @type {SidebarStore} */ (store);
```
robertknight, line 104: Re-reading this PR I think it would make sense to move the `SidebarStore` type and cast to the `store` function, since `SidebarStore` includes all the specific modules in the sidebar's store whereas `createStore` is a utility where the type depends on which modules are passed as input.

src/sidebar/store/create-store.js
LMS007: yes, that makes more sense

src/sidebar/store/use-store.js
```diff
@@ -6,14 +6,14 @@ import shallowEqual from 'shallowequal';
 import warnOnce from '../../shared/warn-once';
 import { useService } from '../util/service-context';
 
-/**
- * @typedef {import("redux").Store} Store
- */
+/** @typedef {import("redux").Store} Store */
+
+/** @typedef {import("./index").SidebarStore} SidebarStore */
```
robertknight, line 11: I believe TS will still parse this correctly these two type imports were folded into a single `/** ... */` block.

PR #2349, opened 2020-07-16
Improve typecheck for MenuSection

LMS007:


src/sidebar/components/menu-section.js
```diff
@@ -15,14 +23,16 @@ import propTypes from 'prop-types';
...
+ * @param {MenuSectionProps} props
  */
 export default function MenuSection({ heading, children }) {
   return (
     <Fragment>
       {heading && <h2 className="menu-section__heading">{heading}</h2>}
       <ul className="menu-section__content">
         {toChildArray(children).map(child => (
-          <li key={child.key}>{child}</li>
+          <li key={/** @type {VNode} **/ (child).key}>{child}</li>
```
LMS007, line 35: I could see a `ts-ignore` here, but forcing this to a VNode<any> also works. I tired a few other types, namely ComponentChild , ComponentChildren, but they were having issues. I think VNode seems reasonable, but it needs the `any` generic passed into it :\

robertknight: `JSX.Element` is an alias for `VNode<any>`. My understanding is that `JSX.Element` is what TS directly uses and this will be an alias for the appropriate type in whatever UI library you are using (Preact, React, Vue etc.).

LMS007: rgr: I'll change to JSXElement

PR #2382, opened 2020-07-27
Improve typechecking (annotation-action-bar, annotation-document-info)

LMS007:
- add props for annotation-action-bar
- add props for annotation-document-info
- dependency fixes:
  - add selector types for groups stores
  - fixup errors discovered in annotation-share-control
  - fixup errors in util/permissions

### DIFF
--- a/src/sidebar/components/annotation-action-bar.js
+++ b/src/sidebar/components/annotation-action-bar.js
@@ -10,9 +10,24 @@ import { withServices } from '../util/service-context';
 import AnnotationShareControl from './annotation-share-control';
 import Button from './button';
 
+/** @typedef {import("../../types/api").Annotation} Annotation */
+
+/** @typedef {import('../../types/config').HostConfig} HostConfig */
+
+/**
+ * @typedef AnnotationActionBarProps
+ * @prop {Annotation} annotation - The annotation in question
+ * @prop {() => any} onReply - Callbacks for when action buttons are clicked/tapped
+ * @prop {Object} annotationsService
+ * @prop {HostConfig} settings
+ * @prop {Object} toastMessenger
+ */
+
 /**
  * A collection of `Button`s in the footer area of an annotation that take
  * actions on the annotation.
+ *
+ * @param {AnnotationActionBarProps} props
  */
 function AnnotationActionBar({
   annotation,
@@ -67,12 +82,12 @@ function AnnotationActionBar({
       .catch(() => toastMessenger.error('Flagging annotation failed'));
   };
 
-  const onReplyClick = event => {
+  const onReplyClick = () => {
     if (!isLoggedIn) {
       openSidebarPanel(uiConstants.PANEL_LOGIN_PROMPT);
       return;
     }
-    onReply(event);
+    onReply();
   };
 
   return (
@@ -109,10 +124,7 @@ function AnnotationActionBar({
 
 AnnotationActionBar.propTypes = {
   annotation: propTypes.object.isRequired,
-  /** Callbacks for when action buttons are clicked/tapped */
   onReply: propTypes.func.isRequired,
-
-  // Injected services
   annotationsService: propTypes.object.isRequired,
   settings: propTypes.object.isRequired,
   toastMessenger: propTypes.object.isRequired,
--- a/src/sidebar/components/annotation-share-control.js
+++ b/src/sidebar/components/annotation-share-control.js
@@ -21,12 +21,9 @@ function AnnotationShareControl({
   group,
   shareUri,
 }) {
-  const annotationIsPrivate = isPrivate(
-    annotation.permissions,
-    annotation.user
-  );
-  const shareRef = useRef();
-  const inputRef = useRef();
+  const annotationIsPrivate = isPrivate(annotation.permissions);
+  const shareRef = useRef(/** @type {HTMLDivElement|null} */ (null));
+  const inputRef = useRef(/** @type {HTMLInputElement|null} */ (null));
 
   const [isOpen, setOpen] = useState(false);
   const wasOpen = useRef(isOpen);
--- a/src/sidebar/store/modules/groups.js
+++ b/src/sidebar/store/modules/groups.js
@@ -206,7 +206,6 @@ const getCurrentlyViewingGroups = createSelector(
  *
  * // Selectors
  * @prop {() => Group[]} allGroups
- * @prop {() => Group[]} allGroups
  * @prop {() => Group|undefined|null} focusedGroup
  * @prop {() => string|null} focusedGroupId
  * @prop {() => Group[]} getFeaturedGroups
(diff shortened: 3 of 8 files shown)


src/sidebar/components/annotation-action-bar.js
```diff
@@ -10,9 +10,22 @@ import { withServices } from '../util/service-context';
 import AnnotationShareControl from './annotation-share-control';
 import Button from './button';
 
+/** @typedef {import("../../types/api").Annotation} Annotation */
+
+/**
+ * @typedef AnnotationActionBarProps
+ * @prop {Annotation} annotation - The annotation in question
+ * @prop {(Event) => any} onReply - Callbacks for when action buttons are clicked/tapped
```
robertknight, line 18: `Event` is interpreted as a parameter name rather than type here. In other words, it behaves like `(Event: any) => any)`. I suggest `(e: Event) => any` instead.

src/sidebar/components/annotation-action-bar.js
```diff
@@ -10,9 +10,22 @@ import { withServices } from '../util/service-context';
...
+ * @prop {Object} annotationsService
+ * @prop {Object} settings
+ * @prop {Object} toastMessenger
+ */
+
 /**
  * A collection of `Button`s in the footer area of an annotation that take
  * actions on the annotation.
+ *
+ * @param {annotation} props
```
robertknight, line 28: The type is wrong here. This should be:

```suggestion
 * @param {AnnotationActionBarProps} props
```

I don't know why this doesn't result in an error, but the same thing appears to happen if you use any property name from the destructuring assignment (eg. `annotation`, `onReply` ...).

Making this change flags up an issue with the `flagged` prop missing from the `Annotation` type.

src/sidebar/components/annotation-action-bar.js
```diff
@@ -77,10 +90,15 @@ function AnnotationActionBar({
 
   return (
     <div className="annotation-action-bar u-layout-row">
-      {showEditAction && <Button icon="edit" title="Edit" onClick={onEdit} />}
+      {showEditAction && (
+        /** @ts-ignore - Fix Button component type */
+        <Button icon="edit" title="Edit" onClick={onEdit} />
```
robertknight, line 95: Given how widely `Button` is used, can I suggest that you update the JSDoc on that component first and then rebase this?

src/sidebar/store/modules/groups.js
```diff
@@ -206,7 +206,11 @@ const getCurrentlyViewingGroups = createSelector(
  *
  * // Selectors
  * @prop {() => Group[]} allGroups
- * // TODO: add rest ...
+ * @prop {() => Group|undefined|null} focusedGroup
```
robertknight, line 209: Separate from this PR, we should fix the issue where `focusedGroup` and `getGroup` can return different falsey values. In practice it doesn't matter that much, but it still creates a small possibility of a bug.

src/sidebar/components/annotation-action-bar.js
```diff
@@ -10,9 +10,22 @@ import { withServices } from '../util/service-context';
...
 import Button from './button';
 
+/** @typedef {import("../../types/api").Annotation} Annotation */
+
+/**
+ * @typedef AnnotationActionBarProps
+ * @prop {Annotation} annotation - The annotation in question
+ * @prop {(Event) => any} onReply - Callbacks for when action buttons are clicked/tapped
+ * @prop {Object} annotationsService
+ * @prop {Object} settings
```
robertknight, line 20: `settings` could use the `HostConfig` type. See the definition of the `isShareable` function.

src/sidebar/components/annotation-action-bar.js
LMS007: thats so odd it didnt fail :\

src/sidebar/store/modules/groups.js
LMS007: Ya, I noticed that, there wa a mis-match there.

src/sidebar/components/annotation-action-bar.js
LMS007: thanks!

src/sidebar/components/annotation-action-bar.js
```diff
@@ -101,6 +118,7 @@ function AnnotationActionBar({
           isPressed={true}
           icon="flag--active"
           title="Annotation has been reported to the moderators"
+          onClick={()=>{}}
```
LMS007, line 121: Funny, we just talked about this. I guess this may be the one and only place we don't have an `onClick`, but it's also something that needs to be fixed. We're going to eventually have to add support for un-flagged an annotation. For now I just added a no-op which seems reasonable. I hesitate to make `onClick` not required. Thoughts?

src/sidebar/components/annotation-action-bar.js
```diff
@@ -10,9 +10,24 @@ import { withServices } from '../util/service-context';
...
 import Button from './button';
 
+/** @typedef {import("../../types/api").Annotation} Annotation */
+
+/** @typedef {import('../../types/config').HostConfig} HostConfig */
+
+/**
+ * @typedef AnnotationActionBarProps
+ * @prop {Annotation} annotation - The annotation in question
+ * @prop {(e: Event) => any} onReply - Callbacks for when action buttons are clicked/tapped
```
robertknight, line 20: The `onReply` callback is used only in one place and the `Event` argument is not used, so we could potentially remove that argument for the time being.

src/sidebar/components/annotation-action-bar.js
robertknight: I agree, passing a no-op callback here seems reasonable. In almost all cases having a `Button` without an `onClick` would be a mistake.

src/sidebar/components/annotation-share-control.js
```diff
@@ -87,6 +84,7 @@ function AnnotationShareControl({
 
   return (
     <div className="annotation-share-control" ref={shareRef}>
+      {/** @ts-ignore - Fix Button component type */}
```
robertknight, line 87: Can this `@ts-ignore` be removed now?

src/sidebar/components/annotation-share-control.js
```diff
@@ -110,6 +108,7 @@ function AnnotationShareControl({
                 readOnly
                 ref={inputRef}
               />
+              {/** @ts-ignore - Fix Button component type */}
```
robertknight, line 111: Can this `@ts-ignore` be removed now?

src/sidebar/util/permissions.js
```diff
@@ -98,9 +98,9 @@ export function isPrivate(perms) {
  *
  * @param {Permissions} perms
  * @param {'update'|'delete'} action
- * @param {string} userid
+ * @param {string|null} userid
  * @return {boolean}
  */
 export function permits(perms, action, userid) {
-  return perms[action].indexOf(userid) !== -1;
+  return perms[action].indexOf(userid ? userid : '') !== -1;
```
robertknight, line 105: This is fine, although I believe a slightly more common pattern in the code is to use `userid || ''`.

LMS007: I'll switch it

PR #2390, opened 2020-07-29
Typechecking for (sort-menu, user-menu)

LMS007:


src/sidebar/components/user-menu.js
```diff
@@ -11,15 +11,32 @@ import MenuItem from './menu-item';
...
  *
  * This menu will contain different items depending on service configuration,
  * context and whether the user is first- or third-party.
+ *
+ * @param {UserMenuProps} props
  */
 function UserMenu({ auth, bridge, onLogout, serviceUrl, settings }) {
   const isThirdParty = isThirdPartyUser(auth.userid, settings.authDomain);
-  const service = serviceConfig(settings);
+  const service = serviceConfig(/** @type {HostConfig} */ (settings));
```
LMS007, line 39: I'm forced to cast here, but I think this is a mistake. I'm pretty sure serviceConfig() only cares about a HostConfig object upon inspecting it, but this is clearly a SidebarConfig object.

src/types/auth.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Type definitions for authentication types
+ *
+ * TODO: this may move to local component files once they are correctly typed
+ */
+
+/**
+ * @typedef AuthState
```
LMS007, line 8: I want to put this in `hypothesis-app.js`, but that file is root level and comes with a mountain of typecheck errors. We'll get there...

src/sidebar/components/user-menu.js
```diff
@@ -11,15 +11,32 @@ import MenuItem from './menu-item';
...
+/** @typedef {import('../../types/config').HostConfig} HostConfig */
+
+/** @typedef {import('../../types/auth').AuthState} AuthState */
+
+/**
+ * @typedef UserMenuProps
+ * @prop {AuthState} auth - object representing authenticated user and auth status
+ * @prop {() => any} onLogout - onClick callback for the "log out" button
+ * @prop {Object} bridge
+ * @prop {(name: string, params?: Object<string, string>) => any} serviceUrl
```
robertknight, line 25: There is a `ServiceUrlGetter` type in `src/sidebar/services/service-url.js` that you can use here for the `serviceUrl` prop.

src/sidebar/components/user-menu.js
```diff
@@ -11,15 +11,32 @@ import MenuItem from './menu-item';
 import MenuSection from './menu-section';
 import SvgIcon from '../../shared/components/svg-icon';
 
+/** @typedef {import('../../types/config').SidebarConfig} SidebarConfig */
+
+/** @typedef {import('../../types/config').HostConfig} HostConfig */
+
+/** @typedef {import('../../types/auth').AuthState} AuthState */
```
robertknight, line 18:
```suggestion
/**
 * @typedef {import('../../types/config').SidebarConfig} SidebarConfig
 * @typedef {import('../../types/config').HostConfig} HostConfig
 * @typedef {import('../../types/auth').AuthState} AuthState 
 */
```

Elsewhere I've taken the approach of combining type imports into a single `/** .. */` comment block to reduce visual noise and the number of empty lines.

src/types/auth.js
robertknight: The `authStateFromProfile` function in `hypothesis-app.js` could also be moved to the `session` module in the store as a selector in future.

Can I suggest moving this type to `user-menu.js` in the meantime and then just importing it from there in other components?

src/sidebar/components/user-menu.js
robertknight: Have a read through the comments in `types/config.js`. The `settings` object's type is `MergedConfig`, which is a combination of `SidebarConfig` and `HostConfig`, rather than `SidebarConfig`.

LMS007: thats exactly what I was looking for, just forgot about it!

src/sidebar/components/user-menu.js
LMS007: I like this, thanks

PR #2392, opened 2020-07-29
Improve typecheck (focus-mode-header, store/selection)

LMS007:


src/sidebar/store/modules/selection.js
```diff
@@ -463,6 +463,9 @@ function focusModeUserPrettyName(state) {
   return state.focusMode.user.displayName;
 }
 
+/**
+ * @return {string[]}
```
robertknight, line 467: `@return` doesn't work with variable declarations in general AFAIK, although it may work if the variable declaration is assigned to a closure (please check, I haven't tested this). Instead you need to use `@type`, eg. `@type {(state: any) => string[]}`.

LMS007: awe... I got confused because a few others here were using @return. you're right, they need to be changed into a function type with a return value.

PR #2393, opened 2020-07-29
Improve typecheck (logged-out-message, login-prompt-panel, sidebar-panel)

LMS007:


src/sidebar/components/logged-out-message.js
```diff
@@ -6,10 +6,18 @@ import { withServices } from '../util/service-context';
 import Button from './button';
 import SvgIcon from '../../shared/components/svg-icon';
 
+/**
+ * @typedef LoggedOutMessageProps
+ * @prop {() => any} onLogin
+ * @prop {(name: string) => any} serviceUrl
```
robertknight, line 12: Can you use the `ServiceUrlGetter` type from `src/sidebar/services/service-url.js` here. This will make it easier to keep in sync if it ever changes.

src/sidebar/components/sidebar-panel.js
```diff
@@ -9,12 +9,26 @@ import Button from './button';
 import Slider from './slider';
 import SvgIcon from '../../shared/components/svg-icon';
 
+/**
+ * @typedef SidebarPanelProps
+ * @prop {Object} [children]
+ * @prop {string} [icon] - An optional icon name for display next to the panel's title
+ * @prop {string} panelName
+ *    - A string identifying this panel. Only one `panelName` may be active at any time.
+ *    Multiple panels with the same `panelName` would be "in sync", opening and closing together.
```
robertknight, line 18:
```suggestion
 * @prop {string} panelName -
 *    A string identifying this panel. Only one `panelName` may be active at any time.
 *    Multiple panels with the same `panelName` would be "in sync", opening and closing together.
```

Small convention nit - when prop docs are moved to multiple lines the ` - ` should still go after the name of the prop. Come to think of it, we could just omit it entirely in this case, assuming it doesn't cause any errors for TS's parsing, as the parameter name is already visually separated from the comments by a new line.

LMS007: If its okay with you, I'll leave it for now, but move it to the correct place. I like the separator.

PR #2394, opened 2020-07-30
Improve typecheck (moderation-banner, timestamp)

LMS007:


src/sidebar/components/moderation-banner.js
```diff
@@ -64,7 +79,7 @@ function ModerationBanner({ annotation, api, toastMessenger }) {
   })();
 
   const bannerClasses = classnames('moderation-banner', {
-    'is-flagged': flagCount > 0,
+    'is-flagged': /** @type {number} */ (flagCount) > 0,
```
robertknight, line 82: This works because `flagCount` is `null` if this information is missing and `Number(null)` is `0`. However I think it would be more correct to explicitly account for the possibility that the flag count is unknown using `flagCount !== null && flagCount > 0`.

PR #2395, opened 2020-07-30
Improve typecheck (annotation-publish-control, annotation-share-control)

LMS007:


src/sidebar/components/annotation-publish-control.js
```diff
@@ -34,7 +49,12 @@ function AnnotationPublishControl({
 
   const isPrivate = draft ? draft.isPrivate : !isShared(annotation.permissions);
 
-  const publishDestination = isPrivate ? 'Only Me' : group.name;
```
robertknight, line 37: Since you can't post an annotation if there is no focused group, I think it would be easier just to render nothing in this case, by adding the following after the last hook call:

```js
  if (!group) {
    return null;
  }
```

The lazier option would be to just assume that the group is successfully fetched and cast the result of the `useStore(store => store.getGroup(...))` call to `Group`. This does create the risk of a runtime error if for any reason this fails though!

robertknight: Looking at https://github.com/hypothesis/client/issues/1542 there was an issue in the past, that may still be there, where an annotation card did somehow end up being rendered when there was no focused group. If that issue still exists, then clicking the "Edit" button in that state would cause an exception here.

src/sidebar/components/annotation-publish-control.js
```diff
@@ -32,6 +47,12 @@ function AnnotationPublishControl({
   const setDefault = useStore(store => store.setDefault);
   const removeAnnotations = useStore(store => store.removeAnnotations);
 
+  if (!group) {
+    // If there is no group, don't render anything. A annotation does not need
+    // controls if it does not belong to a group.
```
robertknight, line 52: The reason `group` is null may be different than what is mentioned in this group. `annotation.group` may be a valid group ID, but it may not be loaded, in which case `store.getGroup(...)` will return undefined/null. In that case we can't tell if the group ID is valid or not.

src/sidebar/components/annotation-share-control.js
```diff
@@ -11,8 +11,27 @@ import useElementShouldClose from './hooks/use-element-should-close';
...
+/**
+ * @typedef {import('../../types/api').Annotation} Annotation
+ * @typedef {import('../../types/api').Group} Group
+ */
+
+/**
+ * @typedef AnnotationShareControlProps
+ * @prop {Annotation} annotation - The annotation in question
+ * @prop {Group} [group] -
+ *  Group that the annotation is in If missing, this component will not render.
```
robertknight, line 23:
```suggestion
 *  Group that the annotation is in. If missing, this component will not render.
```

src/sidebar/components/test/annotation-publish-control-test.js
```diff
@@ -70,6 +70,12 @@ describe('AnnotationPublishControl', () => {
     $imports.$restore();
   });
 
+  it('should not render if group is missing', () => {
+    fakeStore.getGroup.returns(undefined);
+    const wrapper = createAnnotationPublishControl();
+    assert.isFalse(wrapper.find('annotation-publish-control').exists());
```
robertknight, line 76:
```suggestion
    assert.isFalse(wrapper.find('.annotation-publish-control').exists());
```

The CSS selector used here is an element selector, whereas you meant to match the `annotation-publish-control` _class_.

src/sidebar/components/annotation-publish-control.js
LMS007: I'll make the comment change

src/sidebar/components/test/annotation-publish-control-test.js
LMS007: this is a total face palm

PR #2396, opened 2020-07-30
Various fixes for store modules

LMS007:
- Fix typing for createSelector() wrapped selectors. (use @type, not @return)
- Fix `FrameInfo` type and use it. Previously it was orphaned.

### DIFF
--- a/src/sidebar/services/frame-sync.js
+++ b/src/sidebar/services/frame-sync.js
@@ -7,15 +7,6 @@ import * as metadata from '../util/annotation-metadata';
 import { watch } from '../util/watch';
 
 /**
- * @typedef FrameInfo
- * @property {string} uri - Current primary URI of the document being displayed
- * @property {string[]} searchUris - List of URIs that should be passed to the
- *           search API when searching for annotations on this document.
- * @property {string} documentFingerprint - Fingerprint of the document, used
- *                    for PDFs
- */
-
-/**
  * Return a minimal representation of an annotation that can be sent from the
  * sidebar app to a connected frame.
  *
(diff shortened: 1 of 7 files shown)


src/sidebar/store/modules/frames.js
```diff
@@ -7,6 +7,10 @@ import shallowEqual from 'shallowequal';
 
 import * as util from '../util';
 
+/**
+ * @typedef {import('../../services/frame-sync').Frame} Frame
```
robertknight, line 11: Can you move the typedef from `services/frame-sync` to this module to keep the overall dependency structure where modules in `services` depend on modules in `store` but not vice-versa.

src/sidebar/services/frame-sync.js
```diff
@@ -7,12 +7,14 @@ import * as metadata from '../util/annotation-metadata';
 import { watch } from '../util/watch';
 
 /**
- * @typedef FrameInfo
+ * @typedef {import('../../types/annotator').DocumentMetadata} DocumentMetadata
+ */
+
+/**
+ * @typedef Frame
+ * @property {string|null} id
```
robertknight, line 15: I think it would be useful to add a comment explaining what it means for a frame to have an ID vs no ID. That information is currently buried away in a comment: https://github.com/hypothesis/client/blob/240cb4781d43ffa19e2ebac9ec7f3b683d2ff38c/src/sidebar/store/modules/frames.js#L99

src/sidebar/store/modules/frames.js
```diff
@@ -40,20 +44,27 @@ const actions = util.actionTypes(update);
...
  */
 function destroyFrame(frame) {
   return { type: actions.DESTROY_FRAME, frame: frame };
 }
 
 /**
  * Update the `isAnnotationFetchComplete` flag of the frame.
+ *
+ * @param {string} uri
+ * @param {boolean} status
```
robertknight, line 67: This is a pre-existing issue, but `status` is not a great name for a boolean variable. `isFetchComplete` would be better.

PR #2409, opened 2020-08-04
Typecheck most remaining JS files in src/annotator

robertknight:
This PR enables typechecking for most of the remaining JS files in `src/annotator`, fixing errors along the way, with a few exceptions for modules that will require more work. The entry-point `index.js` file is also excluded since that imports many CoffeeScript files that cannot be checked.

There are a few common kinds of change here:

- Casting a generic possibly-null DOM type to a specific subtype (eg. `Node` => `Element`) because we know this is safe but TS doesn't
- Changing implicit int/float => string conversions to explicit ones (eg. when calling `Element.setAttribute`)
- Telling TS to ignore a few uses of deprecated parameters which we needed to set for IE 11 compatibility but which no longer exist according to MDN or TS's DOM type definitions. We could go ahead and remove these now since we've dropped IE 11 support, but I wanted to avoid making any functional changes in this PR

### DIFF
--- a/src/annotator/adder.js
+++ b/src/annotator/adder.js
@@ -41,7 +41,7 @@ const ARROW_H_MARGIN = 20;
  * @return {Element}
  */
 function nearestPositionedAncestor(el) {
-  let parentEl = el.parentElement;
+  let parentEl = /** @type {Element} */ (el.parentElement);
   while (parentEl.parentElement) {
     if (getComputedStyle(parentEl).position !== 'static') {
       break;
@@ -95,15 +95,23 @@ export class Adder {
       zIndex: 999,
     });
 
-    this._view = container.ownerDocument.defaultView;
+    this._view = /** @type {Window} */ (container.ownerDocument.defaultView);
 
-    this._width = () =>
-      this._shadowRoot.firstChild.getBoundingClientRect().width;
-    this._height = () =>
-      this._shadowRoot.firstChild.getBoundingClientRect().height;
+    this._width = () => {
+      const firstChild = /** @type {Element} */ (this._shadowRoot.firstChild);
+      return firstChild.getBoundingClientRect().width;
+    };
+
+    this._height = () => {
+      const firstChild = /** @type {Element} */ (this._shadowRoot.firstChild);
+      return firstChild.getBoundingClientRect().height;
+    };
 
     this._isVisible = false;
+
+    /** @type {'up'|'down'} */
     this._arrowDirection = 'up';
+
     this._onAnnotate = options.onAnnotate;
     this._onHighlight = options.onHighlight;
     this._onShowAnnotations = options.onShowAnnotations;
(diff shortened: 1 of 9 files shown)


src/annotator/adder.js
```diff
@@ -130,7 +138,7 @@ export class Adder {
    * Return the best position to show the adder in order to target the
    * selected text in `targetRect`.
    *
-   * @param {Rect} targetRect - The rect of text to target, in viewport
+   * @param {DOMRect} targetRect - The rect of text to target, in viewport
```
robertknight, line 141: The `Rect` type wasn't defined anywhere. [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) the name of the type returned by various DOM APIs that return rects.

src/annotator/anchoring/text-position.js
```diff
@@ -28,19 +28,21 @@ export function toRange(root, start, end) {
...
+    // @ts-ignore - Ignored parameter required for IE 11
     false // expandEntityReferences
   );
 
   let startContainer;
-  let startOffset;
+  let startOffset = 0;
   let endContainer;
-  let endOffset;
+  let endOffset = 0;
```
robertknight, line 39: This makes it clearer to TS that `startOffset` and `endOffset` will be numbers. They should always get set to numbers before the `while` loop below exits, but TS can't figure that out.

src/annotator/highlighter.js
```diff
@@ -210,19 +237,19 @@ export function highlightRange(normedRange, cssClass = 'hypothesis-highlight') {
  *
  * nb. This is like `ChildNode.replaceWith` but it works in IE 11.
  *
- * @param {Node} node
+ * @param {ChildNode} node
```
robertknight, line 240: The `remove()` method used in this function only exists on `ChildNode`-y nodes (eg. `Element`s)

src/annotator/highlighter.js
```diff
@@ -238,6 +265,8 @@ export function removeHighlights(highlights) {
 }
 
 /**
+ * Subset of `DOMRect` interface.
```
robertknight, line 268: We could potentially rewrite the `getBoundingClientRect` method below to just return the standard `DOMRect` type in future rather than have this additional `DOMRect`-like object.

src/annotator/highlighter.js
LMS007: 👍

src/annotator/highlighter.js
```diff
@@ -210,19 +237,19 @@ export function highlightRange(normedRange, cssClass = 'hypothesis-highlight') {
  *
  * nb. This is like `ChildNode.replaceWith` but it works in IE 11.
  *
- * @param {Node} node
+ * @param {ChildNode} node
  * @param {Node[]} replacements
  */
 function replaceWith(node, replacements) {
-  const parent = node.parentNode;
+  const parent = /** @type {Node} */ (node.parentNode);
```
LMS007, line 244: does this need a null check or is a cast okay?

robertknight: In this case I believe it is safe to assume the node has a parent.

PR #2411, opened 2020-08-05
Improve typechecking

LMS007:
- new-note-btn.js
- search-status-bar.js
- selection-tabs.js
- annotation-action-bar.js
- annotation-quote.js

### DIFF
--- a/src/sidebar/components/new-note-btn.js
+++ b/src/sidebar/components/new-note-btn.js
@@ -8,6 +8,16 @@ import { applyTheme } from '../util/theme';
 
 import Button from './button';
 
+/**
+ * @typedef {import('../../types/config').MergedConfig} MergedConfig
+ */
+
+/**
+ * @typedef NewNoteButtonProps
+ * @prop {Object} annotationsService - Injected service.
+ * @prop {MergedConfig} settings - Injected service.
+ */
+
 function NewNoteButton({ annotationsService, settings }) {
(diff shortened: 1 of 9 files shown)


src/sidebar/components/new-note-btn.js
```diff
@@ -21,9 +31,9 @@ function NewNoteButton({ annotationsService, settings }) {
     }
     const annot = {
       target: [],
-      uri: topLevelFrame.uri,
+      uri: topLevelFrame ? topLevelFrame.uri : '',
```
LMS007, line 34: I wonder if this even a valid state when `mainFrame` returns null? but using a blank string seems reasonable.

robertknight: This component shouldn't be shown when `mainFrame` returns `null`. We definitely can't create an annotation in that situation, so I would suggest to just bail out of the callback in that case, or just make the component render nothing.

src/sidebar/components/new-note-btn.js
```diff
@@ -21,9 +31,9 @@ function NewNoteButton({ annotationsService, settings }) {
...
     };
-    annotationsService.create(annot);
+    annotationsService.create1(annot);
```
robertknight, line 36: This looks like an accidental change.

LMS007: oops!

PR #2413, opened 2020-08-05
Typecheck `DocumentMetadata` class

robertknight:
Typecheck the class that reads metadata from HTML documents and attaches
it to annotations as the `annotation.document` property.

This resulted in a few minor corrections to the `DocumentMetadata` type
in `src/types/annotator.js`.

### DIFF
--- a/src/annotator/plugin/document.js
+++ b/src/annotator/plugin/document.js
@@ -10,9 +10,18 @@
  ** https://github.com/openannotation/annotator/blob/master/LICENSE
  */
 
+/**
+ * nb. The `DocumentMetadata` type is renamed to avoid a conflict with the
+ * `DocumentMetadata` class below.
+ *
+ * @typedef {import('../../types/annotator').DocumentMetadata} Metadata
+ */
+
 import baseURI from 'document-base-uri';
 
+// @ts-expect-error - '../plugin' needs to be converted to JS.
 import Plugin from '../plugin';
+
 import { normalizeURI } from '../util/url';
 
 /**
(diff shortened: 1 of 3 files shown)


src/annotator/plugin/document.js
```diff
@@ -26,14 +35,16 @@ export default class DocumentMeta extends Plugin {
     this.events = {
       beforeAnnotationCreated: 'beforeAnnotationCreated',
     };
+
+    /** @type {Metadata} */
+    this.metadata = { title: document.title, link: [] };
+
+    this.baseURI = options.baseURI || baseURI;
+    this.document = options.document || document;
+    this.normalizeURI = options.normalizeURI || normalizeURI;
```
robertknight, line 44: I moved these assignments from `pluginInit` to here because TS only looks at the constructor of a class to infer what fields it has.

src/types/annotator.js
```diff
@@ -17,8 +17,8 @@
  * @typedef DocumentMetadata
  * @prop {string} title
  * @prop {Object[]} link
- *   @prop {string} link.rel
- *   @prop {string} link.type
+ *   @prop {string} [link.rel]
+ *   @prop {string} [link.type]
  *   @prop {string} link.href
```
robertknight, line 22: Although all `<link>` tags should have a `rel` attribute, it appears that the `link` array contains entries from sources other than `<link>`  and those may only have an `href` property.

src/annotator/plugin/document.js
LMS007: good to know!

PR #2414, opened 2020-08-05
Improve typechecking (group-list-item.js group-list-section.js group-list.js)

LMS007:
group-list-item.js
group-list-section.js
group-list.js


src/sidebar/components/menu-item.js
```diff
@@ -220,10 +220,10 @@ export default function MenuItem({
     <Fragment>
       {menuItem}
       {hasSubmenuVisible && (
-        <Slider visible={isSubmenuVisible}>
+        <Slider visible={/** @type {boolean} */ (isSubmenuVisible)}>
```
LMS007, line 223: I guess it can't see that this _has_ to be a **boolean** due to the `hasSubmenuVisible` check above it. I wonder if thats  a jsx issue with ts or something else...

src/sidebar/components/group-list.js
```diff
@@ -23,9 +28,17 @@ function publisherProvidedIcon(settings) {
   return svc && svc.icon ? svc.icon : null;
 }
 
+/**
+ * @typedef GroupListProps
+ * @prop {(url: string) => any} [serviceUrl]
```
robertknight, line 33: I'd suggest to use the `ServiceUrlGetter` type from `sidebar/services/service-url.js` here.

src/types/api.js
```diff
@@ -120,6 +122,8 @@
  * @prop {boolean} isScopedToUri
  * @prop {string} name
  * @prop {boolean} canLeave
+ * @prop {Object} links
+ *   @prop {string} links.html
```
robertknight, line 126: This link _is_ a property on an API object, so it should be listed higher up with the other properties that are part of the API object.

src/types/config.js
```diff
@@ -11,6 +11,7 @@
  * @prop {string} apiUrl
  * @prop {string} authority
  * @prop {string} grantToken
+ * @prop {string} icon
```
robertknight, line 14: The `icon` property is optional. See https://github.com/hypothesis/client/blob/c8df07a364061d82d24c61da9fcd9ba6d1b67b1f/src/sidebar/components/group-list.js#L23.

src/sidebar/components/group-list-item.js
```diff
@@ -24,7 +41,8 @@ function GroupListItem({
 }) {
   const activityUrl = group.links.html;
   const hasActionMenu = activityUrl || group.canLeave;
-  const isSelectable = !group.scopes.enforced || group.isScopedToUri;
+  const isSelectable =
+    (group.scopes && !group.scopes.enforced) || group.isScopedToUri;
```
robertknight, line 45: 👍 The client should always request scopes so I don't think `group.scopes` will ever actually be null. It seems sensible to check that here though given that this logic far removed from where the group fetching happens.

src/sidebar/components/group-list.js
LMS007: Yep, keep forgetting that 1

LMS007:
Just wanted confirmation about the {Links} type. I assumed you meant to create a new type and then use that in both the `Annotation` and `Groups` api objects

robertknight:
> Just wanted confirmation about the {Links} type. I assumed you meant to create a new type and then use that in both the Annotation and Groups api objects

Ah, no. The type of the `links` property for `Group` and `Annotation` are not the same. Both types have an `html` property but group links do not have an `incontext` property. Can you revert that last change.

What I meant was that `links` property for `Group` was listed alongside a bunch of properties that are not part of the object returned from the API but that are added later by the client after it receives the group. All I asked for here was to move the `links` property a few lines up so that it is listed alongside the other properties that actually come from the API.

PR #2415, opened 2020-08-06
Typecheck annotator/{index, pdf-sidebar} and simplify tsconfig.json

robertknight:
This PR fixes or ignores remaining typechecking errors in `annotator/{index.js, pdf-sidebar.js}`. With that done, the `tsconfig.json` file can be simplified by switching it from listing files to include to only listing files to exclude. This means that any new files that are not explicitly excluded will be checked, other than tests which are currently not checked at all.

As part of this, I added types to `src/types/annotator.js` to document the non-native global variables on the `Window` object which are either set by the client or may influence how the client behaves.

### DIFF
--- a/src/types/annotator.js
+++ b/src/types/annotator.js
@@ -38,9 +38,10 @@
  *
  * @typedef Globals
  * @prop {Object} [PDFViewerApplication] -
- *   PDF.js entry point. If set triggers loading of PDF rather than HTML integration.
+ *   PDF.js entry point. If set, triggers loading of PDF rather than HTML integration.
  * @prop {boolean} [__hypothesis_frame] -
- *   Flag used to indicate that Hypothesis is loaded in the current frame.
+ *   Flag used to indicate that the "annotator" part of Hypothesis is loaded in
+ *   the current frame.
  */
 
 /**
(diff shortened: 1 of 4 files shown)


src/types/annotator.js
```diff
@@ -32,5 +32,20 @@
  * @prop {string} [documentFingerprint]
  */
 
+/**
+ * Global variables which the Hypothesis client looks for on the `window` object
+ * when loaded in a frame that influence how it behaves.
+ *
+ * @typedef Globals
+ * @prop {Object} [PDFViewerApplication] -
+ *   PDF.js entry point. If set triggers loading of PDF rather than HTML integration.
```
LMS007, line 41: "If set, triggers..."

src/types/annotator.js
```diff
@@ -32,5 +32,20 @@
...
+ * @prop {boolean} [__hypothesis_frame] -
+ *   Flag used to indicate that Hypothesis is loaded in the current frame.
```
LMS007, line 43: does Hypothesis mean sidebar or annotator or both?

robertknight: Just the "annotator" part.

robertknight:
@LMS007, I addressed your comments from above. I hope I'm correct in assuming it was good to merge apart from that.

PR #2419, opened 2020-08-06
improve typechecking (help-panel, - stream-search-input, toast-messages, version-info)

LMS007:
- help-panel
- stream-search-input
- toast-messages
- version-info

### DIFF
--- a/src/sidebar/components/help-panel.js
+++ b/src/sidebar/components/help-panel.js
@@ -13,7 +13,15 @@ import Tutorial from './tutorial';
 import VersionInfo from './version-info';
 
 /**
+ * @typedef {import('../components/user-menu').AuthState} AuthState
+ */
+
+/**
  * External link "tabs" inside of the help panel.
+ *
+ * @param {Object} props
+ *   @param {string} props.linkText - What the tab's link should say
+ *   @param {string} props.url - Where the tab's link should go
  */
 function HelpPanelTab({ linkText, url }) {
   return (
@@ -36,14 +44,20 @@ function HelpPanelTab({ linkText, url }) {
 }
 
 HelpPanelTab.propTypes = {
-  /* What the tab's link should say */
   linkText: propTypes.string.isRequired,
-  /* Where the tab's link should go */
   url: propTypes.string.isRequired,
 };
 
 /**
+ * @typedef HelpPanelProps
+ * @prop {AuthState} auth - Object with auth and user information
+ * @prop {Object} session - Injected service
+ */
+
+/**
  * A help sidebar panel with two sub-panels: tutorial and version info.
+ *
+ * @param {HelpPanelProps} props
  */
 function HelpPanel({ auth, session }) {
   const mainFrame = useStore(store => store.mainFrame());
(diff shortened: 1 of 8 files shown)


src/sidebar/components/version-info.js
```diff
@@ -6,8 +6,20 @@ import { withServices } from '../util/service-context';
 
 import Button from './button';
 
+// Type checking imports
+// eslint-disable-next-line no-unused-vars
+import VersionData from '../util/version-data';
```
LMS007, line 11: I may have forgotten the right way to do this, but I want to enforce the class type here.

src/sidebar/util/version-data.js
```diff
@@ -1,30 +1,25 @@
 /**
- * An object representing user info
- *
- * @typedef {Object} UserInfo
- * @property {string=} userid
- * @property {string=} displayName
+ * @typedef {import('../components/user-menu').AuthState} AuthState
```
LMS007, line 2: This looked like a partial `AuthState` type, I felt it was reasonable to just make it the same thing.

src/sidebar/store/modules/toast-messages.js
```diff
@@ -6,6 +6,7 @@ import * as util from '../util';
  * @prop {string} id
  * @prop {string} message
  * @prop {string} moreInfoURL
+ * @prop {boolean} isDismissed
```
LMS007, line 9: this appeared to be missing and I don't know if we intended this to be optional, but I'd like to error on the side of strict.

src/sidebar/components/version-info.js
LMS007: Thats what I tried first and I don't think that worked the way I expected it to.. I'll double check. But I may have looked it wrong.

src/sidebar/util/version-data.js
```diff
@@ -1,30 +1,25 @@
...
 
 /**
- * An object representing document info
+ * An object representing document info.
  *
  * @typedef {Object} DocumentInfo
- * @property {string=} uri - current document URL
- * @property {DocMetadata} metadata - document metadata
+ * @prop {string=} uri - current document URL.
+ * @prop {DocMetadata} metadata - document metadata.
```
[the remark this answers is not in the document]
LMS007: got it!

src/sidebar/components/help-panel.js
```diff
@@ -12,8 +12,20 @@ import SvgIcon from '../../shared/components/svg-icon';
...
 
+/**
+ * @typedef {import('../components/user-menu').AuthState} AuthState
+ */
+
+/**
+ * @typedef HelpPanelTabProps
+ * @prop {string} linkText - What the tab's link should say.
+ * @prop {string} url - Where the tab's link should go.
+ */
```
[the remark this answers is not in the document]
LMS007: okay

src/sidebar/components/version-info.js
robertknight: Ah, I know the confusion here. You need to explicitly use the name "default" for the default export - `import(".../version-data").default`

src/sidebar/components/version-info.js
```diff
@@ -6,8 +6,16 @@ import { withServices } from '../util/service-context';
 
 import Button from './button';
 
+/**
+ * @typedef VersionInfoProps
+ * @prop {import('../util/version-data')} versionData - Object with version information
```
robertknight, line 11:
```suggestion
 * @prop {import('../util/version-data').default} versionData - Object with version information
```

I'm not certain about the reason, but it appears that in this context a default export needs to be referred to explicitly rather than using something like the syntax for an `import` statement.

LMS007: awe, gotcha

PR #2420, opened 2020-08-07
Flush timers after each test

robertknight:
This is a work-in-progress fix for one aspect of https://github.com/hypothesis/client/issues/2249, which is that confusing test failures can occur if a test starts a timer and then the timer doesn't expire until after the test completes. These timers may be scheduled directly by us or implicitly by third-party libraries. In most cases tests should clean up their own timers before they exit, but this may be tricky in some cases (eg. when the timer lives in a third-party library).

If one of these timers expires during another test, Mocha will attribute the error to _that_ test instead of the test that actually started the timer. If the timer expiry occurs "between" tests then there is currently a separate issue in the `mocha`/`karma-mocha` libraries where the test suite aborts with no indication of an error, except for the completed test count being smaller than expected.

This PR ensures that any timers started during the execution of a test are executed before the test completes. It does this by wrapping the `setTimeout` and `requestAnimationFrame` functions and then flushing any not-yet-executed timers in an `afterEach` hook. I'm not sure that this is going to be the best fix long-term, but the code here is set up in such a way that we can easily remove it later.

I also added a Python script in `scripts/run-test-loop.py` which executes the test suite in a loop, to make it easier to flush out flakey test failures.

The issue in the test runner with errors that occur at certain times not being reported needs to be addressed upstream in either the mocha or karma-mocha libraries.

### DIFF
--- a/scripts/run-test-loop.py
+++ b/scripts/run-test-loop.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python3
+
+import re
+import subprocess
+import sys
+
+max_iterations = 100
+for i in range(0, max_iterations):
+    print(f"Running test iteration {i+1} of {max_iterations}")
+
+    try:
+        stdout = None
+        proc = subprocess.run(
+            "node_modules/.bin/gulp test",
+            shell=True,
+            stdout=subprocess.PIPE,
+            check=True,
+        )
+        stdout = proc.stdout.decode()
+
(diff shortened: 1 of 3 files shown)


robertknight:
I don't know why at present, but we haven't seen any test suite failures due to this issue recently. We have upgraded Karma and Preact IIRC and we have also added additional tests. It is possible one of those things made the issue less likely to occur.

robertknight:
I'm going to close this for the moment, per the previous comment, and revisit another time.

PR #2431, opened 2020-08-10
improve typechecking (thread-list, thread-card, thread, timestamp)

LMS007:
- thread-list
- thread-card
- thread
- timestamp

### DIFF
--- a/src/sidebar/components/annotation-header.js
+++ b/src/sidebar/components/annotation-header.js
@@ -27,10 +27,7 @@ export default function AnnotationHeader({
   const isCollapsedReply = isReply(annotation) && threadIsCollapsed;
   const setExpanded = useStore(store => store.setExpanded);
 
-  const annotationIsPrivate = isPrivate(
-    annotation.permissions,
-    annotation.user
-  );
+  const annotationIsPrivate = isPrivate(annotation.permissions);
   const annotationLink = annotation.links ? annotation.links.html : '';
 
   // NB: `created` and `updated` are strings, not `Date`s
--- a/src/sidebar/components/annotation-share-info.js
+++ b/src/sidebar/components/annotation-share-info.js
@@ -7,27 +7,25 @@ import { isPrivate } from '../util/permissions';
 import SvgIcon from '../../shared/components/svg-icon';
 
 /**
+ * @typedef {import('../../types/api').Group} Group
+ */
+
+/**
  * Render information about what group an annotation is in and
  * whether it is private to the current user (only me)
  */
 function AnnotationShareInfo({ annotation }) {
   const group = useStore(store => store.getGroup(annotation.group));
 
-  // We may not have access to the group object beyond its ID
-  const hasGroup = !!group;
-
   // Only show the name of the group and link to it if there is a
   // URL (link) returned by the API for this group. Some groups do not have links
-  const linkToGroup = hasGroup && group.links && group.links.html;
+  const linkToGroup = group?.links.html;
 
-  const annotationIsPrivate = isPrivate(
-    annotation.permissions,
-    annotation.user
-  );
+  const annotationIsPrivate = isPrivate(annotation.permissions);
 
   return (
     <div className="annotation-share-info u-layout-row--align-baseline">
-      {linkToGroup && (
+      {group && linkToGroup && (
         <a
           className="u-layout-row--align-baseline u-color-text--muted"
           href={group.links.html}
--- a/src/sidebar/components/annotation.js
+++ b/src/sidebar/components/annotation.js
@@ -53,7 +53,8 @@ function Annotation({
   const toggleText = `${toggleAction} (${replyCount})`;
 
   const shouldShowActions = !isSaving && !isEditing;
-  const shouldShowLicense = isEditing && !isPrivate && group.type !== 'private';
+  const shouldShowLicense =
+    isEditing && !isPrivate && group && group.type !== 'private';
   const shouldShowReplyToggle = replyCount > 0 && !isReply(annotation);
 
   const onEditTags = ({ tags }) => {
--- a/src/sidebar/components/test/thread-list-test.js
+++ b/src/sidebar/components/test/thread-list-test.js
@@ -103,6 +103,15 @@ describe('ThreadList', () => {
     wrapper.setProps({});
   };
 
+  context('invalid scroll container', () => {
+    it('should throw an error if the scroll container is missing', () => {
+      fakeScrollContainer.remove();
+      assert.throws(() => {
+        createComponent();
+      }, 'Scroll container is missing');
+    });
+  });
+
   context('new annotation created in application', () => {
     it('clears the current selection in the store', () => {
       const wrapper = createComponent();
--- a/src/sidebar/components/thread-list.js
+++ b/src/sidebar/components/thread-list.js
@@ -13,13 +13,24 @@ import {
 
 import ThreadCard from './thread-card';
 
+/** @typedef {import('../util/build-thread').Thread} Thread */
+
 // The precision of the `scrollPosition` value in pixels; values will be rounded
 // down to the nearest multiple of this scale value
 const SCROLL_PRECISION = 50;
 
 function getScrollContainer() {
-  return document.querySelector('.js-thread-list-scroll-root');
+  const container = document.querySelector('.js-thread-list-scroll-root');
+  if (!container) {
+    throw new Error('Scroll container is missing');
+  }
+  return container;
 }
+
+/**
+ * @typedef ThreadListProps
+ * @prop {Thread} thread
+ */
 
 /**
  * Render a list of threads.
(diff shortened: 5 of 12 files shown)


src/sidebar/components/annotation-header.js
```diff
@@ -27,10 +27,7 @@ export default function AnnotationHeader({
   const isCollapsedReply = isReply(annotation) && threadIsCollapsed;
   const setExpanded = useStore(store => store.setExpanded);
 
-  const annotationIsPrivate = isPrivate(
-    annotation.permissions,
-    annotation.user
-  );
+  const annotationIsPrivate = isPrivate(annotation.permissions);
```
LMS007, line 30: extraneous  param

src/sidebar/components/annotation-share-info.js
```diff
@@ -6,12 +6,18 @@ import { isPrivate } from '../util/permissions';
...
+ * @typedef {import('../../types/api').Group} Group
+ */
+
 /**
  * Render information about what group an annotation is in and
  * whether it is private to the current user (only me)
  */
 function AnnotationShareInfo({ annotation }) {
-  const group = useStore(store => store.getGroup(annotation.group));
+  const group = /** @type {Group} */ (useStore(store =>
```
LMS007, line 18: it really barked at me if I took this cast out. It does not seem to be able to parse the jsx below where it checks `hasGroup` first, then `group.blah`. I'm not sure I understand why or perhaps I'm looking at it wrong. A simple solution here is to cast this as it can't fail because we're checking `hasGroup` before any other member referencing of that potentially `undefined` object

src/sidebar/components/thread-list.js
```diff
@@ -13,14 +13,25 @@ import {
...
+
 // The precision of the `scrollPosition` value in pixels; values will be rounded
 // down to the nearest multiple of this scale value
 const SCROLL_PRECISION = 50;
 
 function getScrollContainer() {
-  return document.querySelector('.js-thread-list-scroll-root');
+  const container = document.querySelector('.js-thread-list-scroll-root');
+  if (!container) {
+    throw new Error('Scroll container is missing');
```
LMS007, line 25: I'm not really sure if this can even happen, but its simple enough to throw an error. I suppose we could cast too, but this felt a bit safer.

src/sidebar/components/thread-list.js
```diff
@@ -159,7 +174,9 @@ function ThreadList({ thread }) {
     setThreadHeights(prevHeights => {
       const changedHeights = {};
       for (let { id } of visibleThreads) {
-        const threadElement = document.getElementById(id);
+        const threadElement = /** @type {HTMLElement} */ (document.getElementById(
```
LMS007, line 177: I can't see any of these elements as being null, but I suppose it's possible? However, I felt a cast was a reasonable risk here.

src/types/api.js
```diff
@@ -113,6 +113,8 @@
  * @prop {Organization} organization - nb. This field is nullable in the API, but
  *   we assign a default organization on the client.
  * @prop {GroupScopes|null} scopes
+ * @prop {Object} links
```
LMS007, line 116: note: this is covered in another active PR

src/sidebar/components/annotation-share-info.js
robertknight: > it really barked at me if I took this cast out. It does not seem to be able to parse the jsx below where it checks hasGroup first, then group.blah

It seems null-ness inference is not transitive:

```js
if (group) {
 // `group` is not null here
}
```

But if there is some indirection:

```js
const hasGroup = !!group;
if (hasGroup) {
  // TS thinks `group` may be null here
}
```

In this case a way around this is to remove the `hasGroup` variable and test the group more directly:

```diff
diff --git a/src/sidebar/components/annotation-share-info.js b/src/sidebar/components/annotation-share-info.js
index dbbfb6225..11041e7f7 100644
--- a/src/sidebar/components/annotation-share-info.js
+++ b/src/sidebar/components/annotation-share-info.js
@@ -15,22 +15,17 @@ import SvgIcon from '../../shared/components/svg-icon';
  * whether it is private to the current user (only me)
  */
 function AnnotationShareInfo({ annotation }) {
-  const group = /** @type {Group} */ (useStore(store =>
-    store.getGroup(annotation.group)
-  ));
-
-  // We may not have access to the group object beyond its ID
-  const hasGroup = !!group;
+  const group = useStore(store => store.getGroup(annotation.group));
 
   // Only show the name of the group and link to it if there is a
   // URL (link) returned by the API for this group. Some groups do not have links
-  const linkToGroup = hasGroup && group.links && group.links.html;
+  const linkToGroup = group?.links.html;
 
   const annotationIsPrivate = isPrivate(annotation.permissions);
 
   return (
     <div className="annotation-share-info u-layout-row--align-baseline">
-      {linkToGroup && (
+      {group && linkToGroup && (
         <a
           className="u-layout-row--align-baseline u-color-text--muted"
           href={group.links.html}
diff --git a/src/sidebar/components/test/annotation-share-info-test.js b/src/sidebar/components/test/annotation-share-info-test.js
index d1bd20fd1..4d599d26c 100644
--- a/src/sidebar/components/test/annotation-share-info-test.js
+++ b/src/sidebar/components/test/annotation-share-info-test.js
@@ -108,7 +108,7 @@ describe('AnnotationShareInfo', () => {
       });
 
       it('should show "only me" text for annotation in third-party group', () => {
-        fakeGetGroup.returns({ name: 'Some Name' });
+        fakeGetGroup.returns({ name: 'Some Name', links: {} });
         const wrapper = createAnnotationShareInfo();
 
         const privacyText = wrapper.find(
```

The above diff does feature a redundant test of whether `group` is null in the JSX, but this is very cheap compared to everything else that is going on, and it avoids a problem if `linkToGroup` is changed in future in such a way that it may be set even if `group` is null.

Looking at the `h` backend, the `group.links` object should always exist, but it may be empty. Therefore I adjusted a test fixture accordingly.

src/sidebar/components/test/thread-list-test.js
```diff
@@ -103,6 +103,25 @@ describe('ThreadList', () => {
     wrapper.setProps({});
   };
 
+  context('invalid scroll container', () => {
```
robertknight, line 106: There is a simpler way to make the scroll container invalid that doesn't require installing a `querySelector` mock:

```js
  context('invalid scroll container', () => {
    it('should throw an error if the scroll container is missing', () => {
      fakeScrollContainer.remove();
      assert.throws(() => {
        createComponent();
      }, 'Scroll container is missing');
    });
  });
```

src/sidebar/components/thread-list.js
robertknight: 👍 - It makes sense to check here, since the scroll container is not created by the component itself. Ideally it would be better if the scroll container were part of the component, but when this was set up there were good reasons not to do that.

src/sidebar/components/thread-list.js
```diff
@@ -13,14 +13,25 @@ import {
...
+  const container = document.querySelector('.js-thread-list-scroll-root');
+  if (!container) {
+    throw new Error('Scroll container is missing');
+  }
+  return container;
 }
 
+/**
+ * @typedef ThreadListProps
+ * @prop {Thread} thread - Should annotations render extra document metadata?
```
robertknight, line 32:
```suggestion
 * @prop {Thread} thread
```

The prop documentation refers to something else.

src/sidebar/components/thread-list.js
robertknight: Yes, this seems fine.

src/sidebar/store/modules/annotations.js
```diff
@@ -564,6 +564,7 @@ function savedAnnotations(state) {
  * @prop {(tags: string[]) => string[]} findIDsForTags
  * @prop {() => string[]} focusedAnnotations
  * @prop {() => string[]} highlightedAnnotations
+ * @prop {(tags: string) => boolean} isAnnotationFocused
```
robertknight, line 567:
```suggestion
 * @prop {(tag: string) => boolean} isAnnotationFocused
```

Singular `tag` since we're talking about a single annotation here.

src/types/api.js
```diff
@@ -113,6 +113,8 @@
...
+ *   @prop {string} links.html
```
robertknight, line 117: The `html` property should be optional. It is not currently present for groups in third-party authorities for example.

src/sidebar/components/annotation.js
```diff
@@ -53,7 +53,8 @@ function Annotation({
   const toggleText = `${toggleAction} (${replyCount})`;
 
   const shouldShowActions = !isSaving && !isEditing;
-  const shouldShowLicense = isEditing && !isPrivate && group.type !== 'private';
+  const shouldShowLicense =
+    isEditing && !isPrivate && group && group.type !== 'private';
```
robertknight, line 57: In new code you can also shorten `foo && foo.bar` to `foo?.bar`.

robertknight:
I'm going to merge this so I can review https://github.com/hypothesis/client/pull/2433.

PR #2435, opened 2020-08-12
Improve typechecking (share-annotations-panel sidebar-content-error sidebar-content stream-content)

LMS007:
- share-annotations-panel
- sidebar-content-error
- sidebar-content
- stream-content

### DIFF
--- a/src/sidebar/components/share-annotations-panel.js
+++ b/src/sidebar/components/share-annotations-panel.js
@@ -12,20 +12,23 @@ import SidebarPanel from './sidebar-panel';
 import SvgIcon from '../../shared/components/svg-icon';
 
 /**
+ * @typedef ShareAnnotationsPanelProps
+ * @prop {Object} analytics - Injected service
+ * @prop {Object} toastMessenger - Injected service
+ */
+
+/**
  * A panel for sharing the current group's annotations.
  *
- * Links withinin this component allow a user to share the set of annotations that
+ * Links within this component allow a user to share the set of annotations that
  * are on the current page (as defined by the main frame's URI) and contained
  * within the app's currently-focused group.
+ *
+ * @param {ShareAnnotationsPanelProps} props
  */
 function ShareAnnotationsPanel({ analytics, toastMessenger }) {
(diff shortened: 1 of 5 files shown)


src/sidebar/components/sidebar-content.js
```diff
@@ -44,7 +55,7 @@ function SidebarContent({
   });
   const directLinkedTab = linkedAnnotation
     ? tabForAnnotation(linkedAnnotation)
-    : null;
+    : 'annotation'; // default to 'annotation' tab otherwise
```
LMS007, line 58: I think this is the biggest question mark, 

the method this is passed to is very specific about the tab name and I think thats fine, so I think we should default to one of the 3 possible options rather than null which is somewhat meaning less... unless we have a state in which we should no tabs?

```
/**
 * Set the currently-selected tab to `tabKey`.
 *
 * @param {'annotation'|'note'|'orphan'} tabKey
 */
function selectTab(tabKey) {
  return {
    type: actions.SELECT_TAB,
    tab: tabKey,
  };
}
```

src/sidebar/components/share-annotations-panel.js
```diff
@@ -59,7 +67,7 @@ function ShareAnnotationsPanel({ analytics, toastMessenger }) {
       {shouldRenderPanelContent && (
         <div className="share-annotations-panel">
           <div className="share-annotations-panel__intro">
-            {focusedGroup.type === 'private' ? (
+            {focusedGroup?.type === 'private' ? (
```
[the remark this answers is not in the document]
LMS007: I see, that happened on my previous PR too. I'll keep that in mind.

src/sidebar/components/sidebar-content.js
robertknight, line 58:
```suggestion
    : 'annotation';
```

The comment seems redundant.

LMS007:
> LGTM 👍 . I restarted the Travis build - Travis has encountered a number of sporadic failures over the last few days. Incidentally we'll be switching to GitHub Actions soon - similar to the backend folks.

looking forward to github actions :)

PR #2438, opened 2020-08-14
Fix typecheck on last remaining components

LMS007:
- hypothesis-app
- search-input
- top-bar
- index (root sidebar index)

### DIFF
--- a/src/sidebar/components/top-bar.js
+++ b/src/sidebar/components/top-bar.js
@@ -18,6 +18,24 @@ import StreamSearchInput from './stream-search-input';
 import UserMenu from './user-menu';
 
 /**
+ * @typedef {import('../../types/config').MergedConfig} MergedConfig
+ * @typedef {import('../components/user-menu').AuthState} AuthState
+ * @typedef {import('../../shared/bridge').default} Bridge
+ */
+
+/**
+ * @typedef TopBarProps
+ * @prop {AuthState} [auth]
+ * @prop {Bridge} bridge
+ * @prop {boolean} [isSidebar] - Flag indicating whether the app is the sidebar or a top-level page.
+ * @prop {() => any} [onLogin] - Callback invoked when user clicks "Login" button.
+ * @prop {() => any} [onLogout] - Callback invoked when user clicks "Logout" action in account menu.
+ * @prop {() => any} [onSignUp] - Callback invoked when user clicks "Sign up" button.
+ * @prop {MergedConfig} [settings]
+ * @prop {Object} [streamer]
+ */
+
+/**
  * The toolbar which appears at the top of the sidebar providing actions
  * to switch groups, view account information, sort/filter annotations etc.
  */
@@ -57,8 +75,8 @@ function TopBar({
    * help requests, fire a relevant event instead
    */
   const requestHelp = () => {
-    const service = serviceConfig(settings) || {};
-    if (service.onHelpRequestProvided) {
+    const service = serviceConfig(settings);
+    if (service && service.onHelpRequestProvided) {
       bridge.call(bridgeEvents.HELP_REQUESTED);
     } else {
       togglePanelFn(uiConstants.PANEL_HELP);
@@ -127,7 +145,7 @@ function TopBar({
               }`}
             />
           )}
-          <SearchInput query={filterQuery} onSearch={setFilterQuery} />
+          <SearchInput query={filterQuery || null} onSearch={setFilterQuery} />
           <SortMenu />
           {showSharePageButton && (
             <Button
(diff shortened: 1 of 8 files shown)


src/sidebar/index.js
```diff
@@ -182,7 +184,9 @@ function startApp(config) {
   container.run(setupFrameSync);
 
   // Render the UI.
-  const appEl = document.querySelector('hypothesis-app');
+  const appEl = /** @type {HTMLElement} */ (document.querySelector(
+    'hypothesis-app'
+  ));
   render(
     <ServiceContext.Provider value={container}>
```
LMS007, line 191: This blows up here, I have not looked too deeply into this yet,  but I suspect it has to do with our Injector type.

src/types/api.js
```diff
@@ -84,13 +84,17 @@
...
 /**
- * TODO - Fill out remaining properties
- *
  * @typedef Profile
  * @prop {string|null} userid
  * @prop {Object} preferences
- * @prop {boolean} [preferences.show_sidebar_tutorial]
+ *   @prop {boolean} [preferences.show_sidebar_tutorial]
  * @prop {Object.<string, boolean>} features
+ * @prop {Object[]} [groups]
```
LMS007, line 92: There appears to be group prop here but its different than our `Group` type. Ac-Hoc prop seems reasonable.

src/types/config.js
```diff
@@ -85,6 +85,8 @@
  *   Name of the base theme to use.
  * @prop {string} [usernameUrl]
  *   URL template for username links
+ * @prop {string} [assetRoot]
```
LMS007, line 88: I not 100% sure these 2 go here, I could use feedback on this

src/sidebar/components/top-bar.js
```diff
@@ -52,16 +69,21 @@ function TopBar({
     store => store.getState().sidebarPanels.activePanelName
   );
 
+  const searchQueryProp = {};
```
LMS007, line 72: This is just a trick to not to pass the prop rather than passing an empty string which is a lot less ugly, however,  not  as correct.

src/sidebar/index.js
robertknight: Ah, interesting. The error here is:

```
src/sidebar/index.js(191,30): error TS2322: Type 'Injector' is not assignable to type '{ get(service: any): never; }'.
  The types returned by 'get(...)' are incompatible between these types.
    Type 'any' is not assignable to type 'never'.
```

Breaking that down, it is complaining that the `value` prop is expecting an object with a `get` method that never returns (that's what the `never` type means), but the object you're passing (`container`) has a `get` method that returns _something_ (`any`).

So the problem here is that `ServiceContext` has the wrong type. Looking at the definition, the reason becomes apparent:

```js
const fallbackInjector = {
  get(service) {
    throw new Error(
      `Missing ServiceContext provider to provide "${service}" prop`
    );
  },
};
...
export const ServiceContext = createContext(fallbackInjector);
```

`ServiceContext`'s type is inferred from `fallbackInjector` where `get` never returns. The fix is to override the type:

```diff
diff --git a/src/sidebar/util/service-context.js b/src/sidebar/util/service-context.js
index a9ef08fce..2dfd0d3bf 100644
--- a/src/sidebar/util/service-context.js
+++ b/src/sidebar/util/service-context.js
@@ -12,9 +12,15 @@
  * @typedef {import("redux").Store} Store
  */
 
+/**
+ * @typedef ServiceProvider
+ * @prop {(serviceName: string) => any} get
+ */
+
 import { createContext, createElement } from 'preact';
 import { useContext } from 'preact/hooks';
 
+/** @type {ServiceProvider} */
 const fallbackInjector = {
   get(service) {
     throw new Error(
```

src/sidebar/store/modules/sidebar-panels.js
```diff
@@ -91,7 +91,7 @@ function closeSidebarPanel(panelName) {
  * designated `panelState`
  *
  * @param {string} panelName
- * @param {boolean} panelState - Should the panel be active?
+ * @param {boolean} [panelState] - Should the panel be active?
```
robertknight, line 94: This means "toggle current state" if omitted? If so I think that is worth mentioning in the prop docs.

src/types/api.js
robertknight: The `groups` prop is deprecated and should not be used by the client. I would omit it here and perhaps add a note in the type description that deprecated properties are not included. Alternatively, you could list it separately from the others and use the `unknown` type to enforce that no code uses it:

```js
 * @prop {unknown} [groups] - Deprecated.
```

`unknown` is like `any` but code must check its actual type somehow before using it.

src/types/config.js
robertknight: I would leave these out. These properties may be set in the host config, but they are not used by the sidebar. This type only lists the host config properties which are used by the sidebar - see the comment above. `sidebarAppUrl` is used by "annotator" and `assetRoot` is used by the boot script.

src/sidebar/components/top-bar.js
```diff
@@ -52,16 +69,21 @@ function TopBar({
...
    */
   const requestHelp = () => {
-    const service = serviceConfig(settings) || {};
-    if (service.onHelpRequestProvided) {
+    const service = serviceConfig(settings);
+    if (service && service.onHelpRequestProvided) {
       bridge.call(bridgeEvents.HELP_REQUESTED);
     } else {
-      togglePanelFn(uiConstants.PANEL_HELP);
+      togglePanelFn(uiConstants.PANEL_HELP, false);
```
robertknight, line 86: Passing the value `false` here will hide the sidebar panel every time it is called. Omitting the second parameter toggles the panel, which is the intent.

LMS007: bah! that was a debug thing I forgot to remove, good catch!

LMS007:
Yes, my hunch was it was something to do with fallbackInjector, makes sense!

src/sidebar/store/modules/sidebar-panels.js
```diff
@@ -88,10 +88,11 @@ function closeSidebarPanel(panelName) {
 
 /**
  * Toggle a sidebar panel from its current state, or set it to the
- * designated `panelState`
+ * designated `panelState` if preset.
```
robertknight, line 91:
```suggestion
 * designated `panelState`.
```

IMO the sentence reads better without "if preset".

PR #2456, opened 2020-08-18
Replace Travis with a GitHub Actions CI workflow

robertknight:
The backend Hypothesis projects have switched to using GitHub Actions
for CI. This does the same for the Hypothesis client.

Aside from consistency with other projects, GitHub Actions has the benefit of better integration with GitHub and more flexibility.

This is also a step towards possibly replacing our use of Jenkins for client deployments with GitHub Actions, which the backend and ops folks are also looking at.


robertknight:
WIP bot seems to be having issues. I'm going to merge this as the other status checks have passed.

PR #2459, opened 2020-08-18
WIP - convert xpath.coffee and util.coffee to .js

LMS007:
This is a draft version of the conversion so far. The larger goal for the first bit of work is to move util.coffee, xpath.coffee and range.coffee into a single range module.

Not all functions in these files are tested. There is a fair bit of integration testing already done in `html-test.js`, but it also appears to be lacking any testing of `nodeFromXPath` which is called when the mimi type is `application/xhtml+xml` for a given document. Perhaps that is something we can add to dev-server too?

Currently this is just a plain old conversion. I thought about attempting to remove all jquery bits, but that proved well beyond the scope of the conversion so I think we take that on as a later pass. I did take the liberty and removed some complex bits for a IE9 work around, but pretty much everything else is a 1:1 translation (or so I intended)

The next step will be to merge `range-utils` with `range.coffee` into `range.js`.h


robertknight:
Hi @LMS007 - Can you rebase this branch on master to incorporate the GitHub Actions configuration. This will make the "ci" check work.

src/annotator/anchoring/range-util.js
```diff
@@ -0,0 +1,227 @@
...
+  const paths = nodes.map((index, node) => {
+    return getPathTo(node);
+  });
+  return paths.get();
+}
+
+/**
+ * Public: Flatten a nested array structure
+ */
+function flatten(array) {
```
robertknight, line 112: We could eventually replace this with `Array.prototype.flat` from ES2019, we'd need to set up a polyfill for older browsers for that though. In the interim you could replace `$.isArray` with `Array.isArray` below.

src/annotator/anchoring/range-util.js
```diff
@@ -0,0 +1,227 @@
+import $ from 'jquery';
+
+/**
+ * A simple XPath evaluator using jQuery which can evaluate queries of
+ */
+function simpleXPathJQuery(nodes, relativeRoot) {
```
robertknight, line 6: Since we're in the process of removing jQuery, I would be inclined to remove this code path and just keep the non-jQuery one. It might be easier to do this first before landing the conversion of CoffeeScript => JS.

src/annotator/anchoring/range-util.js
```diff
@@ -0,0 +1,227 @@
...
+  if (prev) {
+    // eslint-disable-next-line no-unused-vars
+    return getLastTextNodeUpTo(prev);
+  } else {
+    return null;
+  }
+}
+
+/**
+ * Public: determine the first text node in or after the given jQuery node.
```
robertknight, line 178: The `Public: ` part can be dropped from the comments - this is implied by the `export` keyword.

PR #2460, opened 2020-08-20
Convert util methods to js (Step #1)

LMS007:
Convert several exposed methods from util.coffee to range-util.js and add a basic set of unit tests for each.

- getTextNodes
- getLastTextNodeUpTo
- getFirstTextNodeNotBefore

This is the first step. I moved a few util methods to a new js file (range-util.js). The next step is to move all the methods out of xpath.coffee and the rest from util.coffee and at least 1 from range.coffee into this new range-util.js file. I'll add basic unit tests as are reasonable but it's hard to guess at all the edge cases for some of these functions so I could use some additional help there.

### DIFF
--- a/src/annotator/anchoring/test/range-util-test.js
+++ b/src/annotator/anchoring/test/range-util-test.js
@@ -1,0 +1,104 @@
+import $ from 'jquery';
+
+import {
+  getTextNodes,
+  getLastTextNodeUpTo,
+  getFirstTextNodeNotBefore,
+} from '../range-util';
+
+describe('annotator/anchoring/range-util', function () {
+  describe('#getTextNodes', () => {
+    let container;
+    before(function () {
+      container = document.createElement('div');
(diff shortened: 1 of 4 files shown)


src/annotator/anchoring/test/range-util-test.js
```diff
@@ -0,0 +1,104 @@
...
+      assert.equal(result.nodeValue, 'text last');
+    });
+
+    it('looks backwards to get the last text node if none are found', () => {
+      container.innerHTML = `<span>text first</span><span>text last</span><div id="too-far"></div>`;
+      const result = getLastTextNodeUpTo(container.querySelector('#too-far'));
+      assert.equal(result.nodeValue, 'text last');
+    });
+
+    it.skip('returns null if no text node exists', () => {
```
LMS007, line 66: These 2 null cases don't seem to be working quite right. I'll give em another look tomorrow

PR #2465, opened 2020-08-21
Upgrade TypeScript and fix various issues related to optional properties

robertknight:
This PR upgrades TypeScript to v4.0 and fixes issues that the new version complains about.

The major change is that optional properties (ie. `@prop {Thing} [aThing]`) are now treated as being `Thing|undefined` rather than just `Thing`, which means code must check that the property exists before trying to use it. This surfaced various issues such as:

- Properties which were incorrectly marked as optional in JSDoc comments
- Places where a missing optional property was not handled correctly
- Cases where we know an optional property must exist, but for non-trivial reasons that TS isn't aware of (eg. we know that we're dealing with an annotation that has been saved to the server, so the `id` field must be set)

There were also a small number of unrelated changes. See individual commits for more details.

In general I expect that most TS updates will require far fewer changes than this.

### DIFF
--- a/src/sidebar/components/annotation-body.js
+++ b/src/sidebar/components/annotation-body.js
@@ -28,7 +28,7 @@ import TagList from './tag-list';
  * @prop {string} text -
  *     The markdown annotation body, which is either rendered as HTML (if `isEditing`
  *     is false) or displayed in a text area otherwise.
- * @prop {MergedConfig} [settings]
+ * @prop {MergedConfig} settings
  */
 
 /**
--- a/src/sidebar/components/menu-item.js
+++ b/src/sidebar/components/menu-item.js
@@ -91,7 +91,7 @@ export default function MenuItem({
   let focusTimer = null;
 
   let renderedIcon = null;
-  if (icon !== 'blank') {
+  if (icon && icon !== 'blank') {
     renderedIcon = iconIsUrl ? (
       <img className={iconClass} alt={iconAlt} src={icon} />
     ) : (
--- a/src/sidebar/components/user-menu.js
+++ b/src/sidebar/components/user-menu.js
@@ -18,11 +18,9 @@ import SvgIcon from '../../shared/components/svg-icon';
 
 /**
  * @typedef AuthState
- * @prop {'logged-in'|'logged-out'} status
- * @prop {string} [displayName]
- * @prop {string} [userid]
- * @prop {string} [username]
- * @prop {string} [provider]
+ * @prop {string} displayName
+ * @prop {string} userid
+ * @prop {string} username
  */
 
 /**
--- a/src/sidebar/store/modules/real-time-updates.js
+++ b/src/sidebar/store/modules/real-time-updates.js
@@ -102,12 +102,13 @@ function receiveRealTimeUpdates({
         ann.group === groups.selectors.focusedGroupId(getState()) ||
         route.selectors.route(getState()) !== 'sidebar'
       ) {
-        pendingUpdates[ann.id] = ann;
+        pendingUpdates[/** @type {string} */ (ann.id)] = ann;
       }
     });
     deletedAnnotations.forEach(ann => {
+      const id = /** @type {string} */ (ann.id);
       // Discard any pending but not-yet-applied updates for this annotation
-      delete pendingUpdates[ann.id];
+      delete pendingUpdates[id];
 
       // If we already have this annotation loaded, then record a pending
       // deletion. We do not check the group of the annotation here because a)
@@ -115,8 +116,8 @@ function receiveRealTimeUpdates({
       // even if the annotation is from the current group, it might be for a
       // new annotation (saved in pendingUpdates and removed above), that has
       // not yet been loaded.
-      if (annotations.selectors.annotationExists(getState(), ann.id)) {
-        pendingDeletions[ann.id] = true;
+      if (annotations.selectors.annotationExists(getState(), id)) {
+        pendingDeletions[id] = true;
       }
     });
     dispatch({
--- a/src/sidebar/util/build-thread.js
+++ b/src/sidebar/util/build-thread.js
@@ -20,7 +20,7 @@
  * @prop {number} totalChildren - Computed count of this thread's immediate
  *       children. This count includes visually-hidden threads.
  * @prop {number} replyCount - Computed count of all replies to a thread
- * @prop {number} [depth] - The thread's depth in the hierarchy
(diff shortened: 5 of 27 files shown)


src/sidebar/store/modules/real-time-updates.js
```diff
@@ -102,21 +102,22 @@ function receiveRealTimeUpdates({
         ann.group === groups.selectors.focusedGroupId(getState()) ||
         route.selectors.route(getState()) !== 'sidebar'
       ) {
-        pendingUpdates[ann.id] = ann;
+        pendingUpdates[/** @type {string} */ (ann.id)] = ann;
```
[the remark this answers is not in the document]
LMS007: Would it be worth having 2 different types of Annotations. 

UnsavedAnnotations vs Annotations for example, or do you think that would make things more complex?

src/sidebar/util/build-thread.js
```diff
@@ -281,8 +282,9 @@ function hasVisibleChildren(thread) {
 const defaultOpts = {
   selected: [],
   expanded: {},
+  forcedVisible: [],
   sortCompareFn: (a, b) => {
-    return a.id < b.id;
+    return a.$tag < b.$tag;
```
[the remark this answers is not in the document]
LMS007: 👍

src/types/annotator.js
```diff
@@ -20,15 +20,17 @@
...
  * @prop {Object.<string, string[]>} [eprints]
  * @prop {Object.<string, string[]>} [facebook]
  * @prop {Object.<string, string[]>} [highwire]
  * @prop {Object.<string, string[]>} [prism]
  * @prop {Object.<string, string[]>} [twitter]
  * @prop {string} [favicon]
- * // pdf files
+ *
+ * // HTML + PDF.
  * @prop {string} [documentFingerprint]
```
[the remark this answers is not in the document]
LMS007: good to know

src/sidebar/components/annotation-body.js
```diff
@@ -28,7 +28,7 @@ import TagList from './tag-list';
  * @prop {string} text -
  *     The markdown annotation body, which is either rendered as HTML (if `isEditing`
  *     is false) or displayed in a text area otherwise.
- * @prop {MergedConfig} [settings]
+ * @prop {MergedConfig} settings
```
LMS007, line 31: Are all injected services in fact required? Or rather, are there any cases of a service prop existing where its not being used?

src/sidebar/components/menu-item.js
```diff
@@ -91,7 +91,7 @@ export default function MenuItem({
   let focusTimer = null;
 
   let renderedIcon = null;
-  if (icon !== 'blank') {
+  if (icon && icon !== 'blank') {
```
LMS007, line 94: Sure a typeof `undefined` is not equal to `string`, but the check (`!==`) is actually asking if the type is not equal in the first place. It just seems like a redundant check and we don't need to check if icon is `undefined` first. Or if we did, then this should be valid:

`if (icon && icon != 'blank')`

but then the linter complains! I feel like TS does not help in cases like this. Granted they are fewer than not, but I think it's important to see the forest for the trees with this typing stuff because it can become a burden at times.

src/sidebar/components/user-menu.js
```diff
@@ -18,11 +18,9 @@ import SvgIcon from '../../shared/components/svg-icon';
 
 /**
  * @typedef AuthState
- * @prop {'logged-in'|'logged-out'} status
```
LMS007, line 21: What happened to status & provider?

src/sidebar/util/thread-annotations.js
```diff
@@ -32,7 +32,7 @@ const sortFns = {
 function buildRootThread(threadState) {
   const selection = threadState.selection;
 
-  /** @type {BuildThreadOptions} */
+  /** @type {Partial<BuildThreadOptions>} */
```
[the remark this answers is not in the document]
LMS007: so its fair to say this is an ad-hoc `BuildThreadOptions` where all its props are optional?

src/types/api.js
```diff
@@ -43,9 +43,12 @@
  * TODO - Fill out remaining properties
  *
  * @typedef Annotation
- * @prop {string} [id]
- * @prop {string} [$tag] - A locally-generated unique identifier for annotations
- *       that have not been saved to the service yet (and thus do not have an id)
+ * @prop {string} [id] -
+ *   The server-assigned ID for the annotation. This is only set once the
+ *   annotation has been saved to the backend
```
LMS007, line 48: missing period at end

src/sidebar/store/modules/real-time-updates.js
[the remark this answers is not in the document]
robertknight: > UnsavedAnnotations vs Annotations for example, or do you think that would make things more complex?

That's a reasonable suggestion. I'm not certain about how the cost/benefit would work out. Something we could consider in future - especially if the unsaved vs. saved types looked very different.

src/sidebar/util/thread-annotations.js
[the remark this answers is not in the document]
robertknight: Yes, `Partial<T>` is like `T` but where all properties are optional. This is a subset of `BuildThreadOptions` where we only customize some of the options and let `buildThread` use defaults for other options.

src/sidebar/components/annotation-body.js
robertknight: > Are all injected services in fact required?

Yes. The service injection helper (`withServices`)  has no notion of optional services. It is however optional whether the consumer of the component overrides the service that `withServices` would otherwise inject. Currently only tests actually override the service.

src/sidebar/components/menu-item.js
robertknight: The check isn't redundant here - if icon is `undefined|string` then inside the `if` statement it could still be `undefined` or any string other than `"blank"`. This value was then being passed to a prop that requires a string. That's what TS was correctly complaining about. I agree that strict equality is not actually necessary here - it's just a convention to use it everywhere.

> Granted they are fewer than not, but I think it's important to see the forest for the trees with this typing stuff because it can become a burden at times.

I agree it is good to be pragmatic about these things and not let too much typing noise creep into the code.

src/sidebar/components/user-menu.js
robertknight: The code in this component doesn't use those props - it is fine if the type that is passed in is a superset of `AuthState` which does have them.

PR #2467, opened 2020-08-21
Copy xpath.coffee & utils.coffee modules to js

LMS007:
_Decided to take a slightly different approach with this conversation and leave all the coffee stuff in place while shoring up the unit tests for the newer js methods first._

Copy several utils required to convert xpath.coffee, util.coffee and eventually range.coffee into javascript. This first step is to just convert the modules straight over and add basic unit tests for them which previously were missing.

A few changes have been made:

- findChild is exposed in part so it can be moved to this module and also be unit tested easily.
- simpleXPathJQuery- will eventually go away, but as a first pass, its being ported
- jquery is kept in this context, but eventually all methods that expect a jquery object will be converted over to pure js.
- typechecking is not complete

--------

range-js.js will eventually replace range.coffee and then rename to range.js once all the methods are converted.

Include 2 initial functions in range-js
- nodeFromXPath
- xpathFromNode

------------

I created a side by side diff of just the methods here to help clarify the differences. (js left, coffee right)
https://kyle-keating.s3.us-west-1.amazonaws.com/xpath_coffee_diff.html

### DIFF
--- a/src/annotator/anchoring/range-js.js
+++ b/src/annotator/anchoring/range-js.js
@@ -1,0 +1,34 @@
+import { findChild, simpleXPathJQuery, simpleXPathPure } from './xpath-util';
+
+/**
+ * Wrapper for simpleXPath. Attempts to call the jquery
+ * version, and if that excepts, then falls back to pure js
+ * version.
+ */
+export function xpathFromNode(el, relativeRoot) {
+  let result;
+  try {
+    result = simpleXPathJQuery(el, relativeRoot);
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.log(
+      'jQuery-based XPath construction failed! Falling back to manual.'
+    );
+    result = simpleXPathPure(el, relativeRoot);
+  }
+  return result;
+}
+
+/**
+ * Finds an element node using an xpath relative to the document root.
+ */
+export function nodeFromXPath(xpath, root) {
+  const steps = xpath.substring(1).split('/');
+  let node = root;
+  steps.forEach(step => {
+    let [name, idx] = step.split('[');
+    idx = idx ? parseInt(idx.split(']')[0]) : 1;
+    node = findChild(node, name.toLowerCase(), idx);
+  });
+  return node;
+}
--- a/src/annotator/anchoring/test/range-js-test.js
+++ b/src/annotator/anchoring/test/range-js-test.js
@@ -1,0 +1,133 @@
+import $ from 'jquery';
+
+import { xpathFromNode, nodeFromXPath, $imports } from '../range-js';
+
+describe('annotator/anchoring/range-js', () => {
+  describe('xpathFromNode', () => {
+    let container;
+    let fakeSimpleXPathJQuery;
+    let fakeSimpleXPathPure;
+
+    beforeEach(() => {
+      container = document.createElement('div');
+      document.body.appendChild(container);
+      fakeSimpleXPathJQuery = sinon.stub().returns('/div[1]');
+      fakeSimpleXPathPure = sinon.stub().returns('/div[1]');
+
+      $imports.$mock({
+        './xpath-util': {
+          simpleXPathJQuery: fakeSimpleXPathJQuery,
+          simpleXPathPure: fakeSimpleXPathPure,
+        },
+      });
+    });
+
+    afterEach(() => {
+      container.remove();
+    });
+
+    it('calls `simpleXPathJQuery`', () => {
+      const xpath = xpathFromNode($(container), document.body);
+      assert.called(fakeSimpleXPathJQuery);
+      assert.equal(xpath, '/div[1]');
+    });
+
+    it('calls `simpleXPathPure` if `simpleXPathJQuery` throws an exception', () => {
+      sinon.stub(console, 'log');
+      fakeSimpleXPathJQuery.throws(new Error());
+      const xpath = xpathFromNode($(container), document.body);
+      assert.called(fakeSimpleXPathPure);
+      assert.equal(xpath, '/div[1]');
+      assert.isTrue(
+        // eslint-disable-next-line no-console
+        console.log.calledWith(
+          'jQuery-based XPath construction failed! Falling back to manual.'
+        )
+      );
+      // eslint-disable-next-line no-console
+      console.log.restore();
+    });
+  });
+
+  describe('nodeFromXPath', () => {
+    let container;
+    let fakeFindChild;
+
+    beforeEach(() => {
+      container = document.createElement('div');
+      document.body.appendChild(container);
+      fakeFindChild = sinon.stub().returns(document.body);
+
+      $imports.$mock({
+        './xpath-util': {
+          findChild: fakeFindChild,
+        },
+      });
+    });
+
+    afterEach(() => {
+      container.remove();
+    });
+
+    it('returns the last node returned from `findChild`', () => {
+      const span = document.createElement('span');
+      container.appendChild(span);
+      fakeFindChild.onFirstCall().returns(container);
+      fakeFindChild.onSecondCall().returns(span);
+      assert.equal(nodeFromXPath('/div[1]/span[1]', document.body), span);
+    });
+
+    [
+      {
+        xpath: '/div[1]',
+        params: [
+          {
+            name: 'div',
+            idx: 1,
+          },
+        ],
+      },
+      {
+        xpath: '/div[1]/span[3]/p[1]',
+        params: [
+          {
+            name: 'div',
+            idx: 1,
+          },
+          {
+            name: 'span',
+            idx: 3,
+          },
+          {
+            name: 'p',
+            idx: 1,
+          },
+        ],
+      },
+      {
+        xpath: '/DIV[2]/TEXT()[3]/SPAN[1]',
+        params: [
+          {
+            name: 'div',
+            idx: 2,
+          },
+          {
+            name: 'text()',
+            idx: 3,
+          },
+          {
+            name: 'span',
+            idx: 1,
+          },
+        ],
+      },
+    ].forEach(test => {
+      it('calls `findChild` with the following node names and indices', () => {
+        nodeFromXPath(test.xpath, document.body);
+        test.params.forEach(call => {
+          assert.calledWith(fakeFindChild, document.body, call.name, call.idx);
+        });
+      });
+    });
+  });
+});
--- a/src/annotator/anchoring/test/xpath-util-test.js
+++ b/src/annotator/anchoring/test/xpath-util-test.js
@@ -1,0 +1,313 @@
+import $ from 'jquery';
+
+import {
+  findChild,
+  getTextNodes,
+  getLastTextNodeUpTo,
+  getFirstTextNodeNotBefore,
+  simpleXPathPure,
+  simpleXPathJQuery,
+} from '../xpath-util';
+
+describe('annotator/anchoring/xpath-util', () => {
+  describe('#findChild', () => {
+    let container;
+
+    beforeEach(() => {
+      container = document.createElement('div');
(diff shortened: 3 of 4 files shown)


src/annotator/anchoring/range-js.js
```diff
@@ -0,0 +1,34 @@
+import { findChild, simpleXPathJQuery, simpleXPathPure } from './xpath-util';
```
LMS007, line 1: This is only called range-js so it does not collide with range.coffee imports. This will be renamed after range.coffee is pulled out.

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,320 @@
...
+    it('throws an error when the desired child is not found', () => {
+      container.innerHTML = `<div>text</div>`;
+      assert.throws(() => {
+        findChild(container, 'fake', 0);
+      }, 'XPath error: wanted child not found.');
+    });
+
+    describe('with children', () => {
+      beforeEach(() => {
+        container = document.createElement('div');
```
robertknight, line 40: Is the override of `container` and the `afterEach` needed here? It seems to me like you could just re-use the container created in the containing block.

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,320 @@
...
+        assert.equal(findChild(container, 'span', 1).innerText, 'text 2');
+      });
+
+      it('finds the desired span node at index 2', () => {
+        assert.equal(findChild(container, 'span', 2).innerText, 'text 3');
+      });
+    });
+  });
+
+  describe('#getTextNodes', () => {
```
robertknight, line 68:
```suggestion
  describe('getTextNodes', () => {
```

src/annotator/anchoring/xpath-util.js
```diff
@@ -0,0 +1,209 @@
...
+/**
+ * Get the node name for use in generating an xpath expression.
+ */
+function getNodeName(node) {
+  const nodeName = node.nodeName.toLowerCase();
+  let result = nodeName;
+  if (nodeName === '#text') {
+    result = 'text()';
+  }
+  if (nodeName === '#comment') {
```
robertknight, line 35: The range selectors that the client actually persists in the API only references elements. We also have some anchoring integration tests that reference `text()`, so the `#comment` and `#cdata-section` cases are not that important.

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,320 @@
...
+        <span>text 2</span>`;
+      const result = getTextNodes($(container));
+      assert.equal(result.length, 3);
+    });
+
+    it('finds basic text nodes and whitespace but ignores comments', () => {
+      container.innerHTML = `<span>text 1</span>
+        <!--span>text 2</span-->`;
+      const result = getTextNodes($(container));
+      assert.equal(result.length, 2);
```
robertknight, line 99: This test only checks the first returned node. One pattern I quite like for checking the length and some aspect of the contents of each item of an array in tests is to do something like the following:

```js
// Helper
const nodeValues = nodes => nodes.map(n => n.nodeValue)

// In individual tests
const result = getTextNodes(...)
assert.deepEqual(nodeValues(result), ['text 1', 'text 2'])
```

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,313 @@
...
+      const span = document.createElement('span');
+      container.appendChild(span);
+      assert.equal(getFirstTextNodeNotBefore(span), null);
+    });
+  });
+
+  describe('xpath', () => {
+    let container;
+    const html = `<div id="root">
+        <h1 id="h1-1">text</h1>
```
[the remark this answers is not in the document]
robertknight: I would suggest to solve the problem by removing the `comment()` and `cdata-section()` logic from the implementation. I don't think it is important.

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,313 @@
...
+        {
+          id: 'em-1',
+          xpath: '/div[1]/p[2]/em[1]',
+        },
+        {
+          id: 'li-3',
+          xpath: '/div[1]/span[1]/ul[1]/li[3]',
+        },
+      ].forEach(test => {
+        it('produces the correct xpath for the provided nodes', () => {
```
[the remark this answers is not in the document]
robertknight: >  It did not appear that xpathFromNode is capable of discovering text nodes in fact.

Looking at the documentation for jQuery's [children function](https://api.jquery.com/children/) it looks like it only deals with elements, so I think you're right.

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,320 @@
...
+          assert.equal(
+            simpleXPathJQuery($(node), document.querySelector('#root')),
+            test.xpath
+          );
+        });
+      });
+    });
+
+    describe('#simpleXPathPure', () => {
+      it('throws an expect if the provided node is not a descendant of root', () => {
```
robertknight, line 248:
```suggestion
      it('throws an error if the provided node is not a descendant of root', () => {
```

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,320 @@
...
+          let node = document.getElementById(test.id);
+          assert.equal(
+            simpleXPathJQuery($(node), document.querySelector('#root')),
+            test.xpath
+          );
+        });
+      });
+    });
+
+    describe('#simpleXPathPure', () => {
```
robertknight, line 247:
```suggestion
    describe('simpleXPathPure', () => {
```

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,320 @@
...
+    });
+
+    it('returns null if no text node exists', () => {
+      const span = document.createElement('span');
+      container.appendChild(span);
+      assert.equal(getLastTextNodeUpTo(span), null);
+    });
+  });
+
+  describe('#getFirstTextNodeNotBefore', () => {
```
robertknight, line 141:
```suggestion
  describe('getFirstTextNodeNotBefore', () => {
```

src/annotator/anchoring/test/xpath-util-test.js
```diff
@@ -0,0 +1,320 @@
...
+    it('finds basic text nodes and whitespace but ignores comments', () => {
+      container.innerHTML = `<span>text 1</span>
+        <!--span>text 2</span-->`;
+      const result = getTextNodes($(container));
+      assert.equal(result.length, 2);
+      assert.equal(result[0].nodeValue, 'text 1');
+    });
+  });
+
+  describe('#getLastTextNodeUpTo', () => {
```
robertknight, line 104:
```suggestion
  describe('getLastTextNodeUpTo', () => {
```

src/annotator/anchoring/xpath-util.js
```diff
@@ -0,0 +1,209 @@
...
+  });
+  return paths.get();
+}
+
+/**
+ * A simple XPath evaluator using only standard DOM methods which can
+ * evaluate queries of the form /tag[index]/tag[index].
+ */
+export function simpleXPathPure(nodes, relativeRoot) {
+  function getPathSegment(node) {
```
robertknight, line 84: `getPathSegment` doesn't reference any captured variables so it could be hoisted to the top level.

src/annotator/anchoring/xpath-util.js
```diff
@@ -0,0 +1,209 @@
...
+      xpath = getPathSegment(elem) + '/' + xpath;
+      elem = elem.parentNode;
+    }
+    xpath = '/' + xpath;
+    xpath = xpath.replace(/\/$/, '');
+    return xpath;
+  }
+  const paths = nodes.map((index, node) => {
+    return getPathTo(node);
+  });
```
robertknight, line 111:
```suggestion
  const paths = nodes.map((index, node) => getPathTo(node));
```

src/annotator/anchoring/xpath-util.js
```diff
@@ -0,0 +1,209 @@
...
+  if (prev) {
+    // eslint-disable-next-line no-unused-vars
+    return getLastTextNodeUpTo(prev);
+  } else {
+    return null;
+  }
+}
+
+/**
+ * Determine the first text node in or after the given jQuery node.
```
robertknight, line 185: From the implementation it looks like `n` is a DOM `Node` and not a jQuery node.

src/annotator/anchoring/xpath-util.js
```diff
@@ -0,0 +1,209 @@
...
+  };
+  return jq.map((index, node) => {
+    return flatten(getTextNodes(node));
+  });
+}
+
+/**
+ * Determine the last text node inside or before the given node.
+ */
+export function getLastTextNodeUpTo(n) {
```
robertknight, line 160:
```suggestion
export function getLastTextNodeUpTo(node) {
```

I realize this was adapted from the previous code. Our convention is to use single letter names only for things like loop variables or short closures where the context makes the meaning very obvious.

src/annotator/anchoring/xpath-util.js
```diff
@@ -0,0 +1,209 @@
...
+    return getLastTextNodeUpTo(prev);
+  } else {
+    return null;
+  }
+}
+
+/**
+ * Determine the first text node in or after the given jQuery node.
+ */
+export function getFirstTextNodeNotBefore(n) {
```
robertknight, line 187:
```suggestion
export function getFirstTextNodeNotBefore(node) {
```

As above, avoid very short names except for local uses within a function where the meaning is obvious.

src/annotator/anchoring/xpath-util.js
```diff
@@ -0,0 +1,209 @@
...
+
+  let rootNode = relativeRoot;
+
+  function getPathTo(node) {
+    let xpath = '';
+    let elem = node;
+    while (elem !== rootNode) {
+      if (!elem) {
+        throw new Error(
+          'Called getPathTo on a node which was not a descendant of @rootNode. ' +
```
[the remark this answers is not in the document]
robertknight: This string is referenced in a unit test so it looks like this is done.

src/annotator/anchoring/range-js.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Finds an element node using an xpath relative to the document root.
+ */
+export function nodeFromXPath(xpath, root) {
+  const steps = xpath.substring(1).split('/');
+  let node = root;
+  steps.forEach(step => {
+    let [name, idx] = step.split('[');
+    idx = idx ? parseInt(idx.split(']')[0]) : 1;
+    node = findChild(node, name.toLowerCase(), idx);
```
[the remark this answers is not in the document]
LMS007: sounds good

src/annotator/anchoring/test/range-js-test.js
```diff
@@ -0,0 +1,133 @@
+import $ from 'jquery';
+
+import { xpathFromNode, nodeFromXPath, $imports } from '../range-js';
+
+describe('annotator/anchoring/range-js', function () {
```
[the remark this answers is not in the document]
LMS007: yep!, I keep forgetting that :(

src/annotator/anchoring/test/xpath-util-test.js
LMS007: for sure!

src/annotator/anchoring/test/xpath-util-test.js
LMS007: done!

PR #2486, opened 2020-08-28
Convert `Host` class to JS

robertknight:
This is a straightforward conversion with decaffeinate, followed by
manual fixups of lint errors etc.


src/annotator/host.js
```diff
@@ -0,0 +1,102 @@
...
+      }
+    }
+
+    // Make a copy of all config settings except `config.sidebarAppUrl`, the app base URL,
+    // and `config.pluginClasses`
+    const configParam =
+      'config=' +
+      encodeURIComponent(
+        JSON.stringify(
+          Object.assign({}, config, {
```
LMS007, line 38: might be able to refactor this a tad more. Is there any reason to initialize sidebarAppUrl or pluginClasses to undefined? seems like that could just be omitted.

either `JSON.stringify(config)` or `JSON.stringify({...config})` if copy is needed

src/annotator/host.js
```diff
@@ -0,0 +1,102 @@
...
+
+    let externalContainer = null;
+
+    if (config.externalContainerSelector) {
+      // Use the native method to also validate the input
+      externalContainer = document.querySelector(
+        config.externalContainerSelector
+      );
+    }
+    super(...arguments);
```
LMS007, line 65: any reason this can't be the first statement in the ctor?

src/annotator/test/host-test.js
```diff
@@ -0,0 +1,122 @@
...
+    );
+    if (!element) {
+      element = document.createElement('div');
+    }
+    return new Host(element, config);
+  };
+
+  beforeEach(() => {
+    // Disable any Host logging.
+    sandbox.stub(console, 'log');
```
LMS007, line 24: Do you have a preference between sandbox or

`sinon.stub(console, 'log');`
`console.log.restore()` 

Just wondering when to use one or the other

src/annotator/test/host-test.js
```diff
@@ -0,0 +1,122 @@
...
+        {
+          $highlight: true,
+        },
+      ]);
+      assert.notCalled(frame.contentWindow.focus);
+    });
+  });
+
+  describe('config', () => {
+    it('disables highlighting if showHighlights: false is given', done => {
```
LMS007, line 91: I don't see showHighlights condition anywhere in this PR

LMS007 (approved):
Just a few tiny suggestions / questions, then good to go!
👍

src/annotator/host.js
robertknight: The comment above this line explains that these parameters should not be included in the config passed in the sidebar's fragment URL. I'll rewrite this to make it clearer.

src/annotator/host.js
robertknight: I can't see any problem with doing so. I'll make this change.

src/annotator/test/host-test.js
robertknight: If you're only stubbing one or two functions, `sinon.stub(...)` is simpler. If you're stubbing a lot of methods, a sandbox provides a convenient way to undo all the stubs at once IIRC. If you find yourself stubbing a lot of methods though, you might want to consider an alternative approach such as injecting a fake version of the object that has the stubbed methods (eg. like we do with `window` in a few places)

src/annotator/test/host-test.js
robertknight: The comment refers to passing `showHighlights: false` in the `createHost(...)` call on the line below.

src/annotator/host.js
robertknight: Update: There _is_ a reason that `super(...)` needs to be called part-way down the function. It turns out that the `super(...)` call results in the `BucketBar` constructor being called and that class in turn assumes that it is called after the bucket bar's container element (`.annotator-frame`) is in the DOM. This needs refactoring, but I'll do that as a separate task.

robertknight:
I've revised a couple of things in response to comments. It turns out that having the `super(...)` call part-way down the constructor is currently needed for non-obvious reasons that will need more substantial refactoring to fix. I added a comment to explain why.

PR #2497, opened 2020-09-02
Range coffee to js (1/3)

LMS007:
- Separate each range class into its own module. There is a circular dependency among the 3 range modules which is handled with an inline require() statement in range-browser.
- Eliminate the custom RangeError class from coffeescript and just use native RangeError
- Add unit tests for range-normalized

--------

This is phase 1/3, The `range-browser` and range-serialized are minimally stubbed out. The next two PRs will fill those modules in & their new unit tests. Note that none of these modules are actually hooked up to any code that runs and only the units tests run. After `range.coffee` is totally converted, I'll create a final PR to hook up this block of work in code. I do have a temporary branch that is indeed hooked up to simply ensure the app does appear to build, run and all tests work (most importantly the `html-test` and `pdf-test` files. So I know this conversion is at least in theory working. But for this PR, only the unit tests matter at this point. Don't worry about totally integration.

**Notes**
- The module names are prefixed with `range-` so they appear in order in the file view. Note that the classes names are inverted for legacy reasons. We could change that I kept it the same for the conversion.
- The code is largely directly ported from coffeescript as are the comments.
- Type checking will be done in a later pass

### DIFF
--- a/src/annotator/anchoring/range-normalized.js
+++ b/src/annotator/anchoring/range-normalized.js
@@ -1,0 +1,170 @@
+import $ from 'jquery';
+
+import { xpathFromNode } from './range-js';
+import { SerializedRange } from './range-serialized';
+import { getTextNodes } from './xpath-util';
+
+/**
+ * Public: A normalized range is most commonly used throughout the annotator.
+ * its the result of a deserialized SerializedRange or a BrowserRange with
+ * out browser inconsistencies.
+ */
+export default class NormalizedRange {
+  /**
+   * Public: Creates an instance of a NormalizedRange.
+   *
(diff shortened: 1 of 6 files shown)


src/annotator/anchoring/range-normalized.js
```diff
@@ -0,0 +1,174 @@
...
+   */
+  toRange() {
+    const range = document.createRange();
+    range.setStartBefore(this.start);
+    range.setEndAfter(this.end);
+    return range;
+  }
+}
+
+export const Normalized = {
```
LMS007, line 172: I wrapped this in a namespace because `require('./range-normalized').default` was not working with the ES6 class in `range-browser.js`

PR #2522, opened 2020-09-11
Convert range classes from coffeescript to js

LMS007:
**WIP**

In short, no unit tests for any of these modules existed. All tests are new and may not all make sense. I did the best I could trying to figure out how all this stuff worked. The goal was make as much sense as possible with 100% code coverage.

**_Note that this code is not hooked up yet so it won't run even if commited._**

It's also possible that once we hook this up to the full integration tests (html-test.js) that something might show up. I don't have 100% confidence this is perfect, but it may not matter because Its close and we should see any issues come up at that time -- I expect them to be minor.

**What is not done yet:**
1. typing.
2. major reworking of comments.
3. various bugs relating to range.
4. remove jquery

Because of the size of this PR, I don't want to complicate it with any of the above issues. They can be done with better confidence in later passes.

TODO:

- [ ] give this a fresh look in the morning. I'm tired right now
- [ ] decide if we should spend the extra time to make 100% of the tests in serialize() and normalize() pure unit tests or continue allow them to use BrowserRange, NormalizedRange, SerializedRange without stubbing the responses entirely.

**(2) Commits:**
---------------

Convert range classes from coffeescript to js.

 1. BrowserRange
 2. NormalizedRange
 3. SerializedRange

Converted classes no longer contain serialize() and normalize() methods. This is done to remove circular dependencies from the classes and break up complex serialization and normalization logic into separate modules and functions which is easier to read and test.

----------------

Add normalize and serialize converted modules

These modules contain only 1 function each and replace the 3 serialize() and normalize() functions previously found in each of the 3 coffeescript range classes. The fundamental difference is that these functions are not bound to a specific range class type and instead take a range instance param and then decided which conversion algorithm to run based on the type of instance.

This removes the circular dependencies each of the 3 range classes had and additionally groups all complex serialize and normalize logic into discrete modules.

### DIFF
--- a/src/annotator/anchoring/normalize.js
+++ b/src/annotator/anchoring/normalize.js
@@ -1,0 +1,224 @@
+import $ from 'jquery';
+
+import { BrowserRange } from './range-browser';
+import { NormalizedRange } from './range-normalized';
+import { nodeFromXPath } from './range-js';
+import { SerializedRange } from './range-serialized';
+import {
+  getFirstTextNodeNotBefore,
+  getLastTextNodeUpTo,
+  getTextNodes,
+} from './xpath-util';
+
+/**
+ * Convert a BrowserRange or SerializedRange instance to a new NormalizedRange instance.
+ * Passing a NormalizedRange instance range param will return itself.
+ */
+export function normalize(range, root) {
+  // BrowserRange
+  if (range instanceof BrowserRange) {
+    // Works around the fact that browsers don't generate
+    // ranges/selections in a consistent manner. Some (Safari) will create
+    // ranges that have (say) a textNode startContainer and elementNode
+    // endContainer. Others (Firefox) seem to only ever generate
+    // textNode/textNode or elementNode/elementNode pairs.
+    if (range.tainted) {
+      throw new Error('You may only call normalize() once on a BrowserRange!');
+    } else {
+      range.tainted = true;
+    }
+    const rangeTemp = {};
+    // Look at the start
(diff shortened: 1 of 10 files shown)


src/annotator/anchoring/normalize.js
```diff
@@ -0,0 +1,224 @@
...
+      // Do we really have to cut?
+      if (
+        !rangeTemp.start.nextSibling ||
+        rangeTemp.start.nodeValue.length > rangeTemp.startOffset
+      ) {
+        // Yes. Cut.
+        normalRange.start = rangeTemp.start.splitText(rangeTemp.startOffset);
+      } else {
+        // Avoid splitting off zero-length pieces.
+        normalRange.start = getFirstTextNodeNotBefore(
```
LMS007, line 96: This was not a straight conversion, the original coffeescript did not call getFirstTextNodeNotBefore(), it just set normalRange.start to rangeTemp.start.nextSibling. but nextSibling may not always be a text node. I discovered this issue while building out unit tests. I'm not 100% I'm right about this, but it could be a long standing bug, or just my misunderstanding.

src/annotator/anchoring/range-browser.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Public: Creates a wrapper around a range object obtained from a DOMSelection.
```
LMS007, line 2: Remove "Public: " here and all places in this commit.

Issue #1968, opened 2020-03-27
Merge coercions into paramWhiteList


LMS007:
We could clean up this file a bit and make the paramWhiteList more complete by making it an object whose keys are allowed parameters and values are coercion functions or validation functions.

Perhaps we also need both. 1. coerce then 2. validate.

For example. `"true"` may need to be `true` down stream, but `"true"` is not necessarily invalid.
So a coerce first, then validation check second would make sense.

We ideally should coerce / validate every config value coming in.

robertknight:
The original context around this issue has been lost, so I'm going to close it.

I will add some context for future reference:

- Via needs to support being passed configuration which is then relayed to the client. This is done by passing query params to Via, with a special `via.` prefix, which then get mapped to client configuration in JSON format
- The developer working on Via didn't want that service to have to know about the types of different client configuration values, instead the idea was that Via could just always pass configuration values as strings.
- Hence we added code in the client to coerce string representations of values to the expected types

The downside of this is all the usual problems that can occur when APIs accept stringly-typed inputs.

Personally I would much rather than we change Via so that either it does know about the types of client config values, or the data is passed to Via in such a way that the type info is included.

PR #2528, opened 2020-09-12
Improve type coercions for paramWhiteList

LMS007:
- Rework coercions so they work more like react props rather than function calls
- Improve type-coercion capability to allow for nullish values. This is similar to propTypes `.isRequired`
- Add coercions to remaining config settings host-config.

partially fixes: #1968

-----------

Example usage..

```
  const paramWhiteList = {
    annotations: coercions.toString.orNull,
    group: coercions.toString.orNull,
    query: coercions.toString.orNull,
    appType: coercions.toString.orNull,
    openSidebar: coercions.toBoolean,
    ...
```
When reviewing this, the diff is hard to follow and makes little sense because the `type-coercions.js` is totally restructured. 

Image diff (master is on the right, this branch on the left) --it's a screenshot of the host-config after its run through the  `paramWhiteList`. They are identical.

Screen Shot 2020-09-12 at 2 49 55 PM

### DIFF
--- a/src/shared/type-coercions.js
+++ b/src/shared/type-coercions.js
@@ -9,12 +9,47 @@
  */
 
 /**
- * Returns a boolean
+ * Returns the value as a string or an empty string if the
+ * value undefined, null or otherwise falsely.
  *
- * @param {any} value - initial value
- * @return {boolean}
+ * @param {any} value - Initial value
  */
-export function toBoolean(value) {
+function toStringShim(value) {
+  if (value && typeof value.toString === 'function') {
+    return value.toString();
+  }
+  return '';
+}
+
+/**
+ * Returns either an string or null.
+ *
+ * @type {(value: any) => string|null}
+ */
+toStringShim.orNull = value => (value === null ? null : toStringShim(value));
+
+/**
+ * Returns either an integer or NaN.
+ *
+ * @param {any} value - Initial value
+ */
+function toIntegerShim(value) {
+  return parseInt(value);
+}
+
+/**
+ * Returns either an integer or null.
+ *
+ * @type {(value: any) => number|null}
+ */
+toIntegerShim.orNull = value => (value === null ? null : toIntegerShim(value));
+
+/**
+ * Returns a boolean.
+ *
+ * @param {any} value - Initial value
+ */
+function toBooleanShim(value) {
   if (typeof value === 'string') {
     if (value.trim().toLocaleLowerCase() === 'false') {
       // "false", "False", " false", "FALSE" all return false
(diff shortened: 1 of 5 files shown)


src/shared/type-coercions.js
```diff
@@ -55,15 +114,61 @@ export function toObject(value) {
...
+  return toObjectShapeShim;
+};
+
+// Exported coercion types
+const coercions = {
+  toString: toStringShim,
+  toInteger: toIntegerShim,
+  toObject: toObjectShim,
+  toBoolean: toBooleanShim,
+  custom: customShim,
```
LMS007, line 170: toObject.shape and custom are similar, the former will ensure the object is in fact an object before the custom shape is called, whereas custom take the raw value of any time. I felt both were different enough to warrant both. Currently, toObject.shape is not used anywhere, but I suspect we will use it if we extend this.

src/shared/type-coercions.js
```diff
@@ -55,15 +114,61 @@ export function toObject(value) {
...
+  any: v => v,
```
LMS007, line 171: This is more or less throwing my hands up because the values need to be re-thunk'ed a bit. 
Perfect example is `showHighlights` where it is either a specific string or a boolean. I think it means we need to refactor the API and then force this into a more strict type of one or the other. Worst case, `showHighlights` could be a custom if we need to still support legacy values.

src/sidebar/test/host-config-test.js
```diff
@@ -65,9 +65,9 @@ describe('hostPageConfig', function () {
     assert.deepEqual(hostPageConfig(window_), {});
   });
 
-  it('ignores `null` values in config', function () {
+  it('ignores `undefined` values in config', function () {
```
LMS007, line 68: a `null` value does now have meaning so only `undefind` will be ignored.

robertknight:
Rebased on master.

robertknight (commented):
Hi @LMS007 - I finally got around to looking at this, sorry for the delay.

The question at the top of my mind here is whether we want to silently _coerce_ values of the "wrong" type or _validate_ them and simply ignore values of the wrong type.

IIRC the original motivation for introducing coercion of client settings was to make it easier to support passing additional configuration from Via to the client, with fewer changes to Via itself.

However a consequence of this is that if a publisher accidentally passes an incorrect value, we will silently accept it and it may or may not do what they expect. We are then required to leave with that loose handling of settings forever - since removing it could potentially break existing web pages.

I have a general preference for being conservative in what is accepted because it is always easier to relax constraints than to introduce them later. Let's catch up about it in Slack.

PR #2546, opened 2020-09-17
Dumb fix for `nzz.ch` site

esanzgar:
I will take a look later at the smart solution (get the compute z-index value of all the parent elements).


esanzgar:
Unintentionally, I pushed this change to master.

PR #2560, opened 2020-09-22
Remove unused code in `Guest` class

robertknight:
This PR removes unused and in some cases untested code from the `Guest` class, which should simplify future refactoring.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -210,23 +210,10 @@ export default class Guest extends Delegator {
   }
 
   addPlugin(name, options) {
-    if (this.plugins[name]) {
-      console.error('You cannot have more than one instance of any plugin.');
-    } else {
-      const Klass = this.options.pluginClasses[name];
-      if (typeof Klass === 'function') {
-        this.plugins[name] = new Klass(this.element[0], options);
-        this.plugins[name].annotator = this;
-        this.plugins[name].pluginInit?.();
-      } else {
-        console.error(
-          'Could not load ' +
-            name +
-            ' plugin. Have you included the appropriate <script> tag?'
-        );
-      }
-    }
-    return this; // allow chaining
+    const Klass = this.options.pluginClasses[name];
+    this.plugins[name] = new Klass(this.element[0], options);
+    this.plugins[name].annotator = this;
+    this.plugins[name].pluginInit?.();
   }
 
   // Get the document info
@@ -329,8 +316,6 @@ export default class Guest extends Delegator {
       $(element).contents().insertBefore(element);
       $(element).remove();
     });
-
-    this.element.data('annotator', null);
 
     for (let name of Object.keys(this.plugins)) {
       this.plugins[name].destroy();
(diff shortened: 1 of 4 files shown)


src/annotator/guest.js
```diff
@@ -67,11 +67,9 @@ const IGNORE_SELECTOR = '[class^="annotator-"],[class^="hypothesis-"]';
 
 export default class Guest extends Delegator {
   constructor(element, config, anchoring = htmlAnchoring) {
-    // TODO - Find out if `defaultConfig` actually does anything and remove it
-    // if not.
     const defaultConfig = {
+      // This causes the `Document` plugin to be initialized.
       Document: {},
-      TextSelection: {},
```
robertknight, line 74: The empty objects here cause various "plugin" modules to be initialized. There is no `TextSelection` plugin any more however.

src/annotator/guest.js
```diff
@@ -210,23 +208,10 @@ export default class Guest extends Delegator {
...
-      const Klass = this.options.pluginClasses[name];
-      if (typeof Klass === 'function') {
-        this.plugins[name] = new Klass(this.element[0], options);
-        this.plugins[name].annotator = this;
-        this.plugins[name].pluginInit?.();
-      } else {
-        console.error(
-          'Could not load ' +
-            name +
-            ' plugin. Have you included the appropriate <script> tag?'
```
robertknight, line 225: This is a vestige of the Annotator project where consumers could specify their own plugins. In Hypothesis this doesn't happen. Any errors here indicate a programming error elsewhere in annotator and I'd prefer that they just fail loudly.

src/annotator/guest.js
```diff
@@ -330,8 +315,6 @@ export default class Guest extends Delegator {
       $(element).remove();
     });
 
-    this.element.data('annotator', null);
```
robertknight, line 333: Nothing used this property.

src/annotator/guest.js
```diff
@@ -538,41 +521,6 @@ export default class Guest extends Delegator {
     return this.createAnnotation({ $highlight: true });
   }
 
-  // Create a blank comment (AKA "page note")
-  createComment() {
```
robertknight, line 542: This method was never used in app code. The "Page Note" button in the sidebar's vertical toolbar calls `createAnnotation` and either creates an annotation or note depending on whether there is a selection.

src/annotator/guest.js
```diff
@@ -538,41 +521,6 @@ export default class Guest extends Delegator {
...
-    return annotation;
-  }
-
-  // Public: Deletes the annotation by removing the highlight from the DOM.
-  // Publishes the 'annotationDeleted' event on completion.
-  //
-  // annotation - An annotation Object to delete.
-  //
-  // Returns deleted annotation.
-  deleteAnnotation(annotation) {
```
robertknight, line 564: This code was never called in the app. Annotations are always deleted from the sidebar rather than the annotator side of the app. This in turn triggers a call to the `detach` method in this class.

src/annotator/pdf-sidebar.js
```diff
@@ -6,15 +6,11 @@ import Sidebar from './sidebar';
...
 const defaultConfig = {
-  TextSelection: {},
   PDF: {},
   BucketBar: {
     container: '.annotator-frame',
     scrollables: ['#viewerContainer'],
   },
-  Toolbar: {
-    container: '.annotator-frame',
-  },
```
robertknight, line 17: There is no `Toolbar` plugin any more

src/annotator/sidebar.js
```diff
@@ -19,8 +19,6 @@ import { ToolbarController } from './toolbar';
 const MIN_RESIZE = 280;
 
 const defaultConfig = {
-  Document: {},
-  TextSelection: {},
```
robertknight, line 23: There is no `TextSelection` plugin any more and the `Document` config is already supplied by the base class.

lyzadanger (approved):
:heart: :heart: :heart:

PR #2562, opened 2020-09-22
display draft annotation while saving on backend

esanzgar:
resolves 2538

I have tested under offline conditions and it's OK. When offline, the draft annotation displays briefly and then it reverts to the editor (plus warning notice).

I added a couple of new tests to check the additional functionality.


lyzadanger (approved):
This is great!

PR #2563, opened 2020-09-22
range.coffee to js conversion

LMS007:
Rename range-js to xpath-evaluate

--------

Convert range.coffee to range.js
- Remove xpath.coffee as it is no longer needed
- Add missing unit tests for range.js module

-------

Per discussion, this is mostly a straight conversion of `range.coffee` unlike the previous 2 attempts. Note that some of the functions from `range.coffee` were moved into `range-js.js` in a previous effort. So if you compare `range.coffee` to `range.js` they won't map exactly 1:1, but that is expected.  I also renamed `range-js.js` (the name of that file was only temporary) to `xpath-evaluate.js` --that name more closely represents what those functions do. Note: we also already have a `range-util.js` file so I don't want to conflate that.

previous PR attempts to convert this module for reference
https://github.com/hypothesis/client/pull/2522
https://github.com/hypothesis/client/pull/2497

TODO for future PRs

- [ ] - Add typing
- [ ] - Rework comments as needed
- [ ] - Major refactor to simply and separate the serialize/normalize logic from range finding/trimming/discovery. In this effort range.js shall hopefully be broken down into separate smaller functionality.


src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,497 @@
...
+      // Do we really have to cut?
+      if (
+        !range.start.nextSibling ||
+        range.start.nodeValue.length > range.startOffset
+      ) {
+        // Yes. Cut.
+        normalRange.start = range.start.splitText(range.startOffset);
+      } else {
+        // Avoid splitting off zero-length pieces.
+        normalRange.start = getFirstTextNodeNotBefore(range.start.nextSibling);
```
LMS007, line 147: This was not a straight conversion, the original coffeescript did not call getFirstTextNodeNotBefore(), it just set normalRange.start to rangeTemp.start.nextSibling. but nextSibling may not always be a text node. I discovered this issue while building out unit tests. I'm not 100% I'm right about this, but it could be a long standing bug, or just my misunderstanding.

src/annotator/anchoring/pdf.js
```diff
@@ -470,8 +469,8 @@ export function describe(root, range, options = {}) {
     );
 
     const quoteRange = document.createRange();
-    quoteRange.setStartBefore(startRange.start);
-    quoteRange.setEndAfter(endRange.end);
+    quoteRange.setStartBefore(startRange?.start);
+    quoteRange.setEndAfter(endRange?.end);
```
robertknight, line 473: What will happen here if `startRange` is in fact nullish? Are we in fact just assuming that it is not?

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+        it('cuts the text node if there is a sibling but text node length > `startOffset`', () => {
+          const browserRange = createBrowserRange({
+            startContainer: container.querySelector('#p-2').firstChild,
+            startOffset: 1,
+          });
+          const normalizedRange = browserRange.normalize();
+          assert.equal(normalizedRange.start.data, 'ext 2a');
+        });
+
+        it('does not cut the text node', () => {
```
robertknight, line 205:
```suggestion
        it('does not cut the text node', () => {
```

Shouldn't there be a "when <condition>" at the end of this test description?

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+            startContainer: container.querySelector('#p-2').firstChild,
+            startOffset: 7,
+            endOffset: 14,
+          });
+          const normalizedRange = browserRange.normalize();
+          assert.equal(normalizedRange.start.data, 'text 2b');
+        });
+      });
+
+      context('the whole selection is inside one text element', () => {
```
robertknight, line 216:
```suggestion
      context('the whole selection is inside one text node', () => {
```

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+              endOffset: test.endOffset,
+            });
+            const normalizedRange = browserRange.normalize();
+            assert.equal(normalizedRange.start.data, test.result);
+          });
+        });
+      });
+
+      context('common ancestor is not ELEMENT_NODE', () => {
+        it('corrects the commons ancestor to the first non text node', () => {
```
robertknight, line 243: Does "first non text node" mean "first non-text ancestor"?

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+          startOffset: 0,
+          end: '/section[1]/span[1]/p[1]',
+          endOffset: 1,
+        });
+      });
+
+      it('converts BrowserRange to SerializedRange instance with `ignoreSelector` condition', () => {
+        const browserRange = createBrowserRange();
+        const result = browserRange.serialize(container, '#p-3');
+        // End xpath shall ignore the provided selector.
```
robertknight, line 272: This seems odd. Why is only the end xpath affected?

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+          bound: '#p-2',
+          textNodes: ['text 2a', 'text 2b'],
+        },
+        {
+          bound: '#section-1',
+          textNodes: ['text 1', 'text 2a', 'text 2b', 'text 3'],
+        },
+      ].forEach(test => {
+        it('limits range to the bounding node', () => {
+          let limit = createNormalizedRange().limit(
```
robertknight, line 307:
```suggestion
          const limit = createNormalizedRange().limit(
```

Doesn't look like this is re-assigned.

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+      });
+
+      afterEach(() => {
+        document.createRange.restore();
+      });
+
+      it('converts normalized range to native range', () => {
+        const range = createNormalizedRange();
+        const nativeRange = range.toRange();
+        assert.deepEqual(nativeRange, document.createRange());
```
robertknight, line 341: Is this assert intended to check the type of the range. If so, I suggest `assert.instanceOf(nativeRange, Range)`?

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+        assert.calledWith(fakeSetStartBefore, range.start);
+        assert.calledWith(fakeSetEndAfter, range.end);
+      });
+    });
+
+    describe('#normalize', () => {
+      it('returns an instance of itself', () => {
```
robertknight, line 348:
```suggestion
      it('returns itself', () => {
```

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+        const initialRange = createNormalizedRange();
+        const normalizedRange = initialRange.normalize();
+        assert.deepEqual(initialRange, normalizedRange);
```
robertknight, line 351:
```suggestion
        assert.equal(initialRange, normalizedRange);
```

More than just "an instance of `NormalizedRange`" it returns just `this`.

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+        it("throws an error if it can't find a valid start offset", () => {
+          const serializedRange = createSerializedRange({
+            startOffset: 99,
+          });
+          assert.throws(() => {
+            serializedRange.normalize(container);
+          }, "Couldn't find offset 99 in element /section[1]/p[1]");
+        });
+
+        it("throws an error if it can't find a valid start offset", () => {
```
robertknight, line 464:
```suggestion
        it("throws an error if it can't find a valid end offset", () => {
```

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+        it('throws a range error if nodeFromXPath() throws an error', () => {
+          fakeNodeFromXPath.throws(new Error('error message'));
+          const serializedRange = createSerializedRange();
+          assert.throws(() => {
+            serializedRange.normalize(container);
+          }, 'Error while finding start node: /section[1]/p[1]: Error: error message');
+        });
+
+        it('throws a range error if nodeFromXPath() does not return a node', () => {
+          fakeNodeFromXPath.returns(null);
```
robertknight, line 500: When does `nodeFromXPath` return `null` as opposed to throwing an error?

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,532 @@
...
+        // The copied instance shall be identical to the initial.
+        assert.deepEqual(result, serializedRange);
+      });
+
+      it('converts a SerializedRange to a new SerializedRange instance with `ignoreSelector` condition', () => {
+        const serializedRange = createSerializedRange();
+        // Mock the normalized return value
+        // fakeNormalize
+        //   .withArgs(serializedRange, container)
+        //   .returns(normalizedRange);
```
robertknight, line 523: This looks like commented-out code that should have been removed.

src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,494 @@
...
+
+import { xpathFromNode, nodeFromXPath } from './xpath-evaluate';
+import {
+  getFirstTextNodeNotBefore,
+  getLastTextNodeUpTo,
+  getTextNodes,
+} from './xpath-util';
+
+const Range = {};
+export default Range;
```
robertknight, line 11: I believe the `Range` object here exists because this code pre-dated modules originally. A more idiomatic conversion would be to make `sniff`, `BrowserRange`, `SerializedRange` and `NormalizedRange` named exports (ie. `export function sniff(...)`) etc.

If you do this you'll run into an ESLint warning about using things before defining them. With `sniff` that is easy to fix by moving it to the bottom of the file. For the other cases you'll need to suppress some of the warnings due to the circular dependencies.

src/annotator/anchoring/range.js
robertknight: What was the test case?

src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,494 @@
...
+        normalRange.start = range.start.splitText(range.startOffset);
+      } else {
+        // Avoid splitting off zero-length pieces.
+        normalRange.start = getFirstTextNodeNotBefore(range.start.nextSibling);
+      }
+    } else {
+      normalRange.start = range.start;
+    }
+
+    // is the whole selection inside one text element ?
```
robertknight, line 153:
```suggestion
    // Is the whole selection inside one text element ?
```

Small casing fixup.

src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,494 @@
...
+    if (!nodes.length) {
+      return null;
+    }
+
+    this.start = nodes[0];
+    this.end = nodes[nodes.length - 1];
+
+    const startParents = $(this.start).parents();
+
+    for (let parent of Array.from($(this.end).parents())) {
```
robertknight, line 252: Can you check whether each of the `Array.from(...)` calls in this file are needed? If you used `decaffeinate` to produce this file it will have generated them quite liberally, including in cases where we know the value is definitely going to be an array.

src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,494 @@
...
+  }
+
+  /**
+   * Creates a concatenated String of the contents of all the text nodes
+   * within the range.
+   *
+   * Returns a String.
+   */
+  text() {
+    return Array.from(this.textNodes())
```
robertknight, line 317: Per above comment, it looks like `Array.from` is redundant here.

src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,494 @@
...
+  }
+
+  /**
+   * Fetches only the text nodes within the range.
+   *
+   * Returns an Array of TextNode instances.
+   */
+  textNodes() {
+    const textNodes = getTextNodes($(this.commonAncestor));
+    const [start, end] = Array.from([
```
robertknight, line 329: The `Array.from` is redundant here, as is the only array literal given that it is immediately destructured.

src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,494 @@
...
+      let length = 0;
+      let targetOffset = this[p + 'Offset'];
+
+      // Range excludes its endpoint because it describes the boundary position.
+      // Target the string index of the last character inside the range.
+      if (p === 'end') {
+        targetOffset--;
+      }
+
+      for (let tn of Array.from(getTextNodes($(node)))) {
```
robertknight, line 417: This `Array.from` looks redundant.

robertknight (commented):
Hi Kyle,

I've had a look through. Although there are many simplifications I think we could make to the code, this looks good for an initial conversion. I flagged up a few minor issues which I'll leave you to fix up. In particular:

- There are redundant uses of `Array.from`. I guess these came from decaffeinate?
- I suggest changing the exports of `range.js` to normal ES exports rather than using the `Range` object. This will necessitate fix-ups in a few other modules.
- I flagged up a few minor issues in the tests

src/annotator/anchoring/range.js
LMS007: Yes, to the best of my knowledge when flushing out the unit tests for complete coverage, I came across a case were the  range.start.nextSibling was not sufficient. I think it was pointing to a <br> tag but it really needed to look before that to the next node.


The test on line 205: "does not cut the text node" will execute this code. If you remove the function call `getFirstTextNodeNotBefore` then the test will fail, but its the way that it failed that I was suspicious of because `normalizedRange.start` (in that case)  points to the <br> tag. I suspect it needed to actually point to a text node which is why I made the change and added the `getFirstTextNodeNotBefore` call.

src/annotator/anchoring/test/range-test.js
LMS007: yes, I noticed a mistake in a previous test description too, fixed both.

src/annotator/anchoring/test/range-test.js
LMS007: yes

src/annotator/anchoring/test/range-test.js
LMS007: Only because the xpath with #p-3 only relates to the end because that is how its set up `createBrowserRange`, but if we had a common node like <span class="foo"> in the DOM and that was found in both start and end xpaths, then it should ignore that in both the start and end xpaths if ".foo" is passed as the ignoreSelector param.

What I will do is change the ignore selector to just `p` and it will affect both the start and end and make it clear.

src/annotator/anchoring/test/range-test.js
LMS007: strangely, `console.log(document.createRange() instanceof Range)` this is false.

So I can't do this test.

I can't make sense of the docs either, they claim createRange returns a Range object.
https://developer.mozilla.org/en-US/docs/Web/API/Range

src/annotator/anchoring/test/range-test.js
LMS007: perhaps it does not, I just converted the code over that checked for null hence this test, but upon closer inspection, it does not ever return a null so I'll rip that code out.

src/annotator/anchoring/range.js
LMS007: I'll make the named export, but I'll still need to set the Range={} object locally because of the 3-way dependency of the Range* classes, unless there is a more appropriate way to do that?

src/annotator/anchoring/pdf.js
LMS007: So limit() can return null and it was never a case being handled here. Its still not handled per se, but we could perhaps throw an error. I'm just not sure what such a case would actually look like in the wild so I did not want to make any assumptions.

src/annotator/anchoring/range.js
robertknight: I see. So it looks like this happens when the range to be normalized starts at the end of a text node? OK, this would be a legitimate range that a browser might create as part of the selection.

src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,490 @@
...
+   */
+  serialize(root, ignoreSelector) {
+    return this.normalize().serialize(root, ignoreSelector);
+  }
+};
+
+/**
+ * A normalized range is most commonly used throughout the annotator.
+ * its the result of a deserialized SerializedRange or a BrowserRange with
+ * out browser inconsistencies.
```
robertknight, line 196:
```suggestion
 * its the result of a deserialized SerializedRange or a BrowserRange
 * without browser inconsistencies.
```

src/annotator/anchoring/range.js
```diff
@@ -0,0 +1,490 @@
...
+   * character offset), which can be easily stored in a database.
+   *
+   * root -           The root Element relative to which XPaths should be calculated
+   * ignoreSelector - A selector String of elements to ignore. For example
+   *                  elements injected by the annotator.
+   *
+   * Returns an instance of SerializedRange.
+   */
+  serialize(root, ignoreSelector) {
+    const serialization = function (node, isEnd) {
```
robertknight, line 271:
```suggestion
    const serialization = (node, isEnd) => {
```

This function doesn't reference `this`, so this can be an arrow function.

src/annotator/anchoring/test/xpath-evaluate-test.js
```diff
@@ -1,8 +1,8 @@
 import $ from 'jquery';
 
-import { nodeFromXPath, xpathFromNode, $imports } from '../range-js';
+import { nodeFromXPath, xpathFromNode, $imports } from '../xpath-evaluate';
 
-describe('annotator/anchoring/range-js', () => {
+describe('annotator/anchoring/xpath-evaluate', () => {
```
robertknight, line 5: `xpath-evaluate` contains functions for both evaluating and generating xpaths. I would be inclined to name it just `xpath.js`.

src/annotator/anchoring/test/range-test.js
```diff
@@ -0,0 +1,525 @@
...
+          startOffset: 0,
+          endContainer: container.querySelector('#p-3').firstChild,
+          endOffset: 1,
+          tainted: false,
+        });
+      });
+    });
+
+    describe('#normalize', () => {
+      it('throws an error BrowserRange instance is normalized more than once', () => {
```
robertknight, line 122:
```suggestion
      it('throws an error if BrowserRange instance is normalized more than once', () => {
```

robertknight (approved):
Hi @LMS007 -

I left some small requests. Otherwise this looks good for an initial conversion so please merge once those are addressed.

Once this has landed there is cleanup we can do to simplify the normalization code, remove the multiple implementations of XPath evaluation, and improve JSDoc comments. It will be easier to review those changes once the initial conversion lands though.

For manual testing, I made some changes in `html.coffee` to disable the text position and text quote selectors and verified that annotations were still anchored on a test page using just the range anchors:

```diff
diff --git a/src/annotator/anchoring/html.coffee b/src/annotator/anchoring/html.coffee
index 532b166a2..a0623ec04 100644
--- a/src/annotator/anchoring/html.coffee
+++ b/src/annotator/anchoring/html.coffee
@@ -63,15 +63,15 @@ exports.anchor = (root, selectors, options = {}) ->
       return querySelector(RangeAnchor, root, range, options)
       .then(maybeAssertQuote)
 
-  if position?
-    promise = promise.catch ->
-      return querySelector(TextPositionAnchor, root, position, options)
-      .then(maybeAssertQuote)
+  # if position?
+  #   promise = promise.catch ->
+  #     return querySelector(TextPositionAnchor, root, position, options)
+  #     .then(maybeAssertQuote)
 
-  if quote?
-    promise = promise.catch ->
-      # Note: similarity of the quote is implied.
-      return querySelector(TextQuoteAnchor, root, quote, options)
+  # if quote?
+  #   promise = promise.catch ->
+  #     # Note: similarity of the quote is implied.
+  #     return querySelector(TextQuoteAnchor, root, quote, options)
 
   return promise
 
```

PR #2570, opened 2020-09-25
Remove unnecessary jQuery usage in several tests

robertknight:
Remove jQuery in places where it can easily be replaced with DOM APIs.

@lyzadanger recently asked me whether there was a general plan to remove jQuery from tests as there is to remove it from the annotator code that ships to users. I suggest the answer to that is "yes", but for different reasons. For end users this will reduce bundle size, which in turn helps startup time. For developers there is a modest consistency benefit in not needing to work with two different APIs for DOM manipulation.


lyzadanger (approved):
👍

PR #2571, opened 2020-09-28
Convert types from coffeescript to js

LMS007:
Convert types.coffee to js

- Convert types to js
- Add missing test coverage in types module

---------

The new coverage tests are purely for codecov and stub out the imported modules so they don't test full logical correctness, but the existing tests, which did not stub the imported modules, are left alone.


src/annotator/anchoring/pdf.js
```diff
@@ -468,9 +465,11 @@ export function describe(root, range, options = {}) {
     startPos += pageOffset;
     endPos += pageOffset;
 
-    const position = new TextPositionAnchor(root, startPos, endPos).toSelector(
-      options
-    );
+    const position = new TextPositionAnchor(
```
LMS007, line 468: `options` is not a param on TextPositionAnchor.toSelector

src/annotator/anchoring/pdf.js
```diff
@@ -480,7 +479,7 @@ export function describe(root, range, options = {}) {
       root,
       quoteRange,
       options
-    ).toSelector(options);
+    ).toSelector();
```
LMS007, line 482: same here

src/annotator/anchoring/pdf.js
```diff
@@ -430,7 +428,6 @@ export function anchor(root, selectors) {
  * @param {Object} options -
  *   Options passed to `TextQuoteAnchor` and `TextPositionAnchor`'s
  *   `toSelector` methods.
- * @return {Promise<[TextPositionSelector, TextQuoteSelector]>}
```
LMS007, line 433: TS was not happy with this return type, removing it and allowing it to infer seems like the most reasonable solution now that types.js is converted.

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * @typedef {import("./range").BrowserRange} BrowserRange}
+ * @typedef {import("./range").NormalizedRange} NormalizedRange}
+ * @typedef {BrowserRange|NormalizedRange|SerializedRange} AnyRangeType
+ *
+ * @typedef {import('../../types/api').RangeSelector} RangeSelector
+ * @typedef {import('../../types/api').TextPositionSelector} TextPositionSelector
+ * @typedef {import('../../types/api').TextQuoteSelector} TextQuoteSelector
+ *
+ * @typedef TextContentNode
```
LMS007, line 31: This is more or less just a Node "light" with only a textContent prop.

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,245 @@
...
+  toPositionAnchor(options) {
+    if (!options) {
+      options = {};
+    }
+    const anchor = toTextPosition(this.root, this.toSelector(), options);
+    if (anchor === null) {
+      throw new Error('Quote not found');
+    }
+    return new TextPositionAnchor(
+      /** @type {Node} */ (this.root),
```
robertknight, line 240: This cast isn't safe if `TextQuoteAnchor` was initialized with a `TextContentNode` that is not a `Node`. It looks like `TextPositionAnchor`'s constructor can be changed to accept a `TextContentNode` instead.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,290 @@ describe('types', function () {
...
+          sniff: fakeSniff,
+          SerializedRange: fakeSerializedRange,
+        },
+      });
+    });
+
+    afterEach(() => {
+      $imports.$restore();
+    });
+    function createRangeAnchor() {
```
robertknight, line 54: This function name is confusing. It creates a DOM `Range` rather than a `RangeAnchor` as the name suggests. Given how simple it is, it looks like it would be just as easy to replace the callers with `new Range()`.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,290 @@ describe('types', function () {
...
+          new RangeAnchor();
+        }, 'missing required parameter "root"');
+      });
+
+      it('throws an error if `range` is missing', () => {
+        assert.throws(() => {
+          new RangeAnchor({});
+        }, 'missing required parameter "range"');
+      });
+    });
```
robertknight, line 70: I suggest these checks and the corresponding code in `RangeAnchor` can be removed.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,290 @@ describe('types', function () {
...
+
+    describe('#fromRange', () => {
+      it('returns a new RangeAnchor instance', () => {
+        const anchor = RangeAnchor.fromRange(container, createRangeAnchor());
+        assert.calledOnce(fakeNormalize);
```
robertknight, line 75: I think it would be better to test that `anchor.range` is the result of whatever `fakeNormalize` returns. This check just verifies that `fakeNormalize` was called but that doesn't mean that `RangeAnchor`'s anchor is actually a normalized range.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,290 @@ describe('types', function () {
...
+        assert.instanceOf(anchor, RangeAnchor);
+      });
+    });
+
+    describe('#fromSelector', () => {
+      it('returns a new RangeAnchor instance', () => {
+        const anchor = RangeAnchor.fromSelector(container, {
+          start: container.firstChild,
```
robertknight, line 83: The argument to `fromSelector` is supposed to be a `RangeSelector` object as defined in `types/api.js`.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,290 @@ describe('types', function () {
...
+        });
+        const range = new RangeAnchor(container, createRangeAnchor());
+        assert.equal(range.toRange(), 'normalized range');
+      });
+    });
+
+    describe('#toSelector', () => {
+      beforeEach(() => {
+        fakeSerialize.returns({
+          start: container.firstChild,
```
robertknight, line 106: Using `container.firstChild` and `container.lastChild` as the values for the start and end points is confusing because in the actual app, these would be XPath-like strings. It might not actually matter for the purpose of these tests if they just copy/pass the values around, but since tests serve to help the reader understand the intent of the code as well as check correctness, I prefer that tests use realistic-looking values.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,290 @@ describe('types', function () {
...
+    });
+
+    function createTextQuoteAnchor() {
+      return new TextQuoteAnchor(container, 'Liberty', {
+        prefix: 'a new nation, conceived in ',
+        suffix: ', and dedicated to the proposition that',
+      });
+    }
+
+    it('creates a TextQuoteAnchor if `context` is missing', () => {
```
robertknight, line 230: If you replace the `context` fallback with a default parameter value then this test will be less necessary.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,290 @@ describe('types', function () {
...
+          exact: 'Liberty',
+          prefix: 'a new nation, conceived in ',
+          suffix: ', and dedicated to the proposition that',
+        });
+      });
+    });
+
+    describe('#toRange', () => {
+      it('returns a valid DOM Range', () => {
+        $imports.$restore();
```
robertknight, line 274: `$restore` undoes all of the active mocks. If for a particular test you just wanted to undo one mock, you can pass an argument to `$restore`. See https://github.com/robertknight/babel-plugin-mockable-imports#restoring-specific-mocks.

src/annotator/anchoring/pdf.js
robertknight, line 433: The reason TypeScript complained about the return type here is because the inferred types of the result of the `toSelector` method are not what you expect. If you fix the `toSelector` return types then no changes are needed in this module, other than removing the unused `options` parameter.

robertknight (changes requested):
Hi Kyle,

I had a look through and left a few comments. One general comment I have is that having type checking does remove or reduce the need for certain kinds of tests, and I think we should take advantage of that to reduce the amount of test code that we need to maintain.

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,245 @@
...
+ * @prop {number} [hint] -
+ *    Offset to prioritize equivalent searches. See https://github.com/tilgovi/dom-anchor-text-quote
+ */
+
+/**
+ * Helper for throwing common errors
+ *
+ * @param {string} name
+ */
+const missingParameter = function (name) {
```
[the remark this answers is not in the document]
LMS007: your 100% right, this is redundant with typechecking

src/annotator/anchoring/test/types-test.js
LMS007: good to know!

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,245 @@
...
+  /**
+   * @param {SelectorOptions} [options]
+   */
+  toSelector(options) {
+    if (!options) {
+      options = {};
+    }
+
+    const range = this.range.serialize(this.root, options.ignoreSelector);
+    /** @type {RangeSelector} */
```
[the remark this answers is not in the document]
LMS007: 👍

src/annotator/anchoring/test/types-test.js
LMS007: This was an oversight on my part

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,218 @@
...
+ * @typedef {import('../../types/api').RangeSelector} RangeSelector
+ * @typedef {import('../../types/api').TextPositionSelector} TextPositionSelector
+ * @typedef {import('../../types/api').TextQuoteSelector} TextQuoteSelector
+ *
+ * @typedef TextContentNode
+ * @prop {string} textContent
+ */
+
+/**
+ * @typedef SelectorOptions
```
robertknight, line 36: The different options here refer to different selector types. I would suggest to remove this and document the supported properties for each `options` usage as appropriate. In other words, something like this:

```js
class RangeAnchor {
  /**
   * @param {Object} [options]
   * @param {string} [options.ignoreSelector]
   */
  toSelector(options = {}) {

  }
}
```

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,218 @@
...
+  toRange as quoteToRange,
+  toTextPosition,
+} from 'dom-anchor-text-quote';
+
+import { SerializedRange, sniff } from './range';
+
+/**
+ * @typedef {import("./range").BrowserRange} BrowserRange}
+ * @typedef {import("./range").NormalizedRange} NormalizedRange}
+ * @typedef {BrowserRange|NormalizedRange|SerializedRange} AnyRangeType
```
robertknight, line 25: Should `AnyRangeType` also accept an ordinary DOM `Range` as you use in several of your tests or does it have to be one of these three classes?

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,284 @@ describe('types', function () {
...
     });
 
-    describe('#toPositionAnchor', function () {
-      it('returns a TextPositionAnchor', function () {
+    describe('#toPositionAnchor', () => {
+      it('returns a TextPositionAnchor', () => {
+        $imports.$restore({
+          'dom-anchor-text-quote': {
+            toTextPosition: true,
+          },
```
robertknight, line 284: I see that the `$imports.$restore` is added here to make the test behave the same as it did before the PR. For a future reader it might be unclear why the `dom-anchor-text-quote` and `dom-anchor-text-position` libraries are mocked in some tests but not others.

Since these modules are third-party code there is some value in having tests that don't mock them because it helps to catch unexpected changes in the third-party library. We could leave that to the integration tests, although debugging failures there will be harder because there is a larger surface area to debug.

I'm OK with leaving this as-is for the moment.

robertknight (changes requested):
Hi Kyle,

I take another pass over this and commented on some minor issues.  Give me a shout when you've had a chance to address them.

src/annotator/anchoring/types.js
LMS007: Actually, after converting html over to js. I think it needs to also needs to support Range. I will double check.

src/annotator/anchoring/test/types-test.js
LMS007: an integration test section is probably important here. I'll work on that.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,284 @@ describe('types', function () {
...
+        assert.instanceOf(anchor, RangeAnchor);
+        assert.deepEqual(anchor.range, fakeNormalize());
+      });
+    });
+
+    describe('#fromSelector', () => {
+      it('returns a new RangeAnchor instance', () => {
+        const anchor = RangeAnchor.fromSelector(container, {
+          type: 'RangeSelector',
+          start: '/div[1]',
```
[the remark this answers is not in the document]
LMS007: good catch, yep, we should clean that up in our refactor!

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,43 +21,284 @@ describe('types', function () {
...
+      });
+    });
+
+    afterEach(() => {
+      $imports.$restore();
+    });
+
+    describe('#fromRange', () => {
+      it('returns a new RangeAnchor instance', () => {
+        const anchor = RangeAnchor.fromRange(container, new Range());
```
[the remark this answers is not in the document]
LMS007: Yes, I confirmed this is the case in practice

src/annotator/anchoring/types.js
LMS007: makes sense

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,218 @@
...
+   */
+  constructor(root, exact, context = {}) {
+    this.root = root;
+    this.exact = exact;
+    this.context = context;
+  }
+  /**
+   * @param {Node} root
+   * @param {Range} range
+   * @param {SelectorOptions} options
```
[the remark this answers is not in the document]
LMS007: indeed, needs to be removed.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,46 +21,335 @@ describe('types', function () {
...
           'some are more equal than others'
         );
-        assert.throws(function () {
+        assert.throws(() => {
           quoteAnchor.toPositionAnchor();
         });
       });
     });
+
+    describe('integration tests', () => {
```
LMS007, line 330: Added a new integration which covers round-tripping a Range()

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,46 +21,335 @@ describe('types', function () {
...
+        assert.deepEqual(anchor.toSelector(), {
+          type: 'TextQuoteSelector',
+          prefix: '',
+          suffix: ' score and seven years ago our f',
+          exact: 'Four',
+        });
+        const newRange = anchor.toRange();
+        // startContainer of anchor.toRange() points to <div>
+        // rather than the text node "Four score ..."
+        // assert.equal(anchor.toRange(), range);
```
LMS007, line 350: Perhaps this is expected, but I was not sure so I left a comment. The two Ranges have different startContainers.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,46 +21,335 @@ describe('types', function () {
...
+    describe('#toRange', () => {
+      it('returns a range object', () => {
+        const anchor = createTextPositionAnchor();
+        fakePosToRange.returns('fake range');
+        assert.equal(anchor.toRange(), 'fake range');
+        assert.calledWith(fakePosToRange, container, { start: 0, end: 0 });
+      });
+    });
+
+    describe('integration tests', () => {
```
LMS007, line 185: Added a new integration test

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,46 +21,335 @@ describe('types', function () {
...
+        range.setStart(container.firstChild, 0);
+        range.setEnd(container.firstChild, 4);
+        const anchor = TextPositionAnchor.fromRange(container, range);
+        assert.deepEqual(anchor.toSelector(), {
+          type: 'TextPositionSelector',
+          start: 0,
+          end: 4,
+        });
+        const newRange = anchor.toRange();
+        assert.deepEqual(range, newRange);
```
robertknight, line 202: The arguments are the wrong way round here. The syntax is `assert.deepEqual(actual, expected)` and `range` is the _expected_ value while _newRange_ is the _actual_ value.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,46 +21,335 @@ describe('types', function () {
...
+
+    function createTextQuoteAnchor() {
+      return new TextQuoteAnchor(container, 'Liberty', {
+        prefix: 'a new nation, conceived in ',
+        suffix: ', and dedicated to the proposition that',
+      });
+    }
+
+    describe('#fromRange', () => {
+      it('returns a range object', () => {
```
robertknight, line 238:
```suggestion
      it('returns a TextQuoteAnchor', () => {
```

`fromRange` takes a range as input rather than returning a range.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,46 +21,335 @@ describe('types', function () {
...
+          suffix: 'brought forth on this continent',
+        });
+        const anchor = TextQuoteAnchor.fromRange(container, new Range());
+        assert.called(fakeQuoteFromRange);
+        assert.instanceOf(anchor, TextQuoteAnchor);
+      });
+    });
+
+    describe('#fromSelector', () => {
+      it('returns a TextQuoteSelector', () => {
```
robertknight, line 250:
```suggestion
      it('returns a TextQuoteAnchor', () => {
```

A `TextQuoteSelector` selector is the input here but the output is a `TextQuoteAnchor`.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,46 +21,335 @@ describe('types', function () {
...
+          exact: 'Liberty',
+          prefix: 'a new nation, conceived in ',
+          suffix: ', and dedicated to the proposition that',
+        });
+        assert.instanceOf(anchor, TextQuoteAnchor);
+      });
+    });
+
+    describe('#toSelector', () => {
+      it('returns a TextQuoteAnchor instance', () => {
```
robertknight, line 262:
```suggestion
      it('returns a TextQuoteSelector selector', () => {
```

The output isn't a `TextQuoteAnchor` instance here but an object with `type: 'TextQuoteSelector'`.

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,210 @@
...
+  /**
+   * @param {Node} root -  A root element from which to anchor.
+   * @param {AnyRangeType} range -  A range describing the anchor.
+   */
+  static fromRange(root, range) {
+    return new RangeAnchor(root, range);
+  }
+
+  /**
+   * Create and anchor using the saved Range selector.
```
robertknight, line 57:
```suggestion
   * Create an anchor from a serialized `RangeSelector` selector
```

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,210 @@
...
+      type: 'TextQuoteSelector',
+      exact: this.exact,
+      prefix: this.context.prefix,
+      suffix: this.context.suffix,
+    };
+  }
+
+  /**
+   * @param {Object} [options]
+   *   @param {Object} [options.hint]
```
robertknight, line 189: A comment here would be useful to clarify what `hint` is. This can point to relevant part of the dom-anchor-text-quote README if there is one.

src/annotator/anchoring/test/types-test.js
```diff
@@ -19,46 +21,335 @@ describe('types', function () {
...
+        fakePosToRange.returns('fake range');
+        assert.equal(anchor.toRange(), 'fake range');
+        assert.calledWith(fakePosToRange, container, { start: 0, end: 0 });
+      });
+    });
+
+    describe('integration tests', () => {
+      beforeEach(() => {
+        // restore dom-anchor-text-position to test third party lib integration
+        $imports.$restore();
```
robertknight, line 188: This will restore _all_ mocks, not just `dom-anchor-text-position`. Use `$imports.$restore({ 'dom-anchor-text-position': true })` for that.

src/annotator/anchoring/types.js
```diff
@@ -0,0 +1,210 @@
...
+ * @typedef {import('../../types/api').RangeSelector} RangeSelector
+ * @typedef {import('../../types/api').TextPositionSelector} TextPositionSelector
+ * @typedef {import('../../types/api').TextQuoteSelector} TextQuoteSelector
+ *
+ * @typedef TextContentNode
+ * @prop {string} textContent
+ */
+
+/**
+ * DOM Range anchor type.
```
robertknight, line 36:
```suggestion
 * Converts between RangeSelector selectors and Range objects
```

This changes the description to be clearer and match the other *Anchor classes.

robertknight (commented):
I had another look over this and found a few minor mistakes in test descriptions and requested a couple of doc updates.

PR #2585, opened 2020-09-29
Find suitable z-index for the adder

### LINKED ISSUE #2009: Annotate/Highlight buttons not visible on nzz.ch
### Steps to reproduce
1. Visit https://www.nzz.ch/feuilleton/niall-ferguson-sechs-fragen-zum-coronavirus-an-xi-jinping-ld.1550428 (or any other page on nzz.ch)
2. Mark text

### Expected behaviour
The Annotate/Highlight speech-bubble-popup should appear.

### Actual behaviour
It appears to not appear, but then I noticed that when you mark text at the leftmost part of the screen, you can see small parts of the popup, so it seems to be opened _behind_ the .pageholder div.

Firefox 47.0.1 on Archlinux, both using via.hypothes.is and the bookmarklet.

esanzgar:
This PR incorporates a solution to programatically find a z-index for the adder. Before this, the z-index of the adder was a hard-coded value, which didn't work for some websites (see for example #2009).

We evaluated the option of finding the greatest z-index in the page. However, this takes too long for heavy pages.

This solution in this PR is not bulletproof, but probably works for the most part of the websites. It uses the greatest z-index of five nearby elements: left-top, left-bottom, middle-center, right-top, right-bottom elements around the adder.
  
Resolves #2009

### DIFF
--- a/dev-server/documents/html/z-index.mustache
+++ b/dev-server/documents/html/z-index.mustache
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="UTF-8">
+  <title>Hypothesis Client Test</title>
+<style>
+  body {
+    font-family: sans-serif;
+  }
+</style>
+</head>
+<body>
+  <h2>z-index test page</h2>
+  <p>This has a z-index of 0</p>
+  <ol style="z-index: 20000; position: relative">
+    <li style="z-index: 25000; position: relative">This has a z-index of 25000</li>
+    <li style="z-index: 30000; position: relative">This has a z-index of 30000</li>
+    <li>This has a z-index of 20000 (from parent)</li>
+    <li>This has a z-index of 20000 (from parent)</li>
+    <li>This has a z-index of 20000 (from parent)</li>
+  </ol>
+  <p>This has a z-index of 0</p>
+
+   {{{hypothesisScript}}}
+  </body>
+</html>
--- a/dev-server/templates/index.mustache
+++ b/dev-server/templates/index.mustache
@@ -18,6 +18,7 @@
     <li><a href="/document/burns">Poems and Songs of Robert Burns</a></li>
     <li><a href="/document/doyle">The Disappearance of Lady Carfax, by Arthur Conan Doyle</a></li>
     <li><a href="/document/tolstoy">Anna Karenina, by Leo Tolstoy</a></li>
+    <li><a href="/document/z-index">z-index page</a></li>
   </ul>
 
   <h2>Test PDF documents</h2>
--- a/src/annotator/adder.js
+++ b/src/annotator/adder.js
@@ -89,10 +89,6 @@ export class Adder {
       // take position out of layout flow initially
       position: 'absolute',
       top: 0,
-
-      // Assign a high Z-index so that the adder shows above any content on the
-      // page
-      zIndex: 10000,
     });
 
     this._view = /** @type {Window} */ (container.ownerDocument.defaultView);
@@ -194,6 +190,51 @@ export class Adder {
   }
 
   /**
+   * Find a Z index value that will cause the adder to appear on top of any
+   * content in the document when the adder is shown at (left, top).
+   *
+   * @param {number} left - Horizontal offset from left edge of viewport.
+   * @param {number} top - Vertical offset from top edge of viewport.
+   * @return {number} - greatest zIndex (default value of 1)
+   */
+  _findZindex(left, top) {
+    if (document.elementsFromPoint === undefined) {
+      // In case of not being able to use `document.elementsFromPoint`,
+      // default to the large arbitrary number (2^15)
+      return 32768;
+    }
+
+    const adderWidth = this._width();
+    const adderHeight = this._height();
+
+    // Find the Z index of all the elements in the screen for five positions
+    // around the adder (left-top, left-bottom, middle-center, right-top,
+    // right-bottom) and use the greatest.
+
+    // Unique elements so `getComputedStyle` is called the minimum amount of times.
+    const elements = new Set([
+      ...document.elementsFromPoint(left, top),
+      ...document.elementsFromPoint(left, top + adderHeight),
+      ...document.elementsFromPoint(
+        left + adderWidth / 2,
+        top + adderHeight / 2
+      ),
+      ...document.elementsFromPoint(left + adderWidth, top),
+      ...document.elementsFromPoint(left + adderWidth, top + adderHeight),
+    ]);
+
+    const zIndexes = [...elements]
+      .map(element => +getComputedStyle(element).zIndex)
+      .filter(Number.isInteger);
+
+    // Make sure the array contains at least one element,
+    // otherwise `Math.max(...[])` results in +Infinity
+    zIndexes.push(0);
+
+    return Math.max(...zIndexes) + 1;
+  }
+
+  /**
    * Show the adder at the given position and with the arrow pointing in
    * `arrowDirection`.
    *
@@ -210,9 +251,12 @@ export class Adder {
     const positionedAncestor = nearestPositionedAncestor(this._container);
     const parentRect = positionedAncestor.getBoundingClientRect();
 
+    const zIndex = this._findZindex(left, top);
+
     Object.assign(this._container.style, {
-      top: toPx(top - parentRect.top),
       left: toPx(left - parentRect.left),
+      top: toPx(top - parentRect.top),
+      zIndex,
     });
 
     this._isVisible = true;
--- a/src/annotator/test/adder-test.js
+++ b/src/annotator/test/adder-test.js
@@ -1,4 +1,6 @@
 import { act } from 'preact/test-utils';
+import { createElement } from 'preact';
+import { mount } from 'enzyme';
 
 import { Adder, ARROW_POINTING_UP, ARROW_POINTING_DOWN } from '../adder';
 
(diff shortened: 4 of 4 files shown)


robertknight:
I have yet to figure out exactly why using `attachTo: document.body` failed, but using the following pattern which is quite common in our tests avoids the issue:

```diff
diff --git a/src/annotator/test/adder-test.js b/src/annotator/test/adder-test.js
index 5ee39cc98..7f0d54cd5 100644
--- a/src/annotator/test/adder-test.js
+++ b/src/annotator/test/adder-test.js
@@ -32,6 +32,17 @@ function revertOffsetElement(el) {
 }
 
 describe('findHighestZindex', () => {
+  let container;
+
+  beforeEach(() => {
+    container = document.createElement('div');
+    document.body.appendChild(container);
+  });
+
+  afterEach(() => {
+    container.remove();
+  });
+
   it('returns zero z-index (default value) if no element is passed)', () => {
     assert.strictEqual(findHighestZindex(undefined), 0);
     assert.strictEqual(findHighestZindex(null), 0);
@@ -44,7 +55,7 @@ describe('findHighestZindex', () => {
         <div className="child1" style={{ zIndex: 5 }} />
         <div className="child2" style={{ zIndex: 15 }} />
       </div>,
-      { attachTo: document.body }
+      { attachTo: container }
     );
 
     const parent = wrapper.find('.parent').getDOMNode();
```

dev-server/templates/index.mustache
```diff
@@ -27,6 +28,15 @@
     <li><a href="/pdf/widdershins">Widdershins (Wikipedia)</a></li>
   </ul>
 
+  <h2>Test z-index</h2>
```
robertknight, line 31: Could you put this on a separate page with a link on the homepage pointing to it? I'd like to avoid cluttering the homepage of the dev server. The test page could be headed "Web page quirks" or something like that.

esanzgar: Sure

src/annotator/adder.js
```diff
@@ -51,6 +51,23 @@ function nearestPositionedAncestor(el) {
   return parentEl;
 }
 
+/**
+ * Recursively find the highest z-index for this element and its parents
+ *
+ * @param {Element|null|undefined } element - element that has the current selection
```
robertknight, line 57: Is the `undefined` needed here? It looks like this function will always be passed an element or `null`.

src/annotator/test/adder-test.js
```diff
@@ -24,6 +31,35 @@ function revertOffsetElement(el) {
...
+    const parent = wrapper.find('.parent').getDOMNode();
+    assert.strictEqual(findHighestZindex(parent), 10);
+
+    const child1 = wrapper.find('.child1').getDOMNode();
+    assert.strictEqual(findHighestZindex(child1), 10);
+
+    const child2 = wrapper.find('.child2').getDOMNode();
+    assert.strictEqual(findHighestZindex(child2), 15);
+
+    wrapper.detach();
```
robertknight, line 59: If you use the suggestion I mentioned in the thread about creating a container and then removing it in an `afterEach` callback, then the `.detach()` call can be removed here. Calling `wrapper.unmount()` is mainly needed when there are effects that need to be cleaned up, but that is not the case with this simple JSX.

esanzgar: I found that rendering components directly into `document.body` is discouraged using enzyme

esanzgar: Both `wrapper.detach()` and `wrapper.unmount()` can be used if the component is not attached directly to `document.body`.

src/annotator/adder.js
esanzgar: Yes it is needed because `focusNode.parentElement` could return `undefined`.
The `focusNode` on a `Selection` objection, in my experience returns a text node.

src/annotator/guest.js
```diff
@@ -557,7 +557,16 @@ export default class Guest extends Delegator {
...
+    // Find the zIndex of two close by element and use the greatest.
+    const zIndex1 =
+      findGreatestZindex(document.elementFromPoint(left, top)) + 1;
+    const zIndex2 = findGreatestZindex(selection.focusNode?.parentElement) + 1;
+    this.adderCtrl.showAt(
+      left,
+      top,
+      arrowDirection,
+      Math.max(zIndex1, zIndex2)
+    );
```
robertknight, line 569: Can you add a test in `guest-test.js` to cover this change. Here is a sketch of what that could look like:

```diff
diff --git a/src/annotator/test/guest-test.js b/src/annotator/test/guest-test.js
index 8ac1aea26..70e4fb573 100644
--- a/src/annotator/test/guest-test.js
+++ b/src/annotator/test/guest-test.js
@@ -37,6 +37,7 @@ describe('Guest', () => {
   const sandbox = sinon.createSandbox();
   let CrossFrame;
   let fakeCrossFrame;
+  let fakeFindGreatestZindex;
   let highlighter;
   let guestConfig;
   let htmlAnchoring;
@@ -81,10 +82,12 @@ describe('Guest', () => {
     CrossFrame = sandbox.stub().returns(fakeCrossFrame);
     guestConfig.pluginClasses.CrossFrame = CrossFrame;
 
+    fakeFindGreatestZindex = sinon.stub().returns(0);
+
     $imports.$mock({
       './adder': {
         Adder: FakeAdder,
-        findGreatestZindex: sinon.stub().returns(0),
+        findGreatestZindex: fakeFindGreatestZindex,
       },
       './anchoring/html': htmlAnchoring,
       './highlighter': highlighter,
@@ -481,6 +484,22 @@ describe('Guest', () => {
       assert.called(FakeAdder.instance.showAt);
     });
 
+    it('shows the adder above the content', () => {
+      createGuest();
+      fakeFindGreatestZindex.onFirstCall().returns(50);
+      fakeFindGreatestZindex.onSecondCall().returns(10);
+
+      simulateSelectionWithText();
+
+      assert.calledWith(
+        FakeAdder.instance.showAt,
+        sinon.match.number,
+        sinon.match.number,
+        sinon.match.any,
+        50 + 1
+      );
+    });
+
     it('sets the annotations associated with the selection', () => {
       createGuest();
       const ann = {};
```

dev-server/documents/html/z-index.mustache
```diff
@@ -0,0 +1,24 @@
...
+  <meta charset="UTF-8">
+  <title>Hypothesis Client Test</title>
+<style>
+  body {
+    font-family: sans-serif;
+  }
+</style>
+</head>
+<body>
+  <h2>Test z-index</h2>
```
robertknight, line 13: On this test page there is an issue when selecting the title:

Title z-index issue

This is because in this instance the location where the adder is shown has higher z-indexes than the locations for which the z-index is determined.

I don't know how common this will be in practice, but we might need to consider other ways of enumerating the elements to be checked.

robertknight (commented):
Hi Eduardo,

I had another look over this. I suggested some minor changes but I also noted a scenario which was highlighted by the test page that you added that might require re-thinking the approach a little.

src/annotator/adder.js
```diff
@@ -193,6 +189,34 @@ export class Adder {
...
+   * @param {number} left - Horizontal offset from left edge of viewport.
+   * @param {number} top - Vertical offset from top edge of viewport.
+   * @return {number} - greatest zIndex
+   */
+  _findZindex(top, left) {
+    const width = this._width();
+    const height = this._height();
+
+    // Unique elements and ancestors
+    const elements = new Set([
```
esanzgar, line 205: Is it OK to use `Set` to get all the unique elements?

src/annotator/adder.js
```diff
@@ -193,6 +189,34 @@ export class Adder {
...
+      ...document.elementsFromPoint(left, top),
+      ...document.elementsFromPoint(left, top + height),
+      ...document.elementsFromPoint(left + width / 2, top + height / 2),
+      ...document.elementsFromPoint(left + width, top),
+      ...document.elementsFromPoint(left + width, top + height),
+    ]);
+
+    const zIndexes = [...elements]
```
esanzgar, line 213: Is it understandable? Or does it need comment?

esanzgar (commented):
I made a more robust and simple implementation.

src/annotator/adder.js
robertknight: In general you can use the `Set` and `Map` classes in our code. We have polyfills for these.

src/annotator/adder.js
```diff
@@ -193,6 +189,34 @@ export class Adder {
...
+    const width = this._width();
+    const height = this._height();
+
+    // Unique elements and ancestors
+    const elements = new Set([
+      ...document.elementsFromPoint(left, top),
+      ...document.elementsFromPoint(left, top + height),
+      ...document.elementsFromPoint(left + width / 2, top + height / 2),
+      ...document.elementsFromPoint(left + width, top),
+      ...document.elementsFromPoint(left + width, top + height),
```
robertknight, line 210: `elementsFromPoint` is [only supported](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementsFromPoint) in quite recent versions of browsers. [`elementFromPoint`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementFromPoint) has better support (eg. Safari < 12). We could also have a simple fallback for browsers that don't support it by returning a fixed z-index value.

src/annotator/adder.js
robertknight: This is fine. Using array spread to convert an iterable into array is fairly common. `Array.from(elements)` would also be fine.

robertknight (changes requested):
Hi Eduardo,

This looks nice and simple. We'll need a fallback for browsers that don't support `document.elementsFromPoint` since it requires, for example, quite recent versions of Firefox and Safari (>= 12). Returning a static value in that case would be fine.

Can you also add a few tests. These can mock `document.elementsFromPoint` to make things easier.

src/annotator/adder.js
```diff
@@ -193,6 +189,34 @@ export class Adder {
...
+    ]);
+
+    const zIndexes = [...elements]
+      .map(element => +getComputedStyle(element).zIndex)
```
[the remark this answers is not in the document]
esanzgar: zIndex is a string: 'auto', '0', '6'

esanzgar:
@robertknight, I fully agree about (1) returning a default static value (as we were doing previously) in case of `document.elementsFromPoint` is not available, and (2) adding more tests.

esanzgar:
On the second commit I provided a fallback option in case `document.elementsFromPoint` is not defined.
IE and Edge supports a non-standard version of this method.

src/annotator/test/adder-test.js
```diff
@@ -217,6 +219,90 @@ describe('Adder', () => {
     });
   });
 
+  describe('#_findZindex', () => {
```
robertknight, line 222: Generally we try to avoid testing code via its private APIs as that can make it more difficult to change the internals of a class.

In this case one way to do that would be to actually check the Z index via the DOM by replacing the `_findZindex` calls with something like:

```js
function getAdderZIndex(left, top) {
  adderCtrl.showAt(left, top);
  return parseInt(adderEl.style.zIndex);
}
```

That would still leave the `_width()` and `_height()` calls but I think that's OK - these would be easy to adapt if the implementation changed.

src/annotator/test/adder-test.js
```diff
@@ -217,6 +219,90 @@ describe('Adder', () => {
...
+              position: 'absolute',
+              width: 1,
+              height: 1,
+              left,
+              top,
+              zIndex,
+            }}
+          />,
+          { attachTo }
+        );
```
robertknight, line 261: Using Enzyme here is OK, although for a simple static piece of DOM like this it would be just as easy to use Preact's `render` function directly.

```js
import { render } from 'preact';
...
render(<div ... />, container);
const rendered = container.firstChild;
```

src/annotator/adder.js
```diff
@@ -193,6 +189,50 @@ export class Adder {
...
+   * @return {number} - greatest zIndex (default value of 1)
+   */
+  _findZindex(left, top) {
+    if (document.elementsFromPoint === undefined) {
+      // In case of not be able to find the zIndex of the surrounding elements,
+      // default to the greatest <integer> allowed in Opera (2^15).
+      // The <integer> value is currently enforced in the specification,
+      // and Opera has one of the smaller ranges.
+      // https://developer.mozilla.org/en-US/docs/Web/CSS/integer#Syntax
+      return 32768;
```
robertknight, line 207: I'm pretty sure that information on MDN is out of date. See https://joeist.com/2012/06/what-is-the-highest-possible-z-index-value/ (note this is from 2012). The actual value chosen here (2^15) should be fine though, as it is similar to what was used before.

I suppose we could have just hard-coded the z-index to 2^31-1. The downside is that there is a small chance that this might interfere with legitimate tools that need to display above the adder.

robertknight (commented):
Hi Eduardo,

This looks good. I think the comment for `_findZindex` could be improved though to explain why it is doing what it does. I believe the comments about Opera on MDN are out of date, although the value of 2^15 is a perfectly reasonable fallback value.

In tests we generally avoid testing private methods, although this may occasionally be OK if it is unlikely to get in the way of refactoring the code in future. I suggested a way that you could test the z-index without calling `_findZindex` directly.

src/annotator/adder.js
esanzgar: OK.

For curiosity, what other legitimate tools are you referring?

src/annotator/test/adder-test.js
esanzgar: Good to know.

src/annotator/test/adder-test.js
```diff
@@ -217,6 +219,95 @@ describe('Adder', () => {
     });
   });
 
+  describe('#_findZindex', () => {
```
robertknight, line 222: Can you update the description here since these tests are no longer calling this private method. Something like `describe('adder Z index', ...` would be fine.

src/annotator/adder.js
robertknight: I didn't have anything specific in mind, I'm just guessing that _maybe_ there are such tools.

src/annotator/adder.js
```diff
@@ -193,6 +189,56 @@ export class Adder {
...
+  _findZindex(left, top) {
+    if (document.elementsFromPoint === undefined) {
+      // In case of not being able to use `document.elementsFromPoint`,
+      // default to the greatest <integer> allowed in Opera (2^15).
+      // The <integer> value is currently enforced in the specification,
+      // and Opera has one of the smaller ranges.
+      // https://developer.mozilla.org/en-US/docs/Web/CSS/integer#Syntax
+      // For a less outdated info see:
+      // https://joeist.com/2012/06/what-is-the-highest-possible-z-index-value/
+      return 32768;
```
robertknight, line 209: I think it would be clearer to remove this comment and just note that the fallback of 2^15 is chosen as an arbitrary large value. The note on MDN applies to a version of Opera (12.1) that is 8 years old and pre-dates Hypothesis!

robertknight (approved):
Hi Eduardo - This looks good. Can you update the description of the `describe(...` block in the test following the change to use only public API of the class. Also, I would suggest to remove the reference to MDN and Opera because that information is long out of date. 2^15 is effectively just "an arbitrary large value that will work almost all web pages". Other than these two small items, this is good to merge.

PR #2587, opened 2020-09-29
Example of how to add Draft type exploiting TS

esanzgar:


robertknight:
Hello @esanzgar ,

Can you clarify what the original problem was that you were trying to solve here?

In future I'm open to the idea of converting code to TypeScript syntax rather than using JSDoc for type annotations, but there is a bunch of existing infrastructure that we have for JS which will need to be adapted and we'll also need to give the rest of the team (ie. other than you and me) a chance to get comfortable with it. It might make sense to start with some smaller project.

For the purposes of whatever you were trying to do here though I expect there will be a solution which is easier to integrate within our current tech stack.

esanzgar:
This was just to showcase that we could exploit more TS.

PR #2588, opened 2020-09-30
Convert html.coffee to JS

LMS007:
Covert html.coffee to js.

- ~~Add extra tests in html-test to cover error handling when fromRange() fails from various anchor objects.~~ (Update: This was moved to a follow-up PR following discussion)
- Remove extraneous options param passed to anchor() in guest.js

-----

- [x] needs to be rebased off master once types is converted
- [x] needs a test for `quote mismatch` error


src/annotator/guest.js
```diff
@@ -357,12 +356,7 @@ export default class Guest extends Delegator {
         return Promise.resolve({ annotation, target });
       }
 
-      // Find a target using the anchoring module.
-      const options = {
-        ignoreSelector: IGNORE_SELECTOR,
-      };
-      return this.anchoring
-        .anchor(root, target.selector, options)
+      return htmlAnchor(root, target.selector)
```
robertknight, line 359: When viewing a PDF the `anchoring` property gets overridden in [src/annotator/plugin/pdf.js](https://github.com/hypothesis/client/blob/129d1ba341736bd5c1848d8e3f94c19fd6ed5463/src/annotator/plugin/pdf.js#L44). Therefore the `Guest` class needs to continue using `this.anchoring` for the moment.

src/annotator/guest.js
```diff
@@ -484,7 +478,7 @@ export default class Guest extends Delegator {
         ignoreSelector: IGNORE_SELECTOR,
       };
       // Returns an array of selectors for the passed range.
-      return this.anchoring.describe(root, range, options);
+      return htmlDescribe(root, range, options);
```
robertknight, line 481: As in the `anchor` method, `this.anchoring` needs to be used to call the appropriate implementation depending on the document type (HTML vs PDF)

src/annotator/anchoring/html.js
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * @param {Node} root
+ * @param {Range} range
+ * @param {Object} [options]
+ *   @param {string} [options.ignoreSelector]
+ */
+export function describe(root, range, options = {}) {
+  const types = [RangeAnchor, TextPositionAnchor, TextQuoteAnchor];
+
+  const selectors = (() => {
```
robertknight, line 110: The `describe` function here can be simplified to:

```js
export function describe(root, range, options = {}) {
  const types = [RangeAnchor, TextPositionAnchor, TextQuoteAnchor];
  const result = [];
  for (let type of types) {
    try {
      const anchor = type.fromRange(root, range);
      result.push(anchor.toSelector(options));
    } catch (error) {
      continue;
    }
  }
  return result;
}
```

Note elimination of `selector` variable and the IIFE it is assigned to.

src/annotator/anchoring/html.js
```diff
@@ -0,0 +1,124 @@
...
+ * @typedef {import("./types").AnyRangeType} AnyRangeType
+ * @typedef {import('../../types/api').Selector} Selector
+ */
+
+/**
+ * @param {RangeAnchor|TextPositionAnchor|TextQuoteAnchor} anchor
+ * @param {Object} [options]
+ *  @param {number} [options.hint]
+ */
+const querySelector = function (anchor, options = {}) {
```
robertknight, line 13: The `querySelector` function can be simplified to:

```js
async function querySelector(anchor, options = {}) {
  return anchor.toRange(options);
}
```

src/annotator/anchoring/html.js
```diff
@@ -0,0 +1,124 @@
...
+    if (quote?.exact && range.toString() !== quote.exact) {
+      throw new Error('quote mismatch');
+    } else {
+      return range;
+    }
+  };
+
+  // From a default of failure, we build up catch clauses to try selectors in
+  // order, from simple to complex.
+  let promise = Promise.reject('unable to anchor');
```
robertknight, line 72: If you widen the type of `promise` then you can remove the `// @ts-ignore` comments below:

```js
  /** @type {Promise<Range>} */
  let promise = Promise.reject('unable to anchor');
```

robertknight (changes requested):
Hi Kyle,

Apart from the bucket bar we're nearly done with work :) - I found an issue with the removal of `this.anchoring` and PDFs. I also suggested some simplifications for several of the converted functions and a way to avoid the need to `@ts-ignore` several lines.

src/annotator/guest.js
LMS007: awe, that was not so obvious. I might add a comment too

src/annotator/anchoring/test/html-test.js
```diff
@@ -419,42 +422,133 @@ describe('HTML anchoring', function () {
...
+                RangeAnchor: {
+                  fromRange: sinon.stub().throws(new Error()),
+                },
+              },
+            });
+          },
+        },
+
+        {
+          name: 'when no TextPositionAnchor.fromRange throws an error',
```
robertknight, line 465:
```suggestion
          name: 'when TextPositionAnchor.fromRange throws an error',
```

src/annotator/anchoring/test/html-test.js
```diff
@@ -419,42 +422,133 @@ describe('HTML anchoring', function () {
...
+              './types': {
+                TextPositionAnchor: {
+                  fromRange: sinon.stub().throws(new Error()),
+                },
+              },
+            });
+          },
+        },
+        {
+          name: 'when no TextQuoteAnchor.fromRange throws an error',
```
robertknight, line 482:
```suggestion
          name: 'when TextQuoteAnchor.fromRange throws an error',
```

src/annotator/anchoring/test/html-test.js
```diff
@@ -419,42 +422,133 @@ describe('HTML anchoring', function () {
...
-
-        const annotationsChecked = annotations.map(async ann => {
-          const root = frame.contentWindow.document.body;
-          const selectors = ann.target[0].selector;
-          const range = await html.anchor(root, selectors);
-          const newSelectors = await html.describe(root, range);
-          assert.deepEqual(sortByType(selectors), sortByType(newSelectors));
+      [
+        {
+          name: 'when no anchors throw an error',
```
robertknight, line 438: It is good to add test cases to cover the current gaps in this module's tests, but I don't think they should be added to "Web baselines" tests. These tests are intended to test anchoring with real webpages and selectors. As a result they are more complex to debug if something fails.

Instead I think it would be better to test this fallback behavior using tests more like the "when anchoring fails" test block above.

robertknight (changes requested):
The converted `html.coffee` looks fine. The `anchor` implementation could be further simplified in future. In the tests it makes sense to add cases for the logic that is not currently covered, but I don't think they should be added to the "Web baselines" tests. See my comments for more details.

It may be helpful to split this task into two steps - convert the existing HTML code but leave the tests as-is in this PR, then add additional tests in a separate PR.

The changes in `guest.js` also conflicted with a PR I merged to remove the `ignoreSelector` option, so this will need rebasing.

LMS007:
Removed changes to html-test.js. That will be done in a follow up

robertknight (approved):
LGTM. I'm going to go ahead and merge this so I can post a follow-up PR to remove CoffeeScript support.

PR #2589, opened 2020-09-30
Simplify XPath generation code

robertknight:
For reasons that AFAIK are no longer relevant, there were two
implementations of XPath generation for use in serializing range
selectors. Since we're removing jQuery from the Hypothesis client, this
removes the jQuery implementation and simplifies the remaining DOM-only
one.

 - Remove jQuery XPath generation

 - Change the non-jQuery implementation to take DOM Nodes as input
   rather than jQuery collections

 - Simplify and add type documentation for the remaining implementation

 - Correct the documentation for `xpathFromNode`. The previous comments
   said that it _evaluated_ XPaths but actually it _generates_ them.

### DIFF
--- a/src/annotator/anchoring/range.js
+++ b/src/annotator/anchoring/range.js
@@ -249,7 +249,7 @@ export class NormalizedRange {
       } else {
         origParent = $(node).parent();
       }
-      const xpath = xpathFromNode(origParent, root)[0];
+      const xpath = xpathFromNode(origParent[0], root ?? document);
       const textNodes = getTextNodes(origParent);
       // Calculate real offset as the combined length of all the
       // preceding textNode siblings. We include the length of the
--- a/src/annotator/anchoring/xpath-util.js
+++ b/src/annotator/anchoring/xpath-util.js
@@ -60,7 +60,7 @@ function getPathSegment(node) {
 }
 
 /**
- * A simple XPath generator using which can generate XPaths of the form
+ * A simple XPath generator which can generate XPaths of the form
  * /tag[index]/tag[index].
  *
  * @param {Node} node - The node to generate a path to
@@ -79,7 +79,7 @@ export function xpathFromNode(node, root) {
     elem = elem.parentNode;
   }
   xpath = '/' + xpath;
-  xpath = xpath.replace(/\/$/, '');
+  xpath = xpath.replace(/\/$/, ''); // Remove trailing slash
 
   return xpath;
 }
--- a/src/annotator/anchoring/xpath-util.js
+++ b/src/annotator/anchoring/xpath-util.js
@@ -1,5 +1,3 @@
-import $ from 'jquery';
-
 /**
  * Finds the child node associated with the provided index and
  * type relative from a parent.
@@ -25,6 +23,8 @@ export function findChild(parent, type, index) {
 
 /**
  * Get the node name for use in generating an xpath expression.
+ *
+ * @param {Node} node
  */
 function getNodeName(node) {
   const nodeName = node.nodeName.toLowerCase();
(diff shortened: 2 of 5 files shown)


src/annotator/anchoring/xpath.js
```diff
@@ -1,23 +1,6 @@
...
-  } catch (e) {
-    // eslint-disable-next-line no-console
-    console.log(
-      'jQuery-based XPath construction failed! Falling back to manual.'
-    );
-    result = simpleXPathPure(el, relativeRoot);
-  }
-  return result;
-}
+export { xpathFromNode } from './xpath-util';
```
robertknight, line 3: This export exists only to keep the module API the same. I think it would make sense to merge the two in future or possibly split out the non-XPath related functions into a separate module.

src/annotator/anchoring/range.js
```diff
@@ -249,7 +249,7 @@ export class NormalizedRange {
       } else {
         origParent = $(node).parent();
       }
-      const xpath = xpathFromNode(origParent, root)[0];
+      const xpath = xpathFromNode(origParent[0], root ?? document);
```
robertknight, line 252: The `root` argument to `xpathFromNode` is not optional. This keeps the tests passing, although I think the `root` argument here should really be required, in which case this fallback can be removed.

src/annotator/anchoring/xpath.js
LMS007: I agree

src/annotator/anchoring/xpath-util.js
```diff
@@ -50,58 +53,35 @@ function getNodePosition(node) {
...
 function getPathSegment(node) {
   const name = getNodeName(node);
   const pos = getNodePosition(node);
   return `${name}[${pos}]`;
 }
 
 /**
- * A simple XPath evaluator using only standard DOM methods which can
- * evaluate queries of the form /tag[index]/tag[index].
+ * A simple XPath generator using which can generate XPaths of the form
```
LMS007, line 63: Something is quit right with this sentence.. previously it was 

```
 * A simple XPath evaluator using only standard DOM methods which can
 * evaluate queries of the form /tag[index]/tag[index].

```

Perhaps

`* A simple XPath evaluator which can evaluate queries of the form /tag[index]/tag[index].`

src/annotator/anchoring/xpath-util.js
```diff
@@ -50,58 +53,35 @@ function getNodePosition(node) {
...
-    xpath = '/' + xpath;
-    xpath = xpath.replace(/\/$/, '');
-    return xpath;
+    xpath = getPathSegment(elem) + '/' + xpath;
+    elem = elem.parentNode;
   }
-  const paths = nodes.map((index, node) => getPathTo(node));
-  return paths.get();
+  xpath = '/' + xpath;
+  xpath = xpath.replace(/\/$/, '');
```
LMS007, line 82: There was not a comment here before but this it warrants one since this regex is a bit confusing.

something like "Removes a trailing slash"

src/annotator/anchoring/xpath-util.js
```diff
@@ -50,58 +53,35 @@ function getNodePosition(node) {
...
-        throw new Error(
-          'Called getPathTo on a node which was not a descendant of @rootNode. ' +
-            rootNode
-        );
-      }
-      xpath = getPathSegment(elem) + '/' + xpath;
-      elem = elem.parentNode;
+  /** @type {Node|null} */
+  let elem = node;
+  while (elem !== root) {
```
LMS007, line 74: way better Rob, no extra function indirection and now operates on just a single node. That entire extra nodes.map(...) call was never used but for an array of 1 :)

LMS007 (approved):
Solid improvement Rob, please change just comments (below) and merge!

src/annotator/anchoring/xpath-util.js
robertknight: The updated comment here is correct - this function generates rather than evaluates XPaths. There is indeed a grammatical error ("generator using which can generate") which I'll correct.

PR #2590, opened 2020-09-30
Remove XPath evaluation fallback

robertknight:
All browsers supported by the Hypothesis client support
`document.evaluate`, and there wasn't a clear case where we expect that
`document.evaluate` might fail and yet the fallback would succeed.

A side effect of this is that a more accurate error is thrown if the
XPath is invalid.

**TODO**

- [ ] Test what happens when annotating content that uses non-HTML namespaces
- [ ] Find out if we have any `RangeSelector`s in existing annotations that rely on namespaces. How common are they?


src/annotator/anchoring/test/xpath-test.js
```diff
@@ -100,14 +100,16 @@ describe('annotator/anchoring/xpath', () => {
...
-        document.evaluate.restore();
-      });
+    it('returns `null` if XPath evaluation fails', () => {
+      assert.equal(nodeFromXPath('/missing/element', container), null);
+    });
+
+    it('throws if XPath is invalid', () => {
+      assert.throws(() => {
+        nodeFromXPath('not-a-valid-xpath', container);
+      }, /The string '.not-a-valid-xpath' is not a valid XPath expression/);
```
robertknight, line 112: The behavior here is the same as before except for the text of the error message that is thrown. Previously it was "wanted child not found".

robertknight:
Reading the DOM specs suggests some cases where `document.evaluate` _might_ fail compared to the previous implementation that should be looked into:

- What happens if content from non-HTML namespaces is annotated, eg. text in an SVG document or MathML?
- What happens if the document is XHTML?
- Can a selection span multiple documents? What happens in that case?

robertknight:
I've done some digging to better understand how HTML document parsing, `document.evaluate` and XML namespaces interact. Here are my notes:

Documents may be parsed as either [_HTML documents_ or _XML documents_](https://dom.spec.whatwg.org/#html-document). This depends upon the [MIME type](https://html.spec.whatwg.org/multipage/introduction.html#html-vs-xhtml) that the document is served with, set by the `Content-Type` header. The document type affects the behavior of certain APIs, including `document.evaluate`, which we use to resolve XPaths (eg. `/div[1]/p[2]`) in `RangeSelector`-type selectors.

Elements in a parsed HTML or XML document have an associated namespace. This can be examined via `element.namespaceURI`. For HTML elements, the namespace is _http://www.w3.org/1999/xhtml_. The document type (XML vs HTML) affects how segments of an XPath with no prefix [are handled](https://html.spec.whatwg.org/#interactions-with-xpath-and-xslt). In an HTML document, `./div[1]` is interpreted as referring to a `div` element in the namespace _http://www.w3.org/1999/xhtml_. In an XML document however `./div[1]` is interpreted as referring to a `div` element which does not have an associated namespace (also known as a _null_ namespace URI).

As a result, passing an XPath like `/html/body` to `document.evaluate` will return the `<body>` element in an _HTML document_ but not an XML document. In an XML document each node in the XPath needs a prefix (eg. `/xhtml:html/xhtml:body`) and a _resolver_ must be specified which maps prefixes in the XPath to namespace URIs (eg `xhtml` => _http://www.w3.org/1999/xhtml_). Resolvers can be created with `document.createNSResolver` or by specifying a function that accepts a prefix and returns the URI.

Within an HTML document, there may be embedded content from non-HTML namespaces, such as SVG and MathML. There is a list of namespaces that might be found in an HTML document here: https://infra.spec.whatwg.org/#namespaces. I'm not clear on whether arbitrary xmlns values can be used or whether other namespaces are used in any significant volume. In order to match elements from these other namespaces (eg. SVG or MathML), the XPath needs to identify which namespace the element comes from.

The client currently always generates XPaths that contain only the local element name.  To see this, go to https://webkit.org/demos/mathml/MathMLDemo.xhtml and create an annotation which highlights part of an equation on the page. The annotation will have a `RangeSelector` like this:

```json
{
  "type": "RangeSelector",
  "startContainer": "/div[1]/p[1]/math[1]/mi[1]",
  "startOffset": 0,
  "endContainer": "/div[1]/p[1]/math[1]/mrow[2]/mfrac[1]/msup[1]/mi[2]",
  "endOffset": 1
}
```

This document is an XML document and the `math` elements come from a different namespace than the `div` and `p` elements, but there is no indication of this in the path. Therefore calls to `document.evaluate` will fail to find the node. When `document.evaluate` is passed a valid XPath but it doesn't match any nodes due to missing namespaces, it will return `null`.

On the `master` branch there is a fallback path that uses a different XPath evaluator that can only process simple XPaths where each segment is of the form `<element name>[index]`. This fallback evaluator ignores namespaces, so unlike `document.evaluate` it can work whether the document is parsed as XML or HTML and whether it references only HTML elements or other content as well. However the fallback evaluator is only triggered if `document.evaluate` throws an exception, not if it returns `null`. `document.evaluate` only throws an exception if passed an invalid XPath or if the XPath contains namespace references that cannot be resolved.

As a result, anchoring `RangeSelector` selectors generated by the client currently fails when:

1. Documents are parsed as XML rather than HTML
2. The selection references a non-HTML node in an HTML document (eg. SVG or MathML).

In practice annotation will still continue to work in most cases because the client will fall back to trying the text position and text quote selectors.

Related to point (2), I discovered a separate issue in Safari 14 where highlighting MathML doesn't work, but it does in Chrome. 

Before highlighting:

mathml-safari
 
After highlighting:

highlighted-mathml-safari

robertknight:
In the investigation above I learned that we do need a non-`document.evaluate` based method of evaluating certain XPaths, but it needs to work differently. To avoid confusion I'm going to close this and open a new PR that fixes that.

PR #2591, opened 2020-10-01
Karma cleanup

esanzgar:
Minor improvement.

This PR combines two consecutive `if` statements into one. The conditions for the `if` statements were identical.


LMS007 (approved):
👍 

look good Eduardo

PR #2594, opened 2020-10-01
Fix range selector anchoring in XHTML documents and embedded SVG/MathML

### LINKED ISSUE #2592: Anchoring `RangeSelector` selectors fails in HTML documents parsed as XML
### Steps to reproduce

1. Go to an HTML document which is served with an XML mime type, such as https://webkit.org/demos/mathml/MathMLDemo.xhtml (`document.contentType` returns `application/xhtml+xml`) on this page
2. Open the browser dev tools and make sure the debugger is configured to stop on unhandled exceptions. It should be by default
3. Attempt to annotate the content (eg. the header "Riemann's Zeta Function")

### Expected behaviour

No errors occur.

### Actual behaviour

When clicking the "Annotate" button, an error is reported in the browser dev tools:

```
Paused on promise rejection
RangeError: Couldn't find offset 0 in element /div[1]/h2[1]
```

This was tested in Chrome 87.

From the user's point of view creating the annotation continues to work because the other types of selector (text position and text quote) still work.

### Notes

This issue occurs because the client generates an XPath of the form `/elementName[index]/elementName[index]/...` and then [tries to resolve it](https://github.com/hypothesis/client/blob/2414a08b6f1152d9875b1950a0cb69d0b66e8dd6/src/annotator/anchoring/xpath.js#L50) using `document.evaluate`. Since this document served with an XML mime type, XPath segments without prefixes are not resolved and `document.evaluate` returns `null`. See https://github.com/hypothesis/client/pull/2590#issuecomment-702132563

### LINKED ISSUE #2595: Anchoring `RangeSelector` selectors fails in SVG or MathML content embedded in HTML
### Steps to reproduce

1. Go to this HTML document which contains embedded MathML: https://developer.mozilla.org/en-US/docs/Web/MathML/Examples/MathML_Pythagorean_Theorem
2. Open the browser dev tools and make sure the debugger is configured to stop on unhandled exceptions. It should be by default
3. Attempt to annotate the equation _a² + b² = c²_ after the sentence "If a and b are the legs and c is the hypotenuse then"

### Expected behaviour

No errors occur.

### Actual behaviour

When clicking the "Annotate" button, an error is reported in the browser dev tools:

```
Paused on promise rejection
RangeError: Couldn't find offset 2 in element /div[1]/main[1]/div[1]/div[1]/article[1]/p[3]/math[1]/mrow[1]/msup[1]/mi[1]
```

This was tested in Chrome 87.

From the user's point of view creating the annotation continues to work because the other types of selector (text position and text quote) still work.

### Notes

This issue occurs because the client generates an XPath of the form `/elementName[index]/elementName[index]/...` and then [tries to resolve it](https://github.com/hypothesis/client/blob/2414a08b6f1152d9875b1950a0cb69d0b66e8dd6/src/annotator/anchoring/xpath.js#L50) using `document.evaluate`. This doesn't work for MathML or SVG nodes (or any non-HTML nodes) because `document.evaluate` compares both the element name and the namespace of elements of elements in the DOM against the XPath. Segments in the xpath with no namespace prefix are assumed to be in the HTML namespace. However all the MathML nodes are in the MathML namespace.

To make this work with `document.evaluate` we'd need to change the XPath generation to add namespaces to non-HTML elements (eg. `/main[1]/article[1]/mathml:math[1]/mathml:mrow[2]`) and then configure a namespace resolver to map the "mathml" string to the MathML namespace (http://www.w3.org/1998/Math/MathML). This won't help for existing `RangeSelector`s however.

Related issue: https://github.com/hypothesis/client/issues/2592
More details on `document.evaluate` and namespaces: https://github.com/hypothesis/client/pull/2590#issuecomment-702132563

robertknight:
In HTML documents that have been served with an XML mime type,
`document.evaluate` handles element names differently. In HTML documents
the XPath segments do not require prefixes on element names. In XHTML
documents however they do. Since the client always generates the same
un-prefixed XPaths regardless of document type, evaluation always
failed. There was a fallback path but it was only executed if
`document.evaluate` threw an error, not if it returned `null` as in this
case.

This commit resolves the issue by first attempting to evaluate the XPath
using custom logic that only handles simple XPaths and then using
`document.evaluate` only for more complex XPaths. The _simple XPath_
logic behaves the same in HTML and XML documents and ignores namespaces.
As a result it works with the XPaths that the client generates
regardless of document type and also regardless of whether the XPath
references only HTML elements or elements from other namespaces (eg.
MathML or SVG).

We could change the way that the client generates XPaths in future in
XML documents or SVG/MathML content within HTML documents, but the
client would still need to handle `RangeSelector` selectors on existing
annotations.

For more details, see https://github.com/hypothesis/client/pull/2590#issuecomment-702132563

Fixes #2592
Fixes #2595


src/annotator/anchoring/xpath.js
```diff
@@ -20,50 +20,113 @@ export function xpathFromNode(el, relativeRoot) {
...
-function nodeFromXPathFallback(xpath, root) {
-  const steps = xpath.substring(1).split('/');
-  let node = root;
-  steps.forEach(step => {
-    let [name, idx] = step.split('[');
-    idx = idx ? parseInt(idx.split(']')[0]) : 1;
-    node = findChild(node, name.toLowerCase(), idx);
-  });
-  return node;
+function nthChildOfType(element, nodeName, index) {
```
robertknight, line 30: The `nthChildOfType` function is very similar to the removed `findChild` function except that it explicitly ignores case and returns `null` rather than throwing an exception if the element is not found, which is more consistent with how DOM APIs like `document.querySelector` or `document.evaluate` work.

src/annotator/anchoring/xpath.js
```diff
@@ -1,52 +1,113 @@
...
+  nodeName = nodeName.toUpperCase();
+
+  let matchIndex = -1;
+  for (let i = 0; i < element.children.length; i++) {
```
LMS007, line 15: I never looked too closely had findChild, but I'm now wondering if we need to directly iterate over the list of child nodes? I'm not sure a query is any faster, but could we reduce some lines of code by doing something like this?

```
try {
    return element.querySelectorAll(`:scope > ${nodeName}`)[index];
  } catch(e) {
    return null;
  }
```

src/annotator/anchoring/xpath.js
```diff
@@ -1,52 +1,113 @@
...
-  steps.forEach(step => {
-    let [name, idx] = step.split('[');
-    idx = idx ? parseInt(idx.split(']')[0]) : 1;
-    node = findChild(node, name.toLowerCase(), idx);
-  });
-  return node;
+function nthChildOfType(element, nodeName, index) {
+  nodeName = nodeName.toUpperCase();
+
+  let matchIndex = -1;
```
LMS007, line 14: `let found` started at `0`, and then it first compared at 1. this starts at -1 and first compares at 0. I guess I'm confused why they are different.

LMS007 (approved):
This looks good Rob. I just had some comments about nthChildOfType. Let me know what you think. I also think we should merge xpath and xpath-util as soon as you get this in.

src/annotator/anchoring/xpath.js
robertknight: I did not know about `:scope` in `querySelector`. Yes, I think this code would work. I expect it would be somewhat slower as it is doing a bunch more work (generating a selector string, parsing it, then executing selector matching). What is the `try...catch` for here?

src/annotator/anchoring/xpath.js
robertknight: In `findChild`, `index` was a 1-based index and `found` is the count of matching elements found. In `nthChildOfType` the `index` param is a 0-based index. The reason for using 1-based indexes previously is probably because XPaths use 1-based indexing (eg. `/div[1]` for the first `div` child). The code that calls `nthChildOfType` has been adjusted to convert the 1-based index from the XPath to an 0-based index during parsing.

robertknight:
I updated PR with a couple of small fixes for edge-case scenarios that won't occur in normal usage of the client but did occur either in a test (anchoring nodes inside another iframe) or could occur with annotations created via the API (`RangeSelector`s with XPaths that refer to non-Element nodes).

PR #2596, opened 2020-10-02
Fix test comment

LMS007:


src/annotator/anchoring/test/types-test.js
```diff
@@ -349,9 +349,9 @@ describe('annotator/anchoring/types', () => {
           exact: 'Four',
         });
         const newRange = anchor.toRange();
-        // startContainer of anchor.toRange() points to <div>
-        // rather than the text node "Four score ..."
-        // assert.equal(anchor.toRange(), range);
+        // Note that startContainer of anchor.toRange() points to <div>
+        // rather than the text node "Four score ..." because it has been
+        // normalized.
```
robertknight, line 354: This comment doesn't make sense to me. Do you mean the range does not start at a text node, unlike many of the other ranges in this file because it has *not* been normalized by one of the `normalize()` methods defined in `range.js`? The `toRange` method that is being called here calls the `toRange` function from `dom-anchor-text-quote` but doesn't do any processing on the result.

robertknight (changes requested):
Hi Kyle,

I don't think the updated comment says what you intended it to say. Can you check?

src/annotator/anchoring/test/types-test.js
LMS007: Awe, yes, that is inverted. I think I'm just gonna get rid of the comment in any case :)

robertknight (approved):

PR #2597, opened 2020-10-05
Simplify annotator plugin initialization

robertknight:
Previously plugin initialization happened in several steps:

 1. The plugin constructor was called with the root element and config
    options
 2. The plugin's `annotator` property was set to a reference to the
    `Guest` instance for the current document
 3. The plugin's `pluginInit` method was called

There was no real need for this multi-step process. This commit
simplifies it by removing the `pluginInit` method and passing the guest
reference as a third argument to the constructor in step 1. Logic that previously used to
be in `pluginInit` has been moved to the constructor.

As a result, the `Plugin` base class is no longer needed. Plugins now
extend the parent `Delegator` class instead.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -133,7 +133,6 @@ export default class Guest extends Delegator {
 
     this.addPlugin('CrossFrame', cfOptions);
     this.crossframe = this.plugins.CrossFrame;
-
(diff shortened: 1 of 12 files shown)


lyzadanger (approved):
Very nice, and glad to see such a useful change made in such a small diff.

PR #2598, opened 2020-10-05
Remove jQuery usage from range.js

robertknight:
As part of the removal of jQuery from the annotator, remove usage from
`range.js`:

 - Replace `parents()` method with a small helper function

 - Rewrite `getTextNodes` function to accept and return DOM nodes rather
   than jQuery collections

 - Use `Node.contains` instead of jQuery's `contains` method. Note that there
   is a semantic difference that `contains(node, node)` returns
   `false` whereas `node.contains(node)` returns `true`. In other words, in the DOM API a node "contains" itself, but not in jQuery. It was possible to leverage this to simplify a condition.

### DIFF
--- a/src/annotator/anchoring/range.js
+++ b/src/annotator/anchoring/range.js
@@ -416,32 +416,8 @@ export class SerializedRange {
       }
     }
 
-    // Here's an elegant next step...
-    //
-    //   range.commonAncestorContainer = $(range.startContainer).parents().has(range.endContainer)[0]
-    //
-    // ...but unfortunately Node.contains() is broken in Safari 5.1.5 (7534.55.3)
-    // and presumably other earlier versions of WebKit. In particular, in a
-    // document like
-    //
-    //   <p>Hello</p>
-    //
-    // the code
-    //
-    //   p = document.getElementsByTagName('p')[0]
-    //   p.contains(p.firstChild)
-    //
-    // returns `false`. Yay.
-    //
-    // So instead, we step through the parents from the bottom up and use
-    // Node.compareDocumentPosition() to decide when to set the
-    // commonAncestorContainer and bail out.
-
-    const contains = (a, b) =>
-      a.compareDocumentPosition(b) & Node.DOCUMENT_POSITION_CONTAINED_BY;
-
     for (let parent of parents(range.startContainer)) {
-      if (contains(parent, range.endContainer)) {
+      if (parent.contains(range.endContainer)) {
         range.commonAncestorContainer = parent;
         break;
       }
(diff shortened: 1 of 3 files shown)


src/annotator/anchoring/range.js
```diff
@@ -423,17 +440,15 @@ export class SerializedRange {
...
-      .each(function () {
-        if (contains(this, range.endContainer)) {
-          range.commonAncestorContainer = this;
-          // bail out of loop
-          return false;
-        }
-        return true;
-      });
+    const contains = (a, b) =>
+      a.compareDocumentPosition(b) & Node.DOCUMENT_POSITION_CONTAINED_BY;
```
robertknight, line 444: See https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition for enum names and associated values.

src/annotator/anchoring/range.js
```diff
@@ -382,7 +396,10 @@ export class SerializedRange {
         targetOffset--;
       }
 
-      for (let tn of getTextNodes($(node))) {
+      for (let tn of getTextNodes(node)) {
+        if (tn.nodeValue === null) {
+          continue;
```
LMS007, line 401: When is a text node's nodeValue null? Is it possible to add a test for this case?

src/annotator/anchoring/range.js
```diff
@@ -423,17 +440,15 @@ export class SerializedRange {
...
-      .parents()
-      .each(function () {
-        if (contains(this, range.endContainer)) {
-          range.commonAncestorContainer = this;
-          // bail out of loop
-          return false;
-        }
-        return true;
-      });
+    const contains = (a, b) =>
```
LMS007, line 443: The comment above still likely applies, but does reference jquery so we should probably at least change that or delete the parts of it that don't make sense anymore.

LMS007 (approved):
If you could just make a change to that comment below and decide on that missing unit test, then we are good. Its nice to get jquery removed!

src/annotator/anchoring/range.js
robertknight: I looked this up in the DOM specs this morning and found that `nodeValue` is always a string for a `Text` node: https://dom.spec.whatwg.org/#dom-node-nodevalue.

In the case of `Text` nodes `nodeValue` is an alias for `data` which is always non-null. I've changed this to use `data` instead.

src/annotator/anchoring/range.js
robertknight: Agreed. I re-tested the test case given in the comment above and was no longer able to reproduce. I couldn't find any relevant bug reports in WebKit's bug tracker either. On that basis I simplified this code with `Node.contains(...)`.

In future I suspect we might be able to remove `Range` normalization from most of the code. The only place where it is really needed AFAIK is before wrapping nodes in highlights.

Issue #2043, opened 2020-04-20
End support for Internet Explorer 11 (IE11)


robertknight:
This issue is about officially ending support for IE 11 in the Hypothesis client, so we can spend our development / testing efforts that go to supporting that browser on improvements for the vast majority of users that are using newer browsers.

Note that the LMS app definitely does not support IE 11, so we don't need to concern ourselves about that.

Things we need to address:

1. What is the time-frame for ending IE 11 support?
2. What will users experience if they visit a page that embeds the Hypothesis client if they are using a browser that the client definitely doesn't support?
3. Should we set up some kind of advance notice such as an in-app banner to tell users that support for IE 11 will end by a certain date?

Re point (1) I need to gather usage stats, but we've accidentally broken IE support on a couple of occasions recently and it took quite a while for someone to notice. This suggests to me that we can afford to make this happen pretty soon. Perhaps end of June 2020?

A thought I had re point (2) is that perhaps we could create a dummy version of the sidebar and the button that opens/closes it, where the sidebar shows a notice that Hypothesis annotations are not available in this browser and a link to view them on the "h" website.

robertknight:
Related Slack thread: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1588062136439800

klemay:
@robertknight checking in on the status of this - do we need to do anything else before we can say we have ended support for IE11, or can we close the issue?

klemay:
Notes from backlog meeting: before closing this issue we'll need to remove IE-specific code from the client. This will have the advantage of making the client smaller + faster; however it's not urgent.

robertknight:
We've now completed the removal of IE 11 support from the code. The boot scripts handles exactly what does happen when the client attempts to start in IE 11. Currently it just logs a console warning with a link to a knowledgebase article. We have discussed showing a banner in when the host page is an environment that we control (eg. Via, LMS, as opposed to a third-party publisher's page).

klemay:
We have an issue (including a screenshot from an IE11 user) for showing a banner here: https://github.com/hypothesis/product-backlog/issues/1141

robertknight:
This issue is resolved. We ended support for IE 11 last year and added code in the boot script to notify users about unsupported browsers.

PR #2599, opened 2020-10-05
Remove IE 11 polyfills

robertknight:
**Depends on https://github.com/hypothesis/client/pull/2627**

Since IE 11 is no longer supported, this PR removes polyfills that were only used to support that browser:

- [Node.baseURI](https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI#Browser_compatibility)
- [CustomEvent constructor](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent#Browser_compatibility)
- [document.evaluate](https://developer.mozilla.org/en-US/docs/Web/API/Document/evaluate#Browser_compatibility)
- Optional-ness of certain arguments to `document.createNodeIterator`

One polyfill I didn't remove in this PR is one for the [URL constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#Browser_compatibility). Although that is available in all the browsers we support, there are comments in the code that it may not work as expected in some browsers and I need to remind myself which ones those are.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -112,7 +112,6 @@ const appBundles = [
 //    the client that need it in `src/boot/boot.js`
 //  - Add the polyfill to the test environment if necessary in `src/karma.config.js`
 const polyfillBundles = [
-  'document.evaluate',
   'es2015',
   'es2016',
   'es2017',
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "diff": "^4.0.1",
     "dom-anchor-text-position": "^5.0.0",
     "dom-anchor-text-quote": "^4.0.2",
-    "dom-node-iterator": "^5.0.0",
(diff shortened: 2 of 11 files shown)


LMS007 (approved):
👍 
I'm guessing we can do-away with `browser-compatibility-utils.js` as well?

robertknight:
Reviewing this again I think we need to put in some kind of graceful failure in browsers that are not supported. I think the minimum acceptable solution would for the boot script to do a feature test and then abort with a console warning if the browser is unsupported. This would minimize disruption for users visiting pages with Hypothesis embedded by the publisher.

A better solution could be to display some kind of button/icon where the Hypothesis sidebar toggle would normally appear with a link to an explanation.

When using Via or the bookmarklet we would want to do something more visible, such as showing a banner at the top of the screen.

PR #2611, opened 2020-10-06
Fix range selector generation in Chrome and Firefox

robertknight:
This fixes a regression introduced in https://github.com/hypothesis/client/pull/2598 in non-Safari browsers.

----

When generating range selectors the client excludes elements from the
Hypothesis UI by filtering out elements that match the CSS selector
defined by `IGNORE_SELECTOR` in guest.js.

In a554058161 this logic was changed from jQuery's selector matching to
`element.matches(...)`. The selector used in the actual application is a
compount selector which is then inverted with `:not`:

```
:not([class^="annotator-"],[class^="hypothesis-"])
```

It turns out that the `:not(...)` pseudo-class only supports _lists_ of
selectors in Safari, but not in Chrome or Firefox [1]. Therefore the call to
`element.matches(...)` would always fail in these browsers and the
resulting annotation would only have `TextQuoteSelector` and
`TextPositionSelector` selectors but no `RangeSelector`.

This commit fixes the issue by removing the use of `:not` and instead
inverting the `element.matches(...)` test. It also adds a test case for
handling of list selectors.

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/:not#Browser_compatibility

### DIFF
--- a/src/annotator/anchoring/range.js
+++ b/src/annotator/anchoring/range.js
@@ -6,15 +6,16 @@ import {
 } from './xpath-util';
 
 /**
- * Return ancestors of `element`, optionally filtered by a CSS `selector`.
+ * Return ancestors of `element`, optionally filtered by a CSS selector.
  *
  * @param {Node} node
- * @param {string} [selector]
+ * @param {string} [ignoreSelector] -
+ *   Any elements that match this CSS selector will be excluded from the result
  */
-function parents(node, selector) {
+function parents(node, ignoreSelector) {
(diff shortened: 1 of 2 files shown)


robertknight:
After further consideration, I decided to drop this in favor of fixing the problem by removing the `ignoreSelector` support entirely: https://github.com/hypothesis/client/pull/2611/commits/b2b823840430dc04220a037e2b1a5cd6a1047da3

PR #2612, opened 2020-10-06
Remove logic to ignore Hypothesis UI when generating range selectors

robertknight:
Remove the logic that was intended to ignore Hypothesis UI when
generating range selectors.

We don't need this any more because all Hypothesis UI elements in the
host page which contain text have `user-select` set to prevent text
selection. This affects the bucket bar, vertical toolbar and adder.

In addition, the logic did not serve its intended purpose as it only applied to
range selector generation. If the user did somehow succeed in selecting text in
the bucket bar for example the client would still generate quote and position
selectors that referred to that content. What the client would need to
do instead is modify the range from which the annotation's selectors
were generated.

As a side effect this fixes a regression introduced in a554058 which caused
range selector generation to fail in Chrome and Firefox due to use of
`Element.matches` with a `:not(<inner selector>)` selector where `<inner selector>` is a
list of selectors (the `IGNORE_SELECTOR` value from `guest.js`). Such selectors are
[not supported](https://developer.mozilla.org/en-US/docs/Web/CSS/:not#Browser_compatibility) in
Chrome and Firefox.

----

**Testing:**

In Chrome or Firefox, open the developer tools and then create an annotation. In the Network tab look at the payload for the request to the server to create the new annotation. The `selector` array should have a `RangeSelector`:

range-selector

### DIFF
--- a/src/annotator/anchoring/range.js
+++ b/src/annotator/anchoring/range.js
@@ -6,17 +6,14 @@ import {
 } from './xpath-util';
 
 /**
- * Return ancestors of `element`, optionally filtered by a CSS `selector`.
+ * Return ancestors of `node`.
  *
  * @param {Node} node
- * @param {string} [selector]
  */
-function parents(node, selector) {
+function parents(node) {
   const parents = [];
   while (node.parentElement) {
-    if (!selector || node.parentElement.matches(selector)) {
-      parents.push(node.parentElement);
-    }
+    parents.push(node.parentElement);
     node = node.parentElement;
   }
   return parents;
@@ -169,13 +166,11 @@ export class BrowserRange {
    * Creates a range suitable for storage.
    *
    * root           - A root Element from which to anchor the serialization.
-   * ignoreSelector - A selector String of elements to ignore. For example
-   *                  elements injected by the annotator.
    *
    * Returns an instance of SerializedRange.
    */
-  serialize(root, ignoreSelector) {
-    return this.normalize().serialize(root, ignoreSelector);
+  serialize(root) {
+    return this.normalize().serialize(root);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/annotator/anchoring/range.js
```diff
@@ -6,17 +6,14 @@ import {
 } from './xpath-util';
 
 /**
- * Return ancestors of `element`, optionally filtered by a CSS `selector`.
+ * Return ancestors of `element`.
```
LMS007, line 9: Return ancestors of `node`.

src/annotator/anchoring/range.js
```diff
@@ -433,13 +421,11 @@ export class SerializedRange {
...
    * root           - A root Element from which to anchor the serialization.
-   * ignoreSelector - A selector String of elements to ignore. For example
-   *                  elements injected by the annotator.
    *
    * Returns an instance of SerializedRange.
    */
-  serialize(root, ignoreSelector) {
-    return this.normalize(root).serialize(root, ignoreSelector);
+  serialize(root) {
+    return this.normalize(root).serialize(root);
```
LMS007, line 428: If this no longer needs an `ignoreSelector`, then I "think" this can now just be

`return this`

if so, then this will also remove the 3-way dependency and we can break each of this ranges objects into separate modules much easier (to-do later of course).

LMS007 (approved):
Looks good. If we can simplify the serialize() method below, then I think we should try and do that now.

src/annotator/anchoring/range.js
robertknight: `return this` would ignore the `root` argument which affects the behavior of `normalize(...)`. I'm not certain that is a safe change yet so I'd say it is out of scope for this PR.

PR #2620, opened 2020-10-08
Remove jQuery from annotator initialization

robertknight:
Remove jQuery from annotator initialization

As one of the final parts of removing the client's jQuery dependency,
use DOM APIs to defer annotator initialization until the document has
loaded.


lyzadanger (approved):
👍

PR #2623, opened 2020-10-09
Show a warning when PDF does not have selectable text

robertknight:
This is an initial solution for https://github.com/hypothesis/product-backlog/issues/1081. It adds a banner to the top of the PDF viewer when the current PDF does not contain any selectable text.

Current design (updated 2020-10-16):

pdf-no-text-v2

The "OCR-optimized" text links to https://web.hypothes.is/help/how-to-ocr-optimize-pdfs/.


src/annotator/anchoring/pdf.js
```diff
@@ -79,16 +79,32 @@ async function getPageView(pageIndex) {
     // a "pdfPage" property.
     pageView = await new Promise(resolve => {
       const onPagesLoaded = () => {
-        document.removeEventListener('pagesloaded', onPagesLoaded);
+        pdfViewer.eventBus.off('pagesloaded', onPagesLoaded);
         resolve(pdfViewer.getPageView(pageIndex));
       };
-      document.addEventListener('pagesloaded', onPagesLoaded);
+      pdfViewer.eventBus.on('pagesloaded', onPagesLoaded);
```
robertknight, line 85: This fixes a compatibility issue with new versions of PDF.js that we hadn't noticed before. PDF viewer events are no longer emitted as DOM events. It should be extracted into a separate PR and shipped first.

lyzadanger:
@robertknight I've pushed some commits that apply styling to the warning message. As of 152212e it looks like this:

image

The border width (borders are top and bottom only) is `2px` currently. We can go to `3px`, but it looks a little chunky.

src/annotator/plugin/pdf.js
```diff
@@ -50,6 +90,69 @@ export default class PDF extends Delegator {
...
+      await this.uri();
+    } catch (e) {
+      return;
+    }
+
+    try {
+      const hasText = await pdfAnchoring.documentHasText();
+      this._showNoSelectableTextWarning(!hasText);
+    } catch (err) {
+      /* istanbul ignore next */
```
lyzadanger, line 108: I suppose I could have used this approach to what we were discussing regarding #2646 and the guarding against arrays being of different lengths...

src/annotator/plugin/pdf.js
```diff
@@ -50,6 +90,69 @@ export default class PDF extends Delegator {
...
+   */
+  _showNoSelectableTextWarning(showWarning) {
+    if (!showWarning) {
+      this._warningBanner?.remove();
+      this._warningBanner = null;
+      return;
+    }
+
+    const mainContainer = /** @type {HTMLElement} */ (document.querySelector(
+      '#mainContainer'
```
lyzadanger, line 127: Is this OK as a literal here (i.e. is this selector used elsewhere and could it be maintained as a constant somewhere?)

src/annotator/plugin/pdf.js
```diff
@@ -50,6 +90,69 @@ export default class PDF extends Delegator {
...
+    ));
+
+    const updateBannerHeight = () => {
+      /* istanbul ignore next */
+      if (!this._warningBanner) {
+        return;
+      }
+      const rect = this._warningBanner.getBoundingClientRect();
```
lyzadanger, line 135: A comment here might not hurt—explaining that we're moving the main container down to accommodate the warning banner, right?

src/annotator/plugin/pdf.js
```diff
@@ -50,6 +90,69 @@ export default class PDF extends Delegator {
...
+      mainContainer.style.top = rect.height + 'px';
+      this._warningBanner.style.top = -rect.height + 'px';
+    };
+
+    this._warningBanner = document.createElement('div');
+    this._warningBanner.className =
```
lyzadanger, line 141: I'll acknowledge that these classnames could be...tidier.

src/annotator/plugin/pdf.js
```diff
@@ -50,6 +90,69 @@ export default class PDF extends Delegator {
...
+      'annotator-pdf-warning-banner annotator-pdf-warning-banner--notice';
+    mainContainer.appendChild(this._warningBanner);
+    render(<WarningBanner />, this._warningBanner);
+
+    // @ts-expect-error - TS is missing `ResizeObserver`
+    if (typeof ResizeObserver !== 'undefined') {
+      // Update the banner when the window is resized or the Hypothesis
+      // sidebar is opened.
+      // @ts-ignore
+      new ResizeObserver(updateBannerHeight).observe(this._warningBanner);
```
lyzadanger, line 151: `ResizeObserver` is supported in, it looks like, all of our target browsers save for Android's Firefox. Historically, we've used `MutationObserver` (broader/older support but more complex API) as a polyfill, see `src/sidebar/util/observe-element-size` — do we have a plan going forward?

lyzadanger (approved):
Looking good.

src/annotator/plugin/pdf.js
robertknight: This selector isn't referenced anywhere else. It could do with a comment explaining where the element comes from though.

src/annotator/plugin/pdf.js
robertknight: `ResizeObserver` was only recently introduced in Safari (13.1) which means it isn't available on older iOS devices. In older browsers the banner height will simply be static based on the window size when the client first loads. I think that's an acceptable failure mode.

PR #2627, opened 2020-10-12
Gracefully abort startup in old browsers

robertknight:
Run a series of feature tests at the very start of the boot process and
abort with a console warning if they fail.

We display a console warning rather than a louder warning to avoid being
disruptive on sites maintained by publishers which embed Hypothesis and
which still want to support older browsers. On those sites the
Hypothesis annotation feature will simply be missing.

In future we may add louder failure modes (eg. a banner or something) in contexts (eg. Via)
where we know that using Hypothesis is the primary intent of the user when they visit the
page.

----

- [x] Test this locally in a VM with IE 11 or via Sauce Labs and verify that it works as expected. _Update: Tested via Sauce Labs using its Tunnel feature_.

### DIFF
--- a/src/boot/browser-check.js
+++ b/src/boot/browser-check.js
@@ -1,0 +1,32 @@
+/**
+ * Run a series of representative feature tests to see if the browser is new
+ * enough to support Hypothesis.
+ *
+ * We use feature tests to try to avoid false negatives, accepting some risk of
+ * false positives due to the host page having loaded polyfills for APIs in order
+ * to support older browsers.
+ *
+ * @return {boolean}
+ */
+export function isBrowserSupported() {
+  const checks = [
+    // ES APIs.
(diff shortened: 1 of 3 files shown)


LMS007 (approved):
This looks good Rob. I'm guessing this is somewhat of a best effort as these checks are generally new features added to newer browsers. Obviously we don't test all the new features, but just see if a few of them don't pass, and then we know to show the message.

I tested on older safari (5.1) and and old Opera -- both failed with the error, as well as IE11. I think the console warn is also good way to present the message as we're really trying to just let the site owners/devs know.

robertknight:
> I'm guessing this is somewhat of a best effort as these checks are generally new features added to newer browsers. Obviously we don't test all the new features, but just see if a few of them don't pass, and then we know to show the message.

Yes, that's exactly the idea.

PR #2637, opened 2020-10-13
Apply `dir="auto"` fix in Edge Legacy

robertknight:
The sidebar app failed to start in Edge Legacy due to `element.dir = "auto` assignments causing an exception during Preact component rendering. We already had a fix for this that applied to IE 11, so the fix for this problem is to change the check to test for the bug specifically rather than testing the user agent.

Although we've officially ended support for IE 11, we do still support the latest version of Edge Legacy (v18, before the recent switch to Chrome in January 2020).

### DIFF
--- a/src/shared/test/renderer-options-test.js
+++ b/src/shared/test/renderer-options-test.js
@@ -1,25 +1,9 @@
 import { createElement } from 'preact';
 
-import { setupIE11Fixes } from '../renderer-options';
-import { $imports } from '../renderer-options';
+import { setupBrowserFixes } from '../renderer-options';
 
 describe('shared/renderer-options', () => {
-  let fakeIsIE11;
-
-  beforeEach(() => {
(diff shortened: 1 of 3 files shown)


src/shared/test/renderer-options-test.js
```diff
@@ -30,44 +14,58 @@ describe('shared/renderer-options', () => {
...
+        setupBrowserFixes(fakeOptions);
         assert.isNotOk(fakeOptions.vnode);
       });
     });
 
-    context('when isIE11 is true', () => {
+    context('when `dir = "auto"` check fails', () => {
+      beforeEach(() => {
+        const fakeElement = {
+          set dir(value) {
```
LMS007, line 27: better test for sure

LMS007 (approved):
👍 
I tested in IE11 and Edge v18 and worked in both cases.
Note, IE11 showed our popup warning and Edge did not  -- as expected.

PR #2638, opened 2020-10-13
Added `document.elementsFromPoint` polyfill

esanzgar:
I added a polyfill for `document.elementsFromPoint` to support old browsers.

The polyfill is based on these two npm packages:
* https://www.npmjs.com/package/polyfill-elements-from-point
* https://www.npmjs.com/package/elementsfrompoint-polyfill

When using the non-standard `msElementsFromPoint` the return value is modified to a standard `Element[]` (instead of `NodeList`)

My concern is that the polyfill assumes all the elements return by `elementFromPoint` have `style` property.

### DIFF
--- a/src/annotator/adder.js
+++ b/src/annotator/adder.js
@@ -248,17 +248,27 @@ export class Adder {
     // Find the Z index of all the elements in the screen for five positions
     // around the adder (left-top, left-bottom, middle-center, right-top,
     // right-bottom) and use the greatest.
+    // In old implementations elementsFromPoint return null.
+    const leftTopElements = document.elementsFromPoint(left, top) ?? [];
+    const leftBottomElements =
+      document.elementsFromPoint(left, top + adderHeight) ?? [];
+    const middleCenterElements =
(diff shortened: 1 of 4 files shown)


robertknight (commented):
The approach looks reasonable. I think the assumption that all elements will have a `style` property is actually OK. You skip the `el.style` assignment if that is not the case. There are a couple more things that need to be done to make the polyfill actually load. Some of these steps are eliminated by https://github.com/hypothesis/client/pull/2636 so I suggest we merge that first and then rebase.

robertknight:
The vast majority of our users have browsers which natively support `elementsFromPoint` (see https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementsFromPoint#Browser_compatibility). It is only really Safari 10 + 11 and perhaps a small number of users on Firefox < 63 that don't. For those we already have a reasonable fallback in place.

On that basis I'm going to suggest that we close this for now rather than try to finish this polyfill PR off. Please let me know if you feel strongly otherwise.

PR #2640, opened 2020-10-14
Fix `document.evaluate` feature test in Edge Legacy

robertknight:
This API is available in Edge Legacy but with a quirk that arguments
that are optional in the spec (and also in TypeScript's knowledge of the API) are actually required in this browser.

This fixes an issue that the client would incorrectly abort startup in
Edge Legacy with a warning that the browser is unsupported.

This is a follow up to https://github.com/hypothesis/client/pull/2627.


robertknight:
I'm going to go ahead and merge this and re-test in QA on Sauce Labs.

PR #2642, opened 2020-10-14
Remove checks for IE 11 in `TagEditor` component

robertknight:
`TagEditor` tested if the browser was IE 11 to determine whether to
check the `inputType` property of `InputEvent`. However it is not only
IE 11 that did not support this property.

From reading the comments in https://github.com/hypothesis/client/pull/1558 it
appears that the `inputType` checks was originally added to resolve an
issue with accepting suggestions by pressing Enter in an earlier version
of `TagEditor` that used the `<datalist>` element. As far as I can tell
this is no longer applicable. `<datalist>` was replaced by a custom UI
and accepting suggestions by pressing Enter works without this logic.
Therefore this commit removes it entirely.

@LMS007 - Please let me know if you can think of something I've overlooked here.

### DIFF
--- a/src/sidebar/components/tag-editor.js
+++ b/src/sidebar/components/tag-editor.js
@@ -1,9 +1,8 @@
 import { createElement } from 'preact';
-import { useMemo, useRef, useState } from 'preact/hooks';
+import { useRef, useState } from 'preact/hooks';
 import propTypes from 'prop-types';
 
 import { withServices } from '../util/service-context';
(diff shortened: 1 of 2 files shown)


LMS007 (approved):

LMS007 (changes requested):
Approval may be premature...
Rob, pause on this for just a bit, need to test a few more things.

LMS007 (approved):
tested in several legacy edge browsers and ff to confirm everything works as expected.

PR #2659, opened 2020-10-20
Update Babel compilation settings to target newer browsers

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/2658**~~

Update Babel compilation targets to match the browsers we currently support,
rather than transpiling everything down to ES5.

This reduces the size of our JS bundles and improves performance by using native
implementations of ES2015+ language features. It should also avoid
occassional issues in development caused by interactions of Babel plugins that
compile native features to older syntax (eg. this happened a few times with `transform-async-to-promises`).

The client's boot script has a different configuration than the rest of the
client because it needs to be able to run in older browsers in order to show a
warning when the browser is not supported.

The production bundle size changes resulting from this are approximately:

Sidebar bundle: 398 KB => 329 KB (-17%)
Annotator bundle: 219 KB => 169 KB (-23%)

In the tests I had to fix a few cases where the `new` operator ended up being used with a value created by an arrow function. This works when the value is compiled into a function expression but not with a real/native arrow function.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,14 +4,17 @@
       "pragma": "createElement"
     }],
     ["@babel/preset-env", {
+       "bugfixes": true,
        "targets": {
-         "browsers": ["> 0.3%"]
+         "chrome": "55",
+         "edge": "17",
+         "firefox": "53",
+         "safari": "10.1",
       }
     }]
   ],
   "plugins": [
-    "angularjs-annotate",
-    "transform-async-to-promises"
+    "angularjs-annotate"
   ],
   "ignore": ["**/vendor/*"],
   "env": {
--- a/src/annotator/plugin/test/pdf-test.js
+++ b/src/annotator/plugin/test/pdf-test.js
@@ -45,7 +45,7 @@ describe('annotator/plugin/pdf', () => {
     };
 
     $imports.$mock({
-      './pdf-metadata': () => fakePDFMetadata,
+      './pdf-metadata': sinon.stub().returns(fakePDFMetadata),
       '../anchoring/pdf': fakePdfAnchoring,
 
       // Disable debouncing of updates.
--- a/src/boot/.babelrc
+++ b/src/boot/.babelrc
@@ -1,0 +1,15 @@
+{
+  "presets": [
+    ["@babel/preset-env", {
+       "targets": {
+         "chrome": "40",
+         "firefox": "38",
+         "safari": "9",
+         "ie": "11"
+      }
+    }]
+  ],
+  "plugins": [
+    "transform-async-to-promises"
+  ]
+}
--- a/src/shared/test/injector-test.js
+++ b/src/shared/test/injector-test.js
@@ -66,7 +66,7 @@ describe('Injector', () => {
 
       container.register('foo', fooFactory);
       container.register('bar', barFactory);
-      container.register('baz', bazFactory);
+      container.register('baz', { factory: bazFactory });
 
       const baz = container.get('baz');
       assert.equal(baz.foo, foo);
--- a/src/sidebar/services/test/oauth-auth-test.js
+++ b/src/sidebar/services/test/oauth-auth-test.js
@@ -78,7 +78,7 @@ describe('sidebar/services/oauth-auth', function () {
       authorize: sinon.stub().returns(Promise.resolve(null)),
     };
 
-    FakeOAuthClient = config => {
+    FakeOAuthClient = function (config) {
       fakeClient.config = config;
       return fakeClient;
     };
--- a/yarn.lock
+++ b/yarn.lock
@@ -154,7 +154,7 @@
   dependencies:
     "@babel/types" "^7.12.1"
 
-"@babel/helper-module-imports@^7.10.4", "@babel/helper-module-imports@^7.12.1":
+"@babel/helper-module-imports@^7.12.1":
   version "7.12.1"
   resolved "https://registry.yarnpkg.com/@babel/helper-module-imports/-/helper-module-imports-7.12.1.tgz#1644c01591a15a2f084dd6d092d9430eb1d1216c"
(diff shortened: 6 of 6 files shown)


.babelrc
```diff
@@ -4,14 +4,17 @@
       "pragma": "createElement"
     }],
     ["@babel/preset-env", {
+       "bugfixes": true,
```
lyzadanger, line 7: Can you help me understand this property?

.babelrc
```diff
@@ -4,14 +4,17 @@
...
        "targets": {
-         "browsers": ["> 0.3%"]
+         "chrome": "55",
```
lyzadanger, line 9: We have this notion of supporting a rolling set of browsers based on release cycles. It's expressed in various ways in various places in our apps/environments. I'm starting to wonder if we need some sort of lightweight management/checklist of places to update semi-regularly over time...any thoughts?

.babelrc
```diff
@@ -4,14 +4,17 @@
...
+         "edge": "17",
+         "firefox": "53",
+         "safari": "10.1",
       }
     }]
   ],
   "plugins": [
-    "angularjs-annotate",
-    "transform-async-to-promises"
+    "angularjs-annotate"
```
lyzadanger, line 17: I happen to know why we use this plugin, but, man, I feel like it could be confusing for future devs. Do we have any options for migrating away from this for our service injections?

src/annotator/plugin/test/pdf-test.js
```diff
@@ -45,7 +45,7 @@ describe('annotator/plugin/pdf', () => {
     };
 
     $imports.$mock({
-      './pdf-metadata': () => fakePDFMetadata,
+      './pdf-metadata': sinon.stub().returns(fakePDFMetadata),
```
lyzadanger, line 48: Could you explain this one to me like I'm 5? I want to understand how the arrow-bound variant fails. It would probably make sense to me if I saw it break the way you did initially.

src/boot/.babelrc
```diff
@@ -0,0 +1,15 @@
+{
```
lyzadanger, line 1: Is this what you were referencing in your PR description about the "client's boot script has a different..."? I just want to be sure I understand. I'm feeling more like we need to have a checklist for updating this later as now the browser targets are duplicated here and I could see missing this config entirely (i.e. only seeing the root `.babelrc` later).

src/shared/test/injector-test.js
```diff
@@ -66,7 +66,7 @@ describe('Injector', () => {
 
       container.register('foo', fooFactory);
       container.register('bar', barFactory);
-      container.register('baz', bazFactory);
+      container.register('baz', { factory: bazFactory });
```
lyzadanger, line 69: Again, I really want to understand this better...

lyzadanger (commented):
From a conceptual and code diff perspective this (well, mostly) makes sense. I'm hoping to grok the arrow-binding differences a little better, but, eh.

I haven't tested this branch yet because of laziness/its dependency on `remove-boot-dependency-on-shared`...

.babelrc
robertknight: It is a bit subtle, but essentially it avoids issues where a natively supported feature gets compiled down to ES5 because some specific browser version had a bug in an edge case. The manual explains it better - https://babeljs.io/docs/en/babel-preset-env#bugfixes

.babelrc
robertknight: I think that's a good idea. Any preferences on where to maintain it?

.babelrc
robertknight: Yes, we could probably create our own plugin which does the same thing but avoids the Angular reference. Coincidentally I had to explain this vestige to Eduardo earlier today.

.babelrc
lyzadanger: Isn't that the $64 question lately...I'll think about it.

src/annotator/plugin/test/pdf-test.js
robertknight: You can see the problem in a browser console:

arrow-function

The latter syntax is what arrow function used to get turned into with the old Babel settings. Now that no longer happens.

src/boot/.babelrc
robertknight: > Is this what you were referencing in your PR description about the "client's boot script has a different..."?

Yes, exactly. We might not ever need to update this - since the amount of code in the boot script is small it doesn't matter so much if we were to compile it down to ES5 for years to come.

src/shared/test/injector-test.js
robertknight: Ah, I should clarify. `bazFactory` is declared with an arrow function above, but `register(...)` by default assumes its second argument is a class and tries to call `new` on it. The easiest way to see the problem is to try reverting this to the old code and running the test.

LMS007 (approved):
👍 

Confirmed that a console warning was shown in these browsers:

```
"chrome": "40",
"firefox": "38",
"ie": "11"
```

I could not confirm in safari v9 because port 3001 was not working with my tunnel -- likely a browserstack issue.

I confirmed the app worked in 

```
chrome": "55",
"edge": "17",
"firefox": "53",
"safari": "10.1",
```

I noticed one visual issue in chrome 55 with the annotator not rendering right. It looked like none of the css took and I suspect it has to to do with the shadow dom. I did look up shadow dom support and it says chrome 53 supported v1. I think this is somewhat minor as the app does still work and very few folks would be using v55. I confirmed this issue is resolved in v58.


Screen Shot 2020-10-21 at 12 26 19 PM

robertknight:
> I think this is somewhat minor as the app does still work and very few folks would be using v55. I confirmed this issue is resolved in v58.

I have seen this issue occasionally during development in Safari, but I've never seen it in the production build. I think it is a sporadic issue with loading stylesheets from the dev server inside the shadow DOM root. I'll do some more testing on QA.

PR #2663, opened 2020-10-21
Remove unused iframe attributes

robertknight:
- Remove `seamless` attribute which has not been supported by major
   browsers for years [1]
 - Remove the `name="hyp_sidebar_frame"` attribute which is not
   referenced anywhere else in the code. I did find a couple of
   references to it in Google, but only in scripts written by Hypothesis
   team members which are not run regularly.

[1] https://caniuse.com/iframe-seamless


LMS007 (approved):
👍

PR #2665, opened 2020-10-22
Enable terser workarounds for Safari 10

### LINKED ISSUE #2664: Production (minified) build does not work in Safari 10
In Safari 10 production builds fail to start with an error:

```
SyntaxError: Cannot declare a let variable twice: 't'.
```

The issue does not occur when using a local development build. The same error is seen in the browser console when visiting pages that are part of the H website. The production build does work in Safari 11 and newer.

Safari 10.1 is currently the oldest version of Safari that we target internally, although we recommend users to use the newest one that is available.

The Terser minifier we use has a specific option to turn on workarounds for Safari 10 bugs (see the [`--safari10`](https://github.com/terser/terser#command-line-options) option here). We don't have this enabled and so I suspect this may be the issue.

robertknight:
This fixes an error in production builds of the client in Safari 10:

```
SyntaxError: Cannot declare a let variable twice: 't'.
```

The bundle is the same size with this flag enabled. From a quick look in the terser source code it seems that the effect of this is to prevent variables from being shadowed in certain situations when that would normally be OK (more details here: https://bugs.webkit.org/show_bug.cgi?id=171041).

Fixes #2664

-----

I tested this locally with Safari 10 on macOS 10.12 by making `make dev` build a production build of the client using:

```diff
diff --git a/Makefile b/Makefile
index 2f7b4f97a..0ae45efa9 100644
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ help:
 
 .PHONY: dev
 dev: build/manifest.json
-	node_modules/.bin/gulp watch
+	env NODE_ENV=production node_modules/.bin/gulp watch
 
 .PHONY: test
 test: node_modules/.uptodate
```

And then accessing the development server at http://localhost:3000 using Safari 10 via Sauce Labs.

### DIFF
--- a/scripts/gulp/minify-stream.js
+++ b/scripts/gulp/minify-stream.js
@@ -31,7 +31,10 @@ function minifyStream() {
       const code = Buffer.concat(this.chunks).toString();
 
       // See https://github.com/terser/terser#minify-options-structure
-      const options = {};
+      const options = {
+        // See https://github.com/hypothesis/client/issues/2664.
+        safari10: true,
+      };
 
       // If the code we're minifying has a sourcemap then generate one for the
       // minified output, otherwise skip it.


esanzgar:
Looks good!

robertknight:
I'll take that as an approval ;)

esanzgar (approved):

PR #2667, opened 2020-10-22
Remove obsolete workaround for a Babel plugin conflict

robertknight:
Remove a workaround in `oauth-auth.js` associated with the transform-async-to-promises Babel plugin that is no longer needed following https://github.com/hypothesis/client/pull/2659.

The plugin itself is still needed because it is used when building the boot script, although we could just ban use of async/await in that module instead.


LMS007 (approved):
+1 for removing the plug-in later on and then putting some comments in the boot script.

PR #2668, opened 2020-10-22
Remove polyfills for features supported by all target browsers

robertknight:
Remove polyfills for features that are supported by all of our target
browsers:

 - `fetch` (https://caniuse.com/fetch)
 - The `URL` constructor (https://caniuse.com/url)
 - `String.prototype.normalize` (https://caniuse.com/mdn-javascript_builtins_string_normalize)


LMS007 (approved):
👍 

https://caniuse.com is great.

PR #2670, opened 2020-10-23
Remove npm dependency

esanzgar:
Rely exclusively on `yarn`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -189,11 +189,8 @@ stage('Publish') {
                     """
                     sh "scripts/create-github-release.js"
 
-                    // Publish the updated package to the npm registry.
-                    // Use `npm` rather than `yarn` for publishing.
-                    // See https://github.com/yarnpkg/yarn/pull/3391.
                     sh "echo '//registry.npmjs.org/:_authToken=${env.NPM_TOKEN}' >> \$HOME/.npmrc"
-                    sh "npm publish --tag ${npmTag}"
+                    sh "yarn publish --no-interactive --tag ${npmTag} --new-version=${newPkgVersion}"
                     sh "scripts/wait-for-npm-release.sh ${npmTag}"
 
                     // Deploy the client to cdn.hypothes.is, where the embedded
--- a/scripts/wait-for-npm-release.sh
+++ b/scripts/wait-for-npm-release.sh
@@ -19,7 +19,7 @@ fi
 expected_version=$(node -p "require('./package.json').version")
 while [ true ]
 do
-  released_version=$(npm show hypothesis dist-tags.$dist_tag)
+  released_version=$(yarn info -s hypothesis dist-tags.$dist_tag)
   if [ "$released_version" = "$expected_version" ]; then
     break
   fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,6 @@ RUN apk update && apk add --no-cache \
   git \
   make \
   nodejs \
-  npm \
   yarn
 
 # Do not download a Chrome build as part of installing the "puppeteer" package,


Jenkinsfile
```diff
@@ -193,7 +193,7 @@ stage('Publish') {
                     // Use `npm` rather than `yarn` for publishing.
                     // See https://github.com/yarnpkg/yarn/pull/3391.
                     sh "echo '//registry.npmjs.org/:_authToken=${env.NPM_TOKEN}' >> \$HOME/.npmrc"
-                    sh "npm publish --tag ${npmTag}"
+                    sh "yarn publish --tag ${npmTag}"
```
robertknight, line 196: There is a comment just above this change stating that the use of `npm publish` rather than `yarn publish` is intentional. In the discussion for a PR there is a link to this commit that explains the reason: https://github.com/hypothesis/client/commit/975843f3ee0d6924479f2f5fa59a66a563db63ea

Does this issue still apply? If so, we should avoid this change. If it no longer applies, we should update the comment above. Also, what about the `.npmrc` line above? Is that the right place to write a token that `yarn publish` uses.

scripts/wait-for-npm-release.sh
```diff
@@ -19,7 +19,7 @@ fi
 expected_version=$(node -p "require('./package.json').version")
 while [ true ]
 do
-  released_version=$(npm show hypothesis dist-tags.$dist_tag)
+  released_version=$(yarn info -s hypothesis dist-tags.$dist_tag)
```
robertknight, line 22: This change looks fine. `yarn info -s hypothesis dist-tags.latest` produces the same result as `npm show`.

robertknight (commented):
I had a couple of queries about the `yarn publish` change given the comments just above the changed line.

scripts/wait-for-npm-release.sh
esanzgar: I confirm it does produce the same result.

Jenkinsfile
```diff
@@ -189,11 +189,7 @@ stage('Publish') {
                     """
                     sh "scripts/create-github-release.js"
 
-                    // Publish the updated package to the npm registry.
```
esanzgar, line 192: Based on this https://github.com/yarnpkg/yarn/pull/5108, I modified the command.

I tested it and I confirm it works as expected. 

I removed the outdated comment and I simplified the way NPM_TOKEN is passed to the yarn command.

Jenkinsfile
esanzgar: Based on this https://github.com/yarnpkg/yarn/pull/5108, I modified the command.

I tested it and I confirm it works as expected. 

I removed the outdated comment and I simplified the way NPM_TOKEN is passed to the yarn command.

Jenkinsfile
```diff
@@ -189,11 +189,7 @@ stage('Publish') {
...
-                    // Use `npm` rather than `yarn` for publishing.
-                    // See https://github.com/yarnpkg/yarn/pull/3391.
-                    sh "echo '//registry.npmjs.org/:_authToken=${env.NPM_TOKEN}' >> \$HOME/.npmrc"
-                    sh "npm publish --tag ${npmTag}"
+                    sh "ENV_TOKEN=${env.NPM_TOKEN} yarn publish --no-interactive --tag ${npmTag} --new-version=${newPkgVersion}"
```
robertknight, line 192: Where does `ENV_TOKEN` come from? I couldn't find a reference to that environment variable as something that yarn would recognize via Google.

robertknight (commented):
I had a query about the `ENV_TOKEN` environment variable.

Jenkinsfile
esanzgar: Sorry, my bad. I correct it to `NPM_TOKEN`.

Alternatively, the previous line could be kept:
`sh "echo '//registry.npmjs.org/:_authToken=${env.NPM_TOKEN}' >> \$HOME/.npmrc"`

esanzgar:
> I had a query about the `ENV_TOKEN` environment variable.

I corrected it to `NPM_TOKEN`. 

More here about the `NPM_TOKEN`:
https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow

Jenkinsfile
```diff
@@ -189,11 +189,7 @@ stage('Publish') {
                     """
                     sh "scripts/create-github-release.js"
 
-                    // Publish the updated package to the npm registry.
-                    // Use `npm` rather than `yarn` for publishing.
-                    // See https://github.com/yarnpkg/yarn/pull/3391.
-                    sh "echo '//registry.npmjs.org/:_authToken=${env.NPM_TOKEN}' >> \$HOME/.npmrc"
-                    sh "npm publish --tag ${npmTag}"
+                    sh "NPM_TOKEN=${env.NPM_TOKEN} yarn publish --no-interactive --tag ${npmTag} --new-version=${newPkgVersion}"
```
robertknight, line 192: Have you confirmed that `yarn publish` definitely uses `NPM_TOKEN` if it is set this way and `.npmrc` is not present?

From what I can tell looking at the yarn issues there is no built-in support for `NPM_TOKEN` in yarn itself. Instead there is a feature where `.npmrc` can contain references to environment variables and there is a convention to write `//registry.npmjs.org/:_authToken=${NPM_TOKEN}` into the `.npmrc` file. See https://github.com/yarnpkg/yarn/pull/1207.

I did however find a reference to `NPM_AUTH_TOKEN` (and also `YARN_AUTH_TOKEN`) in the yarn sources: https://github.com/yarnpkg/yarn/blob/a4708b29ac74df97bac45365cba4f1d62537ceb7/src/cli/commands/login.js#L64

The other issue I see here is that a `.npmrc` file could currently be left over from a previous build and I'm not sure what happens if the auth token is set in both places.

The simplest remedy I can see here is to keep the "sh echo '..." line unless you're sure that setting `NPM_TOKEN` on its own works, and just change the line that invokes `npm publish`.

robertknight (commented):
I did a little more research this morning and I have a query about the `NPM_TOKEN` line.

Jenkinsfile
esanzgar: Yes, I confirm that it works with NPM_TOKEN.

I have reverted the `sh "echo...` line.

robertknight (approved):
LGTM

PR #2685, opened 2020-10-29
Fail more gracefully in unsupported versions of Edge Legacy

robertknight:
Add a test for the [`Element.prepend()` method](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/prepend) so that startup fails more
gracefully in older, unsupported versions of Edge Legacy (<= 16). This
API was chosen as a representative test because it's Chrome/Edge Legacy
support closely matches our minimum versions.

The same change was made for the LMS app in https://github.com/hypothesis/lms/pull/2209.


lyzadanger (approved):
This looks sensible and is well-explained. Let me know if you'd like some form of testing.

PR #2698, opened 2020-11-02
Move browser support list to `browserslist` key in package.json

robertknight:
Move the list of supported browsers from `.babelrc` to the
`browserslist` key in `package.json`. This enables other
browserslist-enabled tools such as autoprefixer to adapt their output to
the target browsers.

This reduces the size of the production CSS bundles slightly by reducing
the number of unnecessary vendor prefixed rules that are generated. There should
be no change in the generated JS bundles.

See https://github.com/postcss/autoprefixer#browsers,
https://github.com/browserslist/browserslist#queries and
https://babeljs.io/docs/en/babel-preset-env#browserslist-integration.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,13 +4,7 @@
       "pragma": "createElement"
     }],
     ["@babel/preset-env", {
-       "bugfixes": true,
-       "targets": {
-         "chrome": "55",
-         "edge": "17",
-         "firefox": "53",
-         "safari": "10.1",
-      }
+       "bugfixes": true
     }]
   ],
   "plugins": [
--- a/.babelrc
+++ b/.babelrc
@@ -3,6 +3,8 @@
     ["@babel/preset-react", {
       "pragma": "createElement"
     }],
+
+    // Compile JS for browser targets set by `browserslist` key in package.json.
     ["@babel/preset-env", {
        "bugfixes": true
     }]
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "wrap-text": "^1.0.7",
     "zen-observable": "^0.3.0"
   },
+  "browserslist": "chrome 55, edge 17, firefox 53, safari 10.1",
   "browserify": {
     "transform": [
       "browserify-versionify",


lyzadanger (approved):
I think in general this is a good move.

I have a few passing thoughts. I note that [browserslist recommends using package.json](https://github.com/browserslist/browserslist#queries) versus a discrete `.browserslistrc` config. OK, but I do find myself wondering _why_. That's trivial...just curious.

If we're going to use `browserslist`, do we have any interest in availing ourselves upon their relative designations (last x versions, > 0.5%, etc.) to keep from having an arms race/having to remember to update the specific version numbers over time? Or are there other needs—e.g. needing to match browser support defined in other configs, needing to know _exactly which browsers we support_—that preclude this? It seems like defining our own specific versions is more precise, but their data would be more meaningful/accurate over time.

robertknight:
>  I note that browserslist recommends using package.json versus a discrete .browserslistrc config. OK, but I do find myself wondering why.

The main benefit I see is avoiding having yet another tiny config file in the repository root.

> If we're going to use browserslist, do we have any interest in availing ourselves upon their relative designations (last x versions, > 0.5%, etc.) to keep from having an arms race/having to remember to update the specific version numbers over time?

We used to use the `> 0.5%` approach, but I now prefer to define specific versions, which then have to be updated periodically, for several reasons:

1. You can cross-reference this against MDN or caniuse more easily to know eg. which APIs need to be polyfilled or which language features are expected to be natively supported
2. It provides a specific baseline that we can test against with Sauce Labs etc.
3. It provides a direct answer to the question of whether Browser X version Y is expected to work with a particular application, should that come up in a support query.
4. I assume the "> X%" metrics refer to global browser usage, which is not necessarily the same as the audience for our particular application ("> 1%" would include IE 11 globally, but not for us for example)

robertknight:
I added a pointer in the `.babelrc` file to where the browser targets are set, since I don't expect all readers will know about the `browserslist` key.

PR #2699, opened 2020-11-03
Enable `strict` option in tsconfig.json

robertknight:
This catches a few minor errors which the previous configuration did
not. See https://www.typescriptlang.org/tsconfig#strict.

`noImplicitAny` is turned off because we still have a lot of
un-annotated variables/parameters in the code.

### DIFF
--- a/src/sidebar/components/group-list-item.js
+++ b/src/sidebar/components/group-list-item.js
@@ -70,7 +70,7 @@ function GroupListItem({
   /**
    * Opens or closes the submenu.
    *
-   * @param {MouseEvent|KeyboardEvent} event
+   * @param {Event} event
    */
   const toggleSubmenu = event => {
     event.stopPropagation();
--- a/src/sidebar/media-embedder.js
+++ b/src/sidebar/media-embedder.js
@@ -396,17 +396,12 @@ function replaceLinkWithEmbed(link) {
  *     the width of the embed.
  */
 export function replaceLinksWithEmbeds(element, { className } = {}) {
-  let links = element.getElementsByTagName('a');
+  // Get a static (non-live) list of <a> children of `element`.
+  // It needs to be static because we may replace these elements as we iterate over them.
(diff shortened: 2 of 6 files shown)


src/sidebar/components/group-list-item.js
```diff
@@ -70,7 +70,7 @@ function GroupListItem({
   /**
    * Opens or closes the submenu.
    *
-   * @param {MouseEvent|KeyboardEvent} event
+   * @param {Event} event
```
robertknight, line 73: `toggleSubmenu` is passed as a prop to a component which doesn't specify what kind of `Event` it passes to the callback.

src/sidebar/media-embedder.js
```diff
@@ -396,17 +396,12 @@ function replaceLinkWithEmbed(link) {
  *     the width of the embed.
  */
 export function replaceLinksWithEmbeds(element, { className } = {}) {
-  let links = element.getElementsByTagName('a');
+  // Get a static (non-live) list of <a> children of `element`.
+  // It needs to be static because we may replace these elements as we iterate over them.
+  const links = Array.from(element.getElementsByTagName('a'));
```
robertknight, line 401: There is no functional change here. I've just rewritten the logic in a more modern way to avoid a spurious TS error.

src/sidebar/services/frame-sync.js
```diff
@@ -126,7 +126,7 @@ export default function FrameSync(annotationsService, bridge, store) {
       annotationsService.create(annot);
     });
 
-    bridge.on('destroyFrame', destroyFrame.bind(this));
+    bridge.on('destroyFrame', frameIdentifier => destroyFrame(frameIdentifier));
```
robertknight, line 129: `destroyFrame` is a plain function, not a method.

esanzgar (approved):
👍🏼

PR #2700, opened 2020-11-04
Replace babel-plugin-angularjs-annotate with babel-plugin-inject-args

robertknight:
It is confusing to have a Babel plugin with an Angular-related name and
Angular-sounding annotations (`@ngInject`) in the code, even though
we're not using AngularJS. Although we're not using AngularJS any more,
we still have services that are instantiated by a dependency injection
container.  The container reads dependency names from a `$inject`
property on service functions/classes. This `$inject` property is added
by `babel-plugin-angularjs-annotate`.

This commit replaces the `angularjs-annotate` Babel plugin with one
maintained by us [1]. The new plugin provides only the functionality that we
need (eg. only processes explicitly annotated functions) and uses a more generic
`@inject` annotation.

[1] https://github.com/hypothesis/babel-plugin-inject-args

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,9 +9,7 @@
        "bugfixes": true
     }]
   ],
-  "plugins": [
-    "angularjs-annotate"
-  ],
+  "plugins": ["inject-args"],
   "ignore": ["**/vendor/*"],
   "env": {
     "development": {
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "autoprefixer": "^10.0.1",
     "aws-sdk": "^2.345.0",
     "axe-core": "^4.0.0",
-    "babel-plugin-angularjs-annotate": "^0.10.0",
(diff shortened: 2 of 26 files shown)


lyzadanger (approved):
Cool. I can build this and it works.

My only feedback is that`@inject`, while totally accurate and appropriate, is not particularly unique. For example, as a dev, I could see myself being like "what magic does this perform" and trying to google for `inject` in the code/dependencies. If it had a more specific, albeit cumbersome, name, like `@babel-inject`, it might be easier to track back the magic to its source. I know you've already had a naming runaround, though...

robertknight:
> My only feedback is that@inject, while totally accurate and appropriate, is not particularly unique.

That's a fair point. `@babel-inject` suggests to me that it is a built-in feature of Babel rather than a custom thing. I'd like to get this merged and then if someone can suggest an alternative which we collectively like it will be easy to update the plugin to support it.

PR #2701, opened 2020-11-05
Speed up `make dev` and add `make build`

robertknight:
This PR fixes a couple of issues with the `Makefile` in this repository:

- `make dev` would unnecessarily create production builds of the client before starting the dev server if `node_modules/` was out of date, rather than just running `yarn install`
- The make command to create a production build had a non-obvious form (`make build/manifest.json`) and incomplete dependency information so it would not create a new build if the contents of `src/` had changed since the command was previously run. I've replaced it with `make build` which just always creates a new production build.

The upshot of the first change is that running `make dev` will get a dev server running 20-30 seconds faster on a typical laptop if `node_modules/` needs updating.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ default: help
 help:
 	@echo "make help              Show this help message"
 	@echo "make dev               Run the app in the development server"
+	@echo "make build             Create a production build of the client"
 	@echo "make lint              Run the code linter(s) and print any warnings"
 	@echo "make checkformatting   Check code formatting"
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
Specifically, on my laptop (2019 MBP):

* On `master`, `make dev` from scratch (`node_modules` nuked): ~50sec
* On this branch, `make dev` from scratch: ~23s

LGTM

esanzgar (approved):
I noticed this before.

Thanks for taking the time to improve this!

PR #2702, opened 2020-11-05
Remove ES 2015 and ES 2016 polyfill sets

robertknight:
All of the browsers we currently support, as specified by the
`browserslist` key in package.json, [fully support](https://caniuse.com/es6) ES 2015 and ES 2016.

The only exception I'm aware of is that Edge Legacy v17 and v18 do not
support using string search/match methods with custom (non-string/regex)
types via [`Symbol.{match, search, replace}`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/search). That's not something we've
ever used AFAIK. I didn't even know it was a thing.


lyzadanger (approved):
👍 Nice housekeeping

Issue #1464, opened 2019-10-28
Creating annotation fails when selection starts or ends in a link in a PDF


LMS007:
Creating an annotation fails if the selection starts or ends inside a link annotation in a PDF - that is, a link which is contained within the `annotationLayer` element for a given PDF page.

**Steps to reproduce:**

Open http://www.orimi.com/pdf-test.pdf in Via or using the browser extension.

1. Select all the text on the last line of the only page in this PDF, including the URL http://www.education.gov.yk.ca/
2. Click the `Highlight` or `Annotate` button 

**Expected result** 

The selected text is highlighted and a new highlight appears in the "Annotations" tab in the client

**Actual result** 

A JS error occurs and no highlight appears
![Screen Shot 2019-10-28 at 4 22 38 PM](https://user-images.githubusercontent.com/3939074/67725322-2eb15280-f99f-11e9-8fb5-17fe9b9a081a.png)

Related to https://github.com/hypothesis/client/issues/1329

**Notes**

A PDF.js page has several layers: A `<canvas>` containing the visible content, a div containing invisible text (`textLayer`) and a div containing external links and possibly other annotations (`annotationLayer`). Our PDF anchoring code currently expects the selection to contain hidden text from the `textLayer` div but on pages with annotations the selection can end up starting or ending in the `annotationLayer`. Hypothesis functionality aside, this DOM structure causes problems when trying to select text in the PDF because links occur after all the text on the page rather than being part of the text flow as they would be on a normal web page.

robertknight:
It looks like the problem here relates not to the selection of the last word in the page but the fact that the selection ends in a link which is in the `annotationLayer` div in the PDF page rather than the `textLayer`:

<img width="969" alt="Annotation layer link" src="https://user-images.githubusercontent.com/2458/99436873-8ce20800-2909-11eb-9121-c3f3db405541.png">

The same issue causes unexpected behavior if you try to select the last line of this paragraph in the http://localhost:3000/pdf/budlong document in the client's dev server:

<img width="843" alt="Screenshot 2020-11-17 at 19 20 26" src="https://user-images.githubusercontent.com/2458/99437191-0a0d7d00-290a-11eb-9f7c-6825f9a584cb.png">

I've tried to select the text "Budlong Woods ... community area" but because of the DOM structure this ended up selecting the text "Early history" as well. If you try to create an annotation from this selection it will fail.

PR #2742, opened 2020-11-17
Hide annotation layers in PDF when there is a selection

robertknight:
Content in a PDF's annotation layer, such as link annotations, can
interfere with text selection and also cause creating annotations to
fail [1]. This commit resolves the problem by temporarily hiding annotation
layers when there is a non-empty selection. This allows the user to
select text inside link annotations (for example) while still enabling
the user to activate the link normally when there is no selection.

Fixes https://github.com/hypothesis/client/issues/1464

[1] This is because the annotation layer is outside the text layer and
    the PDF anchoring code expects the selection to be within the text
    layer.

----

**Testing**

With the client's dev server running go to http://localhost:3000/pdf/budlong and try to make a selection that ends inside a link in the text. On master the text selection jumps to include the whole page when doing this. On this branch it should be possible to make the selection normally, as if you were working with a web page. Also click on one of these links when there is no selection and see that it behaves normally. Note that you can't _start_ the selection inside a link, but that's true of normal web pages as well.

### DIFF
--- a/src/annotator/plugin/pdf.js
+++ b/src/annotator/plugin/pdf.js
@@ -75,9 +75,32 @@ export default class PDF extends Delegator {
     this._warningBanner = null;
 
     this._checkForSelectableText();
+
+    // Hide annotation layer when the user is making a selection. The annotation
+    // layer appears above the invisible text layer and can interfere with text
+    // selection. See https://github.com/hypothesis/client/issues/1464.
+    this._updateAnnotationLayerVisibility = () => {
+      const selection = /** @type {Selection} */ (window_.getSelection());
+
+      // Add CSS class to indicate whether there is a selection. Annotation
+      // layers are then hidden by a CSS rule in `pdfjs-overrides.scss`.
+      this.pdfViewer.viewer.classList.toggle(
+        'is-selecting',
+        !selection.isCollapsed
+      );
+    };
+
+    document.addEventListener(
+      'selectionchange',
+      this._updateAnnotationLayerVisibility
+    );
   }
 
   destroy() {
+    document.removeEventListener(
+      'selectionchange',
+      this._updateAnnotationLayerVisibility
+    );
     this.pdfViewer.viewer.classList.remove('has-transparent-text-layer');
     this.observer.disconnect();
   }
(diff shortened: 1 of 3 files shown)


src/annotator/plugin/test/pdf-test.js
```diff
@@ -61,17 +69,38 @@ describe('annotator/plugin/pdf', () => {
...
+    assert.isFalse(pdfViewerHasClass('is-selecting'));
+
+    selection.selectAllChildren(document.body);
+    await awaitEvent(document, 'selectionchange');
+    assert.isTrue(pdfViewerHasClass('is-selecting'));
+
+    selection.collapseToStart();
+    await awaitEvent(document, 'selectionchange');
+    assert.isFalse(pdfViewerHasClass('is-selecting'));
+  });
```
robertknight, line 102: `awaitEvent` is needed here because the `selectionchange` event is fired asynchronously by the browser. We _could_ forcibly dispatch it ourselves in the test, but it's more faithful to normal application behavior this way.

lyzadanger:
@robertknight First of all, the fix of making it possible to highlight the right thing at all is huge. Is it expected behavior on this branch that one still gets an Exception in the console when trying to annotate such a selection?

Locally, if I make this selection on the Budlong PDF: 

image

And then click `Annotate`, I get:

image

robertknight:
> Is it expected behavior on this branch that one still gets an Exception in the console when trying to annotate such a selection?

No, that's a bug! The annotation still gets created so this must have happened afterwards. This is fixed now.

lyzadanger:
@robertknight Thanks; your latest changes fix the bug for me. I'll take a look at the code soon!

src/annotator/plugin/pdf.js
```diff
@@ -75,9 +75,32 @@ export default class PDF extends Delegator {
     this._warningBanner = null;
 
     this._checkForSelectableText();
+
+    // Hide annotation layer when the user is making a selection. The annotation
```
lyzadanger, line 79: Bravo on comment, thank you.

src/annotator/plugin/pdf.js
```diff
@@ -75,9 +75,32 @@ export default class PDF extends Delegator {
...
+    // layer appears above the invisible text layer and can interfere with text
+    // selection. See https://github.com/hypothesis/client/issues/1464.
+    this._updateAnnotationLayerVisibility = () => {
+      const selection = /** @type {Selection} */ (window_.getSelection());
+
+      // Add CSS class to indicate whether there is a selection. Annotation
```
lyzadanger, line 85: Again, very essential comment—this would have seemed like magic to a future dev otherwise.

lyzadanger (approved):
OK, yeah, this is straightforward. Cool.

PR #2750, opened 2020-11-20
Remove one pixel gap between bucket-bar and iframe

esanzgar:
On mobile devices, reseting the zoom level (pinch gesture) leaves a one pixel gap between the `bucket-bar` and the `iframe`. Adding one extra pixel to the width of the `bucket-bar` eliminates this issue.

This issue sometimes goes away by simple by opening and closing the sidebar. One explanation for that could be that the touch event modifies de zoom level. Otherwise I have not other explanation for this seemingly random behaviour.

I have observed this one-pixel gap in Chrome and Safari. Here two screenshoots from an iPad:

Safari:
IMG-0367

Chrome:
IMG-0368

### DIFF
--- a/src/styles/annotator/bucket-bar.scss
+++ b/src/styles/annotator/bucket-bar.scss
@@ -11,7 +11,9 @@
     pointer-events: none;
     position: absolute;
     height: 100%;
-    width: var.$annotator-bucket-bar-width;
+    // 2020-11-20: interim and pragmatic solution for an apparent glitch on Safari and Chrome.
+    // Adding one pixel resolve this issue: https://github.com/hypothesis/client/pull/2750
+    width: var.$annotator-bucket-bar-width + 1;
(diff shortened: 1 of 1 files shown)


src/styles/annotator/bucket-bar.scss
```diff
@@ -11,7 +11,9 @@
     pointer-events: none;
     position: absolute;
     height: 100%;
-    width: var.$annotator-bucket-bar-width;
+    // Nov-20-2020: interim and pragmatic solution for an apparent glitch on Safari and Chrome.
```
robertknight, line 14: As an aside, if you're ever specifying a date numerically I would always recommend to use ISO dates (YYYY-MM-DD) at Hypothesis to avoid confusion between the Europeans and the Americans on the team who use different ordering in formats where the year comes last (DD-MM-YY vs MM-DD-YY).

robertknight (approved):
LGTM

PR #2783, opened 2020-12-02
Find annotation.css in a more robust way

### LINKED ISSUE #2752: Adder styles sometimes fail to load
(_Update (2020-11-21): The issue is sporadic. The steps below may take several attempts to reproduce the issue - @robertknight_)

### Steps to reproduce
1. activate the browser-extension
2. go to https://www.espn.com/
3. after display the sidebar annotate something on the page

### Expected behaviour

Adder should displayed with styles.

### Actual behaviour

Adder is displayed without styles

### Additional details

* This page has a lot of ads and takes a very long time to fully load.

* After steps 1-3, if the browser-extension is disabled and re-enabled the adder shows correct styles.

Untitled

The above two hints indicate that the issue is caused by the slow loading of the page.

esanzgar:
As indicated by @robertknight  `document.styleSheets` is complete only when stylesheets are processed. This can take a while for certain heavy pages.

The alternative approach presented here to find `annotation.css` is warranted to find the url when the code is run.

Closes #2752

### DIFF
--- a/src/annotator/util/shadow-root.js
+++ b/src/annotator/util/shadow-root.js
@@ -2,9 +2,13 @@
  * Load stylesheets for annotator UI components into the shadow DOM root.
  */
 function loadStyles(shadowRoot) {
-  const adderStyles = Array.from(document.styleSheets)
-    .map(sheet => sheet.href)
-    .filter(url => (url || '').match(/annotator\.css/));
+  const url = /** @type {HTMLLinkElement|undefined} */ (document.querySelector(
+    'link[rel="stylesheet"][href*="/build/styles/annotator.css"]'
+  ))?.href;
+
+  if (!url) {
+    return;
+  }
 
   // Stylesheet <link> elements are inert inside shadow roots [1]. Until
   // Shadow DOM implementations support external stylesheets [2], grab the
@@ -18,7 +22,7 @@ function loadStyles(shadowRoot) {
   // get a usable adder, albeit one that uses browser default styles for the
   // toolbar.
   const styleEl = document.createElement('style');
-  styleEl.textContent = adderStyles.map(url => `@import "${url}";`).join('\n');
+  styleEl.textContent = `@import "${url}";`;
   shadowRoot.appendChild(styleEl);
 }
 
--- a/src/annotator/util/test/shadow-root-test.js
+++ b/src/annotator/util/test/shadow-root-test.js
@@ -29,5 +29,21 @@ describe('annotator/util/shadow-root', () => {
       assert.ok(styleEl);
       assert.match(styleEl.textContent, /@import ".*annotator\.css.*"/);
     });
+
+    it('does not inject stylesheets into the shadow root if style is not found', () => {
+      const container = document.createElement('div');
+
+      const link = document.querySelector(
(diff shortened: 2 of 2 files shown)


src/annotator/util/test/shadow-root-test.js
```diff
@@ -29,5 +29,21 @@ describe('annotator/util/shadow-root', () => {
       assert.ok(styleEl);
       assert.match(styleEl.textContent, /@import ".*annotator\.css.*"/);
     });
+
+    it('does not injects stylesheets into the shadow root if style is not found', () => {
```
robertknight, line 33:
```suggestion
    it('does not inject stylesheets into the shadow root if style is not found', () => {
```

src/annotator/util/test/shadow-root-test.js
```diff
@@ -29,5 +29,21 @@ describe('annotator/util/shadow-root', () => {
...
+        'link[rel="stylesheet"][href*="/build/styles/annotator.css"]'
+      );
+      // Removing the `rel` attribute is enough for the URL to not be found
+      link.removeAttribute('rel');
+
+      createShadowRoot(container);
+
+      const styleEl = container.shadowRoot.querySelector('style');
+      assert.isNull(styleEl);
+      link.setAttribute('rel', 'stylesheet');
```
robertknight, line 46: This could potentially cause unexpected failures in other tests if this test fails. In this case I think we're very likely to be safe since other tests generally depend on the CSS rather than the element and the styles will have already loaded by this point.

src/annotator/util/shadow-root.js
```diff
@@ -2,9 +2,13 @@
  * Load stylesheets for annotator UI components into the shadow DOM root.
  */
 function loadStyles(shadowRoot) {
-  const adderStyles = Array.from(document.styleSheets)
-    .map(sheet => sheet.href)
-    .filter(url => (url || '').match(/annotator\.css/));
+  const url = /** @type {HTMLLinkElement|undefined} */ (document.querySelector(
+    'link[rel="stylesheet"][href*="/build/styles/annotator.css"]'
```
robertknight, line 6: This URL is currently relative to the root of client's npm package (see https://unpkg.com/browse/hypothesis@1.605.0/). Currently that will work with all of the ways that we package the client (development server, https://cdn.hypothes.is, browser extension) but we may want to change that in future.

src/annotator/util/shadow-root.js
```diff
@@ -18,7 +22,7 @@ function loadStyles(shadowRoot) {
   // get a usable adder, albeit one that uses browser default styles for the
   // toolbar.
   const styleEl = document.createElement('style');
-  styleEl.textContent = adderStyles.map(url => `@import "${url}";`).join('\n');
+  styleEl.textContent = `@import "${url}";`;
```
robertknight, line 25: In both the old and new code we were not escaping the URL in any way. That is currently safe because `HTMLLinkElement.href` returns an escaped URL where any quotes are URL-escaped as `%22`.

robertknight (approved):
LGTM, with one trivial correction.

src/annotator/util/shadow-root.js
esanzgar: 👍🏼

PR #2785, opened 2020-12-03
Fix clipboard copy on iOS

esanzgar:
On iOS the element must be an `input` (or `textarea`)  and with attribute `'contenteditable'`

Closes [#858](https://github.com/hypothesis/product-backlog/issues/858)

### DIFF
--- a/src/sidebar/util/copy-to-clipboard.js
+++ b/src/sidebar/util/copy-to-clipboard.js
@@ -10,10 +10,13 @@
  * @param {string} text
  */
 export function copyText(text) {
-  const temp = document.createElement('pre');
-  temp.className = 'copy-text';
-  temp.textContent = text;
+  const temp = document.createElement('input');
+  temp.value = text;
+  temp.setAttribute('data-testid', 'copy-text');
+  // Recipe from https://stackoverflow.com/a/34046084/14463679
+  temp.contentEditable = 'true';
   document.body.appendChild(temp);
+  temp.focus();
 
   try {
     const range = document.createRange();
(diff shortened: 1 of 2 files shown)


esanzgar:
I tried to use the Clipboard API but I encountered the following issue: `DOMException: The Clipboard API has been blocked because of a Feature Policy applied to the current document. See https://goo.gl/EuHzyv for more details.`

src/sidebar/util/copy-to-clipboard.js
```diff
@@ -10,10 +10,13 @@
  * @param {string} text
  */
 export function copyText(text) {
-  const temp = document.createElement('pre');
-  temp.className = 'copy-text';
-  temp.textContent = text;
+  const temp = document.createElement('input');
+  temp.value = text;
+  temp.setAttribute('data-test-id', 'copy-text');
```
LMS007, line 15: minor suggestion

```suggestion
  temp.setAttribute('data-copy-id', 'copy-text');  // used as a selector for testing
```

LMS007, line 15:
```suggestion
  temp.setAttribute('data-test-id', 'copy-text');  // used as a selector for testing
```

LMS007 (approved):
Looks good.

I can't actually repo this bug on my iPad (v13.7). I suppose its different os that iOS but I'm not really sure. In any case, I can confirm this fix does not change the fact that is still works  on my ipod and android devices :)

src/sidebar/util/copy-to-clipboard.js
esanzgar: @robertknight [suggested to us `data-testid`](https://github.com/hypothesis/lms/pull/2146#discussion_r509171217) at it seems it is the standard attribute for testing purposes.

PR #2797, opened 2020-12-07
Fix potential issue when converting `TextRange` to a `Range`

robertknight:
This PR fixes an edge case when converting a `TextRange` (a pair of (element, text offset) points) to a DOM `Range` using `TextRange.toRange` [1].

If the `TextRange`'s `start` or `end` points referred to an element containing no text then `TextRange.toRange()` would always throw an error. This is because there was no text node in the `start`/`end` element to resolve to. When the start or end element has no text AND the offset is zero however, we can do better.

Suppose we have the following DOM structure:

```html
<div>
  <b></b>foo<u></u>
</div>
```

If a `TextRange` is somehow created where the start point is `{ element: <b>, offset: 0 }` and the end point is `{ element: <u>, offset: 0 }` then when calling `toRange` we want it to return a range which uses the start and end points of the `"foo"` text node as its boundary points, even though these points are outside the start/end elements.

This PR implements this by adding a `dir` option to `TextPosition.resolve()` to control how this edge case is handled and then making use of that in `TextRange.toRange()`.

Similar issues came up in the past with the `*Range` classes inherited from AnnotatorJS (`NormalizedRange` etc.) and the correct behavior in these scenarios was never nailed down properly.

[1] This is done when drawing highlights or creating selectors from the current selection in PDFs, and will also become relevant if we change `RangeSelector` anchoring to use `TextRange` in future.

### DIFF
--- a/src/annotator/anchoring/test/text-range-test.js
+++ b/src/annotator/anchoring/test/text-range-test.js
@@ -1,4 +1,9 @@
-import { TextPosition, TextRange } from '../text-range';
+import {
+  TextPosition,
+  TextRange,
+  RESOLVE_FORWARDS,
+  RESOLVE_BACKWARDS,
+} from '../text-range';
 
 import { assertNodesEqual } from '../../../test-util/compare-dom';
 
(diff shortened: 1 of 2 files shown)


src/annotator/anchoring/test/text-range-test.js
```diff
@@ -273,7 +328,7 @@ describe('annotator/anchoring/text-range', () => {
         assert.equal(range.toString(), 'foobar');
       });
 
-      it('throws if start point cannot be resolved', () => {
+      it('throws if start point in same element as end point cannot be resolved', () => {
```
robertknight, line 331: There is no change to the test here. I just called out an aspect of the test setup which is important because `TextRange.toRange` has a specific code path for it. Same for the test below.

src/annotator/anchoring/text-range.js
```diff
@@ -128,15 +131,43 @@ export class TextPosition {
...
+   * determines what happens.
+   *
+   * Offsets at the boundary between two nodes are resolved to the start of the
+   * node that begins at the boundary.
    *
+   * @param {Object} [options]
+   *   @param {RESOLVE_FORWARDS|RESOLVE_BACKWARDS} [options.dir] -
+   *     Specifies in which direction to search for the nearest text node if
+   *     `this.offset` is `0` and `this.element` has no text. If not specified
+   *     an error is thrown.
```
robertknight, line 145: I don't love the `dir` option name but the meaning is tricky to capture in an obvious name.

lyzadanger (approved):
This looks good. Nothing sticking out at all...

PR #2803, opened 2020-12-09
Avoid traversing object tree in `immutable` if already frozen

robertknight:
This makes store updates faster in debug builds when a lot of data is
loaded into the client. In this case most sub-trees in the store are
already frozen after an update (since they are objects from the previous
state) and only the new parts need to be frozen.

As a reminder, the `immutable` utility does nothing in production builds.

### DIFF
--- a/src/sidebar/util/immutable.js
+++ b/src/sidebar/util/immutable.js
@@ -7,6 +7,10 @@
  * in enumerable fields.
  */
 function deepFreeze(object) {
+  if (Object.isFrozen(object)) {
+    return object;
+  }
+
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
Cool. I do wonder if it's time to audit our use of `immutable` (like, at all: What benefits is it currently providing?).

robertknight:
> What benefits is it currently providing?

It is mainly used to freeze the root state object in the store to guard against accidental mutation, which will cause a variety of problems such as components not re-rendering when they should.

PR #2804, opened 2020-12-09
Optimize `sortKeys` selector

robertknight:
This was recomputing frequently even though the result only changes when
the selected tab changes.


lyzadanger (approved):
LGTM. I appreciate you breaking this out of the large `useProxyStore` changeset.

PR #2805, opened 2020-12-09
Replace remaining uses of `store.getState()` in UI components

robertknight:
Replace remaining uses of `store.getState()` in UI components with
selectors. This will avoid unnecessary re-rendering after we change
components to use the new `useStoreProxy` hook for reading from the
store. It also simplifies the tests.


lyzadanger (approved):
👍 Does this mean that we have selectors defined for everything (that components need) at this point?

robertknight:
> Does this mean that we have selectors defined for everything (that components need) at this point?

Yes, we're done now. Perhaps we could somehow "ban" the direct use of `getState()` in components in future.

PR #2806, opened 2020-12-09
Remove unnecessary `requestAnimationFrame` callbacks in `Guest`

robertknight:
Profiling anchoring of large numbers of annotations in Chrome showed
that there was significant overhead [1] for scheduling and
executing `requestAnimationFrame` callbacks. In the case of `Guest.anchor`, the
callback often did nothing (it called `removeHighlights` with an array that is usually empty).

There isn't a need for this any more AFAICS so just invoke the logic
synchronously, which is much cheaper.

[1] On the branch where I tested this originally, which also had the changes in https://github.com/hypothesis/client/pull/2779, it was about ~10% of the total time to anchor annotations.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -510,7 +510,7 @@ export default class Guest extends Delegator {
     }
 
     // Remove all the highlights that have no corresponding target anymore.
-    requestAnimationFrame(() => removeHighlights(deadHighlights));
+    removeHighlights(deadHighlights);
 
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
I agree that the rAF stuff is likely superfluous. See also https://github.com/hypothesis/client/pull/2567/commits/3f58ef1501e2f13ac78c6fd042fcecb6de8aa539

PR #2808, opened 2020-12-09
Replace `useStore` hook with `useStoreProxy` in UI components

robertknight:
This PR switches our UI components to use the new `useStoreProxy` hook introduced in https://github.com/hypothesis/client/pull/2800 to read data from and update the store. See that PR the motivation for this change.

The code changes in this PR are pretty mechanical. In UI components they look like:

```js
// Old
const someData = useStore(store => store.getSomeData()); // Selector
const changeSomething = useStore(store => store.changeSomething); // Action

<button onClick={changeSomething}>...</button>
...


// New
const someData = store.someData(); // Selector

<button onClick={() => store.changeSomething()}>...</button>
....
```

In tests the only code that really need to change was the mocks for the hook:

```js
// Old
$imports.$mock({
  '../store/use-store': callback => callback(fakeStore),
});

// New
$imports.$mock({
  '../store/use-store': { useStoreProxy: () => fakeStore },
});
```

----

One of the tradeoffs of the new API compared to the previous API that can be seen in this PR is that state usage tracking is less fine-grained in some cases. For example, this previous code would only cause a re-render if the logged-in user ID changed:

```js
const userid = useStore(store => store.profile().userid)
```

This replacement code however will cause a re-render if any part of the profile changes:

```js
const userid = store.profile().userid
```

From what I've seen by adding debugging logging this doesn't cause too much unnecessary re-rendering in our current application and so I think this is a good trade-off for the better API. We could add an escape hatch in future to combine the benefits of the old/new API. For example, something like:

```js
// Only re-renders if `userid` property of profile changes
const userid = store.watch(() => store.profile().userid); 
```

What this means right away is that it is good practice to use fine-grained selectors where available. For example, `store.getAnnotationById(id)` rather than `store.allAnnotations().find(ann => ann.id === id)`. This practice also has the advantage that it will tend to make it easier to refactor the internal state of a store module (for example, changing the annotations list to a map keyed by ID).

### DIFF
--- a/src/sidebar/components/test/sidebar-content-error-test.js
+++ b/src/sidebar/components/test/sidebar-content-error-test.js
@@ -26,7 +26,7 @@ describe('SidebarContentError', () => {
       isLoggedIn: sinon.stub().returns(true),
     };
     $imports.$mock({
-      '../store/use-store': callback => callback(fakeStore),
+      '../store/use-store': { useStoreProxy: () => fakeStore },
     });
     $imports.$mock(mockImportedComponents());
   });
@@ -52,7 +52,9 @@ describe('SidebarContentError', () => {
 
     const clearButton = findButtonByText(wrapper, 'Show all annotations');
     assert.isTrue(clearButton.exists());
-    assert.equal(clearButton.props().onClick, fakeStore.clearSelection);
+
+    clearButton.props().onClick();
+    assert.called(fakeStore.clearSelection);
   });
 
   context('unavailable annotation, logged out', () => {
--- a/src/sidebar/components/annotation-action-bar.js
+++ b/src/sidebar/components/annotation-action-bar.js
@@ -2,7 +2,7 @@ import { createElement } from 'preact';
 import propTypes from 'prop-types';
 
 import uiConstants from '../ui-constants';
-import useStore from '../store/use-store';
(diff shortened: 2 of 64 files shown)


src/sidebar/components/test/sidebar-content-error-test.js
```diff
@@ -52,7 +52,9 @@ describe('SidebarContentError', () => {
 
     const clearButton = findButtonByText(wrapper, 'Show all annotations');
     assert.isTrue(clearButton.exists());
-    assert.equal(clearButton.props().onClick, fakeStore.clearSelection);
+
+    clearButton.props().onClick();
+    assert.called(fakeStore.clearSelection);
```
robertknight, line 57: The previous assert didn't work after the other changes in this PR due to the way that the proxy's `clearSelection` is a different function (a wrapper) than the store's `clearSelection`, so I had to change it to check the same thing a different way.

lyzadanger (approved):
In so much that tests pass, stuff works locally, overall script execution timelines on loading the same document (with lots of annotations, FWIW) are equivalent between `master` and this branch: looks good to me.

I'm not a computer, so I'm leaning on our tests to catch anything that my scan here didn't :)

PR #2809, opened 2020-12-10
POC: Extend side-by-side mode to support HTML documents

robertknight:
This is a proof of concept solution for https://github.com/hypothesis/product-backlog/issues/94. It extends the recently introduced "side-by-side" mode for PDFs to also apply to HTML documents.

When the sidebar is opened, the page content is shrunk to make space to see the content alongside the sidebar. The implementation attempts to to choose the amount of shrinking such that the main content of the page is fully visible and allowed as much horizontal space as possible. This means allowing peripheral content such as ads on the side of the page to remain covered by the sidebar. Additionally, the client adjusts the scroll position of the page when opening and closing the sidebar to try and keep the same part of the content visible, despite the layout changes to the page.

Similar to the PDF side-by-side implementation, the page is not shrunk if the amount of horizontal room that would be left for the page's content is below a threshold. This means that on a mobile phone or small tablet for example, the client will just continue to behave as it does today.

### Web page where main content extends to right edge of page

Before this PR:

HTML side-by-side inactive

After this PR:

HTML side-by-side active

### Web page where there is significant horizontal space given to unimportant content (ads) to the right of the main content

Before this PR:

HTML side-by-side inactive - page with ads

After this PR:

(Note that part of the page's content has been allowed to remain under the sidebar to allow more horizontal room for the main content)

HTML side-by-side active - page with ads

## Known limitations

The simple method used to resize pages does not work on all websites. On sites where it fails, the sidebar will continue to overlap the content as it does today. This is something that can be improved over time.

## Implementation notes

We recently improved the experience of annotating PDF documents by
showing the sidebar alongside the content when there is enough
horizontal space available. This commit extends that approach to web pages,
with modifications to allow for differences between web pages and PDFs:

 - Web pages do not have a known layout and response to being resized
   that we can rely on

 - Web pages will often have peripheral content on the side which we
   should not try to squeeze into the available space when the sidebar
   is opened. Instead we should continue to allow the sidebar to cover
   it up to make more room for the main content

This is implemented by first shrinking the `<body>` by the full width of
the sidebar when it opens and then increasing the `<body>`'s width if
there is space between the "main content" of the page and the edge of
the sidebar. Whatever is in this space (eg. related stories, ads) will
then be covered up by the sidebar.

Determining what consitutes the "main content" of the page
is done by heuristics in `guessMainContentArea`. These heuristics assume
that the main content can be inferred by considering the longest
paragraphs of text on the page and looking for common vertical margins
amongst them.

Resizing the document will generally cause the content to shift. To try
and ensure the same content remains visible before and after opening the
sidebar, a `preserveScrollPosition` helper attempts to determine how
much the content shifted after the relayout and adjust the scroll
position to keep, near-enough, the same content visible. To do this, it
chooses an element near the top of the screen as an anchor and checks
how much it shifts following the resize. It then adjusts the document's
scroll position to compensate.

### DIFF
--- a/src/annotator/pdf-sidebar.js
+++ b/src/annotator/pdf-sidebar.js
@@ -22,34 +22,31 @@ export default class PdfSidebar extends Sidebar {
   constructor(element, config) {
     super(element, { ...defaultConfig, ...config });
 
-    this._lastSidebarLayoutState = {
-      expanded: false,
-      width: 0,
-      height: 0,
-    };
-
     this.window = /** @type {HypothesisWindow} */ (window);
     this.pdfViewer = this.window.PDFViewerApplication?.pdfViewer;
     this.pdfContainer = this.window.PDFViewerApplication?.appConfig?.appContainer;
+  }
 
-    // Is the PDF currently displayed side-by-side with the sidebar?
-    this.sideBySideActive = false;
-
-    this.subscribe('sidebarLayoutChanged', state => this.fitSideBySide(state));
-    this.window.addEventListener('resize', () => this.fitSideBySide());
+  /**
+   * Return the minimum width that the PDF viewer requires to be usable.
+   * This determines whether opening the sidebar activates side-by-side mode
+   */
+  minSideBySideWidth() {
+    return MIN_PDF_WIDTH;
   }
 
   /**
    * Set the PDF.js container element to the designated `width` and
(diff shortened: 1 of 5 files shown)


robertknight:
This PR is just a proof of concept and not intended to be merged.

PR #2810, opened 2020-12-10
Use `useStoreProxy` in `useRootThread` hook

robertknight:
Change `useRootThread` hook to use the new API for reading data from the
store.


lyzadanger (approved):
👍

PR #2812, opened 2020-12-10
Remove old `useStore` hook

robertknight:
All consumers now use the replacement `useStoreProxy` hook.


lyzadanger (approved):
Follow-up question: Is the module name `use-store` still appropriate?

robertknight:
> Is the module name use-store still appropriate?

Hmm - Maybe we could name it `hooks.js` instead?

PR #2819, opened 2020-12-14
Infer store module types automatically

robertknight:
Remove the need to define the type of store created by each module
manually by adding a `StoreFromModule` helper in `create-store.js` which
can infer the store type from a store module configuration. Using this
the type of a store composed from several modules can then be created
with:

```js
import fooModule from './modules/foo';
import barModule from './modules/bar';

/** @typedef {StoreFromModule<fooModule> & StoreFromModule<barModule>} AppStore */

const store = /** @type {AppStore} */ (createStore([fooModule, barModule]))

store.getBarById('...'); // Call selector defined in `barModule`
store.addFoo(...); // Dispatch action defined in `fooModule`
store.noSuchMethod(...); // Error - no such method in `fooModule` or `barModule`.
```

Ideally `createStore` would just infer the type based upon its
arguments. I haven't worked out how to do that yet. Nevertheless, this
still removes the need for a lot of manually defined types.

To ensure more useful error messages from TS if a store module's configuration
has the wrong shape, a `storeModule` helper has been added. This wraps
the configuration for each module to check its shape before the
individual modules are combined into one type for the store. This helper could
also perform runtime validation in future.

 - Add `StoreFromModule` type in `create-store.js` and several helpers
   to support it
 - Modify each store module to wrap the export in `storeModule` and
   remove any manually defined store types

### DIFF
--- a/src/sidebar/store/create-store.js
+++ b/src/sidebar/store/create-store.js
@@ -8,6 +8,85 @@ import immutable from '../util/immutable';
 import { createReducer, bindSelectors } from './util';
 
 /**
+ * Helper that strips the first argument from a function type.
+ *
+ * @template F
+ * @typedef {F extends (x: any, ...args: infer P) => infer R ? (...args: P) => R : never} OmitFirstArg
+ */
+
+/**
+ * Helper that converts an object of selector functions, which take a `state`
+ * parameter plus zero or more arguments, into selector methods, with no `state` parameter.
+ *
+ * @template T
+ * @typedef {{ [K in keyof T]: OmitFirstArg<T[K]> }} SelectorMethods
+ */
+
+/**
+ * Map of action name to reducer function.
+ *
+ * @template State
+ * @typedef {{ [action: string]: (s: State, action: any) => Partial<State> }} Reducers
+ */
+
+/**
+ * Configuration for a store module.
+ *
+ * @template State
+ * @template {object} Actions
+ * @template {object} Selectors
+ * @template {object} RootSelectors
+ * @typedef Module
+ * @prop {(...args: any[]) => State} init -
+ *   Function that returns the initial state for the module
+ * @prop {string} namespace -
+ *   The key under which this module's state will live in the store's root state
+ * @prop {Reducers<State>} update -
+ *   Map of action types to "reducer" functions that process an action and return
+ *   the changes to the state
+ * @prop {Actions} actions
+ *   Object containing action creator functions
+ * @prop {Selectors} selectors
+ *   Object containing selector functions
+ * @prop {RootSelectors} [rootSelectors]
+ */
+
+/**
+ * Replace a type `T` with `Fallback` if `T` is `any`.
+ *
+ * Based on https://stackoverflow.com/a/61626123/434243.
+ *
+ * @template T
+ * @template Fallback
+ * @typedef {0 extends (1 & T) ? Fallback : T} DefaultIfAny
+ */
+
+/**
+ * Helper for getting the type of store produced by `createStore` when
+ * passed a given module.
+ *
+ * To get the type for a store created from several modules, use `&`:
+ *
+ * `StoreFromModule<firstModule> & StoreFromModule<secondModule>`
+ *
+ * @template T
+ * @typedef {T extends Module<any, infer Actions, infer Selectors, infer RootSelectors> ?
+ *   Store<Actions,Selectors,DefaultIfAny<RootSelectors,{}>> : never} StoreFromModule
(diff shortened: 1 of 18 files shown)


src/sidebar/store/create-store.js
```diff
@@ -7,6 +7,85 @@ import immutable from '../util/immutable';
...
+ */
+
+/**
+ * Replace a type `T` with `Fallback` if `T` is `any`.
+ *
+ * Based on https://stackoverflow.com/a/61626123/434243.
+ *
+ * @template T
+ * @template Fallback
+ * @typedef {0 extends (1 & T) ? Fallback : T} DefaultIfAny
```
robertknight, line 61: For example:

`DefaultIfAny<string, number>` is `string`
`DefaultIfAny<any, number>` is `number`

This is used below to ensure correct handling of modules which don't define a `rootSelector` key.

src/sidebar/store/create-store.js
```diff
@@ -7,6 +7,85 @@ import immutable from '../util/immutable';
...
+ * passed a given module.
+ *
+ * To get the type for a store created from several modules, use `&`:
+ *
+ * `StoreFromModule<firstModule> & StoreFromModule<secondModule>`
+ *
+ * @template T
+ * @typedef {T extends Module<any, infer Actions, infer Selectors, infer RootSelectors> ?
+ *   Store<Actions,Selectors,DefaultIfAny<RootSelectors,{}>> : never} StoreFromModule
+ */
```
robertknight, line 75: Most modules do not have a `rootSelectors` key. In that case the `RootSelectors` type cannot be inferred and defaults to `any`. The `DefaultIfAny<RootSelectors, {}>` part replaces that with `{}`. If we didn't do this, the `any` would "pollute" the result.

src/sidebar/store/create-store.js
```diff
@@ -7,6 +7,85 @@ import immutable from '../util/immutable';
 
 import { createReducer, bindSelectors } from './util';
 
+/**
+ * Helper that strips the first argument from a function type.
+ *
+ * @template F
+ * @typedef {F extends (x: any, ...args: infer P) => infer R ? (...args: P) => R : never} OmitFirstArg
+ */
```
robertknight, line 15: Credit: https://stackoverflow.com/a/58765199/434243.

This relies on a feature known as [conditional types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-inference-in-conditional-types). To clarify what is going on here, it can be described in words as:

"If the type F is compatible with `(x: any, ...args: P) => R`, where `P` and `R` are to be deduced automatically, then return the type `(...args: P) => R`. Otherwise trigger an error."

src/sidebar/store/create-store.js
```diff
@@ -22,9 +101,10 @@ import { createReducer, bindSelectors } from './util';
  * each action and selector from the input modules as a method which operates on
  * the store.
  *
- * @param {Object[]} modules
+ * @param {Module<any,any,any,any>[]} modules
  * @param {any[]} [initArgs] - Arguments to pass to each state module's `init` function
  * @param {any[]} [middleware] - List of additional Redux middlewares to use
+ * @return Store<any,any,any>
  */
```
robertknight, line 108: Ideally the returned `Store` would have a type that depends upon `modules`. I haven't worked out how to do that yet. The tricky part is that the number of modules is variable and the type of each module is different.

LMS007 (approved):
Okay, I think this is mostly okay to merge. I can see the advantage of not having to redefine unique store types each time, that is a huge plus. I wish TS had a lot less noise in the error messages. At first glance, the error is impossible to understand...

Screen Shot 2020-12-15 at 9 26 30 AM

only until you get to the last line do you see the real error that "namespace is missing"

I guess that's more of a gripe with how the errors are reported in ts.  In light of this, I think we can also remove the code that warns the user is namespace is missing in create-store around line 138. Its not part of the diff so I don't know how to directly comment on it.

```
console.warn('Store module does not specify a namespace', module);
```

Thank Rob

robertknight:
> I can see the advantage of not having to redefine unique store types each time, that is a huge plus. I wish TS had a lot less noise in the error messages.

Yeah. This is an unfortunate case where the argument and parameter types to the function end up being lengthy object literals rather than nice short names. The message would otherwise be quite clear (`Argument of type <long string A> is not assignment to parameter of type <long string B>. Property 'namespace' is missing in type <long string A>`). I don't know of a way to get TS to generate a more readable name in this instance. I guess the best option is to report an issue upstream.

PR #2830, opened 2020-12-14
Convert `RangeAnchor` to use XPath utilities and `TextRange`

robertknight:
Convert `RangeAnchor` to use `TextRange` and the XPath <-> Node mapping
functions in xpath.js directly, rather than the `SerializedRange` and
`NormalizedRange` classes.

This change will mean that all conversion between text positions and
(text node, offset) points in the client will use the same implementation (which
we can ensure is well debugged, tested, documented, optimized etc.).
For `RangeAnchor` selectors this conversion is used for the
`startOffset` and `endOffset` fields.

The new implementation also avoids modifying the DOM, unlike the
previous implementation which would sometimes split text nodes. Avoid
DOM modifications during anchoring opens up the possibility of
adding caching to optimize extraction of text from the document and converting
between text offsets and text node positions.

There is a follow-up PR which removes the old implementation - https://github.com/hypothesis/client/pull/2831.

### DIFF
--- a/src/annotator/anchoring/test/text-range-test.js
+++ b/src/annotator/anchoring/test/text-range-test.js
@@ -204,6 +204,37 @@ describe('annotator/anchoring/text-range', () => {
       });
     });
 
+    describe('fromCharOffset', () => {
+      let el;
+      beforeEach(() => {
+        el = document.createElement('div');
+        el.append('hello', 'world');
+      });
+
+      it('returns TextPosition for offset in Text node', () => {
+        assert.deepEqual(
+          TextPosition.fromCharOffset(el.firstChild, 1),
+          TextPosition.fromPoint(el.firstChild, 1)
+        );
+      });
+
+      it('returns TextPosition for offset in Element node', () => {
+        assert.deepEqual(
+          TextPosition.fromCharOffset(el, 5),
+          new TextPosition(el, 5)
+        );
+      });
+
+      it('throws for an offset in a non-Text/Element node', () => {
+        const comment = document.createComment('This is a test');
+        el.append(comment);
+
+        assert.throws(() => {
+          TextPosition.fromCharOffset(comment, 3);
+        }, 'Node is not an element or text node');
+      });
+    });
+
     describe('fromPoint', () => {
       it('returns TextPosition for offset in Text node', () => {
         const el = document.createElement('div');
--- a/src/annotator/anchoring/test/types-test.js
+++ b/src/annotator/anchoring/test/types-test.js
@@ -28,26 +28,11 @@ describe('annotator/anchoring/types', () => {
   });
 
   describe('RangeAnchor', () => {
-    let fakeSniff;
-    let fakeSerializedRange;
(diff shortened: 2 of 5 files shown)


src/annotator/anchoring/test/types-test.js
```diff
@@ -28,26 +28,11 @@ describe('annotator/anchoring/types', () => {
   });
 
   describe('RangeAnchor', () => {
-    let fakeSniff;
-    let fakeSerializedRange;
-    let fakeNormalize;
-    let fakeSerialize;
+    let container;
```
robertknight, line 31: The updated tests for `RangeAnchor` follow the pattern for the test of the tests in this module of not mocking the `TextRange` or `TextPosition` classes.

src/annotator/anchoring/types.js
```diff
@@ -31,16 +24,16 @@ import { matchQuote } from './match-quote';
 export class RangeAnchor {
   /**
    * @param {Node} root - A root element from which to anchor.
-   * @param {AnyRangeType} range -  A range describing the anchor.
+   * @param {Range} range -  A range describing the anchor.
    */
   constructor(root, range) {
     this.root = root;
-    this.range = sniff(range).normalize(this.root);
+    this.range = range;
```
robertknight, line 31: This is the change that prevents anchoring of `RangeSelector` selectors from modifying the DOM.

src/annotator/anchoring/types.js
```diff
@@ -49,35 +42,55 @@ export class RangeAnchor {
...
   }
 
   /**
    * @return {RangeSelector}
    */
   toSelector() {
-    const range = this.range.serialize(this.root);
+    // "Shrink" the range so that it tightly wraps its text. This ensures more
+    // predictable output for a given text selection.
+    const normalizedRange = TextRange.fromRange(this.range).toRange();
```
robertknight, line 82: Given an HTML structure like:

```html
<div>Foo bar<b></b></div>
```

If a user selects the text "Foo bar" then the browser's selection (`window.getSelection().getRangeAt(0)`) could represent that in several different ways (eg. it could select the whole `<div>`, just the text "Foo bar", or the text "Foo bar" plus the empty `<b>` tag). This normalization results in a range that tightly wraps the text.

src/annotator/anchoring/test/text-range-test.js
```diff
@@ -204,6 +204,37 @@ describe('annotator/anchoring/text-range', () => {
...
 
+    describe('fromCharOffset', () => {
+      let el;
+      beforeEach(() => {
+        el = document.createElement('div');
+        el.append('hello', 'world');
+      });
+
+      it('returns TextPosition for offset in Text node', () => {
+        assert.deepEqual(
```
LMS007, line 215: I see, Node.TEXT_NODE calls

LMS007 (approved):
Good to good. I will not miss SerializedRange

robertknight:
Thanks!

PR #2831, opened 2020-12-15
Remove old range selector anchoring code

robertknight:
**Depends on https://github.com/hypothesis/client/pull/2830**

This removes the code in `src/annotator/anchoring/range.js` and `src/annotator/anchoring/xpath-util.js` which is no longer used for range selector anchoring, as well as the associated tests. It also moves the code from `xpath-util.js` which was still being used into `src/annotator/anchoring/xpath.js`, consolidating the XPath <-> Node conversion code in one module.


LMS007 (approved):
👍

LMS007 (approved):

PR #2834, opened 2020-12-16
Preload `/api/` and `/api/links` responses in boot script

robertknight:
https://github.com/hypothesis/h/pull/5862 introduced an optimization to startup of the sidebar app, which also applies to the upcoming "Notebook" app, by preloading the `/api/` and `/api/links` API responses.

Because this change was made in h it only benefitted the embedded client, but not the browser extension.

This commit moves the logic to add the `<link rel="preload">` links to the client's boot script. The upside of this is that it works with all distributions of the client and can be changed in future without needing to modify h and the browser extension etc. separately.

The downside is that because the `<link>` is no longer in the sidebar app's HTML but in the boot script the prefetch will start a little later. Since Cloudflare caches the `/api/` and `/api/links` responses, these requests are still very likely to have completed by the time the client code starts executing.

There is a counterpart change in https://github.com/hypothesis/h/pull/6420 which removes the `<link rel="preload">` elements from https://hypothes.is/app.html. They can be deployed in either order.

### DIFF
--- a/src/boot/boot.js
+++ b/src/boot/boot.js
@@ -16,6 +16,7 @@ const commonPolyfills = [
  * @prop {string} assetRoot - The root URL to which URLs in `manifest` are relative
  * @prop {Object.<string,string>} manifest -
  *   A mapping from canonical asset path to cache-busted asset path
+ * @prop {string} apiUrl
  */
 
 /**
@@ -81,6 +82,27 @@ function injectLink(doc, rel, type, url) {
   link.rel = rel;
   link.href = url;
   link.type = `application/annotator+${type}`;
+
+  tagElement(link);
+  doc.head.appendChild(link);
+}
+
+/**
+ * Preload a URL using a `<link rel="preload" as="<type>" ...>` element
+ *
+ * This can be used to preload an API request or other resource which we know
+ * that the client will load.
+ *
+ * @param {Document} doc
+ * @param {string} type - Type of resource
+ * @param {string} url
+ */
+function preloadUrl(doc, type, url) {
+  const link = doc.createElement('link');
+  link.rel = 'preload';
+  link.as = type;
+  link.crossOrigin = 'anonymous';
+  link.href = url;
 
   tagElement(link);
   doc.head.appendChild(link);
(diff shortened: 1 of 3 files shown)


src/boot/boot.js
```diff
@@ -163,6 +185,10 @@ export function bootHypothesisClient(doc, config) {
  * @param {SidebarAppConfig} config
  */
 export function bootSidebarApp(doc, config) {
+  // Preload `/api/` and `/api/links` API responses.
+  preloadUrl(doc, 'fetch', config.apiUrl);
+  preloadUrl(doc, 'fetch', config.apiUrl + 'links');
```
esanzgar, line 190:
```suggestion
  preloadUrl(doc, 'fetch', `${config.apiUrl}links`);
```

esanzgar (approved):
👍🏼 

For those, like me, that are not familiar this a description of what these two endpoint return:
- `/api`: is a description of the API, endpoints and type of request it support
- `/api/links`: list other endpoints that the client depend on but are not in the above resources.

I wonder if these two could or should be combined together, they seem very similar.

robertknight:
> /api/links: list other endpoints that the client depend on but are not in the above resources.

Yes. The difference is that everything in `/api/` is an API endpoint that you can call and get a JSON response. The links from `/api/links` are URL templates for web pages that the client may want to direct the user to. It also includes the authorization endpoints for getting an access token, which return form-encoded rather than JSON responses because they follow the OAuth 2 specs.

> I wonder if these two could or should be combined together, they seem very similar.

Maybe. I think the different usage and response format of the endpoints was the main reason for the separation originally. It could things a bit simpler in practical terms for clients if they could get both sets of data with one request though.

robertknight:
Thanks for the review. I'll leave deploying this until I'm back in the new year.

esanzgar:
Just a comment:

> Since Cloudflare caches the /api/ and /api/links responses, these requests are still very likely to have completed by the time the client code starts executing.

Cloudflare expires the cache every 5 min (300s) according to the cache-control header.

`Cache-Control: max-age=300, public`

src/boot/boot.js
robertknight: We don't have a firm convention about when to use string interpolation vs. concatenation in our code. I think the informal approach up to this point has been "use interpolation when it makes obvious sense, concatenation is fine for simple cases (like adding a prefix or suffix)". Happy to discuss more in Slack.

robertknight:
> Cloudflare expires the cache every 5 min (300s) according to the cache-control header.

This is true, although the Cloudflare cache should be shared between all users (or at least those in a particular location - I'm not sure how sharing works between different Cloudflare regions). So as long as at least one user (perhaps "at least one user from the same region") has used the client in the last 5 minutes, Cloudflare should have a copy of these resources in its cache.

PR #2836, opened 2020-12-17
Move scroll container measuring into a layout effect

robertknight:
`ThreadList` was measuring the height and scroll offset of the scroll
container inside the render function instead of a layout effect. As a result
it could fail if the scroll container is rendered by a parent component
and its DOM has not yet been created.

This issue was uncovered when trying to change the app to call
`router.sync()` before waiting for `groups.load()` to complete.

A test expectation had to be changed becaused it turned out that the
scroll container had the wrong height when it was measured in the
previous way (it was measured as 0px instead of the expected 350px).

### DIFF
--- a/src/sidebar/components/thread-list.js
+++ b/src/sidebar/components/thread-list.js
@@ -1,5 +1,5 @@
 import { createElement } from 'preact';
-import { useEffect, useMemo, useState } from 'preact/hooks';
+import { useEffect, useLayoutEffect, useMemo, useState } from 'preact/hooks';
 import propTypes from 'prop-types';
 import debounce from 'lodash.debounce';
 
@@ -27,6 +27,11 @@ function getScrollContainer() {
   return container;
 }
 
+/** @param {number} pos */
+function roundScrollPosition(pos) {
+  return Math.max(pos - (pos % SCROLL_PRECISION), 0);
+}
+
 /**
  * @typedef ThreadListProps
  * @prop {Thread} thread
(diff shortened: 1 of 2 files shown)


src/sidebar/components/thread-list.js
```diff
@@ -46,16 +51,20 @@ function getScrollContainer() {
 function ThreadList({ thread }) {
   // Height of the visible area of the scroll container.
   const [scrollContainerHeight, setScrollContainerHeight] = useState(
-    getScrollContainer().clientHeight
+    window.innerHeight
```
robertknight, line 54: `window.innerHeight` is just used as an initial default here which doesn't depend on anything rendered by a parent component.

src/sidebar/components/thread-list.js
```diff
@@ -46,16 +51,20 @@ function getScrollContainer() {
...
-  const [scrollPosition, setScrollPosition] = useState(
-    getScrollContainer().scrollTop
-  );
+  const [scrollPosition, setScrollPosition] = useState(0);
+
+  useLayoutEffect(() => {
+    const scrollContainer = getScrollContainer();
+    setScrollContainerHeight(scrollContainer.clientHeight);
+    setScrollPosition(roundScrollPosition(scrollContainer.scrollTop));
+  }, []);
```
robertknight, line 67: This layout effect sets the initial values of `scrollContainerHeight` and `scrollPosition` after the first render.

lyzadanger (approved):
Cool.

PR #2838, opened 2020-12-17
Defer initial WebSocket connection until profile is fetched

robertknight:
Following https://github.com/hypothesis/client/pull/2837 the
`SidebarView` component may now be rendered before the profile has been
fetched. This component contains an effect which triggers the initial
WebSocket connection. Since the WebSocket reconnects after the profile userid
changes, this was causing the initial connection to almost immediately
be disconnected if the user was logged in. Fix this by deferring the initial
connection until after the profile has been fetched.


lyzadanger (approved):
👍 Sorry I didn't notice anything in the console in my review of the triggering changes!

PR #2841, opened 2020-12-18
Do not clear filters after initial groups fetch completes

robertknight:
https://github.com/hypothesis/client/pull/2837 caused a regression where
the initial annotation filters would be cleared after groups where
fetched for the first time. This is because `SidebarView` had logic to
clear the selection/filters when switching groups. This used to work
because `SidebarView` was only rendered after groups were initially
rendered. This is no longer the case after #2837.

Slack thread: https://hypothes-is.slack.com/archives/C1M8NH76X/p1608299842322000


lyzadanger (approved):
Thanks for the speedy fix.

robertknight:
This affected all client features which caused a selection/filter to apply on startup. This includes:

- Focusing on a specific user in the grading view in the LMS app
- Direct links to specific annotations
- Direct links that pre-fill the filter input

PR #2849, opened 2020-12-21
Improve Adder location for touch screen devices

LMS007:
- Ensure that the adder location is always below the highlighted selection on a touch devices so that it does not compete with the copy/paste bar.
- Add a little extra margin (10px) to the adder on touch devices to give room for the selection handles
- Add helper method `isTouchDevice` for touch devices
- Add comment in isSelectionBackwards to explain that iOS does not have the same selection object as non-iOS devices
- Add 1 missing test coverage case in adder.js

------

fixes https://github.com/hypothesis/client/issues/2626

This ticket was initially created to address issues with the events dealing with drag and drop on a touch device.  Ideally, it would be nice if there were events that gave a "end" to a "drag" event, but no such event exists. I presume that is because you can re-adjust the selection handles after the fact. For all intents and purposes we're no longer going to try and hide the adder when a selection is being made.

That leads the second set of problems outlined in the ticket: The adder on touch devices tends to get in the way of the copy & paste bar when you make a selection from bottom to top (backwards) as you can see here.

(master branch)
102418112-f8270480-3fb1-11eb-9ba6-992729f9e60b

More so, I discovered that android and iOS are not equal in this behavoir. For reasons I don't yet understand, window.getSelection() does not return the same object on iOS that is returned on safari or chrome on mac desktop or android. In other words, it appears iOS is the outlier. For that reason, the condition in the last line of `isSelectionBackwards` always returns false in iOS because those two nodes are never the same. 

For that reason, iOS will never render the add near the start of a range selection and so the adder is almost always below the highlighted text. The only times it flips is when it's close to the bottom of the screen. 

With android, its different and will in fact render above the highlight selection if you create it backwards. This leads to problems because it cover the  build-in copy bar. This fix presented here is to force the adder to only render below a selection on all touch devices (except when off the bottom of screen). This is already the case on iOS (But I suspect it was inadvertent). This PR further forces both iOS and android (or rather, all touch devices) to render it below the highlighted text.

Additionally, I added a 10px margin to create separation between the adder the the highlighted text to give more room for the grab handles. These blue handles can impede the adder esp. on android where they are large.

Per best practice, I avoid using being specific with the user-agent. It's my hope we don't get into specific conditions based on iOS vs android, but I think it's very reasonable to have a condition for touch devices in general. 

In this next gif  you can see the adder never blocks the copy & paste bar.

(this branch)
android-adder

The adder can still render above a selection if it falls off the bottom of the screen! In this case, it will render under the copy and paste bar, but.

1. We can't know for sure where the copy & past bar is. In some cases it renders below when the selection is made near the top of the screen so we're always guessing in a way.
2. It's totally reasonable to assume users know how to, and can easily scroll to center the text better in the viewport. Scrolling on mobile is just part of the nature of the device and its reasonable to assume users can make scrolling corrections in these edge cases.

There is no way to guess where the copy & paste bar shows up, but these changes here fix the bulk of the problem. The remaining edge cases only show up when near the bottom or top of the screen.

### DIFF
--- a/src/annotator/adder.js
+++ b/src/annotator/adder.js
@@ -1,6 +1,7 @@
 import { createElement, render } from 'preact';
 
 import AdderToolbar from './components/adder-toolbar';
+import { isTouchDevice } from '../shared/user-agent';
 import { createShadowRoot } from './util/shadow-root';
 
 /**
@@ -180,9 +181,12 @@ export class Adder {
     // Set the initial arrow direction based on whether the selection was made
     // forwards/upwards or downwards/backwards.
     /** @type {ArrowDirection} */ let arrowDirection;
-    if (isRTLselection) {
+    if (isRTLselection && !isTouchDevice()) {
       arrowDirection = ARROW_POINTING_DOWN;
     } else {
+      // Render the adder below the selection for touch devices due to competing
+      // space with the native copy/paste bar that typical (not always) renders above
+      // the selection.
       arrowDirection = ARROW_POINTING_UP;
     }
     let top;
@@ -192,6 +196,9 @@ export class Adder {
     // and close to the end.
     const hMargin = Math.min(ARROW_H_MARGIN, selectionRect.width);
     const adderWidth = this._width();
+    // Render the adder a little lower on touch devices to provide room for the native
+    // selection handles so that the interactions with selection don't compete with the adder.
+    const touchScreenOffset = isTouchDevice() ? 10 : 0;
     const adderHeight = this._height();
     if (isRTLselection) {
       left = selectionRect.left - adderWidth / 2 + hMargin;
@@ -212,7 +219,11 @@ export class Adder {
     }
 
     if (arrowDirection === ARROW_POINTING_UP) {
-      top = selectionRect.top + selectionRect.height + ARROW_HEIGHT;
+      top =
+        selectionRect.top +
+        selectionRect.height +
+        ARROW_HEIGHT +
+        touchScreenOffset;
     } else {
       top = selectionRect.top - adderHeight - ARROW_HEIGHT;
     }
--- a/src/annotator/range-util.js
+++ b/src/annotator/range-util.js
@@ -10,6 +10,8 @@ export function isSelectionBackwards(selection) {
   }
 
   const range = selection.getRangeAt(0);
+  // Does not work correctly on iOS when selecting nodes backwards.
+  // https://bugs.webkit.org/show_bug.cgi?id=220523
   return range.startContainer === selection.focusNode;
 }
 
--- a/src/annotator/test/adder-test.js
+++ b/src/annotator/test/adder-test.js
@@ -2,7 +2,12 @@ import { act } from 'preact/test-utils';
 import { createElement } from 'preact';
 import { mount } from 'enzyme';
 
-import { Adder, ARROW_POINTING_UP, ARROW_POINTING_DOWN } from '../adder';
+import {
+  Adder,
+  ARROW_POINTING_UP,
+  ARROW_POINTING_DOWN,
+  $imports,
+} from '../adder';
 
 function rect(left, top, width, height) {
   return { left, top, width, height };
@@ -50,6 +55,7 @@ describe('Adder', () => {
   afterEach(() => {
     adderCtrl.hide();
     adderEl.remove();
+    $imports.$restore();
   });
 
   function windowSize() {
@@ -199,6 +205,12 @@ describe('Adder', () => {
       assert.equal(target.arrowDirection, ARROW_POINTING_UP);
     });
 
+    it('does not position the adder above the top of the viewport even when selection is backwards', () => {
+      const target = adderCtrl._calculateTarget(rect(100, -100, 100, 20), true);
+      assert.isAtLeast(target.top, 0);
+      assert.equal(target.arrowDirection, ARROW_POINTING_UP);
+    });
+
     it('does not position the adder below the bottom of the viewport', () => {
       const viewSize = windowSize();
       const target = adderCtrl._calculateTarget(
@@ -220,6 +232,22 @@ describe('Adder', () => {
     it('does not positon the adder beyond the left edge of the viewport', () => {
       const target = adderCtrl._calculateTarget(rect(-100, 100, 10, 10), false);
       assert.isAtLeast(target.left, 0);
+    });
+
+    context('touch device', () => {
+      it('positions the adder below the selection even if the selection is backwards', () => {
+        $imports.$mock({
+          '../shared/user-agent': {
+            isTouchDevice: sinon.stub().returns(true),
+          },
+        });
+        const target = adderCtrl._calculateTarget(
+          rect(100, 200, 100, 20),
+          true
+        );
+        assert.isAbove(target.top, 220);
+        assert.equal(target.arrowDirection, ARROW_POINTING_UP);
+      });
     });
   });
 
--- a/src/shared/test/user-agent-test.js
+++ b/src/shared/test/user-agent-test.js
@@ -1,4 +1,4 @@
-import { isMacOS, isIOS } from '../user-agent';
+import { isMacOS, isIOS, isTouchDevice } from '../user-agent';
 
 describe('shared/user-agent', () => {
   describe('isMacOS', () => {
@@ -16,6 +16,23 @@ describe('shared/user-agent', () => {
           'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.92 Safari/537.36 Edg/80.0.361.109'
         )
       );
+    });
+  });
+
+  describe('isTouchDevice', () => {
+    let matchMedia;
+
+    beforeEach(() => {
+      matchMedia = sinon.spy(window, 'matchMedia');
+    });
+
+    afterEach(() => {
+      window.matchMedia.restore();
+    });
+
+    it('calls `window.matchMedia` with the query string "(pointer: coarse)"', () => {
+      isTouchDevice(window);
+      assert.calledWith(matchMedia, '(pointer: coarse)');
     });
   });
 
--- a/src/shared/user-agent.js
+++ b/src/shared/user-agent.js
@@ -35,3 +35,14 @@ export const isIOS = (
     (_navigator.userAgent.includes('Mac') && _ontouchend)
   );
 };
+
+/**
+ * Returns true when the device is a touch device such
+ * as android or iOS.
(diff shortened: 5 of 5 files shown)


src/annotator/test/adder-test.js
```diff
@@ -199,6 +204,12 @@ describe('Adder', () => {
       assert.equal(target.arrowDirection, ARROW_POINTING_UP);
     });
 
+    it('does not position the adder above the top of the viewport even when selection is backwards', () => {
```
LMS007, line 207: just adding a test for prexisting missing codecov

esanzgar:
@LMS007 Do you know if Android's built-in copy bar always appears on the top of the selected text? For all languages? In all situations?

robertknight:
I've just read through the PR description. The proposed approach generally sounds good to me. I was confused by this sentence:

> For reasons I don't yet understand, document.getSelection() does not return the same object on iOS that is returned on safari or chrome on mac desktop or android.

There is no `document.getSelection()` API. I guess you mean [window.getSelection()](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection)? Can you clarify what you mean by "does not return the same object"? The API directly returns a `Selection` instance.

src/annotator/range-util.js
```diff
@@ -10,6 +10,7 @@ export function isSelectionBackwards(selection) {
   }
 
   const range = selection.getRangeAt(0);
+  // Does not work in iOS because these two nodes are never equal.
   return range.startContainer === selection.focusNode;
```
robertknight, line 14: What was the relation between `range.startContainer` and `selection.focusNode` in your iOS testing? According to [the spec](https://w3c.github.io/selection-api/#dfn-focus):

> Each selections also have an anchor and a focus. If the selection's range is null, its anchor and focus are both null. If the selection's range is not null and its direction is forwards, its anchor is the range's start, and its focus is the end. Otherwise, its focus is the start and its anchor is the end.

So it seems from a reading of the spec that this check should be valid and if not, might indicate a bug that we should report.

src/annotator/adder.js
```diff
@@ -180,9 +181,11 @@ export class Adder {
     // Set the initial arrow direction based on whether the selection was made
     // forwards/upwards or downwards/backwards.
     /** @type {ArrowDirection} */ let arrowDirection;
-    if (isRTLselection) {
+    if (isRTLselection && !isTouchDevice()) {
       arrowDirection = ARROW_POINTING_DOWN;
     } else {
+      // Render the adder below the selection for touch devices due to competing
+      // space with the copy/paste bar.
```
robertknight, line 188: I think this sentence could be clarified to explain that "the copy/paste bar" is a native UI which typically appears above the selection on touch devices.

src/annotator/adder.js
```diff
@@ -192,6 +195,9 @@ export class Adder {
     // and close to the end.
     const hMargin = Math.min(ARROW_H_MARGIN, selectionRect.width);
     const adderWidth = this._width();
+    // Render the adder a little lower on a touch device to provide room for the
+    // selection grab handles.
```
robertknight, line 199: By this comment you mean that we're moving the adder lower so that interactions with it are not confused with interactions with the selection grab handles?

src/annotator/adder.js
```diff
@@ -192,6 +195,9 @@ export class Adder {
...
+    const touchScreenMargin = isTouchDevice() ? 10 : 0;
```
robertknight, line 200:
```suggestion
    const touchScreenOffset = isTouchDevice() ? 10 : 0;
```

"Offset" seems to me like a more appropriate term here than margin since we're talking about a position (the Y coordinate of the top of the adder) rather than a box (which has a margin around it).

robertknight (changes requested):
The proposed UI changes sound good to me - thank-you for the detailed PR description. I had a query regarding the issue with `isSelectionBackwards` on iOS and also had a suggestion regarding how we test whether the device is a touch device or not, to make it consistent with how we detect in CSS.

LMS007:
> @LMS007 Do you know if Android's built-in copy bar always appears on the top of the selected text? For all languages? In all situations?

I did not test other langs, thats a good question, but I do know it will render below if its near the top edge of the page, but how it calculates that is somewhat of a blackbox. I wouldn't want to guess and be wrong.

esanzgar:
> I do know it will render below if its near the top edge of the page

In these situations where the built-in copy bar appears below the selection, will the built-in copy bar still obstruct the view of the adder?

src/annotator/test/adder-test.js
```diff
@@ -221,6 +232,23 @@ describe('Adder', () => {
...
+            isTouchDevice: sinon.stub().returns(true),
+          },
+        });
+        const target = adderCtrl._calculateTarget(
+          rect(100, 200, 100, 20),
+          true
+        );
+        assert.isAbove(target.top, 220);
+        assert.equal(target.arrowDirection, ARROW_POINTING_UP);
+        $imports.$restore();
```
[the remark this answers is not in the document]
LMS007: A quick search in the code, I don't see any usage of params in $restore() so I'll move it to the top level

src/shared/user-agent.js
```diff
@@ -35,3 +35,13 @@ export const isIOS = (
...
 };
+
+/**
+ * Returns true when the device is a touch device such
+ * as android or iOS.
+ *
+ * @param _document {Document} - Test seam
+ */
+export const isTouchDevice = (_document = document) => {
+  return 'ontouchend' in _document;
```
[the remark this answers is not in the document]
LMS007: Sure, also, I didnt think the file name made sense anymore as this is now a device type detection, but I didn't have a better idea.

src/annotator/range-util.js
LMS007: I need to get the exact objects so I can be more specific, its a bit hard to debug on my ipad, I need to find a better way.

src/annotator/adder.js
LMS007: correct. I will re-word that a bit

src/annotator/adder.js
LMS007: agreed

LMS007:
> > I do know it will render below if its near the top edge of the page
> 
> In these situations where the built-in copy bar appears below the selection, will the built-in copy bar still obstruct the view of the adder?

Yes it will and with that, this solution is mostly fixing ~80% of the problem. My justification for ignoring these edge cases are simply:

- its somewhat more rare.
- I would expect users to generally be able to scroll on a touch device and center the selecting text they are trying to annotate. 

Having said that, there are 2 edge cases and both require us to guess when the copy and paste bar renders in a different place.

1. At the bottom when we render the adder above the selection -- on touch devices we could perhaps add a  lot of offset and to force the adder to be above the presumed copy/paste bar. Something like ~80px maybe required. 
2.  At the top, we do the same thing, but push it lower by that amount. 

We also (very likely) have to find a best fix model for both ios and android. I suggest we consider doing this if we still continue to see trouble on touch devices. We could land a follow up PR after this one to flush out these edge cases as best as we can.

LMS007:
> I've just read through the PR description. The proposed approach generally sounds good to me. I was confused by this sentence:
> 
> > For reasons I don't yet understand, document.getSelection() does not return the same object on iOS that is returned on safari or chrome on mac desktop or android.
> 
> There is no `document.getSelection()` API. I guess you mean [window.getSelection()](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection)? Can you clarify what you mean by "does not return the same object"? The API directly returns a `Selection` instance.

yes, window.getSelection, i typed it wrong. I will pull details between iOS and non-iOS and report back.

src/shared/test/user-agent-test.js
```diff
@@ -19,6 +19,14 @@ describe('shared/user-agent', () => {
     });
   });
 
+  describe('isTouchDevice', () => {
+    it('calls `window.matchMedia` with the query string "(pointer: coarse)"', () => {
+      const matchMedia = sinon.spy(window, 'matchMedia');
```
robertknight, line 24: This spy is not reset anywhere. I would suggest to stub `matchMedia` in a `beforeEach` and reset in an `afterEach` block.

src/shared/user-agent.js
```diff
@@ -35,3 +35,13 @@ export const isIOS = (
...
 };
+
+/**
+ * Returns true when the device is a touch device such
+ * as android or iOS.
+ *
+ * @param _window {Window} - Test seam
+ */
+export const isTouchDevice = (_window = window) => {
+  return _window.matchMedia('(pointer: coarse)').matches;
```
robertknight, line 46: From https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer#browser_compatibility the only browser that is theoretically supported but may not support `(pointer: coarse)` is Firefox < 64. These devices will simply report as non-touch, which is fine. The same will happen with CSS rules that use this query.

src/annotator/range-util.js
robertknight: This comment was marked as resolved but I didn't get an answer on the query above. Did you figure out what happens here on iOS?

robertknight (changes requested):
Hi Kyle,

The changes look good except for a sinon stub that is not being reset. Also, did you figure out what was happening with `isSelectionBackwards` on iOS? What are the values of `range.startContainer` and `selection.focusNode` when the selection is backwards?

src/shared/test/user-agent-test.js
LMS007: oops, good catch

src/shared/user-agent.js
LMS007: I will add that link in a comment.. I think its valuable.

src/annotator/range-util.js
```diff
@@ -10,6 +10,8 @@ export function isSelectionBackwards(selection) {
   }
 
   const range = selection.getRangeAt(0);
+  // Does not work correctly on iOS when selecting nodes backwards.
+  // https://bugs.webkit.org/show_bug.cgi?id=220523
```
robertknight, line 14: Thanks for filing this bug report. I notice that it links to a URL in an S3 bucket that you host. Just in case you remove this bucket in future for any reason I would suggest adding a copy of the HTML file as an attachment to the bug report. Also it may be worth mentioning the URL of this PR as additional context for anyone looking at the issue in future.

robertknight (approved):
LGTM. Sorry for the delay in re-reviewing this Kyle. Please do feel free to chase me up if you're waiting on a re-review that I've missed. I added a small request regarding the bug report that you filed upstream with WebKit.

src/annotator/range-util.js
LMS007: done!

PR #2860, opened 2021-01-04
Hide the flag icon if user is unauthenticated

esanzgar:
Previously, all users were able to click the flag icon. However, unauthenticated users were shown the following toast message if they click the icon: `You must be logged in to report an annotation`.

The behaviour has been changed so that the flag button is only shown (and clickable) for authenticated users.

Closes https://github.com/hypothesis/product-backlog/issues/1037

### DIFF
--- a/src/sidebar/components/annotation-action-bar.js
+++ b/src/sidebar/components/annotation-action-bar.js
@@ -50,9 +50,9 @@ function AnnotationActionBar({
   const showDeleteAction = userIsAuthorizedTo('delete');
   const showEditAction = userIsAuthorizedTo('update');
 
-  // Anyone may flag an annotation except the annotation's author.
-  // This option is even presented to anonymous users
-  const showFlagAction = userProfile.userid !== annotation.user;
+  //  Only authenticated users can flag an annotation, except the annotation's author.
+  const showFlagAction =
+    !!userProfile.userid && userProfile.userid !== annotation.user;
   const showShareAction = isShareable(annotation, settings);
 
   const onDelete = () => {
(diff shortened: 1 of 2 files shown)


esanzgar:
A comment regarding flagged annotation: the red flag icon is only shown to the authenticated user that clicked on the icon. Other users do not see the red flag. This behaviour has not changed.

src/sidebar/components/annotation-action-bar.js
```diff
@@ -50,9 +50,9 @@ function AnnotationActionBar({
   const showDeleteAction = userIsAuthorizedTo('delete');
   const showEditAction = userIsAuthorizedTo('update');
 
-  // Anyone may flag an annotation except the annotation's author.
-  // This option is even presented to anonymous users
-  const showFlagAction = userProfile.userid !== annotation.user;
+  //  Only authenticated users can flag an annotation, except the annotation's author.
+  const showFlagAction =
+    !!userProfile.userid && userProfile.userid !== annotation.user;
```
robertknight, line 55: Since `userid` is either `null` or a non-empty string, the boolean cast isn't strictly needed here.

robertknight (approved):
The code change looks correct to me for the intended change. The decision to still show the flag action to logged-out users must have been intentional in the past though. I think the rationale is to make logged-out users aware of the affordance, even though they can't actually use it without logging in. The fact that someone filed a report about the current behavior suggests that this intention wasn't very clear though. Also the UI looked terrible - it would have been better to show a panel (like when you try to annotate while not logged in), rather than an error.

esanzgar:
Another alternative would have be to disable the button for logged-out users. The problem with this approach is to explain to the user way a button is disabled. A common approach for communicating that is via a tooltip.

I personally prefer the idea proposed by @robertknight  above.

src/sidebar/components/annotation-action-bar.js
esanzgar: You are correct. I intended to make the type of  `showFlagAction` always `boolean`, no `boolean | null`, although in this particular case doesn't make a difference.

PR #2885, opened 2021-01-15
Fix page overflow caused by highlights close to borders

### LINKED ISSUE #2811: On small screens the client makes some websites horizontally scrollable
### Steps to reproduce
1. On a small window (mobile devices included), , go to https://via.hypothes.is/https://example.com
2. The `annotator-bar` appears on the right margin and then shortly it disappears
3. The client makes the page to not fit in the screen and and is now scrollable horizontally 
4. The `annotation-bar` can be found by scrolling to the right side of the screen 

Dec-10-2020 14-13-20

### Expected behaviour
example.com has a responsive layout and it fits any screen size without scrolling

### Actual behaviour
Addition of the client to example.com makes the document scrollable.

I have tested this behaviour on Android/Chrome, iOS/Chrome, iOS/Safari and Mac/Chrome (making the window small or by activating the chrome devtools/device toolbar)

esanzgar:
Highlights contains two pseudo-elements (`::before` and `::after`) with
text for assistive technologies. Because these elements have dimensions,
on RTL languages a highlight near the right border of the document,
makes the `::after` pseudo-element to overflow the page to the right.

image

Based on similar classes from other popular CSS frameworks, I made these
pseudo-elements to have
`1 x 1px` dimensions. This does not affect negatively the screen readers.

Example:

- [Bootstrap](https://github.com/twbs/bootstrap/blob/c2f67d72e8010307f657b2a55ae93fc5c9cab2a5/scss/mixins/_visually-hidden.scss#L9-L17)

- [Foundation](https://github.com/foundation/foundation-sites/blob/258d9eb3432d7e497f41b4554f4116d5c59c92a8/scss/util/_mixins.scss#L217-L224)

Closes #2811

### DIFF
--- a/src/styles/annotator/highlights.scss
+++ b/src/styles/annotator/highlights.scss
@@ -34,7 +34,7 @@
       @include a11y.screen-reader-only;
 
       // nb. The leading/trailing spaces are intended to ensure the text is treated
-      // as separate words by assisitve technologies from the content before/after it.
+      // as separate words by assistive technologies from the content before/after it.
       content: ' annotation start ';
     }
     &::after {
--- a/src/styles/mixins/a11y.scss
+++ b/src/styles/mixins/a11y.scss
@@ -6,6 +6,9 @@
 @mixin screen-reader-only {
   // Take the content out of the normal layout flow.
   position: absolute;
+  width: 1px;
+  height: 1px;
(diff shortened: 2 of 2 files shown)


robertknight (approved):
👍

PR #2886, opened 2021-01-15
Load notebook view correctly in extension

robertknight:
When the app launches the `router` service sets the initial route based on the path. When the notebook view is loaded from the extension the path is `/client/notebook.html` rather than `/notebook`. This PR updates the `router` service to match
the `/client/notebook.html` path to the `notebook` route.

In the process of fixing this I found an issue where the notebook's outer `<div>` was mis-positioned due to generic styles on https://example.com. I fixed this by changing the tag to `<hypothesis-notebook>` following the example of other UI elements created by the annotator.

Fixes https://github.com/hypothesis/client/issues/2866.

----

**Before:**

Extension notebook before

**After:**

Extension notebook after

### DIFF
--- a/src/annotator/notebook.js
+++ b/src/annotator/notebook.js
@@ -34,7 +34,7 @@ export default class Notebook extends Delegator {
     this._groupId = null;
     this._prevGroupId = null;
 
-    this.container = document.createElement('div');
+    this.container = document.createElement('hypothesis-notebook');
     this.container.style.display = 'none';
     this.container.className = 'notebook-outer';
 
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -86,8 +86,8 @@ function autosave(autosaveService) {
 }
 
 // @inject
-function setupFrameSync(frameSync, isSidebar) {
-  if (isSidebar) {
+function setupFrameSync(frameSync, store) {
+  if (store.route() === 'sidebar') {
     frameSync.connect(true);
   }
 }
@@ -135,12 +135,6 @@ import { Injector } from '../shared/injector';
 function startApp(config) {
   const container = new Injector();
 
-  const isSidebar = !(
-    window.location.pathname.startsWith('/stream') ||
-    window.location.pathname.startsWith('/a/') ||
-    window.location.pathname.startsWith('/notebook')
-  );
-
   // Register services.
   container
     .register('analytics', analyticsService)
@@ -172,7 +166,6 @@ function startApp(config) {
   // that use them, since they don't depend on instances of other services.
   container
     .register('$window', { value: window })
-    .register('isSidebar', { value: isSidebar })
     .register('settings', { value: config });
 
   // Initialize services.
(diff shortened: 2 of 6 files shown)


robertknight:
In testing this I noticed errors in the browser console about the existence of multiple `Discovery` servers. I think that is because the extension is trying to inject itself into the notebook. That is not fixed in this PR:

Extension multiple discovery

In testing with a production build of the extension I found an issue where the initial attempt to open the notebook in a group would result in the notebook appearing briefly and then hiding itself. Opening the notebook a second time would leave it on screen. This is something I'll look into separately.

robertknight:
I pushed an additional commit which fixes the issues mentioned in the previous comment. See commit for explanation.

src/annotator/notebook.js
```diff
@@ -34,7 +34,7 @@ export default class Notebook extends Delegator {
     this._groupId = null;
     this._prevGroupId = null;
 
-    this.container = document.createElement('div');
+    this.container = document.createElement('hypothesis-notebook');
```
lyzadanger, line 37: I'm the HTML validation pedant, I know: I believe this will leave us in a situation in which we wouldn't pass HTML validation; that can be problematic with a11y assessments, etc. Can any of the work we're doing to more aggressively use shadow DOM help here? This might be OK temporarily...

src/sidebar/index.js
```diff
@@ -135,12 +135,6 @@ import { Injector } from '../shared/injector';
 function startApp(config) {
   const container = new Injector();
 
-  const isSidebar = !(
```
lyzadanger, line 138: I'm kind of glad to see this go...

src/sidebar/services/router.js
```diff
@@ -14,9 +14,21 @@ export default function router($window, store) {
     const pathSegments = path.slice(1).split('/');
     const params = queryString.parse($window.location.search);
 
+    // The extension puts client resources under `/client/` to separate them
```
lyzadanger, line 17: I'm glad this is captured here. I had no idea!

src/sidebar/services/router.js
```diff
@@ -14,9 +14,21 @@ export default function router($window, store) {
...
+    // from extension-specific resources. Ignore this part.
+    if (pathSegments[0] === 'client') {
+      pathSegments.shift();
+    }
+
+    // Routes loaded from h have no file extensions (eg. `/a/:id`, `/notebook`).
```
lyzadanger, line 23: Good comments, thanks.

lyzadanger (approved):
Rock on! Looks great. Tried it out locally against production data. 👍 

I just had one wariness about the custom HTML element...nothing new here.

src/annotator/notebook.js
robertknight: > I believe this will leave us in a situation in which we wouldn't pass HTML validation; that can be problematic with a11y assessments

We are replacing a non-semantic element with one that follows the rules for custom element names, so I don't envision problems with tools that understand modern HTML. We've also been using this naming scheme (`hypothesis-xyz`) for a number of other elements for quite a while now. In the event that we do get a report of this nature I suggest we deal with it as and when it arises.

More importantly though we should make time to check that the notebook app and navigation to/from it works well with assistive tech.

PR #2898, opened 2021-01-18
Memoize `Annotation` and `Thread` components

robertknight:
Typing in replies in large threads (eg. those that already have hundres
of replies) was slow because every keystroke would cause every
reply to be re-rendered. The critical thread list virtualization
optimization in the client only applies to top-level threads and doesn't
benefit the case of a single thread with a large number of replies.

Implement some easy performance optimizations for large threads (ie.
with hundreds of replies) by memoizing:

 1) `Thread` components so that they don't re-render if something
    changes in a parent but the thread itself is unchanged.
 2) `Annotation` components so that they don't re-render if something
    changes elsewhere in the thread list but the current annotation is not
    changed.

As a result, creating and editing replies in large threads is faster. The
initial action to show a large thread list by clicking "Show replies" is still
slow because this change doesn't help the initial render. However,
editing performance is the most important issue from a user's
perspective.

Partial fix for #2897

### DIFF
--- a/src/sidebar/components/thread-card.js
+++ b/src/sidebar/components/thread-card.js
@@ -56,6 +56,13 @@ function ThreadCard({ frameSync, thread }) {
     return !!target.closest('button') || !!target.closest('a');
   };
 
+  // Memoize threads to reduce avoid re-rendering when something changes in a
+  // parent component but the `Thread` itself has not changed.
+  const threadContent = useMemo(
+    () => <Thread thread={thread} showDocumentInfo={showDocumentInfo} />,
+    [thread, showDocumentInfo]
+  );
+
   return (
     /* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
Good work. I have a wonder if this pattern might be useful in some other places.

The syntax is a bit fugly (the nested anonymous arrow/implicit returns with JSX), IMHO, but that's just an aesthetic snobbery.

I'm glad there was such a straightforward way to alleviate this!

robertknight:
> The syntax is a bit fugly (the nested anonymous arrow/implicit returns with JSX), IMHO, but that's just an aesthetic snobbery.

Yeah. Fortunately this is something that you only have to apply strategically in a few places to reap most of the benefit. There is also the possibility in future that a JS processor (eg. a Babel plugin) could generate this for you, so as an author you'd just put some kind of annotation above a block of JSX to indicate that it should be processed this way. Maybe it already exists - I haven't looked into it recently.

PR #2900, opened 2021-01-19
Fix JS browser compatibility on dev

esanzgar:
Because the code in `dev-server/static/scripts/util.js` is not
transpiled by Babel, it needs to be compatible with all the browsers
versions in `browserslist` (`package.json`).

Optional chaining `(?.)` is for example not support in Chrome versions <
80.

This is only an issue when testing `localhost:3000` on those supported,
but older browser versions.

### DIFF
--- a/dev-server/serve-dev.js
+++ b/dev-server/serve-dev.js
@@ -29,11 +29,12 @@ function renderScript(context) {
   const scriptTemplate = `
     {{{hypothesisConfig}}}
 
-    <script type="module">
-      import { loadClient } from '/scripts/util.js';
-
-      const clientUrl = '{{{clientUrl}}}'.replace('{current_host}', document.location.hostname);
-      loadClient(clientUrl);
+    <script src="/scripts/util.js"></script>
+    <script>
+      (function(){
+        let clientUrl = '{{{clientUrl}}}'.replace('{current_host}', document.location.hostname);
+        loadClient(clientUrl);
+      })();
     </script>
   `;
   return Mustache.render(scriptTemplate, context);
--- a/dev-server/static/scripts/util.js
+++ b/dev-server/static/scripts/util.js
@@ -1,13 +1,21 @@
+ /**
+  * Because the code in this file is not transpiled by Babel, it must be compatible
+  * with all the supported browsers version (see `browserlist` in `package.json`)
+  * without transpilation. Do not include latest EcmaScript features as these
+  * will cause exceptions while working on dev (`localhost:3000`) on slightly
+  * older, yet supported browser versions.
+  */
+
 /** @type {string|null} */
-export let activeClientUrl;
+let activeClientUrl;
 
 /**
  * Load the Hypothesis client into the page.
  *
  * @param {string} clientUrl
  */
-export function loadClient(clientUrl) {
-  const embedScript = document.createElement('script');
+function loadClient(clientUrl) {
+  let embedScript = document.createElement('script');
   embedScript.src = clientUrl;
   document.body.appendChild(embedScript);
   activeClientUrl = clientUrl;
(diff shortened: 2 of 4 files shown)


dev-server/static/scripts/util.js
```diff
@@ -19,7 +19,9 @@ export function loadClient(clientUrl) {
  * This uses the same method as the browser extension does to deactivate the client.
  */
 export function unloadClient() {
-  const annotatorLink = document.querySelector('link[type="application/annotator+html"]');
-  annotatorLink?.dispatchEvent(new Event('destroy'));
+  const annotatorLink = document.querySelector('link[type="application/annotatror+html"]');
+  // Do not use optional chaining (?.) below. Because this code is not bundled by babel
+  // we need to make compatible with older browsers 
+  annotatorLink && annotatorLink.dispatchEvent(new Event('destroy'));
```
robertknight, line 25:
```suggestion
if (annotatorLink) {
  annotatorLink.dispatchEvent(new Event('destroy'));
}
```

What you have written works and probably what something like Terser would generate, but I think using a conditional is more idiomatic if we're restricted to not using optional chaining.

dev-server/static/scripts/util.js
```diff
@@ -19,7 +19,9 @@ export function loadClient(clientUrl) {
  * This uses the same method as the browser extension does to deactivate the client.
  */
 export function unloadClient() {
-  const annotatorLink = document.querySelector('link[type="application/annotator+html"]');
-  annotatorLink?.dispatchEvent(new Event('destroy'));
+  const annotatorLink = document.querySelector('link[type="application/annotatror+html"]');
+  // Do not use optional chaining (?.) below. Because this code is not bundled by babel
```
robertknight, line 23: I'd suggest to add a general note at the top of the file that it is not transpiled, so future readers are aware, then remove this more specific comment.

robertknight (changes requested):
Hello - The change works but I suggested to replace the comment with a general note at the top of the file and use a conditional rather than `&&`.

dev-server/static/scripts/util.js
esanzgar: Good idea.

esanzgar:
I accepted your suggestion and added the general comment on top of the file.

esanzgar:
I added a second commit to replace the module import in `<script>` tags.

dev-server/serve-dev.js
```diff
@@ -29,11 +29,12 @@ function renderScript(context) {
...
 
-    <script type="module">
-      import { loadClient } from '/scripts/util.js';
-
-      const clientUrl = '{{{clientUrl}}}'.replace('{current_host}', document.location.hostname);
-      loadClient(clientUrl);
+    <script src="/scripts/util.js"></script>
+    <script>
+      (function(){
+        var clientUrl = '{{{clientUrl}}}'.replace('{current_host}', document.location.hostname);
```
robertknight, line 35: Given our browser support policy you can still use `const`/`let` etc. It is only modules and optional chaining which aren't natively supported.

dev-server/static/scripts/util.js
```diff
@@ -1,13 +1,21 @@
...
+  * Because the code in this file is not bundled by babel, it must be compatible
+  * with all the supported browsers version (see `browserlist` in `package.json`)
+  * without transpilation. Do not include latest EcmaScript features as these
+  * will cause exceptions while working on dev (`localhost:3000`) on slightly
+  * older, yet supported browser versions.
+  */
+
 /** @type {string|null} */
-export let activeClientUrl;
+var activeClientUrl;
```
robertknight, line 10: You can still use `let`/`const` rather than `var`.

robertknight (approved):
Thanks for taking care of this. The revised code works fine, although I noted that reverting from `const`/`let` back to `var` was not strictly necessary.

dev-server/static/scripts/util.js
esanzgar: Yes, I agree.

I will change the all the `var` for `let` and merge the PR.

PR #2902, opened 2021-01-20
Fix incorrect use of `hasClass` in `Button` test

robertknight:
The Enzyme `hasClass` method was being called on a non-DOM node by
mistake. This didn't cause any problems, but it just meant that the test
was not checking what it intended to check.


lyzadanger (approved):
Heh. Nice catch. Thanks.

PR #2903, opened 2021-01-20
Enable progress reporting in test runner

robertknight:
This PR enables the built-in progress reporting in the Karma test runner. This was helpful yesterday while debugging an issue in https://github.com/hypothesis/client/pull/2892 where the the browser disconnected from Karma after running the test suite.

I also removed some obsolete verbose logging that was added a while back to help debug a problem on Jenkins. We haven't seen that problem much (if at all?) on Jenkins recently.

### DIFF
--- a/src/karma.config.js
+++ b/src/karma.config.js
@@ -129,7 +129,7 @@ module.exports = function (config) {
 
     // Use https://www.npmjs.com/package/karma-mocha-reporter
     // for more helpful rendering of test failures
-    reporters: ['mocha', 'coverage-istanbul'],
+    reporters: ['progress', 'mocha', 'coverage-istanbul'],
 
     // web server port
     port: 9876,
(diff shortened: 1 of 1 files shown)


esanzgar (approved):
In this PR, you are removing obsolete verbose logging options that are not needed except for debugging purposes. On the other hand, you are adding another verbose logging for the same reason. I wonder if someone would find the new verbose option unnecessary in the future.

My suggestion would be to leave and comment all these debugging options in configuration file. If we encounter issues, then we can easily uncomment those.

robertknight:
> On the other hand, you are adding another verbose logging for the same reason.

The logging I'm removing was extremely verbose (as the comments said, tens of megabytes in size). What I've enabled is just a progress reporter that is built into Karma so you can see how far through the test run it is. This is useful aside from debugging Karma because you can see how long you've got to wait before the test run completes.

> My suggestion would be to leave and comment all these debugging options in configuration file. If we encounter issues, then we can easily uncomment those.

I am really not a fan of large blocks of commented out code in files, as I find it distracts from the configuration / code which is actively used. There is also is a decent chance that this stuff is made obsolete by future changes in Chrome. What I would much prefer to do instead is make sure any relevant knowledge is captured somewhere that is easy to find. Small blocks of comments about enabling additional debugging (eg. the one above the `logLevel` config) are OK.

PR #2904, opened 2021-01-20
Use shadow DOM to encapsulate sidebar's style

esanzgar:
A side effect of using shadow DOM for the sidebar is that the BucketBar
'plugin' cannot be injected easily using a current query mechanism.
After consulting with @robertknight, we decided to avoid using the
normal plugin injection mechanism and instead instantiate the BucketBar
from the sidebar.

This PR also includes:

- a (non documented) configuration option to disable the shadow DOM
  encapsulation. This can be removed in the future if not needed.
- more strict types
- simplification of the logic in the sidebar

I have tested these changes in the following browsers:

Brower\OS    | MacOS              | Windows
------------ | ------------------ | -----------------
Chrome 57    | :heavy_check_mark: | :heavy_check_mark:
Chrome beta  | :heavy_check_mark: | :heavy_check_mark:
Edge 17      |                    | :heavy_check_mark:
Edge beta    | :heavy_check_mark: | :heavy_check_mark:
Firefox 53   | :heavy_check_mark: | :heavy_check_mark:
Firefox beta | :heavy_check_mark: | :heavy_check_mark:
Safari 10    | :heavy_check_mark: |
Safari 14    | :heavy_check_mark: |

### DIFF
--- a/src/annotator/delegator.js
+++ b/src/annotator/delegator.js
@@ -28,7 +28,7 @@ export default class Delegator {
    * Construct the `Delegator` instance.
    *
    * @param {HTMLElement} element
-   * @param {Object} [config]
+   * @param {Record<string, any>} [config]
    */
   constructor(element, config) {
     this.options = { ...config };
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -8,6 +8,8 @@ import features from './features';
 
 import Guest from './guest';
 import { ToolbarController } from './toolbar';
+import { createShadowRoot } from './util/shadow-root';
+import BucketBar from './plugin/bucket-bar';
 
 /**
  * @typedef LayoutState
@@ -18,12 +20,6 @@ import { ToolbarController } from './toolbar';
 
 // Minimum width to which the frame can be resized.
 const MIN_RESIZE = 280;
-
-const defaultConfig = {
-  BucketBar: {
-    container: '.annotator-frame',
-  },
-};
 
 /**
  * Create the iframe that will load the sidebar application.
@@ -53,24 +49,20 @@ function createSidebarIframe(config) {
  * as well as the adjacent controls.
  */
 export default class Sidebar extends Guest {
+  /**
+   * @param {HTMLElement} element
+   * @param {Record<string, any>} config
+   */
   constructor(element, config) {
-    if (config.theme === 'clean' || config.externalContainerSelector) {
-      delete config.pluginClasses.BucketBar;
-    }
-
-    let externalContainer = null;
-
-    if (config.externalContainerSelector) {
-      externalContainer = document.querySelector(
-        config.externalContainerSelector
-      );
-    }
+    super(element, config);
 
     let externalFrame;
     let frame;
+    let hypothesisSidebar; // refers to <hypothesis-sidebar> element
 
-    if (externalContainer) {
-      externalFrame = externalContainer;
+    if (config.externalContainerSelector) {
+      externalFrame =
+        document.querySelector(config.externalContainerSelector) || element;
     } else {
       frame = document.createElement('div');
       frame.style.display = 'none';
@@ -78,18 +70,32 @@ export default class Sidebar extends Guest {
 
       if (config.theme === 'clean') {
         frame.classList.add('annotator-frame--theme-clean');
+      } else {
+        // BucketBar is a "plugin" for legacy reasons and is now constructed here so
+        // that the parent element can be passed into the constructor.
+        this.plugins.BucketBar = new BucketBar(frame, {}, this);
       }
 
-      element.appendChild(frame);
+      // Undocumented switch to enable/disable the wrapping of the sidebar inside a shadow DOM
+      // 2021-01-22: remove this switch after the 2021-02-05
+      if (config.disableShadowSidebar) {
+        element.appendChild(frame);
+      } else {
+        // Wrap up the 'frame' element into a shadow DOM so it is not affected by host CSS styles
+        hypothesisSidebar = document.createElement('hypothesis-sidebar');
+        const shadowDom = createShadowRoot(hypothesisSidebar);
+        shadowDom.appendChild(frame);
+
+        element.appendChild(hypothesisSidebar);
+      }
     }
 
     const sidebarFrame = createSidebarIframe(config);
-
-    super(element, { ...defaultConfig, ...config });
+    (frame || externalFrame).appendChild(sidebarFrame);
 
     this.externalFrame = externalFrame;
     this.frame = frame;
-    (frame || externalFrame).appendChild(sidebarFrame);
+    this.hypothesisSidebar = hypothesisSidebar;
 
     this.subscribe('panelReady', () => {
       // Show the UI
@@ -171,6 +177,7 @@ export default class Sidebar extends Guest {
   destroy() {
     this._hammerManager?.destroy();
     this.frame?.remove();
+    this.hypothesisSidebar?.remove();
     super.destroy();
   }
 
--- a/src/annotator/test/sidebar-test.js
+++ b/src/annotator/test/sidebar-test.js
@@ -15,6 +15,7 @@ describe('Sidebar', () => {
 
   // `Sidebar` instances created by current test.
   let sidebars;
+  let sidebarContainer;
 
   let FakeToolbarController;
   let fakeToolbar;
@@ -25,6 +26,7 @@ describe('Sidebar', () => {
 
   after(() => {
     window.requestAnimationFrame.restore();
+    sidebarContainer?.remove();
   });
 
   const createSidebar = (config = {}) => {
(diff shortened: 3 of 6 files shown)


src/annotator/sidebar.js
```diff
@@ -48,48 +44,58 @@ function createSidebarIframe(config) {
...
-    }
-
-    let externalContainer = null;
-
-    if (config.externalContainerSelector) {
-      externalContainer = document.querySelector(
-        config.externalContainerSelector
-      );
-    }
+    super(element, { ...defaultConfig, ...config });
```
robertknight, line 59: `defaultConfig` is empty, so I believe this is equivalent to `super(element, config)`.

src/annotator/sidebar.js
```diff
@@ -48,48 +44,58 @@ function createSidebarIframe(config) {
...
       frame = document.createElement('div');
       frame.style.display = 'none';
       frame.className = 'annotator-frame annotator-outer';
 
       if (config.theme === 'clean') {
         frame.classList.add('annotator-frame--theme-clean');
+      } else {
+        // The instantiation of the BucketBar plugin can't be done as other plugin.
+        // The sidebar must be inserted inside a shadow DOM which is not accessible
+        // via a common querySelector.
```
robertknight, line 77: I understand what this comment means but I think it may not make sense to someone in future who wasn't aware of the way the code used to work.

I think it would be better to say that the `BucketBar` is a "plugin" for legacy reasons and is now constructed here so that the parent element can be passed into the constructor.

src/annotator/sidebar.js
```diff
@@ -48,48 +44,58 @@ function createSidebarIframe(config) {
...
+        this.plugins.BucketBar = new BucketBar(frame, {}, this);
       }
 
-      element.appendChild(frame);
+      // Undocumented switch to enable/disable the wrapping of the sidebar inside a shadow DOM
```
robertknight, line 81: Do you have a use case for this switch or is it effectively a "feature flag" that we can deploy quickly if we find that the change causes problems? If it is the latter, I suggest we schedule a task to remove it in the next sprint, once the change has been deployed for a week or two and we're happy that there are no major issues.

src/annotator/delegator.js
```diff
@@ -28,7 +28,7 @@ export default class Delegator {
    * Construct the `Delegator` instance.
    *
    * @param {HTMLElement} element
-   * @param {Object} [config]
+   * @param {Record<string, any>} [config]
```
robertknight, line 31: 👍 - It is good to iteratively improve these.

src/annotator/sidebar.js
```diff
@@ -19,12 +21,6 @@ import { ToolbarController } from './toolbar';
 // Minimum width to which the frame can be resized.
 const MIN_RESIZE = 280;
 
-const defaultConfig = {
-  BucketBar: {
-    container: '.annotator-frame',
```
robertknight, line 24: The `container` option is no longer set anywhere but `BucketBar` still has code to handle it. Can you open a follow-up PR to remove that code?

robertknight (commented):
Hi @esanzgar - I had a few minor pieces of feedback but overall this looks good. The main issues are:

- There is some obsolete code in `BucketBar` that can be removed either here or in a follow-up
- I had a query about the `disableShadowSidebar` flag
- There is an empty `defaultConfig` variable which I think can be removed

src/annotator/sidebar.js
esanzgar: Yes, you are correct.

At some point I removed the `defaultConfig`, I added back to make it consistent to the `pdf-sidebar.js` class. I thought maybe in the future we want to add configuration at this level.

I am happy to remove it if you wish.

src/annotator/sidebar.js
esanzgar: Done

src/annotator/sidebar.js
esanzgar: It is a `feature flag`. Yes, I agree it should be removed once we are happy.

src/annotator/sidebar.js
esanzgar: Yes, I will. In addition, we may want to remove anything that make it work as a 'plugin'.

src/annotator/sidebar.js
esanzgar: You are correct.

At some point I remove the `defaultConfig`, then I put it back to make if consistent with the pdf-sidebar.js. I thought that maybe in the future it could be handy.

I am happy to remove it entirely.

src/annotator/sidebar.js
esanzgar: https://github.com/hypothesis/client/issues/2907

src/annotator/sidebar.js
esanzgar: Done

src/annotator/sidebar.js
```diff
@@ -171,6 +176,7 @@ export default class Sidebar extends Guest {
   destroy() {
     this._hammerManager?.destroy();
     this.frame?.remove();
+    this.hypothesisSidebar?.remove();
```
robertknight, line 179: If `this.hypothesisSidebar` is set then it contains `this.frame` doesn't it? If so then `this.frame` doesn't need to be removed as well. In other words, I think this logic have the same effect:

```js
if (this.hypothesisSidebar) {
  this.hypothesisSidebar.remove();
} else {
  this.frame?.remove();
}
```

I'm OK with keeping the existing logic as is, I'm just noting that we could simplify it when the flag is removed.

src/annotator/test/sidebar-test.js
```diff
@@ -724,9 +738,19 @@ describe('Sidebar', () => {
 
     it('does not have the BucketBar if an external container is provided', () => {
       const sidebar = createSidebar({
-        externalContainerSelector: '.' + EXTERNAL_CONTAINER_SELECTOR,
+        externalContainerSelector: `.${EXTERNAL_CONTAINER_SELECTOR}`,
       });
       assert.isUndefined(sidebar.plugins.BucketBar);
     });
+
+    it('disables shadow DOM', () => {
```
robertknight, line 746:
```suggestion
    it('disables shadow DOM if `disableShadowSidebar` flag is set', () => {
```

This just makes the test description more obvious.

src/annotator/sidebar.js
robertknight: OK. Can you add a note here that we should remove this flag once the change has been deployed for a week or two?

src/annotator/sidebar.js
```diff
@@ -53,43 +49,52 @@ function createSidebarIframe(config) {
...
-    if (config.externalContainerSelector) {
-      externalContainer = document.querySelector(
-        config.externalContainerSelector
-      );
-    }
+    super(element, config);
 
     let externalFrame;
     let frame;
+    let hypothesisSidebar;
```
robertknight, line 61: Can you add a comment here that this refers to the `<hypothesis-sidebar>` element?

Once we've confirmed that the Shadow DOM approach is working then we can look to simplify some of the inner DOM structure and class names.

src/annotator/sidebar.js
```diff
@@ -53,43 +49,52 @@ function createSidebarIframe(config) {
...
+        this.plugins.BucketBar = new BucketBar(frame, {}, this);
       }
 
-      element.appendChild(frame);
+      // Undocumented switch to enable/disable the wrapping of the sidebar inside a shadow DOM
+      if (config.disableShadowSidebar) {
+        element.appendChild(frame);
+      } else {
+        // Wrap up the 'frame' element into a shadow DOM so it is not affected by host CSS styles
+        hypothesisSidebar = document.createElement('hypothesis-sidebar');
```
robertknight, line 84: Currently this element has no styles and just functions as a container. That means that the browser shows it as having zero size in devtools which is perhaps surprising. Once we're satisfied that the introduction of a Shadow DOM is working and hasn't caused any unexpected issues I think it would make sense to move some of the frame's styles onto this element.

robertknight (approved):
This looks to be working as intended in my testing. There is a bunch of follow-up work we can do here to clean up the sidebar's DOM and CSS once we are satisfied that moving the sidebar frame into a shadow root doesn't cause any unexpected issues. For example we can remove many of the specificity hacks in the annotator CSS.

src/annotator/sidebar.js
esanzgar: 👍🏼

src/annotator/sidebar.js
esanzgar: Yes, it is true, your suggested logic has the same effect.

I will simplify once we remove the flag.

esanzgar: Interestingly, the `this.externalFrame` is never removed. I don't know if that was intentional or not.

src/annotator/sidebar.js
esanzgar: Added comment and issue to track follow up actions (#2908)

src/annotator/sidebar.js
esanzgar: In my opinion `<hypothesis-sidebar>` should act purely as a container and should not have associated styles.

PR #2928, opened 2021-01-26
Add github action to publish frontend-shared package

LMS007:
Add custom github action to publish frontend-shared package …

Action will perform the following:
- Test if any changes to the frontend-shared/ folder have been made since the last release of the client (parent).
- If they have, then it will get the last published version of the frontend-shared package, add 1 to the number, and save new version to package.json file
- Build the frontend-shared for release
- Publish the frontend-shared package to npm
----------------

_updated_

depends on https://github.com/hypothesis/client/pull/2875

This PR is based on the github action docs for publishing a package to npm. https://docs.github.com/en/actions/guides/publishing-nodejs-packages 

The core difference is that this kicks off its own custom action that invokes nodejs and runs a file `index.js` inside the actions folder.  

It first checks to see if there are any actual changes to frontend-shared by looking at the diff. If there are, then it
bumps the minor number of the verison similar to our `Jenkins` file. 

I did test this out a number of times on a different event `[push]` and it did publish several versions to npmjs which is why the number is currently sitting at 1.4.0. I assume we could force this back to 1.0.0 if we wanted, but it didn't actually seem straightforward so I left it for now. -- TBD.

_I have not actually tested this on the "release" event the action is currently targeted to fire for, so that remains to be tested._ 

Foring the action to run on [push] will yield this result. To get it to actually publish a package, we will likely need to test this out on a release with an actual change in frontend-shared/. 

Screen Shot 2021-01-28 at 11 15 13 AM

### DIFF
--- a/.github/actions/frontend-shared-version/action.yml
+++ b/.github/actions/frontend-shared-version/action.yml
@@ -1,0 +1,18 @@
+name: 'Get Updated Package Version'
+inputs:
+  # The shared package name, "@hypothesis/frontend-shared"
+  shared-package:
+    required: true
+    default: ''
+    description: 'The npm package name'
+  # The client package name, "hypothesis"
+  parent-package:
+    required: true
+    default: ''
+    description: 'The npm package name'
+outputs:
+  updated_version:
+    description: 'The new version of the npm package'
+runs:
+  using: 'node12'
+  main: 'index.js'
--- a/.github/actions/frontend-shared-version/index.js
+++ b/.github/actions/frontend-shared-version/index.js
@@ -1,0 +1,63 @@
+'use strict'
+
+const core = require('@actions/core');
+const util = require('util');
+
+const exec = util.promisify(require('child_process').exec);
+
+/**
+ * Returns true is there is a non-zero length diff between the last published version
+ * and master within the frontend-shared/ folder.
+ */
+async function haveChangesOccurred() {
+  const parentPackage = core.getInput('parent-package');
+  const versionResult = await exec(`npm view ${parentPackage}@latest version`);
+  const version = versionResult.stdout.trim();
+  const diffResult = await exec(`git diff --stat v${version}..master frontend-shared/`)
+  return diffResult.stdout.length > 0;
+}
+
+/**
+ * Increment the minor part of a `MAJOR.MINOR.PATCH` semver version.
+ */
+function bumpMinorVersion(version) {
+  const parts = version.split('.')
+  if (parts.length !== 3) {
+    throw new Error(`${version} is not a valid MAJOR.MINOR.PATCH version`);
+  }
+  const majorVersion = parseInt(parts[0]);
+  const newMinorVersion = parseInt(parts[1]) + 1;
+  const patchVersion = parseInt(parts[2]);
+  if (isNaN(majorVersion) || isNaN(newMinorVersion) || isNaN(patchVersion)) {
+    throw new Error(`${version} does not have valid integer parts`);
+  }
+  return `${parts[0]}.${newMinorVersion}.${parts[2]}`
+}
+
+/**
+ * Get the current version from npm and bump the minor part by 1.
+ */
+async function getNewVersion() {
+  const sharedPackage = core.getInput('shared-package');
+  const result = await exec(`npm view ${sharedPackage} version`);
+  const newVersion = bumpMinorVersion(result.stdout);
+  return newVersion;
+}
+
+async function main() {
+  if (await haveChangesOccurred()) {
+    // eslint-disable-next-line no-console
+    console.log('Changes detected in frontend-shared/, publishing new version...');
+    const newVersion = await getNewVersion();
+    core.setOutput("updated_version", newVersion);
+  } else {
+    // eslint-disable-next-line no-console
+    console.log('No changes detected in frontend-shared/');
+    core.setOutput("updated_version", null);
+  }
+}
+
+main().catch(error => {
+  core.setFailed(error.message);
+})
+
--- a/.github/workflows/frontend-shared.yml
+++ b/.github/workflows/frontend-shared.yml
@@ -1,0 +1,44 @@
+name: frontend-shared package
+on:
+  release:
+    types:
+      - published
+jobs:
+  update-frontend-shared-version:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    # Additionally fetch master to be able to diff for changes in the frontend-shared/ folder
+    - name: Fetch origin
+      run: git fetch origin master:master
+    # Setup .npmrc file to publish to npm
+    - name: Setup Nodejs
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+        registry-url: 'https://registry.npmjs.org'
+    - name: Install client
+      run: yarn install --frozen-lockfile
+    - name: Get new version
+      uses: ./.github/actions/frontend-shared-version
+      id: version
+      with:
+        shared-package: '@hypothesis/frontend-shared'
+        parent-package: 'hypothesis'
+    - name: Set up frontend-shared
+      run: yarn setup-frontend-shared
+      # If updated_version is null, skip this step
+      if: ${{ steps.version.outputs.updated_version }}
+    - name: Set new version in package.json
+      working-directory: frontend-shared
+      run: yarn version --no-git-tag-version --new-version ${{ steps.version.outputs.updated_version }}
+      # If updated_version is null, skip this step
+      if: ${{ steps.version.outputs.updated_version }}
+    - name: Publish package
+      working-directory: frontend-shared
+      run: npm publish --access public
+      # If updated_version is null, skip this step
+      if: ${{ steps.version.outputs.updated_version }}
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/.eslintrc
+++ b/.github/.eslintrc
@@ -1,0 +1,3 @@
+{
+  "parserOptions": {}
+}
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "repository": "hypothesis/client",
   "dependencies": {},
   "devDependencies": {
+    "@actions/core": "^1.2.6",
     "@babel/core": "^7.1.6",
     "@babel/preset-env": "^7.1.6",
     "@babel/preset-react": "^7.0.0",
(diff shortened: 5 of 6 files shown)


.github/workflows/frontend-shared.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    # Setup .npmrc file to publish to npm
+    - name: Setup Nodejs
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+        registry-url: 'https://registry.npmjs.org'
+    - name: Install client
```
LMS007, line 18: An install of the client is needed for the imports to work in the .github/actions/.../index.js file(s).

Because we're already installing the client, it made more sense to bundle the package from the client (using `yarn setup-frontend-shared`) rather than changing into the frontend-shared dir and running `npm run build` 
That means the `build` command in frontend-shared will likely be unused until the whole package moves into its own repository.


`yarn setup-frontend-shared and`
vs
`cd setup-frontend-shared && npm run build`

run on completely different code paths but should produce the same result.

.github/workflows/frontend-shared.yml
```diff
@@ -0,0 +1,34 @@
+name: frontend-shared package
+on:
+  release:
+    types:
+      - published
+jobs:
+  update-frontend-shared-version:
```
LMS007, line 7: This is redundant with CI in some respects, so there may be ways to optimize these jobs.

robertknight:
> I assume we could force this back to 1.0.0 if we wanted, but it didn't actually seem straightforward so I left it for now. -- TBD.

AFAIK you can't overwrite an existing package with a new release, so we'll have to start from whatever version is currently released. For testing purpose what I would generally recommend doing is adding a prerelease suffix to the version number. That's OK - the initial version number is not that important. What normally matters with a library package is that you have a clear compatibility policy (usually semver) and follow it once the initial release is out.

.github/actions/frontend-shared-version/index.js
```diff
@@ -0,0 +1,33 @@
...
+ * Increment the minor part of a `MAJOR.MINOR.PATCH` semver version.
+ */
+function bumpMinorVersion(version) {
+  const parts = version.split('.')
+  if (parts.length !== 3) {
+    throw new Error(`${version} is not a valid MAJOR.MINOR.PATCH version`);
+  }
+  const newMinorVersion = parseInt(parts[1]) + 1;
+  return `${parts[0]}.${newMinorVersion}.${parts[2]}`
+}
```
robertknight, line 16: It might be possible to do this more succinctly and with additional validation using the `semver` package, but I think this is fine for the moment.

.github/workflows/frontend-shared.yml
```diff
@@ -0,0 +1,34 @@
+name: frontend-shared package
+on:
+  release:
```
robertknight, line 3: To check my understanding - the frontend-shared package gets updated when there is a _client release_? That will might work OK since we ship new client releases all the time. However it is also likely to mean there are a lot of releases of this package with no changes in them. Perhaps this could bail out early if there are no changes in `frontend-shared/` since the previous release?

.github/workflows/frontend-shared.yml
```diff
@@ -0,0 +1,34 @@
...
+  update-frontend-shared-version:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    # Setup .npmrc file to publish to npm
+    - name: Setup Nodejs
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
```
robertknight, line 16: You might as well use the latest Node version for this task. I can't think of any features that we immediately _require_ from it though, so skip this change if it is difficult. You'll have to change the `node12` config in `action.yml` as well.

robertknight (commented):
This looks good from an initial look through. I had a query though about whether we'll end up publishing a lot of releases with no changes in them. This will then trigger a cascade of no-op Dependabot updates in other repos. I suggest skipping the package update if nothing has changed.

.github/workflows/frontend-shared.yml
LMS007: Indeed. Off the top of my head, something like this could work.

`npm view hypothesis version `  -- assuming thats the version before the client that was just published. We may have subtract one depending on order.

Then, use that number and run this command.

`git diff v<that version #>..master frontend-shared/`

and if the stdout is anything but null/empty, then we should do a release.

.github/workflows/frontend-shared.yml
LMS007: sure, I think this was just copy pasted from examples. :)

.github/actions/frontend-shared-version/index.js
LMS007: I think we can do that, perhaps as a follow up?

LMS007:
I changed both yml files to use node 15, but not with success.
Screen Shot 2021-01-28 at 11 25 01 AM

.github/workflows/frontend-shared.yml
```diff
@@ -0,0 +1,44 @@
...
+    types:
+      - published
+jobs:
+  update-frontend-shared-version:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    # Additionally fetch master to be able to diff for changes in the frontend-shared/ folder
+    - name: Fetch origin
```
LMS007, line 13: This stumped me for a while, we can't do any `diff` without updating master. I think actions/checkout@v2 only gets the minimal git data it needs to be speedy.

.github/workflows/frontend-shared.yml
LMS007: No go on node15 :(

.github/actions/frontend-shared-version/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Returns true is there is a non-zero length diff between the last published version
+ * and master within the frontend-shared/ folder.
+ */
+async function haveChangesOccurred() {
+  const parentPackage = core.getInput('parent-package');
+  const versionResult = await exec(`npm view ${parentPackage} version`);
+  const version = versionResult.stdout.trim();
+  const diffResult = await exec(`git diff v${version}..master frontend-shared/`)
```
LMS007, line 16: This ends up looking like 

`git diff v1.647.0..master frontend-shared/`

for example.

You can see a non-zero diff if you look back 1 more version

`git diff v1.646.0..master frontend-shared/`

.github/actions/frontend-shared-version/index.js
```diff
@@ -0,0 +1,58 @@
...
+    const newVersion = await getNewVersion();
+    core.setOutput("updated_version", newVersion);
+  } else {
+    // eslint-disable-next-line no-console
+    console.log('No changes detected in frontend-shared/');
+    core.setOutput("updated_version", null);
+  }
+}
+try {
+  main();
```
robertknight, line 55: `main` is an async function so you need to use `main().catch(...)` rather than try..catch here.

.github/workflows/frontend-shared.yml
robertknight: > I think actions/checkout@v2 only gets the minimal git data it needs to be speedy.

I strongly suspect you're right. We could probably optimize this fetch by only fetching changes going back for some period of time (N days?). The repository is not too large so fetching the whole thing should work to begin with.

.github/actions/frontend-shared-version/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+const exec = util.promisify(require('child_process').exec);
+
+/**
+ * Returns true is there is a non-zero length diff between the last published version
+ * and master within the frontend-shared/ folder.
+ */
+async function haveChangesOccurred() {
+  const parentPackage = core.getInput('parent-package');
+  const versionResult = await exec(`npm view ${parentPackage} version`);
```
robertknight, line 14: The package tag to look at isn't specified here. According to https://docs.npmjs.com/cli/v6/commands/npm-view it defaults to `latest`, which is what we want.

.github/actions/frontend-shared-version/action.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Get Updated Package Version'
+inputs:
+  shared-package:
```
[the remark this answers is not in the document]
robertknight: From looking at the code it looks like it is. Could you add a comment to that effect.

.github/actions/frontend-shared-version/index.js
robertknight: You could potentially simplify the output here using `--stat` (`git diff --stat ...`) to make the output in log files more readable.

.github/workflows/frontend-shared.yml
robertknight: No problem. We can stick with Node 12 for now.

.github/workflows/frontend-shared.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Get new version
+      uses: ./.github/actions/frontend-shared-version
+      id: version
+      with:
+        shared-package: '@hypothesis/frontend-shared'
+        parent-package: 'hypothesis'
+    - name: Set up frontend-shared
+      run: yarn setup-frontend-shared
+      # If updated_version is null, skip this step
+      if: ${{ steps.version.outputs.updated_version }}
```
robertknight, line 32: I wonder if there is a way to exit early if `updated_version` is null rather than repeating the `if:` on each subsequent step?

robertknight (approved):
I found a couple of minor issues - in particular attempting to use `try..catch` around an async `main` function. Otherwise this looks quite straightforward. Can I leave it to you to fix those issues and merge/test this?

.github/workflows/frontend-shared.yml
LMS007: I noticed it only took about 1 or 2 seconds, but we would probably add a depth=100 which surely would cover to the previous tag.

.github/actions/frontend-shared-version/index.js
LMS007: will do

.github/actions/frontend-shared-version/index.js
LMS007: ahh, yes I forgot to clean that up, the try / catch was from an example that I never refactored after I added async bits.
good catch.

.github/actions/frontend-shared-version/action.yml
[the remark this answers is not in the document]
LMS007: package name, will do

.github/workflows/frontend-shared.yml
LMS007: I thought the same thing, I looked into it for about 5mins, but did not find much. I'll do some more searches about early exit.

.github/actions/frontend-shared-version/index.js
LMS007: I'll add @latest so its clear

.github/actions/frontend-shared-version/index.js
LMS007: parseInt(T10)  will be NaN but parseInt(10T) turns into 10 so I think it just lobs off letters without complaining.  Probably not big of a deal.

PR #2937, opened 2021-02-01
Update sidebar left margin when browser window is resized

esanzgar:
This PR is a variant of #2899.

Original description:

> Currently, if the client's sidebar is opened it doesn't behave
> graciously when the browser window is resized. Sometimes the sidebar
> appears floating on the middle of the window.
> 
> We discussed several alternative solutions: when the sidebar is opened
> and the window is resized, then...
> 
> 1. close the sidebar and the user opens it manually
> 2. maintain the sidebar open, but scale it properly
> 3. close the sidebar and open it after X milliseconds of the last resize
>    event
> 4. close the sidebar if the width of the window is reduced, but
>    maintain it open if the window's width is increased

In #2899, we implemented solution 1, while in this PR we implemented solution 2, because it gives a more consistent behaviour for both sidebars, PDF and non-PDF.

In addition, I added a mechanism to register and unregister events, to avoid resource leaks.

Both `sidebar` and `pdf-sidebar` register a listeners for the `window.resize` event. `sidebar` listener is triggered first because it is registered first via the `super` constructor. Although it is not very clean to have two listener registered to the same event, they do not clash with each other.

Update 2021-02-12:
- On the second commit I removed the registration of a click event that was never triggered because the element had a `pointer-events: none`

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -18,8 +18,15 @@ import BucketBar from './plugin/bucket-bar';
  * @prop {number} height
  */
 
+/**
+ * @typedef RegisteredListener
+ * @prop {Window|HTMLElement} eventTarget
+ * @prop {string} eventType
+ * @prop {(event: any) => void} listener
+ */
+
 // Minimum width to which the frame can be resized.
-const MIN_RESIZE = 280;
+export const MIN_RESIZE = 280;
 
 /**
  * Create the iframe that will load the sidebar application.
@@ -93,6 +100,9 @@ export default class Sidebar extends Guest {
     const sidebarFrame = createSidebarIframe(config);
     (frame || externalFrame).appendChild(sidebarFrame);
 
+    /** @type {RegisteredListener[]} */
+    this.registeredListeners = [];
+
     this.externalFrame = externalFrame;
     this.frame = frame;
     this.hypothesisSidebar = hypothesisSidebar;
@@ -123,7 +133,7 @@ export default class Sidebar extends Guest {
     }
 
     if (this.plugins.BucketBar) {
-      this.plugins.BucketBar.element.addEventListener('click', () =>
+      this._registerEvent(this.plugins.BucketBar.element, 'click', () =>
         this.show()
       );
     }
@@ -146,6 +156,8 @@ export default class Sidebar extends Guest {
       // not shown.
       this.toolbarWidth = 0;
     }
+
+    this._registerEvent(window, 'resize', () => this._onResize());
 
     this._gestureState = {
       // Initial position at the start of a drag/pan resize event (in pixels).
@@ -175,10 +187,28 @@ export default class Sidebar extends Guest {
   }
 
   destroy() {
+    this._unregisterEvents();
     this._hammerManager?.destroy();
     this.frame?.remove();
     this.hypothesisSidebar?.remove();
     super.destroy();
+  }
+
+  /**
+   * @param {Window|HTMLElement} eventTarget
+   * @param {string} eventType
+   * @param {(event: any) => void} listener
+   */
+  _registerEvent(eventTarget, eventType, listener) {
+    eventTarget.addEventListener(eventType, listener);
+    this.registeredListeners.push({ eventTarget, eventType, listener });
+  }
+
+  _unregisterEvents() {
+    this.registeredListeners.forEach(({ eventTarget, eventType, listener }) => {
+      eventTarget.removeEventListener(eventType, listener);
+    });
+    this.registeredListeners = [];
   }
 
   _setupSidebarEvents() {
@@ -222,11 +252,13 @@ export default class Sidebar extends Guest {
     const toggleButton = this.toolbar.sidebarToggleButton;
     if (toggleButton) {
       // Prevent any default gestures on the handle.
-      toggleButton.addEventListener('touchmove', e => e.preventDefault());
+      this._registerEvent(toggleButton, 'touchmove', e => e.preventDefault());
 
-      this._hammerManager = new Hammer.Manager(toggleButton)
-        // eslint-disable-next-line no-restricted-properties
-        .on('panstart panend panleft panright', this._onPan.bind(this));
+      this._hammerManager = new Hammer.Manager(toggleButton).on(
+        'panstart panend panleft panright',
+        /* istanbul ignore next */
+        event => this._onPan(event)
+      );
       this._hammerManager.add(
         new Hammer.Pan({ direction: Hammer.DIRECTION_HORIZONTAL })
       );
@@ -313,6 +345,19 @@ export default class Sidebar extends Guest {
       this.onLayoutChange(layoutState);
     }
     this.publish('sidebarLayoutChanged', [layoutState]);
+  }
+
+  /**
+   *  On window resize events, update the marginLeft of the sidebar by calling hide/show methods.
+   */
+  _onResize() {
+    if (this.toolbar.sidebarOpen === true) {
+      if (window.innerWidth < MIN_RESIZE) {
+        this.hide();
+      } else {
+        this.show();
+      }
+    }
   }
 
   _onPan(event) {
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -132,12 +132,6 @@ export default class Sidebar extends Guest {
       this.subscribe('panelReady', () => this.show());
     }
 
-    if (this.plugins.BucketBar) {
-      this._registerEvent(this.plugins.BucketBar.element, 'click', () =>
-        this.show()
-      );
-    }
-
     // Set up the toolbar on the left edge of the sidebar.
     const toolbarContainer = document.createElement('div');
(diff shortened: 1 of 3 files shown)


esanzgar:
Both `sidebar` and `pdf-sidebar` has events that listen for window resize events. The event in `pdf-sidebar` is triggered first because it is registered first. Having two events responding to the same events is not clean and it could lead to unusual behaviour. At the moment, it seems that the two don't interfere with each other.

src/annotator/sidebar.js
```diff
@@ -123,7 +134,7 @@ export default class Sidebar extends Guest {
     }
 
     if (this.plugins.BucketBar) {
-      this.plugins.BucketBar.element.addEventListener('click', () =>
+      this._registerEvent(this.plugins.BucketBar.element, 'click', () =>
```
lyzadanger, line 137: It's not reported here, but this line is a test miss (you can see it in local test coverage information; let me know if you need any help finding that!).

src/annotator/sidebar.js
```diff
@@ -18,8 +18,16 @@ import BucketBar from './plugin/bucket-bar';
  * @prop {number} height
  */
 
+/**
```
lyzadanger, line 21: Can you help me understand the alternate `@typdef` syntax here? I find it a little harder to read and it's formatted funky in my editor (VSCode), but perhaps there is a reason to do it like this?

image

Can we use our normal conventions if there isn't a reason to do this one differently?

src/annotator/sidebar.js
```diff
@@ -315,6 +346,19 @@ export default class Sidebar extends Guest {
     this.publish('sidebarLayoutChanged', [layoutState]);
   }
 
+  /**
+   * Minimize the sidebar on window resize event
```
lyzadanger, line 350: This comment doesn't seem to quite describe what this callback actually does: it only closes the sidebar if the resulting width is too narrow, right?

src/annotator/sidebar.js
```diff
@@ -18,8 +18,16 @@ import BucketBar from './plugin/bucket-bar';
...
+ * @typedef {{
+ *    eventTarget: Window| HTMLElement,
+ *    eventType: string,
+ *    listener: (event: any) => void;
+ * }} RegisteredListener
+ */
+
 // Minimum width to which the frame can be resized.
-const MIN_RESIZE = 280;
+export const MIN_RESIZE = 280;
```
lyzadanger, line 30: It looks like the reason for exporting this might be a test seam? If so, there might be other ways to provide the test affordance without having to export the constant. Don't tie yourself in knots if you can't find a way to do it within a few minutes...

src/annotator/test/sidebar-test.js
```diff
@@ -184,7 +184,7 @@ describe('Sidebar', () => {
 
   describe('toolbar buttons', () => {
     it('shows or hides sidebar when toolbar button is clicked', () => {
-      const sidebar = createSidebar({});
```
lyzadanger, line 187: I presume removing the empty objects here is just cleanup?

src/annotator/sidebar.js
```diff
@@ -18,8 +18,16 @@ import BucketBar from './plugin/bucket-bar';
  * @prop {number} height
  */
 
+/**
+ * @typedef {{
+ *    eventTarget: Window| HTMLElement,
```
lyzadanger, line 23: There's an extra space before `HTMLElement` here, per our current conventions.

src/annotator/test/sidebar-test.js
```diff
@@ -559,10 +559,53 @@ describe('Sidebar', () => {
...
+    const sidebar = createSidebar();
+    sidebar.show();
+    assert.isNotEmpty(sidebar.frame.style.marginLeft);
+
+    window.innerWidth = MIN_RESIZE - 1;
+    window.dispatchEvent(new Event('resize'));
+    assert.isEmpty(sidebar.frame.style.marginLeft);
+  });
+
+  it('updates marginLeft CSS style of the sidebar if sidebar is open and window width is resized > MIN_RESIZE', () => {
```
lyzadanger, line 576: I'm having trouble understanding this test description (and what's in the test). It might help to make the test description describe what the expected outcome is, not the implementation of that outcome, if that makes sense?

lyzadanger (changes requested):
The user-facing behavior of this PR seems spot-on. Well done.

I had a confusion over a test, and there is a test miss or two that could use attention.

I left a few other notes, but they are all minor.

At present, we're not debouncing `window.resize` events. (I know there's precedent from other parts of the code not to debounce). When you resize the window, an awful lot of `resize` events get dispatched. Just something to keep in mind...

src/annotator/sidebar.js
esanzgar: Thanks, I will add a test for that.

FYI, I tried to add a test for src/annotator/sidebar.js#L260 and I couldn't. It was extremely difficult to trigger hammer.js event.

src/annotator/sidebar.js
esanzgar: No problem, I will change that.

src/annotator/sidebar.js
esanzgar: Yes, you are right. These comments described a previous implementation.

src/annotator/test/sidebar-test.js
esanzgar: Yes, cleanup

src/annotator/sidebar.js
esanzgar: Fixed.

Are the spaces here OK? `(event: any) => void` Or should I remove the spaces?

src/annotator/sidebar.js
esanzgar: Yes, testing is the only reason why I import the value.
I could create a static member of the class, like this `Sidebar.MIN_SIZE`, but that would make it more difficult, if in the future, we want to refactor the class as a function component.

I would keep it as it is. If you have other suggestion, please let me know.

src/annotator/test/sidebar-test.js
esanzgar: Yes, I get your point.

Ideally, it should be something like this:

```
    it('updates marginLeft of the sidebar if window is resized', () => {
      const sidebar = createSidebar();
      sidebar.show();

      // Make the window very small
      window.innerWidth = MIN_RESIZE;
      window.dispatchEvent(new Event('resize'));
      const marginLeftInitial = sidebar.frame.style.marginLeft;

      // Make the window very large
      window.innerWidth = MIN_RESIZE * 10;
      window.dispatchEvent(new Event('resize'));
      const marginLeftFinal = sidebar.frame.style.marginLeft;

      // Compare marginLeft values
      assert.notEqual(marginLeftInitial, marginLeftFinal);
    });
```

However, it is a little bit more complicated because the media query styles are not applied when for example the screen become very small.

src/annotator/test/sidebar-test.js
```diff
@@ -98,6 +98,7 @@ describe('Sidebar', () => {
     sidebars.forEach(s => s.destroy());
     sandbox.restore();
     $imports.$restore();
+    global.viewport.reset();
```
lyzadanger, line 101: One of the things about magical things in scope (i.e. without importing directly) is that they're magical. I can imagine a future dev happening upon this line and thinking "My goodness, where did `global.viewport` come from?" It may well be a slight ignorance on my part to `karma`'s internals, but I've not seen a pattern of using `global` like this before. Is this a standard karma convention for plugins? I happen to _know_ this comes from `karma-viewport` because of the context of this PR, but I'm not sure how I would have tied the two together without that context. 

Is there any kind of hinting we can give future devs? Is there any kind of karma configuration that would allow us to be more explicit about when we're using plugins in tests (e.g. importing something or flipping some sort of config "switch" on)? 

As a quick aid, perhaps adding comments to the tests where you use this plugin would be immediately helpful.

yarn.lock
```diff
@@ -3908,6 +3923,15 @@ fs-constants@^1.0.0:
   resolved "https://registry.yarnpkg.com/fs-constants/-/fs-constants-1.0.0.tgz#6be0de9be998ce16af8afc24497b9ee9b7ccd9ad"
   integrity sha512-y6OAwoSIf7FyjMIv94u+b5rdheZEjzR63GTyZJm5qh4Bi+2YgwLCcI/fPFZkL5PSixOt6ZNKm+w+Hfp/Bciwow==
 
+fs-extra@^7.0.1:
```
lyzadanger, line 3926: `fs-extra`, really? Huh, OK... :)

lyzadanger (approved):
This feels good to go, though I still have a few reservations about `karma-viewport`...

src/annotator/test/sidebar-test.js
esanzgar: The reason why I use `global.` was because eslint complaints that `viewport` it is not defined. One way of fixing this is to add `karma-viewport` types to `tsconfig.json`, however test files are excluded, so it triggers the error in eslint anyway.

PR #2946, opened 2021-02-01
Pascal case frontend-shared components

LMS007:
_Updated_

### Commits

Pascal case frontend-shared components

------

Add /components folder in frontend-shared …

------

### Discussion

This will also serve as the first test of versioning the shared package, version should change to 1.4 to 1.5, so we will need to watch the actions to ensure that goes smoothly.

### Update

This is simplified now that it is rebased with recent changes.


frontend-shared/src/components/test/SvgIcon-test.js
```diff
@@ -15,8 +15,8 @@ describe('SvgIcon', () => {
 
     registerIcons(
       {
-        'arrow-left': require('../../images/icons/arrow-left.svg'),
-        'arrow-right': require('../../images/icons/arrow-right.svg'),
+        'arrow-left': require('../../../images/icons/arrow-left.svg'),
```
lyzadanger, line 18: Can you help me understand this change? Did something move?

LMS007: Ya, I thought we could use the /components folder here too to make it consistent with client

scripts/gulp/frontend-shared.js
```diff
@@ -15,7 +15,7 @@ const buildFrontendSharedJs = () => {
 
   return (
     gulp
-      .src('frontend-shared/src/*')
+      .src('frontend-shared/src/**/!(*-test).js')
```
lyzadanger, line 18: I was a little confused about the relationship of this to this set of changes until I read your commit message. Might I suggest breaking this out into a separate commit? This change doesn't feel like it belongs with the PascalCasing/component structure changes.

lyzadanger (commented):
Hi Kyle,

These changes overall _look_ reasonable. A couple of things:

* Can you help me understand how to test these changes locally in a way that's meaningful?
* Could you consider splitting out a separate commit for the test-compilation changes?

src/tsconfig.json
```diff
@@ -20,6 +20,6 @@
     "**/test/**/*.js",
     "test-util/**/*.js",
     "karma.config.js",
-    "../frontend-shared/src/test/**/*.js",
+    "../frontend-shared/src/**/*-test.js",
```
robertknight, line 23: We have occassionally put non-test utility modules inside test directories. Hence the other exclusion patterns exclude directories named `test` rather than files with a `-test.js` suffix. I'd suggest to do the same here, but just allow the `test` directory to be anywhere under `../frontend-shared/`.

robertknight (commented):
Aside from Lyza's comments, this looks fine to me apart from a change I suggested to the glob used to exclude test files in `frontend-shared/` from typechecking.

LMS007:
@lyzadanger , I think I still need to write up some quick docs on the fronted-shared folder that would make things a bit more clear, but in the meantime, I broke this PR up into 2 commits.

1. rename svg-icon to SvgIcon
2. move that into a new folder /components


To test.

`yarn typecheck` will be needed, but also you need to test a false positive. I would recommend this change to break both typecheck and lint _(Note: you don't have to test lint, because it's not changed in this PR)_

Screen Shot 2021-02-02 at 11 39 31 AM

Then running `yarn typecheck` should both produce errors. This is good because that means those operations are looking at that file.

Secondly, we need to make sure sure typecheck is ignoring the `SvgIcon-test.js` file. Comment out this line in `tsconfig` and then run typecheck again. You should see the terminal blow up with lots of errors in `SvgIcon-test.js` -- thats good. Add the line back and the errors should go away.

Screen Shot 2021-02-02 at 11 41 32 AM

Finally, we need to make sure babel is work right. Delete /lib in `frontend-shared` and run `gulp build-frontend-shared-js`

You should see the lib/ folder re-created and it should have `/components/SvgIcon.js` and no test files too! if that all works, then this PR **should** be solid.

scripts/gulp/frontend-shared.js
robertknight: I made a change with the same functional effect in https://github.com/hypothesis/client/pull/2951 following Kyle's feedback. However I used a negative glob in a glob array in order to exclude _all_ files in `test` directories and also because the "negative glob in a glob array" approach is documented at https://gulpjs.com/docs/en/getting-started/explaining-globs/, whereas this more advanced minmatch syntax is not.

lyzadanger:
> @lyzadanger , I think I still need to write up some quick docs on the fronted-shared folder that would make things a bit more clear, but in the meantime, I broke this PR up into 2 commits.

> rename svg-icon to SvgIcon
> move that into a new folder /components

The module renaming and move into a different location is fine to be in a single commit. The change that seemed unrelated was the test glob in `frontend-shared.js`.

lyzadanger:
As discussed in a sync meeting: I'm going to wait for this to be rebased before reviewing.

LMS007:
@lyzadanger it was originally not a single commit because of tooling work that needed to be done to fix the subdir so it works, but that was redundant with Rob's prev. PR so I'll merge these into a single commit now.

lyzadanger (approved):
OK! On my local:

* Icons show up, so that seems like a good sign
* `lib` rebuilds correctly if I nuke it and then run the associated `gulp` task
* Type-checking seems to be getting applied correctly

PR #2953, opened 2021-02-03
Set a custom :focus outline style 

LMS007:
### Commits

- Remove existing customized styling on text and input fields
- Override default outline style and instead, replace with a box-shadow that is styled similar to the default outline but with a fixed color
- This allows the outline to still retain a small border radius, and a custom color at the same time

--------------

Change markdown editor :focus styling to look more like button …

- This makes the focus outline behave with a rounded radius.

-------------

Research details
https://docs.google.com/document/d/1iO4uIxp7ECXZnZjnsXQqyxikQKxdtTuZrsbeRNcFba0/edit

fixes https://github.com/hypothesis/lms/issues/2296
fixes https://github.com/hypothesis/client/issues/2226

### Changes Propose

I decided to go with #3 box-shadow effect (see doc) because is honors border-radius and it won't muck with the border styling.  I choose a width of 2px because it was similar to safai / chrome's native styling and a color we already had in our variables.scss file. _Ideally, I would have rather just made overrides to the default outline color, but this was not possible on all browsers (details in google doc above)_

### branch
focus_2

### master
outline-current

### DIFF
--- a/frontend-shared/styles/mixins/focus.scss
+++ b/frontend-shared/styles/mixins/focus.scss
@@ -1,3 +1,25 @@
+@use '../variables' as var;
+
+/**
+ * Add stylized focus to interactive elements such <input> or <textarea>
+ *
+ * @param {boolean} $inset -
+ *   The focus style is implemented with a box-shadow and this parameter determines whether
+ *   the box-shadow should be inset on the element. Set this to true when the element
+ *   may be adjacent to another in order to prevent part of the outline from being obscured.
+ */
+@mixin outline($inset: false) {
+  &:focus {
+    outline: none;
+    box-shadow: 0 0 0 2px var.$color-focus-outline if($inset, inset, null);
+  }
+}
+
+@mixin outline--hide {
+  outline: none;
+  box-shadow: none;
+}
+
 /**
  * Display an outline on an element only when it has keyboard focus.
  *
@@ -6,18 +28,21 @@
  *
  * [1] https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
  * [2] https://github.com/WICG/focus-visible
+ *
+ * @param {boolean} $inset - Does the outline render inset or not
  */
-@mixin outline-on-keyboard-focus {
+@mixin outline-on-keyboard-focus($inset: false) {
+  @include outline($inset);
   // Selector for browsers using JS polyfill, which adds the "focus-visible"
   // class to a keyboard-focused element.
   &:focus:not(.focus-visible) {
-    outline: none;
+    @include outline--hide;
   }
 
   // Selector for browsers with native :focus-visible support.
   // (Do not combine with above, as an unsupported pseudo-class disables the
   // entire selector)
   &:focus:not(:focus-visible) {
-    outline: none;
+    @include outline--hide;
   }
 }
--- a/frontend-shared/styles/variables.scss
+++ b/frontend-shared/styles/variables.scss
@@ -1,0 +1,2 @@
+// Other colors
+$color-focus-outline: #51a7e8;
--- a/src/styles/mixins/forms.scss
+++ b/src/styles/mixins/forms.scss
@@ -16,17 +16,9 @@
   }
 }
 
-@mixin focus-outline {
-  border-color: var.$color-focus-outline;
-  box-shadow: 0px 1px 2px var.$color-shadow--base inset,
-    0px 0px 5px var.$color-focus-shadow;
-}
-
 @mixin form-input-focus {
-  outline: none;
   background-color: var.$white;
 
-  @include focus-outline;
   @include placeholder {
     color: var.$color-text--light;
   }
@@ -47,6 +39,7 @@
   }
 
   @include utils.border;
+  @include focus.outline;
   border-radius: var.$border-radius;
   color: var.$color-text--light;
   background-color: var.$grey-0;
--- a/src/styles/sidebar/components/MarkdownEditor.scss
+++ b/src/styles/sidebar/components/MarkdownEditor.scss
@@ -28,26 +28,24 @@
 }
 
 .MarkdownEditor__toolbar-button {
-  @include buttons.reset-native-btn-styles;
-  @include layout.row(center, center);
-  appearance: none;
+  @include buttons.button--icon-only($with-active-state: false);
   min-width: 24px;
   min-height: 24px;
 
-  color: var.$grey-5;
+  padding-top: 0;
+  padding-bottom: 0;
 
-  &:hover,
-  &:focus {
-    color: var.$grey-7;
-  }
+  color: var.$grey-5;
+  transition: none;
 
   &:disabled {
     color: var.$grey-3;
   }
+}
 
-  &-icon {
-    @include utils.icon--xsmall;
-  }
+svg.MarkdownEditor__toolbar-button-icon {
+  // Stronger specificity to override mixin button styling
+  @include utils.icon--xsmall;
 }
 
 .MarkdownEditor__preview {
@@ -91,9 +89,10 @@
   .MarkdownEditor__toolbar-button {
     min-width: var.$touch-target-size;
     min-height: var.$touch-target-size;
+  }
 
-    &-icon {
-      @include utils.icon--small;
-    }
+  svg.MarkdownEditor__toolbar-button-icon {
+    // Stronger specificity to override mixin button styling
+    @include utils.icon--small;
   }
 }
--- a/src/styles/sidebar/components/MenuItem.scss
+++ b/src/styles/sidebar/components/MenuItem.scss
@@ -18,7 +18,7 @@ a.MenuItem:hover {
 }
 
 .MenuItem {
-  @include focus.outline-on-keyboard-focus;
+  @include focus.outline-on-keyboard-focus($inset: true);
   @include layout.row($align: center);
   align-self: stretch;
   flex-grow: 1;
--- a/src/styles/sidebar/components/SelectionTabs.scss
+++ b/src/styles/sidebar/components/SelectionTabs.scss
@@ -33,6 +33,8 @@
   cursor: pointer;
   min-width: 85px;
   min-height: 18px;
+  // Give the tab a radius to allow :focus styling to appear similar to that of buttons
+  border-radius: var.$border-radius;
 
   user-select: none;
 
--- a/src/styles/sidebar/elements.scss
+++ b/src/styles/sidebar/elements.scss
@@ -1,9 +1,16 @@
+@use "@hypothesis/frontend-shared/styles/mixins/focus";
+
 @use "../variables" as var;
 
 // basic standard styling for elements
 a {
+  @include focus.outline-on-keyboard-focus;
+  // Ensure the tag has width in order for :focus styling to render correctly.
+  display: inline-block;
   color: var.$color-link;
   text-decoration: none;
+  // Give links a radius to allow :focus styling to appear similar to that of buttons
+  border-radius: var.$border-radius;
 }
 
 a:active,


frontend-shared/styles/mixins/focus.scss
```diff
@@ -8,16 +27,17 @@
  * [2] https://github.com/WICG/focus-visible
  */
 @mixin outline-on-keyboard-focus {
+  @include outline;
```
LMS007, line 30: Here is the major difference, we do the styling

frontend-shared/styles/mixins/focus.scss
```diff
@@ -1,3 +1,22 @@
...
+@use '../variables' as var;
+
+/**
+ * Add stylized focus to interactive elements such <input> or <textarea>
+ */
+@mixin outline {
+  @include utils.transition--fast(box-shadow);
+  &:focus {
+    outline: 0;
+    box-shadow: 0 0 0 2px var.$color-focus-outline inset;
```
LMS007, line 11: `inset` is critical because otherwise the box-shadow can get overlapped by adject elements. That is one aspect that native styling has a clear advantage. This also makes its very hard to see the 2px border radius on small elements, but it is there, just less pronounced.

frontend-shared/styles/mixins/utils.scss
```diff
@@ -0,0 +1,11 @@
+/**
+ * Fast transition effect with fixed speed and ease.
+ *
+ * @param {string[]} $properties - CSS properties, such as background-color, to apply the
+ *   transition too.
+ */
+@mixin transition--fast($properties...) {
```
LMS007, line 7: I decided to go with a rapid transition on the box-shadow so this mixin helps manage these transitions because many elements also need to apply it on background-color and color.

frontend-shared/styles/variables.scss
```diff
@@ -0,0 +1,2 @@
+// Other colors
+$color-focus-outline: #51a7e8;
```
LMS007, line 2: plucked from original variables.scss file. I think we will need a two copies of this file, ideally with no overlap.

src/styles/mixins/buttons.scss
```diff
@@ -18,6 +18,7 @@
  * And pass 'my-component-button' as the `className` prop to `Button`.
  */
 @use "@hypothesis/frontend-shared/styles/mixins/focus";
+@use "@hypothesis/frontend-shared/styles/mixins/utils" as u;
```
LMS007, line 21: collides with @use "./utils"; so I named space with `u`, probably a bad choice, but it was terse.

src/styles/mixins/forms.scss
```diff
@@ -16,17 +16,9 @@
   }
 }
 
-@mixin focus-outline {
```
LMS007, line 19: gets rid of the `<input>` and `<textarea>` custom :focus styling

src/styles/sidebar/components/AnnotationPublishControl.scss
```diff
@@ -52,6 +54,7 @@
   // Safari. See https://github.com/hypothesis/client/issues/2302
   &__menu-wrapper {
     @include layout.row;
+    @include u.transition--fast('background-color');
```
LMS007, line 57: I think this may be a long standing bug... the arrow next to the publish button does not animate on hover. It's actually unrelated to this change, but I noticed it doing these changes. I can create follow PR for it.

src/styles/sidebar/components/AnnotationUser.scss
```diff
@@ -5,4 +5,8 @@
     color: var.$color-text;
     font-weight: bold;
   }
+
+  &__link {
+    display: inline-block;
```
LMS007, line 10: This box model needs to have width in order for box-shadow :focus outline to work. Previously the native outline worked without issue -- another subtle advantage of not going with a heavy handed approach; however, I think this is the only element in our UI where I noticed this issue.

src/styles/sidebar/components/SelectionTabs.scss
```diff
@@ -33,6 +33,7 @@
   cursor: pointer;
   min-width: 85px;
   min-height: 18px;
+  border-radius: var.$border-radius;
```
LMS007, line 36: The background color is transparent, so this has no visual effect, but it does effect the :focus outline `box-shadow`, and makes it look slightly less squared. The native focus outline has a curved radius and so do our buttons so I applied the same radius to `<a>` tags which do require keyboard focus styling.

src/styles/sidebar/elements.scss
```diff
@@ -1,9 +1,14 @@
...
+
 @use "../variables" as var;
 
 // basic standard styling for elements
 a {
+  @include focus.outline-on-keyboard-focus;
   color: var.$color-link;
   text-decoration: none;
+  // Give links a radius so the :focus outline effect renders the same here as it does on a button.
+  border-radius: var.$border-radius;
```
LMS007, line 11: Same comment here as above.

src/styles/sidebar/elements.scss
```diff
@@ -1,9 +1,14 @@
+@use "@hypothesis/frontend-shared/styles/mixins/focus";
+
 @use "../variables" as var;
 
 // basic standard styling for elements
 a {
+  @include focus.outline-on-keyboard-focus;
```
LMS007, line 7: Same comment here as above.

robertknight:
Here are my initial thoughts on the functional / aesthetic aspects, I haven't looked at the CSS closely yet. I think visually the choice of styling looks good. It feels like a part-way house between Safari and Chrome's native styling and I think that's a pretty good choice for many elements. In terms of accessibility, I think we can trust that something like this will work for a majority of users. Although I had hoped a while back that we could use the browser default focus styling everywhere, the issues we've seen with Firefox and Linux and dark buttons on Chrome suggest that it might be better to do what you've done here and pick something neutral and consistent everywhere. The Firefox issue confused our third-party accessibility auditor for example.

The styling on this branch works well on buttons with a light background and inputs:

focus-light-button

focus-icon-button

focus-input

On some controls I did see a problem where it hugs the button label very tightly, making the label a bit harder to read:

focus-clipped-text

This happens with Chrome's native styling too, but not Safari's:

chrome-preview

safari-preview

Dark buttons are better in Chrome and Firefox on this branch, but I still think Safari's native styling is easier to see due to the gap between the focus outline and the inner border. On this branch:

focus-dark-button

Chrome's native styling:

chrome-dark-button

Safari:

safari-dark-button

I'm open to whether we try adding a gap between the focus ring and input like Safari or whether we have a specific adjustment for dark buttons. I think I'm leaning towards a specific adjustment for dark buttons as the effect here looks good elsewhere .

Regarding the animated transition as the focus ring appears and disappears, I honestly think it is better to remove it, for two reasons:

1. The interface just feels faster to me when navigating around via the keyboard without the transition.
2. It feels more consistent to me with the behavior of other controls when the transitions are off. We don't have written-down guidance on when to use or not use transitions, but I think the place we've arrived at naturally is that we use animations to reveal or hide content (eg. panels, submenus, the sidebar) but not for color or contrast changes, unless very subtle (eg. the toolbar buttons on hover).

robertknight:
The rounding in the outline adapts to the control, which is nice flexibility to have. I think we should probably adjust the toolbar buttons so that their outline rounding matches other buttons rather than being square:

toolbar-button-outline

src/styles/mixins/buttons.scss
lyzadanger: How about `frontend-utils` or similar? Single-letter references are a pain :)

src/styles/sidebar/components/AnnotationPublishControl.scss
lyzadanger: Why is `'background-color'` quoted here but in another usage it is like so

`@include u.transition--fast(color, background-color, box-shadow);`

src/styles/sidebar/components/AnnotationUser.scss
lyzadanger: If this change is needed to support focus outlines, can you add a comment here to indicate that?

frontend-shared/styles/mixins/utils.scss
```diff
@@ -0,0 +1,11 @@
+/**
+ * Fast transition effect with fixed speed and ease.
+ *
+ * @param {string[]} $properties - CSS properties, such as background-color, to apply the
```
lyzadanger, line 4: This might be better documented as

`@param {...string} $properties`

per https://jsdoc.app/tags-param.html#parameters-with-properties

Though I just Googled that myself so it's not like I'm any kind of expert. But in any case: these strings are not passed in an Array...

frontend-shared/styles/mixins/utils.scss
```diff
@@ -0,0 +1,11 @@
+/**
```
lyzadanger, line 1: I suppose sticking transitions in a `utils` SASS module for now is OK...we'll see how this module grows.

lyzadanger (changes requested):
This is a review of the implementation (SCSS). Rob has taken the time to play around with this branch in enough depth to have solid aesthetic feedback; I have not done so. I agree with  Rob that I'd like to see the transitions removed for now. We can easily add them, but let's keep this pared as thin as possible. I defer to Rob's review on the visible _results_ of these changes.

SCSS structure looks overall solid. I added a few comments.

LMS007:
Updated the styling to not be `inset` in the default case. I have overridden that case for menu items where they tend to be adjacent and cause part of the outline to be hidden if not `inset`

Outline masking is still apparent in several places in the UI where buttons are adjacent (most notably, the "Post to..." button, but there is little we can do about that if we don't inset the styling or rely on the native outline.

We don't have the ability to control the offset for a box-shadow (as far as I know) so for that reason, the best we can do is render the box shadow tightly around buttons and not use `inset`. This helps make it a bit more visible on dark buttons compared to `inset`, but, it's still not as good as having a little negative space -- I have some ideas that we can try later on to help fix that.

src/styles/mixins/buttons.scss
LMS007: Yes, for now, since I don't be adding the transition, we don't ned to cross that bridge, but when we do thats a reasonable name.

src/styles/sidebar/components/AnnotationPublishControl.scss
LMS007: a mistake, it should just be text w/o quote (i think)

lyzadanger:
@LMS007 

> We don't have the ability to control the offset for a box-shadow (as far as I know)

Can you clarify? You can set both an x and y offset for `box-shadow`: https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow

LMS007:
> @LMS007
> 
> > We don't have the ability to control the offset for a box-shadow (as far as I know)
> 
> Can you clarify? You can set both an x and y offset for `box-shadow`: https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow


Sure, I guess I meant offset like `outline-offset` which scales the entire outline. The box-shadow controls only moves the x and y positions.

frontend-shared/styles/mixins/focus.scss
```diff
@@ -1,3 +1,26 @@
+@use '../variables' as var;
+
+/**
+ * Add stylized focus to interactive elements such <input> or <textarea>
+ *
+ * @param {boolean} $inset - Does the outline render inset or not
```
lyzadanger, line 6: Can this comment be expanded to explain why it matters if an outline is `inset` or not? What information would a user need to make a decision one way or the other?

frontend-shared/styles/mixins/focus.scss
```diff
@@ -1,3 +1,26 @@
...
+ */
+@mixin outline($inset: false) {
+  &:focus {
+    outline: none;
+    @if $inset {
```
lyzadanger, line 11: Might check out the SASS `if` function for alternate (potentially more concise) syntax here, but it's not a big deal: https://sass-lang.com/documentation/modules#if

src/styles/sidebar/components/MarkdownEditor.scss
```diff
@@ -28,26 +28,24 @@
...
 
   &:disabled {
     color: var.$grey-3;
   }
+}
 
-  &-icon {
-    @include utils.icon--xsmall;
-  }
+svg.MarkdownEditor__toolbar-button-icon {
```
lyzadanger, line 46: I believe this rule could be nested under `.MarkdownEditor__toolbar-button` and the selector could be:

`&svg`

but don't quote me on that. Just a thought.

lyzadanger (changes requested):
Hi! I think we're just about there...my comments are surgical. If you can address them, I think this is good to go (I'm relying on Rob's initial review in terms of what this treatment actually _looks_ like—my review is on the code/implementation).

src/styles/sidebar/components/MenuItem.scss
```diff
@@ -18,7 +18,7 @@ a.MenuItem:hover {
 }
 
 .MenuItem {
-  @include focus.outline-on-keyboard-focus;
+  @include focus.outline-on-keyboard-focus(true);
```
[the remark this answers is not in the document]
LMS007: yes, and I agree with that suggestion for all mixins. much more readable.

src/styles/sidebar/components/MarkdownEditor.scss
LMS007: roger!

LMS007: I could not get that to work, but this will

svg {
   ...
}

However, that breaks our pattern of styling svg's via classname and makes MarkdownEditor__toolbar-button-icon classname irrelevant. I think we just leave this or make change to the button mixin so it won't set the icon size. (or perhaps can be overridden)

lyzadanger: Sure, leave it. NBD.

frontend-shared/styles/mixins/focus.scss
```diff
@@ -1,3 +1,25 @@
+@use '../variables' as var;
+
+/**
+ * Add stylized focus to interactive elements such <input> or <textarea>
+ *
+ * @param {boolean} $inset -
+ *   Does the outline render inset or not. The focus style is created with a box-shadow and this
+ *   parameter sets the box-shadow to be inset on the element. Set this to true  when the focusable
+ *   element may be adjacent to another and prevent part of the outline from being covered.
```
lyzadanger, line 9:
```suggestion
 *   The focus style is implemented with a box-shadow and this parameter determines whether
 *   the box-shadow should be inset on the element. Set this to true when the element
 *   may be adjacent to another in order to prevent part of the outline from being obscured.
```

Did I understand that correctly ^^ ?

frontend-shared/styles/mixins/focus.scss
```diff
@@ -1,3 +1,25 @@
...
+ */
+@mixin outline($inset: false) {
+  &:focus {
+    outline: none;
+    box-shadow: 0 0 0 2px var.$color-focus-outline if($inset, inset, null);
```
lyzadanger, line 14: Sweet! I wasn't sure what the proper combination of syntax was here before.

lyzadanger (approved):
The comment about the `$inset` parameter was a little unclear to me, but other than that, I think this is ready to go. Go ahead and land it when you're ready.

frontend-shared/styles/mixins/focus.scss
LMS007: Thats more clear, thank you!

PR #2955, opened 2021-02-03
Generate `.d.ts` files as part of the frontend-shared package

robertknight:
Generate TypeScript definition files as part of the frontend-shared package so that TS can typecheck code that uses the package in other projects. This creates `.d.ts` files alongside each of the `.js` files in `frontend-shared/lib`.

Within the client repository we don't need to generate these files because TypeScript will read the JSDoc comments. When the package is consumed from other projects however, it seems that it does not read the JSDoc comments. Note that when these `.d.ts` files exist, they are used in preference to JSDoc comments by other code in the client repository. Therefore they
need to be kept up to date when the files are recompiled in watch mode.

In the process of adding this I discovered an issue that the method of running CLI commands in gulp tasks in `scripts/gulp/frontend-shared.js` did not fail if the exit status was non-zero. I created a `run` utility that handles this correctly along with providing the desired defaults.

This PR is a prerequisite for https://github.com/hypothesis/lms/pull/2384.

### DIFF
--- a/scripts/gulp/frontend-shared.js
+++ b/scripts/gulp/frontend-shared.js
@@ -1,10 +1,12 @@
 'use strict';
 
-const cp = require('child_process');
 const fs = require('fs');
+
 const gulp = require('gulp');
 const babel = require('gulp-babel');
 const sourcemaps = require('gulp-sourcemaps');
+
+const { run } = require('./run');
 
 const buildFrontendSharedJs = () => {
   // There does not appear to be a simple way of forcing gulp-babel to use a config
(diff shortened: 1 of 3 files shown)


scripts/gulp/frontend-shared.js
```diff
@@ -25,22 +27,29 @@ const buildFrontendSharedJs = () => {
...
+    '--emitDeclarationOnly',
+    '--outDir',
+    'frontend-shared/lib',
+    'frontend-shared/src/index.js',
+  ]);
+};
+
+const linkFrontendShared = async () => {
+  // Make @hypothesis/frontend-shared available for linking in other projects.
+  await run('yarn', ['link'], { cwd: './frontend-shared' });
```
robertknight, line 45: A side effect of running `yarn link` like this is that some warnings that were previously hidden are now visible. One warning is about a missing license, which is addressed in https://github.com/hypothesis/client/pull/2954. The other is about running `yarn link` being redundant if the package is already linked, which I suggest we deal with separately.

scripts/gulp/frontend-shared.js
```diff
@@ -25,22 +27,29 @@ const buildFrontendSharedJs = () => {
...
-      env: process.env,
-      stdio: 'inherit',
-    }).on('exit', () => {
-      done();
-    });
-  });
+const buildFrontendSharedTypes = async () => {
+  // nb. If the options get significantly more complex, they should be moved to
+  // a `tsconfig.json` file.
+  await run('node_modules/.bin/tsc', [
```
LMS007, line 33: Would this be more modern or a dependency we don't want to force?

`await run('npx', [
    'tsc', ....`

LMS007 (approved):
Looks good Rob, thanks for linking this up.

scripts/gulp/frontend-shared.js
robertknight: Using `npx` might work for this purpose. It is bundled with npm >= 5.2 apparently, so we can probably rely on it. I'm not sure how much overhead it has vs. just invoking the script directly.

scripts/gulp/run.js
```diff
@@ -0,0 +1,33 @@
...
+ * or make.
+ *
+ * This function is useful for running CLI tools as part of a gulp command.
+ *
+ * @param {string} cmd - Command to run
+ * @param {string[]} args - Command arguments
+ * @param {object} options - Options to forward to `spawn`
+ * @return {Promise<void>}
+ */
+function run(cmd, args, options = {}) {
```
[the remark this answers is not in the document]
robertknight: Yes, possibly. We could put things like these in a package in the `frontend-toolkit` repo. There are numerous npm packages which do something similar, but its debatable whether it is better to just have the 10 lines of code vs the work of vetting and managing a third-party dependency.

PR #2961, opened 2021-02-05
Changes in configuration files: prettierc and babel.config.json

esanzgar:
* moved prettierc to package.json. This is a minimal configuration file with no comments.
* moved frontend-shared/babel.config.json to frontend-shared/.babelrc.cjs. This changes allows the configuration to have comments and continue to be loaded via CommonJS `require`.


LMS007 (approved):
👍

PR #2962, opened 2021-02-05
Apply applyFocusVisiblePolyfill to shadow DOM elements

### LINKED ISSUE #2689: PDF.js CSS disables focus outline for sidebar toolbar buttons
The PDF.js viewer's CSS disables focus outlines for buttons. This makes it near-impossible to tell which button in the sidebar's left toolbar has focus when using keyboard navigation:

pdf-focus

(Note the arrow above is just part of the PDF content, not something that I'm trying to point to)

There are several possible ways we can fix this:

- Override the `button { outline: none }` rule within the annotator frame CSS
- Switch the sidebar to use shadow DOM, like we do for the adder toolbar, so that it isn't affected by the page's styles

### LINKED ISSUE #2959: Focus ring missing on sidebar's vertical toolbar after moving elements into shadow root
We recently moved all of the DOM elements in the sidebar's vertical toolbar into a shadow root. An overlooked side effect of that was that buttons in the sidebar's vertical toolbar no longer show a focus ring when these buttons are focused using the keyboard.

It looks to me like this is an issue with our polyfill for the `:focus-visible` pseudo-class not working because it relies on a `js-focus-visible` class that is applied to the document body.

**Steps to reproduce:**

1. Go to https://hypothes.is/docs/help
2. Focus the toggle button for the sidebar using the keyboard.

**Expected result:**

A visible focus ring should be shown on the button.

**Actual result:**

Testing in Chrome 90 (dev), Firefox 85 and Safari 14, no focus ring is visible:

Sidebar toolbar focused

See also https://github.com/hypothesis/client/issues/2689.

esanzgar:
Follow instructions from: https://github.com/WICG/focus-visible#shadow-dom

Closes #2959,  closes #2689

### DIFF
--- a/src/annotator/components/toolbar.js
+++ b/src/annotator/components/toolbar.js
@@ -98,7 +98,7 @@ export default function Toolbar({
   useMinimalControls = false,
 }) {
   return (
-    <div>
+    <>
       {useMinimalControls && isSidebarOpen && (
         <ToolbarButton
           className="annotator-toolbar__sidebar-close"
@@ -134,7 +134,7 @@ export default function Toolbar({
           />
         </div>
       )}
-    </div>
+    </>
   );
 }
 
--- a/src/annotator/util/shadow-root.js
+++ b/src/annotator/util/shadow-root.js
@@ -47,6 +47,7 @@ export function createShadowRoot(container) {
   loadStyles(shadowRoot);
 
   // @ts-ignore The window doesn't know about the polyfill
+  // applyFocusVisiblePolyfill comes from the `focus-visible` package.
   const applyFocusVisible = window.applyFocusVisiblePolyfill;
   if (applyFocusVisible) {
     applyFocusVisible(shadowRoot);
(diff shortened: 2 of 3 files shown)


src/annotator/components/toolbar.js
```diff
@@ -98,7 +98,7 @@ export default function Toolbar({
   useMinimalControls = false,
 }) {
   return (
-    <div>
```
esanzgar, line 101: Spurious, non-semantic div

src/annotator/util/shadow-root.js
```diff
@@ -45,5 +45,12 @@ export function createShadowRoot(container) {
 
   const shadowRoot = container.attachShadow({ mode: 'open' });
   loadStyles(shadowRoot);
+
+  // @ts-ignore The window doesn't know about the polyfill
+  const applyFocusVisible = window.applyFocusVisiblePolyfill;
```
robertknight, line 50: Ideally it would be nicer to import the `applyFocusVisiblePolyfill` function from the `focus-visible` package. However, it doesn't look like the function is exported :(

Can you add a comment here about which package this function comes from.

robertknight (approved):
LGTM, with one suggestion to add a comment.

src/annotator/util/shadow-root.js
esanzgar: I agree, it would have been better to add the function directly.

I added the comment.

PR #2963, opened 2021-02-08
Post Slack deployment updates to #eng-ci

robertknight:
Post Slack messages about client deployment events to #eng-ci. Messages
with a non-success status are also posted to #eng-frontend for
visibility.

This change is intended to reduce the amount of noise in the #eng-frontend channel

Fixes https://github.com/hypothesis/playbook/issues/517


robertknight:
I verified that messages are being posted to the expected places.

Unhappy events (#eng-frontend and #eng-ci): https://hypothes-is.slack.com/archives/C1M8NH76X/p1612779655000500
Happy events (#eng-ci only): https://hypothes-is.slack.com/archives/C6GCG1Q9G/p1612779654000100

PR #2989, opened 2021-02-10
Added and unified options for the gulp test command

esanzgar:
The purpose of this PR is to be able to easily run karma tests in other
browsers without manually editing or making local copies of the karma
configuration.

Additions:

- `yarn gulp test --no-browser` to avoid launching the default browser.
  It is up to the developer to run the tests in whatever browser she/he
  chooses by navigating to http://localhost:9876/.

- `yarn gulp test --browser <browser>` run the tests in a different
  browser/s instead of the default browser. Chrome launcher comes by
  default with karma. Firefox, Safari and others can be installed
  independently.

Modifications:

- `yarn gulp test-watch` has been substituted by `yarn gulp test --watch` for consistency.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -34,17 +34,34 @@ const IMAGES_DIR = 'build/images';
 function parseCommandLine() {
   commander
     .option(
-      '--grep [pattern]',
-      'Run only tests where filename matches a pattern'
+      '--grep <pattern>',
+      'Run only tests where filename matches a regex pattern'
+    )
+    .option('--watch', 'Continuously run tests (default: false)', false)
+    .option('--browser <browser>', 'Run tests in browser of choice.')
+    .option(
+      '--no-browser',
+      "Don't launch default browser. Instead, navigate to http://localhost:9876/ to run the tests."
     )
     .parse(process.argv);
 
-  return {
-    grep: commander.opts().grep,
+  const { grep, watch, browser } = commander.opts();
+  const karmaOptions = {
+    grep: grep,
+    singleRun: !watch,
   };
+
+  // browser option can be either false | undefined | string
+  if (browser === false) {
+    karmaOptions.browsers = null;
+  } else if (browser) {
+    karmaOptions.browsers = [browser];
+  }
+
+  return karmaOptions;
 }
 
-const taskArgs = parseCommandLine();
+const karmaOptions = parseCommandLine();
 
 /** A list of all modules included in vendor bundles. */
 const vendorModules = Object.keys(vendorBundles.bundles).reduce(function (
@@ -352,13 +369,12 @@ gulp.task(
   )
 );
 
-function runKarma({ singleRun }, done) {
+function runKarma(done) {
   const karma = require('karma');
   new karma.Server(
     {
       configFile: path.resolve(__dirname, './src/karma.config.js'),
-      grep: taskArgs.grep,
-      singleRun,
+      ...karmaOptions,
     },
     done
   ).start();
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -34,13 +34,22 @@ const IMAGES_DIR = 'build/images';
...
     .option(
-      '--grep [pattern]',
+      '--grep [pattern...]',
       'Run only tests where filename matches a pattern'
     )
+    .option('--watch', 'Continuously run tests (default: false)', false)
+    .option(
+      '--browser [browsers...]',
+      'Run test with this browsers (default "ChromeHeadless_Custom")',
+      ['ChromeHeadless_Custom']
```
robertknight, line 44: Unfortunately with this option you can only specify `ChromeHeadless` or `Chrome` since those are the only launchers installed. What I was going to suggest doing instead is to have an option that doesn't launch any browser automatically but instead just launches Karma and waits for the user to load the Karma server in a browser of their choice. When calling the Karma API, this just requires setting the `browsers` argument to `null`.

I'm sure what the best name for such an option would be - `--no-launch-chrome` ("Don't launch headless Chrome to run the tests. Navigate to http://localhost:9876/ to run the tests.")?

robertknight (commented):
Hi @esanzgar - I added a couple of thoughts about supplying multiple patterns to grep and a possible alternative to the `--browser` option which will enable running tests in other browsers without installing additional packages. The `--watch` change I'm fine with.

gulpfile.js
```diff
@@ -34,13 +34,22 @@ const IMAGES_DIR = 'build/images';
 function parseCommandLine() {
   commander
     .option(
-      '--grep [pattern]',
+      '--grep [pattern...]',
       'Run only tests where filename matches a pattern'
```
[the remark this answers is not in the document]
robertknight: Related to this, one annoying issue at the moment is that this help documentation only exists in the code. If you run `gulp test --help` you get Gulp's help output, not the output for the specific command. It would be nice to fix that somehow.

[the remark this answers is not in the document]
esanzgar: I have changed the help text as you suggested.

I will remove the options to allow multiple `--grep` options.

Regarding adding our own help documentation:
- it is been discussed in the gulp project: https://github.com/gulpjs/gulp/issues/2383
- another alternative: https://www.npmjs.com/package/gulp-display-help

gulpfile.js
esanzgar: I think it is a very good idea to allow developers to run the test in whatever browser they wish. I added the option `--no-browser` for that.

esanzgar:
@robertknight, thank you for reviewing my PR and your comments.

* I removed the optional `--grep` options
* I added `--no-browser` option
* I kept `--browser` option, just for convenience

gulpfile.js
```diff
@@ -34,17 +34,36 @@ const IMAGES_DIR = 'build/images';
...
+    grep: grep,
+    singleRun: !watch,
   };
+
+  // browser option can be either undefined | false | string
+  if (browser === false) {
+    karmaOptions.browsers = null;
+  }
+
+  if (browser) {
```
robertknight, line 59: I would suggest to combine this with the above `if` as:

```js
if (browser === false) {
  ...
} else if (browser) {
  ...
}
```

To make it a little more immediately obvious that these are mutually exclusive options.

robertknight (approved):
LGTM. A useful improvement to our dev tools 👍

gulpfile.js
esanzgar: Yes, thanks.

Issue #2930, opened 2021-01-26
Check process.env.NODE_ENV when bundling frontend-shared


LMS007:
Currently `yarn setup-frontend-shared` outputs sourcemap files, but does not make sense for a release. Possibly check `process.env.NODE_ENV === 'production'` and then don't output source maps or have a different code path for doing a bundle for an npm release

LMS007:
Closing this out, sourcemaps still have usefulness in prod for Sentry.

Issue #2984, opened 2021-02-09
Sourcemaps for frontend-shared js files don't work


LMS007:
When looking at a source file in the debugger, it appears its looking at the transpiled code, rather than the original. Observe `SvgIcon.js` in the debugger and you won't see the original source, but rather a much more difficult to follow transpiled version.

`SvgIcon.js.map` does exist but it appears Chrome / Safari are not using that sourcemap but rather the one found inside `sidebar.bundle.js.map`. If you search for "SvgIcon.js.map" inside `sidebar.bundle.js.map`, you'll see there is indeed a sourcemap but it does appear to be different than the original file `SvgIcon.js.map` in the `frontend-shared/lib` folder. 
Looking closely, you can see that the string "nObject.defineProperty" (a transpiled artifact) appear in the `sourcesContent` string related to SvgIcon.js.map. You won't find that is the original map file in `frontend-shared/lib/`. I belive what is happening is that a secondary map file is being generated by the bundle process and its assuming that the transpiled /lib/component/SvgIcon.js file is the source of truth, but it needs to be /src/components/SvgIcon

PR #2992, opened 2021-02-10
Omit sourcemaps for production

LMS007:
- *.js.map files are no longer generated for production releases.
- Sourcemaps still don't fully work in the browser's debugger -- an existing issue that needs to be fixes separately

-----------

fixes https://github.com/hypothesis/client/issues/2930

relates to https://github.com/hypothesis/client/issues/2984

### DIFF
--- a/scripts/gulp/frontend-shared.js
+++ b/scripts/gulp/frontend-shared.js
@@ -6,21 +6,28 @@ const sourcemaps = require('gulp-sourcemaps');
 
 const { run } = require('./run');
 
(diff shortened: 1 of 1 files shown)


robertknight (commented):
Hi Kyle - Sourcemaps can still be useful in production. We include them in the generated package for the client. Sentry for example can use them to map frames in crash reports back to the original source files.

LMS007:
> Hi Kyle - Sourcemaps can still be useful in production. We include them in the generated package for the client. Sentry for example can use them to map frames in crash reports back to the original source files.

awe..okay, I'll close this out

PR #2993, opened 2021-02-11
Clean-up after running test

esanzgar:
Remove a few container elements left-over from testing the `sidebar`.

### DIFF
--- a/src/annotator/test/sidebar-test.js
+++ b/src/annotator/test/sidebar-test.js
@@ -13,9 +13,9 @@ describe('Sidebar', () => {
   let fakeCrossFrame;
   const sidebarConfig = { pluginClasses: {} };
 
-  // `Sidebar` instances created by current test.
+  // Containers and Sidebar instances created by current test.
+  let containers;
   let sidebars;
-  let sidebarContainer;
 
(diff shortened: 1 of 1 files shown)


src/annotator/test/sidebar-test.js
```diff
@@ -48,10 +48,11 @@ describe('Sidebar', () => {
   };
 
   const createExternalContainer = () => {
-    const externalFrame = document.createElement('div');
+    externalFrame = document.createElement('div');
```
robertknight, line 51: This assumes that the test only calls `createExternalContainer` once. That might be a reasonable assumption - but it would be worth adding a check to ensure that this assumption isn't violated. Alternatively, you could record all the elements added to the DOM in tests in an array and then remove them all in the cleanup.

robertknight (approved):
This looks good, although I did raise a possible issue if a test ever calls `createExternalContainer` multiple times.

src/annotator/test/sidebar-test.js
esanzgar: I went for your second suggestions because the array pattern was already in the test for `sidebars`. I added `containers`

PR #2996, opened 2021-02-11
Improve focus style for adjacent buttons

LMS007:
This avoids using $inset for adjacent buttons which is ideal. I did not see any side effects so far of setting z-index to 1, but it may require a bit more testing. Note that this is more of an issue in the LMS grader where we are either going to have to use $inset or this trick. I propose this is a better solution and we should only use inset in the  menu where the z-index trick fails because the items are not direct sibling.

**Before**
before

**After**
after


esanzgar (approved):
Not a bulletproof solution, but for common scenarios it will work.

Issue #2979, opened 2021-02-09
SPIKE: assess the possibility of not injecting `annotator.css` to the host page


esanzgar:
Now that we have shadow-DOMed the sidebar, we could experiment with the possibility of not injecting `annotator.css` to the host page.

Pros:

- Host page's look and feel won't be disrupted by hypothesis's CSS
- Less backdoors to change certain aspects of the application. For example, https://github.com/hypothesis/client/issues/1853 (https://elifesciences.org/for-the-press/81d42f7d/elife-enhances-open-annotation-with-hypothesis-to-promote-scientific-discussion-online).

Cons:

- Annotation highlight must be shadow DOMed.

robertknight:
Currently we have the following elements in the host page:

- Sidebar container (`<hypothesis-sidebar>`)
- Adder toolbar (`<hypothesis-adder>`)
- Notebook container (`<hypothesis-notebook>`)
- Highlights (`<hypothesis-highlight>`)

Following https://github.com/hypothesis/client/pull/2997, all of these except highlights will be in shadow DOM in the vast majority of browsers. What we could do is split the annotator.css bundle into a tiny bundle that contains just the highlight styles and no global CSS, which is loaded in the host page, and the rest of the annotator CSS which is only loaded inside shadow roots, assuming that the browser supports shadow DOM.

PR #2997, opened 2021-02-12
Move notebook container into shadow DOM

robertknight:
Follow the example of the `<hypothesis-sidebar>` element by making the `<hypothesis-notebook>` element an unstyled shadow-host which serves as the root container for the notebook, isolating it from the page's styles. Inside this is a styled element which serves as the visual container.

As well as isolating the notebook from the host page's styles, this will also make it possible to avoid loading annotator styles into the host page. See https://github.com/hypothesis/client/issues/2979.

To reduce resource usage a little when the notebook is not used, only the notebook's shadow host is created initially. The styled inner container is created when the notebook is shown for the first time.

### DIFF
--- a/src/annotator/notebook.js
+++ b/src/annotator/notebook.js
@@ -1,5 +1,6 @@
 import Delegator from './delegator';
 import { createSidebarConfig } from './config/sidebar';
+import { createShadowRoot } from './util/shadow-root';
 
 /**
  * Create the iframe that will load the notebook application.
@@ -34,9 +35,21 @@ export default class Notebook extends Delegator {
     this._groupId = null;
     this._prevGroupId = null;
 
-    this.container = document.createElement('hypothesis-notebook');
-    this.container.style.display = 'none';
-    this.container.className = 'notebook-outer';
+    /**
+     * Un-styled shadow host for the notebook content.
+     *
+     * This isolates the notebook from the page's styles.
+     */
+    this._outerContainer = document.createElement('hypothesis-notebook');
+    this.element.appendChild(this._outerContainer);
+
+    /**
+     * Lazily-initialized container for the notebook iframe. This is only created
+     * when the notebook is actually used.
+     *
+     * @type {HTMLElement|null}
+     */
+    this.container = null;
 
     this.subscribe('showNotebook', groupId => {
       this._groupId = groupId;
@@ -48,6 +61,8 @@ export default class Notebook extends Delegator {
   }
 
   _update() {
+    const container = this._initContainer();
+
     // Create a new iFrame if we don't have one at all yet, or if the
     // groupId has changed since last use
     const needIframe =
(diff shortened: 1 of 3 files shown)


src/styles/annotator/notebook.scss
```diff
@@ -2,7 +2,6 @@
 @use '../mixins/molecules';
 
 .notebook-outer {
-  // TODO: CSS namespace conflicts?
```
robertknight, line 5: I believe this comment referred to conflicts with host page styles. This is no longer something we need to worry about with annotator CSS, assuming that the styles are applied to an element in shadow DOM. For the small percentage of browsers that can load Hypothesis but don't support shadow DOM (mainly legacy Edge and possibly a few users on old versions of Firefox), we just accept a small risk of style conflicts.

src/annotator/notebook.js
```diff
@@ -57,23 +72,39 @@ export default class Notebook extends Delegator {
...
+      this.container.style.display = 'none';
+    }
   }
 
   destroy() {
-    this.frame?.remove();
+    this._outerContainer.remove();
+  }
+
+  _initContainer() {
```
robertknight, line 97: To be truthful I'm not quite sure how much resource usage is saved by lazily creating this container. The main possible benefit is not loading an extra copy of the annotator CSS inside the shadow root. I believe that some browsers are able to share parsed stylesheets between different shadow DOM elements. If we later learn that the stylesheet is effectively shared in most browsers, we might want to simplify this code a little by removing the lazy initialization.

src/annotator/notebook.js
```diff
@@ -57,23 +72,39 @@ export default class Notebook extends Delegator {
...
+    container.style.display = '';
   }
 
   hide() {
-    this.container.classList.remove('is-open');
-    this.container.style.display = 'none';
+    if (this.container) {
+      this.container.classList.remove('is-open');
+      this.container.style.display = 'none';
+    }
```
esanzgar, line 90:
```suggestion
    if (!this.container) {
      return;
    }
    this.container.classList.remove('is-open');
    this.container.style.display = 'none';
```

src/annotator/notebook.js
esanzgar: I agree.

esanzgar (approved):
One observation: on browsers that don't cache parsed CSS, the lazy loading approach will result on displaying initially the component unstylised while parsing the CSS, and then it will update the style.

A future improvement would be to add one single shadow DOM at the root of the application. That's the ideal, but it may take a while to get there. Therefore, right now this is a good approach.

lyzadanger (approved):
Thanks for the ping! I have read and understood these changes and defer to @esanzgar 's approval.

robertknight:
> A future improvement would be to add one single shadow DOM at the root of the application. That's the ideal, but it may take a while to get there. Therefore, right now this is a good approach.

It may not be possible to put all of the elements that we render under a single root element. There may be cases where we need to position an element somewhere else within the tree.

One way I would suggest we can solve this specific problem if we stop loading annotator.css in the root document would be to preload it in the boot script with `<link rel="preload">`. This will make it extremely likely that it has been fetched by the time it is used within a shadow root somewhere.

PR #2998, opened 2021-02-12
Debounce on window.resize event

esanzgar:
Based on @lyzadanger [suggestion](https://github.com/hypothesis/client/pull/2937#pullrequestreview-585794221) I added debouncing to the listener associated to the `window.resize` event.

Although I am waiting only 10ms to fire the event, it feels sluggish.

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -1,4 +1,5 @@
 import Hammer from 'hammerjs';
+import debounce from 'lodash.debounce';
 
 import annotationCounts from './annotation-counts';
 import sidebarTrigger from './sidebar-trigger';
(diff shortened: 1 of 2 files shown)


esanzgar:
We have tried to overcome the sluggish behaviour by playing with the option debounce option `maxWait`. However the amount of calls to the listener was almost identical to the version without debouncing.

https://hypothes-is.slack.com/archives/C1M8NH76X/p1613128895046500

esanzgar:
We have tried to overcome the sluggish behaviour by experimenting with the debounce option `maxWait`. We made the behaviour snappier but the amount of call to the listener was almost identical to the version without debouncing.

https://hypothes-is.slack.com/archives/C1M8NH76X/p1613128895046500

PR #2999, opened 2021-02-12
Switch to using `<link>` elements for stylesheets in shadow DOM

robertknight:
When Shadow DOM was first considered for use in the client `<link rel="stylesheet">` was not widely supported as a way to load external styles into shadow roots. This was resolved in the spec in https://github.com/whatwg/html/pull/1572
and has now long been supported by all browsers.

Using `<link>` simplifies the code and also avoids a possible problem with sites that have a strict CSP policy that disallows inline styles. In the browser extension context a `<link>` can still work in that scenario as long as the stylesheet is loaded from a `chrome-extension://` URL.

### DIFF
--- a/src/annotator/util/shadow-root.js
+++ b/src/annotator/util/shadow-root.js
@@ -10,20 +10,10 @@ function loadStyles(shadowRoot) {
     return;
   }
 
-  // Stylesheet <link> elements are inert inside shadow roots [1]. Until
-  // Shadow DOM implementations support external stylesheets [2], grab the
-  // relevant CSS files from the current page and `@import` them.
-  //
-  // [1] http://stackoverflow.com/questions/27746590
-  // [2] https://github.com/w3c/webcomponents/issues/530
-  //
-  // This will unfortunately break if the page blocks inline stylesheets via
-  // CSP, but that appears to be rare and if this happens, the user will still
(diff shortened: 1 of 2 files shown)


esanzgar (commented):
~~In production, as far as I understand, the style is served from hypothesis' CDN (even for the browser-extension). Probably, we should make sure this works in the browser-extension context.~~

On a follow up note, I would suggest an alternative way to find the annotator CSS href. Could we device a way that the href for the annotator.css could we replaced in the code once the SCSS is compiled and minimised?

esanzgar (approved):
I realised it is safe to apply this in the context of browser-extension (ignore my previous comment).

robertknight:
> On a follow up note, I would suggest an alternative way to find the annotator CSS href. 

We could follow the example of the way that the `src` attribute is discovered for the sidebar and notebook iframes. Basically a `<link>` tag is rendered into the page with special attribute to identify it, and other code looks for this. Alternatively we could perhaps render a JSON `<script>` tag into the page which other code can then use. The reason for writing the data into DOM elements, rather than using a pure JS solution such as global variables is because in the browser extension context the boot script runs in a separate JS "world" to the rest of the client code. These worlds share the same DOM, but not the same JS globals.

PR #3024, opened 2021-02-16
Added close button to the notebook

esanzgar:
This PR improves the UX for opening and closing of the notebook by:

1. Hiding the controls of the sidebar when the notebook is opened
2. Adding a button in the notebook (but not in the iframe, notebook main app) to close the notebook. 

The PR contains 2 commits:

- first commit is non-esential and can be dropped:
   1. It renames all the `hide/show` methods in the sidebar and notebook to `open/close`.
   2. For consistency, it renames associated events.
   
- second commit is the essential one:
  1. added the main features described at the top

### DIFF
--- a/src/annotator/notebook.js
+++ b/src/annotator/notebook.js
@@ -1,6 +1,7 @@
 import Delegator from './delegator';
 import { createSidebarConfig } from './config/sidebar';
 import { createShadowRoot } from './util/shadow-root';
+import { render } from 'preact';
 
 /**
  * Create the iframe that will load the notebook application.
@@ -51,13 +52,13 @@ export default class Notebook extends Delegator {
      */
     this.container = null;
 
-    this.subscribe('showNotebook', groupId => {
+    this.subscribe('openNotebook', groupId => {
       this._groupId = groupId;
-      this.show();
+      this.open();
     });
-    this.subscribe('hideNotebook', () => this.hide());
+    this.subscribe('closeNotebook', () => this.close());
     // If the sidebar has opened, get out of the way
-    this.subscribe('sidebarOpened', () => this.hide());
+    this.subscribe('sidebarOpened', () => this.close());
   }
 
   _update() {
@@ -76,14 +77,14 @@ export default class Notebook extends Delegator {
     }
   }
 
-  show() {
+  open() {
     const container = this._initContainer();
     this._update();
     container.classList.add('is-open');
     container.style.display = '';
   }
 
-  hide() {
+  close() {
     if (this.container) {
       this.container.classList.remove('is-open');
       this.container.style.display = 'none';
--- a/src/annotator/notebook.js
+++ b/src/annotator/notebook.js
@@ -3,6 +3,9 @@ import { createSidebarConfig } from './config/sidebar';
 import { createShadowRoot } from './util/shadow-root';
 import { render } from 'preact';
 
+// FIXME: use the button from the frontend shared package once this is stable.
+import Button from '../sidebar/components/Button';
+
 /**
  * Create the iframe that will load the notebook application.
  *
@@ -33,7 +36,9 @@ export default class Notebook extends Delegator {
     super(element, config);
     this.frame = null;
 
+    /** @type {null|string} */
     this._groupId = null;
+    /** @type {null|string} */
     this._prevGroupId = null;
 
     /**
@@ -56,8 +61,7 @@ export default class Notebook extends Delegator {
       this._groupId = groupId;
       this.open();
     });
-    this.subscribe('closeNotebook', () => this.close());
-    // If the sidebar has opened, get out of the way
+    // Defensive programming: if the sidebar has opened, get out of the way
     this.subscribe('sidebarOpened', () => this.close());
   }
 
@@ -105,6 +109,25 @@ export default class Notebook extends Delegator {
     this.container.style.display = 'none';
     this.container.className = 'notebook-outer';
     shadowRoot.appendChild(this.container);
+
+    render(
+      <div className="Notebook__controller-bar">
+        <Button
+          icon="cancel"
+          className="Notebook__close-button"
+          buttonText="Close"
+          title="Close the Notebook"
+          onClick={event => {
+            // Guest 'component' captures all click events in the host page and opens the sidebar.
+            // We stop the propagation of the event to prevent the sidebar to be opened.
+            event.stopPropagation();
+            this.close();
+            this.publish('closeNotebook');
+          }}
+        />
+      </div>,
+      this.container
+    );
 
     return this.container;
   }
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -129,14 +129,14 @@ export default class Sidebar extends Guest {
       config.query ||
       config.group
     ) {
-      this.subscribe('panelReady', () => this.show());
+      this.subscribe('panelReady', () => this.open());
     }
 
     // Set up the toolbar on the left edge of the sidebar.
     const toolbarContainer = document.createElement('div');
     this.toolbar = new ToolbarController(toolbarContainer, {
       createAnnotation: () => this.createAnnotation(),
-      setSidebarOpen: open => (open ? this.show() : this.hide()),
+      setSidebarOpen: open => (open ? this.open() : this.close()),
       setHighlightsVisible: show => this.setAllVisibleHighlights(show),
     });
     this.toolbar.useMinimalControls = config.theme === 'clean';
@@ -161,7 +161,7 @@ export default class Sidebar extends Guest {
       final: /** @type {number|null} */ (null),
     };
     this._setupGestures();
-    this.hide();
+    this.close();
 
     // Publisher-provided callback functions
     const [serviceConfig] = config.services || [];
@@ -207,21 +207,21 @@ export default class Sidebar extends Guest {
 
   _setupSidebarEvents() {
     annotationCounts(document.body, this.crossframe);
-    sidebarTrigger(document.body, () => this.show());
+    sidebarTrigger(document.body, () => this.open());
     features.init(this.crossframe);
 
-    this.crossframe.on('showSidebar', () => this.show());
-    this.crossframe.on('hideSidebar', () => this.hide());
+    this.crossframe.on('openSidebar', () => this.open());
+    this.crossframe.on('closeSidebar', () => this.close());
 
     // Re-publish the crossframe event so that anything extending Delegator
     // can subscribe to it (without need for crossframe)
-    this.crossframe.on('showNotebook', groupId => {
-      this.hide();
-      this.publish('showNotebook', [groupId]);
+    this.crossframe.on('openNotebook', groupId => {
+      this.close();
+      this.publish('openNotebook', [groupId]);
     });
-    this.crossframe.on('hideNotebook', () => {
-      this.show();
-      this.publish('hideNotebook');
+    this.crossframe.on('closeNotebook', () => {
+      this.open();
+      this.publish('closeNotebook');
     });
 
     const eventHandlers = [
@@ -347,9 +347,9 @@ export default class Sidebar extends Guest {
   _onResize() {
     if (this.toolbar.sidebarOpen === true) {
       if (window.innerWidth < MIN_RESIZE) {
-        this.hide();
+        this.close();
       } else {
-        this.show();
+        this.open();
       }
     }
   }
@@ -386,9 +386,9 @@ export default class Sidebar extends Guest {
           this._gestureState.final === null ||
           this._gestureState.final <= -MIN_RESIZE
         ) {
-          this.show();
+          this.open();
         } else {
-          this.hide();
+          this.close();
         }
         this._resetGestureState();
         break;
@@ -407,7 +407,7 @@ export default class Sidebar extends Guest {
     }
   }
 
-  show() {
+  open() {
     this.crossframe.call('sidebarOpened');
     this.publish('sidebarOpened');
 
@@ -426,7 +426,7 @@ export default class Sidebar extends Guest {
     this._notifyOfLayoutChange(true);
   }
 
-  hide() {
+  close() {
     if (this.frame) {
       this.frame.style.marginLeft = '';
       this.frame.classList.add('annotator-collapsed');
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -215,13 +215,12 @@ export default class Sidebar extends Guest {
 
     // Re-publish the crossframe event so that anything extending Delegator
     // can subscribe to it (without need for crossframe)
-    this.crossframe.on('openNotebook', groupId => {
-      this.close();
+    this.crossframe.on('openNotebook', (/** @type {string} */ groupId) => {
+      this.hide();
       this.publish('openNotebook', [groupId]);
     });
-    this.crossframe.on('closeNotebook', () => {
-      this.open();
-      this.publish('closeNotebook');
+    this.subscribe('closeNotebook', () => {
+      this.show();
     });
 
     const eventHandlers = [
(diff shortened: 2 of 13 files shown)


lyzadanger (changes requested):
I think the overall behavior is solid, which I'd summarize as:

* When opening the notebook, entirely hide the sidebar;
* Provide a close button to close the notebook;
* Clicking the close button hides the notebook and re-opens the sidebar;

There are two Notebook contexts I could imagine coming into play in the future:

* The Notebook is opened in its own tab. This doesn't seem problematic with the current implementation, as your approach renders the close control outside of the Notebook itself.
* The Notebook is opened from within a context that includes the Sidebar, but the Sidebar isn't open/active when the Notebook is opened. Right now, the Notebook is opened from a control that is only available when the Sidebar is open. However, it's possible there might be a control to open the Notebook from, say, the Bucket Bar.
    
    In that case, would we want to force-open the Sidebar on Notebook close, or is it more appropriate to "restore" the Sidebar to whatever state it was in before the Notebook was open (ensuring it is not _entirely_ hidden) ?

Is there any case that could lead to both the Notebook and the Sidebar being entirely hidden? Or, put another way, is the combination of "Notebook:hidden and Sidebar:hidden" valid? If it is, can we make it invalid/prevent that?

I see that you used the `Button` component to render the close button, however, styles are not being applied for me (I'm in Chrome 88 on MacOS).

It may also be beyond the scope of this set of changes (which is fine), but the scrolling situation can be confusing as it's possible to scroll both the Notebook and the underlying page (_this predates these changes!_). We can ignore this for now if it's not intended to be addressed in these changes.

Here's an animated GIF that shows the lack of styles (including the focus ring and lack of pointer cursor) and the scrolling confusing in one go:

notebook-close-button

For reference, the button should look like:

image

Given that the button will add vertical height, I think we should also tighten the top margin/padding/whatever for the Notebook view in both wide and narrow viewports, that is, less blank space at top:

image

Note: I have not yet done a full technical dive on implementation but from a quick scan nothing jumps out at me. I wanted to surface these conceptual things first to keep you unblocked.

esanzgar:
Thanks for your review, @lyzadanger.

> In that case, would we want to force-open the Sidebar on Notebook close, or is it more appropriate to "restore" the Sidebar to whatever state it was in before the Notebook was open (ensuring it is not entirely hidden) ?

I agree it makes sense to restore the sidebar to the prior UI state (just before opening the notebook). This is what it does currently, but only because there is only one way to open the notebook.

I will make sure the UI state can be easily restored by making the sidebar component orchestrates the process of opening the notebook and decoupling the closing/opening the sidebar from the hiding/showing the sidebar's controls.

> I see that you used the Button component to render the close button, however, styles are not being applied for me (I'm in Chrome 88 on MacOS).

I will take a look at this as well as the scrolling issue.

esanzgar:
I discovered that the issue with the scrolling is displayed on smaller viewports. I suggest we assess that in connection to https://github.com/hypothesis/client/issues/2730.

src/annotator/notebook.js
```diff
@@ -105,6 +108,22 @@ export default class Notebook extends Delegator {
     this.container.className = 'notebook-outer';
     shadowRoot.appendChild(this.container);
 
+    render(
+      <div className="Notebook__controller-bar">
+        <Button
```
lyzadanger, line 113: This `Button` might benefit from using an extra prop; I think it might be `title`, that can allow a more verbose `aria-label` to be set. In this case, I might suggest setting `title` (double-check that prop name, please) to `Close the Notebook`.

src/annotator/sidebar.js
```diff
@@ -449,4 +450,22 @@ export default class Sidebar extends Guest {
   setAllVisibleHighlights(shouldShowHighlights) {
     this.crossframe.call('setVisibleHighlights', shouldShowHighlights);
   }
+
+  /**
+   * Shows the sidebar's controls
+   */
+  showControls() {
```
lyzadanger, line 457: This and `hideControls` are well-named 👍🏻 

I'm having trouble drawing a table of possible states in my head. Does anything prevent controls from being hidden when the sidebar is open?

src/styles/annotator/notebook.scss
```diff
@@ -1,16 +1,16 @@
 @use '../variables' as var;
 @use '../mixins/molecules';
+@use "../mixins/buttons";
 
 .notebook-outer {
   box-sizing: border-box;
   position: fixed;
+  z-index: 2147483647;
```
lyzadanger, line 8: This does fix the PDF viewer toolbar interference (though I do get twitchy when I see CSS trying to out-z-index things). Can you add a comment here as to why this `z-index` rule was added?

src/styles/annotator/notebook.scss
```diff
@@ -24,4 +24,23 @@
   padding: 0;
   width: 100%;
   height: 100%;
+  border: none;
+}
+
+.Notebook__controller-bar {
```
lyzadanger, line 30: A comment or two here about what the intended purpuse/layout is for this element would be helpful for the future.

src/styles/annotator/notebook.scss
```diff
@@ -24,4 +24,23 @@
...
+  position: absolute;
+  width: calc(100vw - #{var.$layout-space * 2});
+  display: flex;
+  justify-content: flex-end;
+  background-color: var.$grey-2;
+}
+
+.Notebook__close-button {
+  position: absolute;
+  @include buttons.button--labeled(
```
lyzadanger, line 40: Ah! I presume this is to adapt for not being on white background? 👍🏻 I will capture this, as I'd like to bake this kind of adaptation into the new `Button*` work.

lyzadanger (changes requested):
This is looking good! Some minor comments in the code.

One thing I'd like to do is add some margin around the close button. Perhaps `0.5em`? It's currently flush against the edges of the Notebook:

Screen Shot 2021-02-19 at 7 54 26 AM

Aesthetically I think it's fine, but it feels out of pattern.

src/annotator/notebook.js
```diff
@@ -1,6 +1,8 @@
 import Delegator from './delegator';
 import { createSidebarConfig } from './config/sidebar';
 import { createShadowRoot } from './util/shadow-root';
+import { render } from 'preact';
+import Button from '../sidebar/components/Button';
```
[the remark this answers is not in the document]
esanzgar: I'll do.

src/annotator/notebook.js
esanzgar: You are right, it is `title`. I added that.

src/styles/annotator/notebook.scss
esanzgar: Yes, I agree it is not a very clean solution. The adder has a mechanism to compute the zIndex of the 'neighbouring' elements so it is not a hard coded value. Would you like me to investigate using the same mechanism here?

I am using the same zIndex that the sidebar uses. Do you know if there has been issues with sidebar not been visible or partially hidden because the zIndex was hard coded (and no sufficiently high)?

I added the comment you suggested.

src/annotator/sidebar.js
esanzgar: Yesterday, I realised there is an alternative way to hide the controls in the sidebar that is used when the `externalContainer` option is used. It maybe better to investigate to reuse that options so we don't end up with two alternative ways to achieve the same thing.

src/styles/annotator/notebook.scss
esanzgar: Correct.

esanzgar:
I wonder if an `X` symbol would be enough? Or is not ideal for accessibility reasons?

image

lyzadanger:
> I wonder if an X symbol would be enough? Or is not ideal for accessibility reasons?

An "X" with appropriate button attributes could be accessible. At narrow-viewport scale, I think an "X" alone could be nice, provided its tap target is ample. For wider screens like you've posted in your screenshot, I feel like `X Close` or similar is easier to spot—the "X" alone feels itty-bitty.

What do you think about: going with the "standard" `X Close` treatment for the moment, and revisiting this for tuning after we have a little more sanity and flexibility with our `Button` components?

src/styles/annotator/notebook.scss
esanzgar: Added.

src/annotator/sidebar.js
esanzgar: I tried to reuse the previous mechanism, using `this.toolbarWidth`, but it serves a different purpose and it is not reusable.

esanzgar: >Does anything prevent controls from being hidden when the sidebar is open?

I can't think on a state where the sidebar is open but the controls are hidden. The logic to show/hide controls and open/close the sidebar are only in two functions and triggered by two events `open/closeNotebook`.

On  `openNotebook` the sidebars (1) hides its controls and (2) closes the bar. These two actions always should happen together.

On `closeNotebook` event, the sidebar (1) shows its controls and (2) opens the bar. (1) must always happen. (2) maybe optional (in the future we may want to restore the UI state [previous to the opening of the notebook] if we device a different way to open the notebook that doesn't require the sidebar to be opened).

lyzadanger (changes requested):
Hi! I think The styling for the button and its containing element could be simplified and made more resilient; I've left some suggestions. The result also gives a more "standard" `0.5em` margin around the close button (vs. the tighter `0.25em` here).

We still need to get to the bottom of the `Close` button being "sticky" on narrow-screen viewports (but, oddly, not on wider ones from what I can tell). 

EDITED: There was a big, associated comment that explains a possible strategy for the styling and it appears to have been _completely eaten_ by GitHub. It took me a long while to write, so I'll try to figure out how to...communicate my intent once I've stopped flipping tables.

src/styles/annotator/notebook.scss
```diff
@@ -24,4 +27,29 @@
   padding: 0;
   width: 100%;
   height: 100%;
+  border: none;
```
lyzadanger, line 30: Can you help me understand this rule? Was a border showing up? That is, is this rule necessary?

src/styles/annotator/notebook.scss
```diff
@@ -24,4 +27,29 @@
...
+  );
+  padding: var.$layout-space--xsmall;
+  font-size: var.$font-size--large;
+  margin-top: var.$layout-space--xxsmall;
+  margin-right: var.$layout-space--xxsmall;
+
+  // Center the button if icon-only
+  // .SvgIcon > svg {
+  // margin: 0;
+  // }
```
lyzadanger, line 54: If I'm understanding the intent here correctly, I'm guessing (but just guessing, could be wrong) that the `position:absolute` rule is intended to give the `.Notebook__controller-bar` no height? This causes a few weird side effects (and is a little hard to understand). Instead, might I suggest simplifying the rules here (and, while we're at it, handling spacing _around_ the button and font-sizing in the parent element instead of overriding button styles?

```suggestion
// This container element has the purpose of pushing children to the bottom-right end.
.Notebook__controller-bar {
  font-size: var.$font-size--large;
  display: flex;
  justify-content: flex-end;
  background-color: var.$grey-2;
  padding: 0.5em;
}

.Notebook__close-button {
  @include buttons.button--labeled(
    $background-color: var.$grey-2,
    $active-background-color: var.$grey-3
  );
```

Then, you can tighten up the vertical dead space on the top of the view by editing styles in `.HypothesisApp.scss`, something like this:

```
  &--notebook {
    padding: 0 var.$layout-space;
    @include responsive.respond-to(tablets desktops) {
      padding: 0 var.$layout-space--xlarge;
    }
  }
}
```

(In this case, zeroing out the top/bottom padding).

After, narrow screen:

image

After, wider screen:

image

lyzadanger (changes requested):
Posting another review in the hopes that GitHub doesn't eat the pertinent comment this time about styling adjustments.

src/styles/annotator/notebook.scss
esanzgar: There was a 1px grey border showing up. It came from `@include molecules.panel` in the `.notebook-inner`. 

In the `.Notebook__close-buton` if `position: absolute`, it will show the border between the iframe and the parent container, in this way:
image

src/styles/annotator/notebook.scss
esanzgar: The problem of removing the top/bottom padding for the `.HypothesisApp--notebook` is that it also affects the notebook when displayed standalone, leaving no vertical space on top or bottom:

image

lyzadanger:
> The problem of removing the top/bottom padding for the .HypothesisApp--notebook is that it also affects the notebook when displayed standalone, leaving no vertical space on top or bottom:

I believe we can address this. Currently, there is no standalone variant, correct? We don't precisely know how this view will be "wrapped" in standalone mode and we can apply padding as needed to the containing element.

src/styles/annotator/notebook.scss
esanzgar: Alternatively we could remove the bottom padding from `.Notebook__controller-bar` and a little bit from top paddings from `.HypothesiApp--notebook`, as a compromise. This way it would still looks acceptable on both as standalone and within `<hypothesis-notebook>`.

Standalone | Within `<hypothesis-notebook>`
------------|----------------------------------|
image | image
 image | image

src/annotator/sidebar.js
```diff
@@ -449,4 +450,22 @@ export default class Sidebar extends Guest {
...
+    if (this.frame) {
+      this.frame.classList.remove('controls-hide');
```
lyzadanger, line 459: I think this className could use some adjustment to better match our conventions. This is a state-related class name, and we usually use the form `.is-<something>`

 The presence of this class causes the entire annotator frame to be hidden, not just the controls. So even though the net effect is that controls get hidden, it's really the entire frame, right?

As such, perhaps `.is-hidden` would be more appropriate?

lyzadanger (approved):
Thanks for your perseverance on this! I think we're just about ready to land here.

Go ahead and address the comments/questions here, and then merge when you feel ready. Good work!

src/styles/annotator/notebook.scss
```diff
@@ -24,4 +27,30 @@
...
   height: 100%;
+
+  // FIXME: these properties produce a visual break between the Notebook__controller-bar and the iframe.
+  // A better approach would be if the Notebook__controller-bar and iframe could be children of notebook-inner.
+  border: none;
+  border-top-left-radius: 0;
+  border-top-right-radius: 0;
+}
+
+// This container element has the purpose of pushing children to the bottom-right end.
```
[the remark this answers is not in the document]
esanzgar: I correct this, thanks.

src/annotator/notebook.js
```diff
@@ -105,6 +110,23 @@ export default class Notebook extends Delegator {
...
 
+    render(
+      <div className="Notebook__controller-bar">
+        <Button
+          icon="cancel"
+          className="Notebook__close-button"
+          buttonText="Close"
+          title="Close the Notebook"
+          onClick={event => {
+            event.stopPropagation();
```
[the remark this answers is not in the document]
esanzgar: I have added an explanation in the code.

src/annotator/notebook.js
```diff
@@ -105,6 +110,23 @@ export default class Notebook extends Delegator {
     this.container.className = 'notebook-outer';
     shadowRoot.appendChild(this.container);
 
+    render(
+      <div className="Notebook__controller-bar">
```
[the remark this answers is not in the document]
esanzgar: True. 

This could become a preact component, in the future.

src/annotator/sidebar.js
esanzgar: No, it only hides the controls.

`.is-controls-hidden` a better className?

lyzadanger: > No, it only hides the controls.

So, if the sidebar is open/visible, it would _only_ hide the controls?

esanzgar: That was the original intention... but actually works hiding the _whole_ sidebar. This has the advantage that the it keeps the original state of the sidebar (either open or close).

I renamed the methods to `show` and `hide`.

PR #3032, opened 2021-02-19
Replace custom DOM element names in `AdderToolbar` component

robertknight:
These custom names were a method of reducing the likelihood of these
elements being affected by the page's CSS. Since Shadow DOM support is
now much more ubiquitous this is not an issue, so switch back to
ordinary `<div>`s.

Note that the shadow hosts still use custom element names to prevent
them being impacted by page styles.  The elements being changed here are
inside the shadow DOM.


lyzadanger (approved):
I AM SO HAPPY

PR #3034, opened 2021-02-19
Remove unneeded functionality from the bucket-bar

### LINKED ISSUE #2907: Remove the use of 'container' from BucketBar
As part of #2904, @robertknight  [suggested to remove the use of `container` property in the BucketBar component](https://github.com/hypothesis/client/pull/2904#discussion_r561832993).

As BucketBar is not longer used as a 'plugin' we may take this opportunity to further simplify this component.

esanzgar:
Because of recent changes in the sidebar (#2904), the BucketBar
component is no longer used as a plugin. @robertknight suggested to
remove all unneeded functionality.

Closes #2907

### DIFF
--- a/src/annotator/plugin/bucket-bar.js
+++ b/src/annotator/plugin/bucket-bar.js
@@ -6,7 +6,7 @@ import Buckets from '../components/Buckets';
 import { anchorBuckets } from '../util/buckets';
 
 export default class BucketBar extends Delegator {
-  constructor(element, options, annotator) {
+  constructor(container, options, annotator) {
     const defaultOptions = {
       // Selectors for the scrollable elements on the page
       scrollables: [],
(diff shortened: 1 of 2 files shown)


robertknight (changes requested):
Unfortunately this PR has a problem that it breaks updating of the bucket bar in the PDF viewer. To reproduce:

1. Start the client's dev server
2. Go to http://localhost:3000/pdf/nils-olav and annotate different parts of the document
3. Scroll up and down

~~On master, the buckets move as you scroll. On this branch they do not.~~

**Edit:** Correction: I see the same issue on master as well. I'm currently looking into this.

**Edit 2:** I filed the issue with updating the bucket bar in the PDF viewer here: https://github.com/hypothesis/client/issues/3047.

robertknight:
The fix for the issue mentioned in the previous review is https://github.com/hypothesis/client/pull/3049. It doesn't conflict with the changes here, so this is good to merge.

robertknight (approved):

PR #3035, opened 2021-02-19
Refactor: added a method to register events

esanzgar:
This refactor brings the following advantage:

- no need to run the same checks when adding and removing events
- better ergonomics

Corrected a few things wrong in the tests.

### DIFF
--- a/src/annotator/plugin/bucket-bar.js
+++ b/src/annotator/plugin/bucket-bar.js
@@ -5,10 +5,18 @@ import Buckets from '../components/Buckets';
 
 import { anchorBuckets } from '../util/buckets';
 
+/**
+ * @typedef RegisteredListener
+ * @prop {Window|HTMLElement} eventTarget
+ * @prop {string} eventType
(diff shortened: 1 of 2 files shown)


robertknight (commented):
This needs a rebase due to other changes in `BucketBar`. I guess this is where your comments about introducing a utility in the dev sync last week came from? I'd agree that might be useful. Maybe with an API something like:

```js
class ListenerCollection {
  register(target, event, callback);
  // Unregister all listeners. The `destroy` name is more conventional
  // in our code but perhaps `dispose` would be better here.
  destroy(); 
}
```

esanzgar:
Yes, that's what I was thinking...

Would you mind if that's not a class but a function?

robertknight:
> Would you mind if that's not a class but a function?

Sure - if you don't the affordances that a class provides then a function is fine (or better perhaps).

esanzgar:
Replaced by #3099 and #3100

PR #3046, opened 2021-02-22
Reposition notebook's user-filter on narrow viewports

esanzgar:
Moved the user-filter below the heading and added a little bit of top
and bottom space.

image

### DIFF
--- a/src/styles/sidebar/components/NotebookView.scss
+++ b/src/styles/sidebar/components/NotebookView.scss
@@ -4,18 +4,21 @@
 
 .NotebookView {
   display: grid;
-  row-gap: var.$layout-space--xsmall;
+  row-gap: var.$layout-space--small;
   grid-template-areas:
     'heading'
+    'filters'
     'results'
     'items';
 
   &__heading {
+    line-height: 1;
     grid-area: heading;
   }
 
   &__filters {
     grid-area: filters;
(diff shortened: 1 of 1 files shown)


lyzadanger:
The introduced margin here causes wonkiness in wider-screen layouts:

image

You'll need to account for this within the wider-screen media query.

lyzadanger (changes requested):
This PR's changes adds vertical margins to all widths, not just narrow screens, resulting in alignment wonkiness on wider screens:

image

Perhaps look into the `row-gap` property for `grid` to establish a vertical rhythm?

lyzadanger:
Updated such that, wider screens:

image

and narrower:

image

src/styles/sidebar/components/NotebookView.scss
```diff
@@ -4,18 +4,21 @@
...
-  row-gap: var.$layout-space--xsmall;
+  row-gap: var.$layout-space--small;
   grid-template-areas:
     'heading'
+    'filters'
     'results'
     'items';
 
   &__heading {
+    line-height: 1;
```
lyzadanger, line 15: This was needed to normalize the row height so that there wasn't dead space above and below the heading text.

esanzgar (commented):
Looks good!

PR #3049, opened 2021-02-22
Pass bucket bar options to the constructor

### LINKED ISSUE #3047: Bucket bar does not update when PDF is scrolled
### Steps to reproduce

1. Go to https://via.hypothes.is/https://latimerslj.weebly.com/uploads/1/3/3/7/133773423/sweat_znhurston.pdf and select the Public group. This should have ~150 annotations.
2. Wait for all the annotations to load and anchor
3. Scroll down the page at a moderate pace

### Expected behaviour

The buckets should scroll smoothly as the corresponding highlights change their position relative to the viewport.

### Actual behaviour

The buckets do not update when scrolling

Tested in Safari 14.

### Additional details

Slack thread: https://hypothes-is.slack.com/archives/C1M8NH76X/p1614004139008000

esanzgar:
On PR #2904 it was forgotten to pass the bucket bar options to the constructor.

Closes #3047

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -80,7 +80,7 @@ export default class Sidebar extends Guest {
       } else {
         // BucketBar is a "plugin" for legacy reasons and is now constructed here so
         // that the parent element can be passed into the constructor.
-        this.plugins.BucketBar = new BucketBar(frame, {}, this);
+        this.plugins.BucketBar = new BucketBar(frame, config.BucketBar, this);
       }
 
       // Undocumented switch to enable/disable the wrapping of the sidebar inside a shadow DOM


src/annotator/sidebar.js
```diff
@@ -80,7 +80,11 @@ export default class Sidebar extends Guest {
       } else {
         // BucketBar is a "plugin" for legacy reasons and is now constructed here so
         // that the parent element can be passed into the constructor.
-        this.plugins.BucketBar = new BucketBar(frame, {}, this);
+        this.plugins.BucketBar = new BucketBar(
+          frame,
+          this.options.BucketBar,
```
robertknight, line 85: This could be changed to `config.BucketBar` to make it a little more obvious about where the value comes from. It will also be a step towards potentially removing `this.options` in future, which isn't used in that many remaining places.

robertknight (approved):
LGTM, with one suggestion.

src/annotator/sidebar.js
esanzgar: Thanks for the advice. Implemented.

PR #3051, opened 2021-02-22
Update annotator toolbar CSS class names to match conventions

robertknight:
Rename CSS classes  for `Toolbar` component to match the component itself. This component name is rather generic but at least this brings the CSS and Preact components into line.

One of the toolbar's classes was applied to a container element created by `ToolbarController`. This class has been moved to an element created by the `Toolbar` component for consistency. This means there is an apparently redundant (no styling, one child) container div when looking at the DOM structure for the toolbar + bucket bar. See `<div>` child of `.annotator-frame` with no attributes here:

Redundant div in toolbar

This will go away if the whole sidebar container is made into a single Preact component in future.

See also https://github.com/hypothesis/client/pull/3033.

### DIFF
--- a/src/annotator/components/Toolbar.js
+++ b/src/annotator/components/Toolbar.js
@@ -14,7 +14,7 @@ import propTypes from 'prop-types';
 function ToolbarButton({
   buttonRef,
   expanded,
-  className = 'annotator-toolbar-button',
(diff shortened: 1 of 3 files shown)


lyzadanger (approved):
Looks good.

PR #3055, opened 2021-02-22
Cleanup bucket bar

esanzgar:
- Added types
- Moved out of the plugin directory
- Avoid extending from Delegator (unnecessary)

### DIFF
--- a/src/annotator/bucket-bar.js
+++ b/src/annotator/bucket-bar.js
@@ -1,31 +1,32 @@
-import Delegator from '../delegator';
-
 import { render } from 'preact';
-import Buckets from '../components/Buckets';
+import Buckets from './components/Buckets';
 
-import { anchorBuckets } from '../util/buckets';
+import { anchorBuckets } from './util/buckets';
 
-export default class BucketBar extends Delegator {
-  constructor(container, options, annotator) {
-    const defaultOptions = {
-      // Selectors for the scrollable elements on the page
-      scrollables: [],
-    };
+/**
+ * @typedef BucketBarOptions
+ * @prop {string[]} [scrollables] - Selectors for the scrollable elements on the page
+ */
 
-    const opts = { ...defaultOptions, ...options };
+export default class BucketBar {
+  /**
+   * @param {HTMLElement} container
+   * @param {Pick<import('./guest').default, 'anchors'|'selectAnnotations'>} guest
+   * @param {BucketBarOptions} [options]
+   */
+  constructor(container, guest, options = {}) {
+    this.options = options;
+    this.element = document.createElement('div');
+    this.element.className = 'annotator-bucket-bar';
 
-    const el = document.createElement('div');
-    el.className = 'annotator-bucket-bar';
-    super(el, opts);
-
-    this.annotator = annotator;
+    this.guest = guest;
     container.appendChild(this.element);
 
     this.updateFunc = () => this.update();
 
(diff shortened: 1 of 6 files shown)


src/annotator/bucket-bar.js
```diff
@@ -1,24 +1,30 @@
...
 import { render } from 'preact';
-import Buckets from '../components/Buckets';
+import Buckets from './components/Buckets';
+
+import { anchorBuckets } from './util/buckets';
 
-import { anchorBuckets } from '../util/buckets';
+/**
+ * @typedef BucketBarOptions
+ * @prop {string[]} scrollables  // Selectors for the scrollable elements on the page
```
robertknight, line 8:
```suggestion
 * @prop {string[]} scrollables - Selectors for the scrollable elements on the page
```

src/annotator/bucket-bar.js
```diff
@@ -1,24 +1,30 @@
...
+   * @param {Pick<import('./guest').default, 'anchors'|'selectAnnotations'>} guest
+   */
+  constructor(container, options, guest) {
     const defaultOptions = {
-      // Selectors for the scrollable elements on the page
       scrollables: [],
     };
 
-    const opts = { ...defaultOptions, ...options };
+    this.options = { ...defaultOptions, ...options };
```
robertknight, line 22: There is no real need to have `this.options` any more. You could just have a `this.scrollables` argument which is initialized to an empty array if there is no `scrollables` property in `options`.

robertknight (commented):
I think this is is going in the right direction. There are some remaining references to `BucketBar` being a "plugin" in other parts of the code, eg. in `guest.js`, `sidebar.js` and tests. Do you have a plan for those?

src/annotator/bucket-bar.js
esanzgar: I addressed this as you suggested.

esanzgar:
> I think this is is going in the right direction. There are some remaining references to BucketBar being a "plugin" in other parts of the code, eg. in guest.js, sidebar.js and tests. Do you have a plan for those?

I addressed that in a second commit in this PR.

src/annotator/guest.js
```diff
@@ -151,6 +151,7 @@ export default class Guest extends Delegator {
     });
 
     this.plugins = {};
+    this.BucketBar = null;
```
robertknight, line 154:
```suggestion
    /** @typedef {BucketBar|null} */
    this.bucketBar = null;
```

No reason not to use our normal public field naming conventions here. Also its helpful for readers to specify the type.

src/annotator/test/bucket-bar-test.js
```diff
@@ -10,7 +10,7 @@ describe('BucketBar', () => {
 
   const createBucketBar = function (options) {
     const element = document.createElement('div');
-    return new BucketBar(element, options || {}, fakeAnnotator);
+    return new BucketBar(element, fakeAnnotator, options);
```
robertknight, line 13: 👍 - swapping the options object to be the last parameter is nicer.

robertknight (approved):
Hi Eduardo - The changes here look good. I suggested to change the name of the `BucketBar` field to `bucketBar` and document its type to fit with our normal conventions for public fields. Please make sure you give this a test with both PDF and HTML documents.

src/annotator/guest.js
esanzgar: Awesome.

PR #3056, opened 2021-02-22
Convert meta tag attributes to lowercase for case insensitive comparison

### LINKED ISSUE #671: doi equivalence fails for dc.Identifier
At http://journals.sagepub.com/doi/full/10.1177/1363461517725224, this DOI is declared:

```
<meta name="dc.Identifier" scheme="doi" content="10.1177/1363461517725224">
```

We don't report a DOI link because of a case mismatch. Here -- https://github.com/hypothesis/client/blob/c8737ff229360b6909687dae74deb9bf19dfb430/src/annotator/plugin/document.coffee#L161 -- we match "identifier" but miss "Identifier"

According to https://hyp.is/yCX5MhKZEeiiaqsrkJO-tg/dublincore.org/documents/2000/07/16/usageguide/, the preferred capitalization is actually `DC.Identifier`. 

We'd catch the `DC` prefix because we do case-insensitive matching here: https://github.com/hypothesis/client/blob/c8737ff229360b6909687dae74deb9bf19dfb430/src/annotator/plugin/document.coffee#L93 

But we'll miss camelCased names.

esanzgar:
This PR converts the following attributes in the meta tag:

```
name=CITATION_[Y]
property="OG:[Y]
name="TWITTER:[Y]
name="DC[X][Y]
name="PRISM[X][Y]
name="EPRINTS[X][Y]
```

to:

```
name=citation_[y]
property="og:[y]
name="twitter:[y]
name="dc[x][y]
name="prism[x][y]
name="eprints[x][y]
```

Closes #671

### DIFF
--- a/src/annotator/integrations/html-metadata.js
+++ b/src/annotator/integrations/html-metadata.js
@@ -130,46 +130,46 @@ export class HTMLMetadata {
   }
 
   _getHighwire() {
-    this.metadata.highwire = this._getMetaTags('citation', 'name', '_');
+    this.metadata.highwire = this._getMetaTags('name', 'citation', '_');
   }
 
   _getFacebook() {
-    this.metadata.facebook = this._getMetaTags('og', 'property', ':');
+    this.metadata.facebook = this._getMetaTags('property', 'og', ':');
   }
 
   _getTwitter() {
-    this.metadata.twitter = this._getMetaTags('twitter', 'name', ':');
+    this.metadata.twitter = this._getMetaTags('name', 'twitter', ':');
   }
 
   _getDublinCore() {
-    this.metadata.dc = this._getMetaTags('dc', 'name', '.');
+    this.metadata.dc = this._getMetaTags('name', 'dc', '.');
   }
 
   _getPrism() {
-    this.metadata.prism = this._getMetaTags('prism', 'name', '.');
+    this.metadata.prism = this._getMetaTags('name', 'prism', '.');
   }
 
   _getEprints() {
-    this.metadata.eprints = this._getMetaTags('eprints', 'name', '.');
+    this.metadata.eprints = this._getMetaTags('name', 'eprints', '.');
   }
 
   /**
    * Return an array of all the `content` values of `<meta>` tags on the page
    * where the attribute named `attribute` begins with `<prefix><delimiter>`.
    *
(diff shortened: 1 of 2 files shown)


src/annotator/plugin/test/document-test.js
```diff
@@ -64,11 +64,11 @@ describe('DocumentMeta', function () {
         <meta name="citation_title" content="Foo">
         <meta name="citation_pdf_url" content="foo.pdf">
         <meta name="dc.identifier" content="doi:10.1175/JCLI-D-11-00015.1">
-        <meta name="dc:identifier" content="foobar-abcxyz">
+        <meta name="dc:Identifier" content="foobar-abcxyz">
```
robertknight, line 67: Rather than modifying an existing test I think it would be better to add a separate test to cover this change more explicitly. As it is this particular test is now checking two things - handling of `:` vs `.` and lower-casing of the name.

src/annotator/plugin/document.js
```diff
@@ -177,10 +177,10 @@ export default class DocumentMeta extends Delegator {
     /** @type {Object.<string,string[]>} */
     const tags = {};
     for (let meta of Array.from(this.document.querySelectorAll('meta'))) {
-      const name = meta.getAttribute(attribute);
+      const name = meta.getAttribute(attribute)?.toLowerCase();
```
robertknight, line 180: This is going to affect the metadata that is stored with the annotation and sent via the API when it is created. I suspect it is probably fine to always lower case the name value and not store the original version. It would be worth checking whether the backend references any of these values explicitly and whether it does so in a case-insensitive way.

robertknight (changes requested):
I think the basic change here is going to be fine. However the tests currently conflate checks for several different aspects of parsing in the same test case. I think it would be better to add a separate, explicit test for the lower-casing of meta names. I also suggested having a look in the backend in case the same issue occurs there and would need dealing with separately.

src/annotator/plugin/test/document-test.js
esanzgar: I agree. I added a separated test.

src/annotator/plugin/document.js
esanzgar: I think I am going to need a little help to check the backend code.

PR #3059, opened 2021-02-24
Spike buttons ergonomics

esanzgar:
I am proposing to have only one interface for the `<Button>` component with several props that switches the use of the different versions of the button.


LMS007 (commented):
I see, more or less a wrapper around the proposed new components. 
This might be a good middle ground as it does not affect the sass structure and we can still easily expand to new button components when/if needed.

`<IconButton>` -> `<Button iconButton>`
`<CompactIconButton>`  -> `<Button compactIconButton>`
`<LinkButton>` -> `<Button linkButton>`

PR #3060, opened 2021-02-25
Revert "Added close button to the notebook"

esanzgar:
Reverts hypothesis/client#3024

This PR introduced a couple of regressions:
* pagination is not visible
* if the notebook is opened and  the window is resized from a narrow to wide width, the notebook closes and you have no sidebar controls


lyzadanger (approved):
👍🏻 Thanks!

PR #3061, opened 2021-02-25
Added a button to close the notebook

esanzgar:
This PR introduces a button to close the notebook app.

* the buttons is outside the notebook application. That way it doesn't interfere in case we ever want to render the notebook in a different context or standalone.
 
* the button sticks to the top-right position without taking space from the notebook. However this creates a funny behaviour when the viewport is very narrow.

image

* opening the notebook hides the _whole_ sidebar, leaving their current UI state of the sidebar (open or close) untouched. Thus, when the the notebook is closed the previous UI state of the sidebar is preserved.

* sets the zIndex of the container element `<hypothesis-notebook>` to a very high number so it is not hidden by any element in the host page.

* opening of the notebook happens through the sidebar who listens to the `openNotebook` event. After the notebook closes, the sidebar 'regains' control.

### DIFF
--- a/src/annotator/notebook.js
+++ b/src/annotator/notebook.js
@@ -1,6 +1,7 @@
 import Delegator from './delegator';
 import { createSidebarConfig } from './config/sidebar';
 import { createShadowRoot } from './util/shadow-root';
+import { render } from 'preact';
 
 /**
  * Create the iframe that will load the notebook application.
@@ -51,13 +52,13 @@ export default class Notebook extends Delegator {
      */
     this.container = null;
 
-    this.subscribe('showNotebook', groupId => {
+    this.subscribe('openNotebook', groupId => {
       this._groupId = groupId;
-      this.show();
+      this.open();
     });
-    this.subscribe('hideNotebook', () => this.hide());
+    this.subscribe('closeNotebook', () => this.close());
     // If the sidebar has opened, get out of the way
-    this.subscribe('sidebarOpened', () => this.hide());
+    this.subscribe('sidebarOpened', () => this.close());
   }
 
   _update() {
@@ -76,14 +77,14 @@ export default class Notebook extends Delegator {
     }
   }
 
-  show() {
+  open() {
     const container = this._initContainer();
     this._update();
     container.classList.add('is-open');
     container.style.display = '';
   }
 
-  hide() {
+  close() {
     if (this.container) {
       this.container.classList.remove('is-open');
       this.container.style.display = 'none';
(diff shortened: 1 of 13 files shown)


src/annotator/notebook.js
```diff
@@ -61,8 +61,6 @@ export default class Notebook extends Delegator {
       this._groupId = groupId;
       this.open();
     });
-    // Defensive programming: if the sidebar has opened, get out of the way
```
esanzgar: The left-over defensive programming caused the issue 😄

lyzadanger:
Hi, @esanzgar 

Might I suggest taking a step back and regrouping a little before continuing? I noticed a couple of issues while trying this branch out again locally:

* There's no spacing at all at the top of the Notebook view, for any viewport width:
image
* Closing the Notebook no longer restores the previous state of the sidebar (you have to re-open it).
* The situation with the Close button and the "stickiness" factor in narrow viewports is unfortunate, and I think we need to consider options for how to alleviate it. We seem to be bumping up against a few hurdles based on the button existing outside of the Notebook's `iframe`. Please don't be discouraged (I know I'd be feeling frustrated by this point in the cycle!) but I'm wondering what the level of effort would be to render the Close button within the client side of the app and trigger a `crossframe` even similar to the "Open Notebook" item in the user menu. I know that we won't want to render this button when the Notebook is standalone, but there are ways of communicating that to the app (such that the view doesn't render the button) via config or other mechanisms. If this is just a non-starter, I understand. But I thought I'd ask...

esanzgar:
I corrected the 3 issues that were reported by @lyzadanger:

- spacing has been restored
- closing the notebook on touch devices restored the previous state state of the sidebar, however immediately after it closed the sidebar. This was because the `Guest` classes listens for all touch events and closes the sidebar. I had to stop the propagation of the touch event.

Regarding the las point, I am happy to integrate the close button within the notebook application, if that's what we want.

LMS007:
Thanks Eduardo. I would probably try the darker primary button background color.  The button is really washed out on the grey panel. I also think it may need to sit a little bit more inside the panel so it does not cover the browser scroll bar.

Screen Shot 2021-03-02 at 9 59 19 AM
Screen Shot 2021-03-02 at 9 59 30 AM

If we don't like the color, perhaps just the simple "X" icon button w/ transparent background like we use in LMS dialogs.

lyzadanger (approved):
Thanks for all of this work! Let's get it landed as-is. While there's further work that needs to be done to make the close button placement ideal, these changes introduce some useful behavior. Let's move forward and regroup about how to make the close button a little better.

esanzgar:
@LMS007, yes I agree, it looks better with a darker background. Let's implement that on a follow up PR.

Thanks @lyzadanger and @LMS007 .

PR #3063, opened 2021-02-25
Keep the content within the grid boundaries

esanzgar:
This small trick borrowed from
https://css-tricks.com/preventing-a-grid-blowout/ keeps the grid content
within the grid boundaries, thus improving the responsive design.

## Before

https://user-images.githubusercontent.com/8555781/109205164-8257ca00-77a6-11eb-9bd2-4ec5ab031740.mov

## After

https://user-images.githubusercontent.com/8555781/109205169-85eb5100-77a6-11eb-82d6-b953986cf974.mov

### DIFF
--- a/src/styles/sidebar/components/NotebookView.scss
+++ b/src/styles/sidebar/components/NotebookView.scss
@@ -5,6 +5,8 @@
 .NotebookView {
   display: grid;
   row-gap: var.$layout-space--small;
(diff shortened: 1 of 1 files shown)


lyzadanger (approved):
I think this seems like a nice safety valve. Just one thing: can you add a comment with a link to that CSS Tricks post in the SASS source here?

PR #3065, opened 2021-02-26
Move "no selectable text" PDF warning banner into shadow DOM

robertknight:
For consistency with the rest of the UI elements rendered by the annotator code, move the warning banner at the top of the PDF viewer into shadow DOM. This is also another step towards not needing to load any global CSS (eg. our CSS reset) into the host page.

 - Change the container element for the banner from `<div>` to
   `<hypothesis-banner>` to match other `<hypothesis-*>` UI container
   elements
 - Move the `WarningBanner` component into `components/WarningBanner.js`
   and rename the CSS classes to match
   
The new DOM structure for the banner looks as follows:

pdf-warning-banner-dom

**Testing:**

With the client's dev server running, go to http://localhost:3000/pdf/painting and verify that the banner shows up correctly at the top of the viewer.

### DIFF
--- a/src/annotator/plugin/pdf.js
+++ b/src/annotator/plugin/pdf.js
@@ -113,54 +113,38 @@ export default class PDF extends Delegator {
    * @param {boolean} showWarning
    */
   _showNoSelectableTextWarning(showWarning) {
+    // Get a reference to the top-level DOM element associated with the PDF.js
+    // viewer.
+    const outerContainer = /** @type {HTMLElement} */ (document.querySelector(
+      '#outerContainer'
+    ));
+
     if (!showWarning) {
       this._warningBanner?.remove();
       this._warningBanner = null;
+
+      // Undo inline styles applied when the banner is shown. The banner will
+      // then gets its normal 100% height set by PDF.js's CSS.
+      outerContainer.style.height = '';
+
       return;
     }
 
-    // Get a reference to the root element of the main content area in
-    // PDF.js.
-    const mainContainer = /** @type {HTMLElement} */ (document.querySelector(
-      '#mainContainer'
-    ));
-
-    // Update the position of the warning banner and `mainContainer` element
-    // below it when the size of the banner changes, eg. due to a window resize.
-    const updateBannerHeight = () => {
-      /* istanbul ignore next */
-      if (!this._warningBanner) {
-        return;
-      }
-      const rect = this._warningBanner.getBoundingClientRect();
-      mainContainer.style.top = rect.height + 'px';
-      this._warningBanner.style.top = -rect.height + 'px';
-    };
-
     this._warningBanner = document.createElement('hypothesis-banner');
-    Object.assign(this._warningBanner.style, {
-      // Position the banner at the top of the viewer and make it span the full width.
-      position: 'absolute',
-      left: '0',
-      right: '0',
-
-      // The Z-index is necessary to raise the banner above the toolbar at the
-      // top of the PDF.js viewer.
-      zIndex: '10000',
-    });
-    mainContainer.appendChild(this._warningBanner);
+    document.body.prepend(this._warningBanner);
 
     const warningBannerContent = createShadowRoot(this._warningBanner);
     render(<WarningBanner />, warningBannerContent);
 
-    // nb. In browsers that don't support `ResizeObserver` the banner height
-    // will simply be static and not adjust if the window is resized.
-    if (typeof ResizeObserver !== 'undefined') {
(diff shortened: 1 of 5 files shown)


src/annotator/plugin/pdf.js
```diff
@@ -164,11 +137,21 @@ export default class PDF extends Delegator {
...
+    this._warningBanner = document.createElement('hypothesis-banner');
+    Object.assign(this._warningBanner.style, {
+      // Position the banner at the top of the viewer and make it span the full width.
+      position: 'absolute',
+      left: '0',
+      right: '0',
+
+      // The Z-index is necessary to raise the banner above the toolbar at the
+      // top of the PDF.js viewer.
+      zIndex: '10000',
```
robertknight, line 149: These styles used to be applied by the `.annotator-pdf-warning-banner` styles in `annotator.scss`. Since we dynamically adjust the `top` style of `_warningBanner` above, it is necessary to apply these styles to the `_warningBanner` container rather than its content (now in Shadow DOM). Applying this via inline styles was the easiest thing to do, although we could potentially move this to Shadow DOM in future as well, with some changes to the code above this that sets `this._warningBanner.style.top`.

esanzgar, line 149: This `zIndex` is not needed, I believe.

esanzgar: I suggests a different DOM structure that does not need the `top` value. Instead of appending to `#mainContainer` prepend to the body:

```
    this._warningBanner = document.createElement('hypothesis-banner');
    const warningBannerContent = createShadowRoot(this._warningBanner);
    render(<WarningBanner />, warningBannerContent);
    document.body.prepend(this._warningBanner);
```

esanzgar (commented):
I suggests a different DOM structure that doesn't require to dynamically calculate the `top` value:

```diff
 
-    // Get a reference to the root element of the main content area in
-    // PDF.js.
-    const mainContainer = /** @type {HTMLElement} */ (document.querySelector(
-      '#mainContainer'
-    ));
-
-    // Update the position of the warning banner and `mainContainer` element
-    // below it when the size of the banner changes, eg. due to a window resize.
-    const updateBannerHeight = () => {
-      /* istanbul ignore next */
-      if (!this._warningBanner) {
-        return;
-      }
-      const rect = this._warningBanner.getBoundingClientRect();
-      mainContainer.style.top = rect.height + 'px';
-      this._warningBanner.style.top = -rect.height + 'px';
-    };
-
     this._warningBanner = document.createElement('hypothesis-banner');
-    Object.assign(this._warningBanner.style, {
-      // Position the banner at the top of the viewer and make it span the full width.
-      position: 'absolute',
-      left: '0',
-      right: '0',
-
-      // The Z-index is necessary to raise the banner above the toolbar at the
-      // top of the PDF.js viewer.
-      zIndex: '10000',
-    });
-    mainContainer.appendChild(this._warningBanner);
-
     const warningBannerContent = createShadowRoot(this._warningBanner);
     render(<WarningBanner />, warningBannerContent);
-
-    // nb. In browsers that don't support `ResizeObserver` the banner height
-    // will simply be static and not adjust if the window is resized.
-    //
-    // @ts-expect-error - TS is missing `ResizeObserver`
-    if (typeof ResizeObserver !== 'undefined') {
-      // Update the banner when the window is resized or the Hypothesis
-      // sidebar is opened.
-      // @ts-ignore
-      new ResizeObserver(updateBannerHeight).observe(this._warningBanner);
-    }
-    updateBannerHeight();
+    document.body.prepend(this._warningBanner);
   }
 
   // This method (re-)anchors annotations when pages are rendered and destroyed.
```

robertknight:
> I suggests a different DOM structure that doesn't require to dynamically calculate the top value

Thanks for the suggestion. I tried this and indeed it looks like it can simplify things quite a bit. I've rebased this PR and added a second commit which implements your suggestion. See commit message for full details.

esanzgar (approved):
I would suggest to enable `ResizeObserver` in a future PR. Otherwise, if the window is resized and the size of the warning banner changes it leads to displaying a double scrolling bar.

For now, this is fine.

esanzgar:
In case we want to `ResizeObserver`, I wonder if covering the PDF toolbar by the warning banner would be acceptable (`position: fixed` and a high `z-Index`):

image

robertknight:
> I wonder if covering the PDF toolbar by the warning banner would be acceptable (position: fixed and a high z-Index):

Maybe. The PDF can't really be used at all with Hypothesis as long as this problem exists, so it may not matter if a small amount of the content is obscured by the banner.

Issue #2976, opened 2021-02-09
SPIKE: remove inheritance hierarchy in the client (Delegator -> Guest ->   Sidebar)


esanzgar:
The client contains the following inheritance hierarchy:

```
Delegator
 |- Guest
      |- Sidebar
```

This hierarchy is used to include functionality from base classes. Apart from this, there is very little use of the the hierarchal system. It can be perceived as overly complex by developers. In addition, there are other potential issues with the use of this hierarchical pattern, such adding the same plugin several times at different levels (see https://hypothes-is.slack.com/archives/C1M8NH76X/p1610632193014200)

Is there an alternative and simpler arrangements?

esanzgar:
#3076 removes the inheritance between `Guest` and `Sidebar`. The inheritance from `Delegator` can also be removed.

PR #3076, opened 2021-03-01
Make Sidebar collaborate with rather than inherit from Guest

robertknight:
Change the relationship between the `Guest` and `Sidebar` classes from
one of inheritance to one where `Sidebar` receives a reference to the
`Guest` instance and calls methods on it or subscribes to events from
it. The `Guest` also no longer "owns" references to the bucket bar /
toolbar objects that logically belong to the sidebar. Instead it emits
events which the sidebar responds to.

This change makes the interface of the `Guest` used by the `Sidebar`
more explicit and ensures a better boundary between the two. This also
makes it easier for the `Sidebar` tests to be concerned only with the
interface of the `Guest` and not its implementation details. In future
this change will also make it possible to have a frame which does not
contain a sidebar but is not annotateable. We had a need for this
historically when integrating with epub viewers, although it was never
implemented.

Updating the tests for `PdfSidebar` was complicated by the fact that the
`PdfSidebar` tests are not pure unit tests. They instantiate the
`Sidebar` base class and so depend on many implementation details of it.
To make this change and others easier, the `PdfSidebar` tests have been
changed to mock the `Sidebar` base class. The steps involved in this are
non-obvious so I extracted the logic into a utility function.

Fixes https://github.com/hypothesis/client/issues/2976

 - Change `Sidebar` to no longer inherit `Guest` but accept it as a
   constructor argument
 - Remove direct references to the `BucketBar` and `ToolbarController`
   instances from the `Guest` class and instead emit events from the
   `Guest` which the `Sidebar` responds to.
 - Add a `mockBaseClass` testing helper in `src/test-util/mock-base.js`
   and change the `PdfSidebar` tests to use it, so that they are less
   coupled to `Sidebar` implementation details.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -21,7 +21,6 @@ import { normalizeURI } from './util/url';
  * @typedef {import('../types/annotator').AnnotationData} AnnotationData
  * @typedef {import('../types/annotator').Anchor} Anchor
  * @typedef {import('../types/api').Target} Target
- * @typedef {import('./toolbar').ToolbarController} ToolbarController
  */
 
 /**
@@ -120,9 +119,6 @@ export default class Guest extends Delegator {
 
     this.visibleHighlights = false;
 
-    /** @type {ToolbarController|null} */
-    this.toolbar = null;
-
     this.adderToolbar = document.createElement('hypothesis-adder');
     this.adderToolbar.style.display = 'none';
     this.element.appendChild(this.adderToolbar);
@@ -151,9 +147,6 @@ export default class Guest extends Delegator {
     });
 
     this.plugins = {};
-
-    /** @typedef {import('./bucket-bar').default|null} */
-    this.bucketBar = null;
 
     /** @type {Anchor[]} */
     this.anchors = [];
@@ -483,10 +476,9 @@ export default class Guest extends Delegator {
       annotation.$orphan = hasAnchorableTargets && !hasAnchoredTargets;
 
       // Add the anchors for this annotation to instance storage.
-      this.anchors = this.anchors.concat(anchors);
+      this._updateAnchors(this.anchors.concat(anchors));
 
       // Let plugins know about the new information.
-      this.bucketBar?.update();
       this.plugins.CrossFrame?.sync([annotation]);
 
       return anchors;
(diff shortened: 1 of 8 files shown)


src/annotator/index.js
```diff
@@ -66,11 +66,16 @@ function init() {
 
   config.pluginClasses = pluginClasses;
 
-  const annotator = new Klass(document.body, config);
+  const guest = new Guest(document.body, config);
+  const sidebar = SidebarClass
+    ? new SidebarClass(document.body, config, guest)
```
esanzgar, line 71: I would probably invert the last two arguments, making `config` the last one and optional.

src/annotator/index.js
```diff
@@ -66,11 +66,16 @@ function init() {
 
   config.pluginClasses = pluginClasses;
 
-  const annotator = new Klass(document.body, config);
+  const guest = new Guest(document.body, config);
```
esanzgar, line 69: Would it be better to instantiate `guest` inside `sidebar`?

src/annotator/guest.js
```diff
@@ -658,9 +652,7 @@ export default class Guest extends Delegator {
     }
 
     this.selectedRanges = [range];
-    if (this.toolbar) {
-      this.toolbar.newAnnotationType = 'annotation';
-    }
+    this.publish('hasSelectionChanged', [true]);
```
esanzgar, line 656: Similar to `highlightsVisibleChanged`:

```suggestion
    this.publish('selectionChanged', [true]);
    
```

esanzgar (approved):
I really like this cleanup/decoupling.

src/annotator/guest.js
robertknight: Hmm - I debated this. In the end I chose `hasSelectionChanged` rather than `selectionChanged` to make it more obvious that the argument was a boolean (whether there is a selection) rather than the current selection itself.

src/annotator/index.js
robertknight: There may be cases in future where the guest is specialized or configured depending on the document type and it will probably make sense to do that outside the sidebar. There may also be scenarios where the Sidebar is instantiated without a guest, if the content to be annotated is inside a frame.

src/annotator/index.js
robertknight: Yes, I think that makes sense. If its easy to do I'll make a follow-up commit in this PR. If its more complex I might separate that out into a follow-up PR.

PR #3079, opened 2021-03-01
Ensure PDF integration is loaded in PDF documents

robertknight:
The [recent refactor](https://github.com/hypothesis/client/pull/3076) to change the relationship between the `Guest` and `Sidebar`
classes broke PDF.js integration because the `PdfSidebar` class is no
longer able to modify the config that is passed to the `Guest` class.
Previously it would pass `PDF: {}` in the config to the `super()` call
which would then result in the `Guest` class loading the PDF
integration.

This now has to be done in the `index.js` module. We already had to do
this previously for ebooks. Now it has to be done for all scenarios.

**Testing:**

Go to http://localhost:3000/pdf/nils-olav and click Help => About this version in the sidebar. On master there is a line in the panel that reads: "Fingerprint: N/A". With this change it now reads: "Fingerprint: <hex string>" as it should.

### DIFF
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -52,10 +52,6 @@ function init() {
   let SidebarClass = isPDF ? PdfSidebar : Sidebar;
 
   if (config.subFrameIdentifier) {
-    // Make sure the PDF plugin is loaded if the subframe contains the PDF.js viewer.
-    if (isPDF) {
-      config.PDF = {};
-    }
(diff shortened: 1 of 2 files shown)


robertknight:
I went ahead and merged this since it was urgent to get the fix out. See https://hypothes-is.slack.com/archives/C4K6M7P5E/p1614622548295900 for details.

PR #3080, opened 2021-03-01
Remove frontend-shared subfolder

LMS007:
### Commits

-----------

Update shared package …

There was a version issue between the current installed version 1.4.0 and the local copy being using in the subfolder (/frontend-shared). The newest shared package exports an _index.scss file which is required by the client, but it was not formally importing a version of the shared package which exported the _index.scss file. The client now imports 1.9.0 which has _index.scss and now the frontend-shared folder can be fully removed from the client.

TL;DR, This is just more reason to move the shared package to its own repo because this issue was being masked.

-----------

Remove frontend-share subfolder …

- Remove github actions related to frontend-shared
- Remove scripts and gulp commands related to frontend-shared
- frontend-shared is now managed from its own repository.

------------

Remove dependabot exception for frontend-shared

----------

### Discussion

This pulls frontend-shared out of the client and simple treats it as an imported dependency.  I think we are close enough on the shared repo now that we can safely do the switch. There are still a few manual steps involved when doing a release in the frontend-shared repo, but nothing is very unique or out of the ordinary.


robertknight (approved):
LGTM. I've checked this out locally and verified that I can build the client with `@hypothesis/frontend-shared` installed from npm.

I did have some confusion caused by an existing `yarn link`-ed installation of @hypothesis/frontend-shared from the now-removed frontend-shared/ directory. There is a `yarn unlink <package>` command but I found that it doesn't work if the symlinked destination has been removed. The easiest way to resolve this was to remove node_modules and re-run `yarn install`.

I suggest it might be a good idea to use the frontend sync tomorrow to go through the process of publishing new versions of the frontend-shared package from the separate repo.

LMS007:
Right, I forgot about the local link clean up, which does have to be done manually for each dev. I'll mention this tomorrow.

PR #3081, opened 2021-03-02
Fix test warnings from missing prop

LMS007:
The ThreadList component has a required felid  `threads` but its value was undefined in tests. The value comes from the mocked version of `useRootThread.children`

--------------

I'm not actually sure when this started but I noticed it for the first time today. 

Screen Shot 2021-03-01 at 3 54 31 PM


robertknight (approved):
LGTM. Thanks for this.

PR #3082, opened 2021-03-02
Replace the "plugin" functionality in the Guest

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/3083** (_This contains a bug fix necessary to make sure that correct document metadata is sent to the server for PDFs after the changes in this branch_)~~

Replace the "plugin" functionality in the Guest that indirectly
instantiates the `DocumentMeta`, `PDF` and `CrossFrame` classes with
direct instantiation of those classes.

The annotator part of the Hypothesis client is no longer a pluggable
toolbox which can be arbitrarily extended by user-provided plugins.
Removing the "plugins" logic and replacing it with direct construction
of the necessary classes removes a lot of indirection that made the code
harder to follow and also makes it easier for TypeScript to statically
check the code. It will also enable the `config` object that is passed
to `Guest` to be paired down to just the fields that are actually needed
and properly typed.

In future we will need some kind of plugin/integration facility to
support different document viewers. For we will build a more specific
document type integration facility when the time comes.

 - Remove `this.plugins` field from `Guest` class and `pluginClasses`
   config and replace with direct construction of `PDFIntegration`,
   `DocumentMeta` and `CrossFrame` classes

 - Update the `AnchoringImpl` typedef to correctly reflect that some anchoring
   implementations (HTML) have a synchronous `describe` method

 - Remove unnecessary `config` argument to `PDF` constructor

 - Replace the `PDF: {}` content in the guest config with `documentType:
   'pdf'` as a more obvious way for the annotator entry point to
   configure which document viewer integration is loaded. In future we
   can generalize this to support different types of viewer integration.
   
   ----
   
   **TODO:**
   - [x] Add tests for `documentType: "pdf"` config
   - [x] Verify that all aspects of HTML / PDF support are correctly used depending on the document type (anchoring implementation, document metadata, viewer integration such as side-by-side behavior)

### DIFF
--- a/src/annotator/config/sidebar.js
+++ b/src/annotator/config/sidebar.js
@@ -33,9 +33,7 @@ export function createSidebarConfig(config) {
   // nb. We don't currently strip all the annotator-only properties here.
   // That's OK because validation / filtering happens in the sidebar app itself.
   // It just results in unnecessary content in the sidebar iframe's URL string.
-  ['notebookAppUrl', 'sidebarAppUrl', 'pluginClasses'].forEach(
-    key => delete sidebarConfig[key]
-  );
+  ['notebookAppUrl', 'sidebarAppUrl'].forEach(key => delete sidebarConfig[key]);
 
   return sidebarConfig;
 }
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -463,7 +463,7 @@ export default class Guest extends Delegator {
       // Add the anchors for this annotation to instance storage.
       this._updateAnchors(this.anchors.concat(anchors));
 
-      // Let the sidebar know about the new annotation.
+      // Let other frames (eg. the sidebar) know about the new annotation.
       this.crossframe.sync([annotation]);
 
       return anchors;
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -2,6 +2,9 @@ import scrollIntoView from 'scroll-into-view';
 
 import Delegator from './delegator';
 import { Adder } from './adder';
+import CrossFrame from './plugin/cross-frame';
+import DocumentMeta from './plugin/document';
+import PDFIntegration from './plugin/pdf';
 
 import * as htmlAnchoring from './anchoring/html';
 import { TextRange } from './anchoring/text-range';
@@ -106,16 +109,11 @@ export default class Guest extends Delegator {
    * @param {HTMLElement} element -
    *   The root element in which the `Guest` instance should be able to anchor
    *   or create annotations. In an ordinary web page this typically `document.body`.
-   * @param {Record<string, any>} config
-   * @param {typeof htmlAnchoring} anchoring - Anchoring implementation
+   * @param {Record<string, any>} [config]
+   * @param {typeof htmlAnchoring} [anchoring] - Anchoring implementation
    */
-  constructor(element, config, anchoring = htmlAnchoring) {
-    const defaultConfig = {
-      // This causes the `Document` plugin to be initialized.
-      Document: {},
-    };
-
-    super(element, { ...defaultConfig, ...config });
+  constructor(element, config = {}, anchoring = htmlAnchoring) {
+    super(element, config);
 
     this.visibleHighlights = false;
 
@@ -146,7 +144,14 @@ export default class Guest extends Delegator {
       }
     });
 
-    this.plugins = {};
+    // Set the initial anchoring module. Note that for PDFs this is replaced by
+    // `PDFIntegration` below.
+    this.anchoring = anchoring;
+
+    this.documentMeta = new DocumentMeta(this.element);
+    if (config.documentType === 'pdf') {
+      this.pdfIntegration = new PDFIntegration(this.element, this);
+    }
 
     /** @type {Anchor[]} */
     this.anchors = [];
@@ -154,8 +159,6 @@ export default class Guest extends Delegator {
     // Set the frame identifier if it's available.
     // The "top" guest instance will have this as null since it's in a top frame not a sub frame
     this.frameIdentifier = config.subFrameIdentifier || null;
-
-    this.anchoring = anchoring;
 
     const cfOptions = {
       config,
@@ -167,22 +170,13 @@ export default class Guest extends Delegator {
       },
     };
 
-    this.addPlugin('CrossFrame', cfOptions);
-    this.crossframe = this.plugins.CrossFrame;
+    this.crossframe = new CrossFrame(this.element, cfOptions);
     this.crossframe.onConnect(() => this._setupInitialState(config));
 
     // Whether clicks on non-highlighted text should close the sidebar
     this.closeSidebarOnDocumentClick = true;
     this._connectAnnotationSync();
     this._connectAnnotationUISync(this.crossframe);
-
-    // Load plugins
-    for (let name of Object.keys(this.options)) {
-      const opts = this.options[name];
-      if (!this.plugins[name] && this.options.pluginClasses[name]) {
-        this.addPlugin(name, opts);
-      }
-    }
 
     // Setup event handlers on the root element
     this._elementEventListeners = [];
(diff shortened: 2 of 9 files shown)


src/annotator/guest.js
```diff
@@ -2,6 +2,9 @@ import scrollIntoView from 'scroll-into-view';
 
 import Delegator from './delegator';
 import { Adder } from './adder';
+import CrossFrame from './plugin/cross-frame';
+import DocumentMeta from './plugin/document';
+import PDFIntegration from './plugin/pdf';
```
robertknight, line 7: I plan to move and rename these files in future so that they no longer live in a "plugin" directory. That's out of scope for this PR though.

src/annotator/guest.js
```diff
@@ -106,16 +109,11 @@ export default class Guest extends Delegator {
    * @param {HTMLElement} element -
    *   The root element in which the `Guest` instance should be able to anchor
    *   or create annotations. In an ordinary web page this typically `document.body`.
-   * @param {Record<string, any>} config
-   * @param {typeof htmlAnchoring} anchoring - Anchoring implementation
+   * @param {Record<string, any>} [config]
```
robertknight, line 112: In future it will be possible to replace the generic `Record<string,any>` type with a more specific one which lists only the fields we actually use. This is for a follow-up PR though.

src/annotator/guest.js
```diff
@@ -146,7 +144,14 @@ export default class Guest extends Delegator {
       }
     });
 
-    this.plugins = {};
+    // Set the initial anchoring module. Note that for PDFs this is replaced by
+    // `PDFIntegration` below.
```
robertknight, line 148: The way that the `anchoring` property is set in PDFs is pretty confusing. The `anchoring` property is initially set to the HTML anchoring implementation and then `PDFIntegration` replaces it. I plan to replace this in future but didn't want to include too many changes in this PR.

src/annotator/test/guest-test.js
```diff
@@ -93,7 +84,6 @@ describe('Guest', () => {
     }
 
     CrossFrame = sandbox.stub().returns(fakeCrossFrame);
-    guestConfig.pluginClasses.CrossFrame = CrossFrame;
```
robertknight, line 96: This was a method of mocking `CrossFrame` in these tests. Replaced by a more conventional method of mocking below.

src/annotator/test/guest-test.js
```diff
@@ -605,13 +572,18 @@ describe('Guest', () => {
...
+      }
+      $imports.$mock({ './plugin/document': FakeDocumentMeta });
+
       const guest = createGuest();
-      sinon.stub(guest, 'getDocumentInfo').returns(
-        Promise.resolve({
-          metadata: { title: 'hello' },
-          uri: 'http://example.com/',
-        })
-      );
```
robertknight, line 614: This changes the mocking of document metadata extraction in these tests to use our normal approach rather than monkey-patching a method of the class being tested.

src/annotator/test/integration/anchoring-test.js
```diff
@@ -2,6 +2,7 @@
 // with various combinations of selector are anchored.
 
 import Guest from '../../guest';
+import { $imports as guestImports } from '../../guest';
```
robertknight, line 5: As a convention `$imports` in a test module refers to the imports for the module being tested. This however is an integration test so it doesn't have a 1:1 association with a particular module. Hence I've given this a more specific name.

src/types/annotator.js
```diff
@@ -61,7 +61,7 @@
  *
  * @typedef AnchoringImpl
  * @prop {(root: HTMLElement, selectors: Selector[], options: any) => Promise<Range>} anchor
- * @prop {(root: HTMLElement, range: Range, options: any) => Promise<Selector[]>} describe
+ * @prop {(root: HTMLElement, range: Range, options: any) => Selector[]|Promise<Selector[]>} describe
```
robertknight, line 64: The HTML anchoring implementation has a synchronous `describe` method, whereas PDF anchoring has an async one. We might want to force both implementations to use the same interface in future. For now I've simply updated the types to match the implementation. Previously this was wrong but TypeScript wasn't able to see that.

src/annotator/config/sidebar.js
```diff
@@ -33,9 +33,7 @@ export function createSidebarConfig(config) {
   // nb. We don't currently strip all the annotator-only properties here.
   // That's OK because validation / filtering happens in the sidebar app itself.
   // It just results in unnecessary content in the sidebar iframe's URL string.
-  ['notebookAppUrl', 'sidebarAppUrl', 'pluginClasses'].forEach(
-    key => delete sidebarConfig[key]
-  );
+  ['notebookAppUrl', 'sidebarAppUrl'].forEach(key => delete sidebarConfig[key]);
```
robertknight, line 36: The `pluginClasses` key in the config no longer exists.

src/annotator/guest.js
```diff
@@ -146,44 +144,39 @@ export default class Guest extends Delegator {
...
+    //
+    // In future the plan is to put these behind a common interface so that the
+    // `Guest` class doesn't need different code paths depending on the document type.
+
+    // nb. The `anchoring` field defaults to HTML anchoring and in PDFs is replaced
+    // by `PDFIntegration` below.
+    this.anchoring = anchoring;
+    this.documentMeta = new DocumentMeta(this.element);
+    if (config.documentType === 'pdf') {
+      this.pdfIntegration = new PDFIntegration(this.element, this);
```
robertknight, line 161: The way that `PDFIntegration` immediately overwrites the `this.anchoring` property of the `Guest` that is passed in is roundabout, but fixing that is out of scope for this PR.

src/annotator/guest.js
```diff
@@ -247,26 +240,19 @@ export default class Guest extends Delegator {
...
+
+    if (this.pdfIntegration) {
+      metadataPromise = this.pdfIntegration.getMetadata();
+      uriPromise = this.pdfIntegration.uri();
     } else {
-      uriPromise = Promise.reject();
-      metadataPromise = Promise.reject();
+      uriPromise = Promise.resolve(this.documentMeta.uri());
+      metadataPromise = Promise.resolve(this.documentMeta.metadata);
     }
```
robertknight, line 256: The plan in future is that metadata extraction for different document types will have a common interface. In this PR I've just kept the existing structure.

src/annotator/guest.js
```diff
@@ -355,9 +341,8 @@ export default class Guest extends Delegator {
 
     removeAllHighlights(this.element);
 
-    for (let name of Object.keys(this.plugins)) {
-      this.plugins[name].destroy();
-    }
+    this.documentMeta.destroy();
```
robertknight, line 344: The `DocumentMeta` class doesn't actually have a `destroy` implementation, but since it currently inherits from `Delegator`, invoking `destroy` is the proper thing to do. In future we can change `DocumentMeta` to no longer inherit from `Delegator` and then this can be removed.

src/annotator/plugin/pdf.js
```diff
@@ -46,8 +46,8 @@ export default class PDF extends Delegator {
   /**
    * @param {Annotator} annotator
    */
-  constructor(element, config, annotator) {
-    super(element, config);
+  constructor(element, annotator) {
+    super(element);
```
robertknight, line 50: The `config` argument was unused, as was the `this.options` field under which `Delegator` exposes it.

src/annotator/test/guest-test.js
```diff
@@ -44,22 +35,29 @@ const timeoutPromise = (millis = 0) =>
...
   let notifySelectionChanged;
 
+  let CrossFrame;
+  let fakeCrossFrame;
+
+  let DocumentMeta;
+  let fakeDocumentMeta;
+
+  let PdfIntegration;
+  let fakePdfIntegration;
```
robertknight, line 51: I've refactored the Guest tests to use a consistent approach to mocking these classes. The `CapitalCase` var is the constructor mock. The `fakeCamelCase` var is the mock instance.

src/annotator/guest.js
```diff
@@ -478,8 +463,8 @@ export default class Guest extends Delegator {
       // Add the anchors for this annotation to instance storage.
       this._updateAnchors(this.anchors.concat(anchors));
 
-      // Let plugins know about the new information.
-      this.plugins.CrossFrame?.sync([annotation]);
+      // Let the sidebar know about the new annotation.
```
esanzgar, line 466: This comment could be more general, isn't? Now that we have the notebook, which is in another iframe, it could communicate with that as well.

src/annotator/test/guest-test.js
esanzgar: Thanks for the notes. I helps me learn about the conventions.

esanzgar (approved):
I have tested and all seems to work well.

I had a few general questions, mostly for my information.

src/annotator/guest.js
```diff
@@ -146,44 +144,39 @@ export default class Guest extends Delegator {
...
+    }
+
     // Set the frame identifier if it's available.
     // The "top" guest instance will have this as null since it's in a top frame not a sub frame
     this.frameIdentifier = config.subFrameIdentifier || null;
 
-    this.anchoring = anchoring;
-
-    const cfOptions = {
+    this.crossframe = new CrossFrame(this.element, {
```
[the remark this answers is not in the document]
robertknight: > It misses a word to describe that it is an event emitter or a communication system.

I agree. I'd avoid using the term iframe specifically though because some of the frames (usually just the main frame) listening in won't be the main frame.

src/annotator/guest.js
robertknight: Yes, that's true.

src/annotator/guest.js
```diff
@@ -146,44 +144,39 @@ export default class Guest extends Delegator {
...
-        this.subscribe(event, handler);
-      },
-      emit: (event, ...args) => {
-        this.publish(event, args);
-      },
-    };
-
-    this.addPlugin('CrossFrame', cfOptions);
-    this.crossframe = this.plugins.CrossFrame;
+      on: (event, handler) => this.subscribe(event, handler),
```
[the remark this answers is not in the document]
robertknight: The `CrossFrame` implementation only listens for `message` events that are specially formatted, so this won't fire if you `window.postMessage` arbitrary data from another frame to the current one.

PR #3093, opened 2021-03-02
Refactor Guest `createAnnotation` method and add missing tests

robertknight:
I noticed it was quite hard to follow the async flow of the `Guest#createAnnotation` method and see which properties of the annotation are guaranteed to be populated before it is sent to the sidebar. Also the method is inherently async but returns synchronously, so there is no way to properly wait for it to complete. This PR refactors `createAnnotation` using async/await to fix that. It also adds some missing tests for this important method.

Changes in detail:

- Refactor `createAnnotation` method of `Guest` to use async/await, which simplifies the control flow. I also renamed a misleading `selectors` variable
- Add tests for triggering this method by clicking on the "Annotate" or "Highlight" buttons in the adder toolbar, which is how this method gets called in the actual application
- Add tests for the `target` field of new annotations, which contains the selectors for the annotation

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -546,15 +546,16 @@ export default class Guest extends Delegator {
     annotation.document = info.metadata;
     annotation.uri = info.uri;
 
-    // `selectors` is an array of arrays: each item is an array of selectors
-    // identifying a distinct target.
     const root = this.element;
-    const selectors = await Promise.all(
+    const rangeSelectors = await Promise.all(
       ranges.map(range => this.anchoring.describe(root, range))
     );
-    annotation.target = selectors.map(selector => ({
+    annotation.target = rangeSelectors.map(selectors => ({
       source: info.uri,
-      selector,
+
+      // In the Hypothesis API the field containing the selectors is called
+      // `selector`, despite being a list.
+      selector: selectors,
     }));
 
     this.publish('beforeAnnotationCreated', [annotation]);
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -122,13 +122,13 @@ export default class Guest extends Delegator {
     this.element.appendChild(this.adderToolbar);
 
     this.adderCtrl = new Adder(this.adderToolbar, {
-      onAnnotate: () => {
-        this.createAnnotation();
+      onAnnotate: async () => {
+        await this.createAnnotation();
         /** @type {Selection} */ (document.getSelection()).removeAllRanges();
       },
-      onHighlight: () => {
+      onHighlight: async () => {
         this.setVisibleHighlights(true);
-        this.createHighlight();
+        await this.createHighlight();
         /** @type {Selection} */ (document.getSelection()).removeAllRanges();
       },
       onShowAnnotations: anns => {
@@ -572,6 +572,8 @@ export default class Guest extends Delegator {
    *
    * This flag indicates that the sidebar should save the new annotation
    * automatically and not show a form for the user to enter a comment about it.
+   *
+   * @return {Promise<AnnotationData>}
    */
   createHighlight() {
     return this.createAnnotation({ $highlight: true });
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -128,7 +128,7 @@ export default class Guest extends Delegator {
       },
       onHighlight: async () => {
         this.setVisibleHighlights(true);
-        await this.createHighlight();
+        await this.createAnnotation({ highlight: true });
         /** @type {Selection} */ (document.getSelection()).removeAllRanges();
       },
       onShowAnnotations: anns => {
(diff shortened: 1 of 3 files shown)


src/annotator/guest.js
```diff
@@ -122,13 +122,13 @@ export default class Guest extends Delegator {
     this.element.appendChild(this.adderToolbar);
 
     this.adderCtrl = new Adder(this.adderToolbar, {
-      onAnnotate: () => {
-        this.createAnnotation();
+      onAnnotate: async () => {
+        await this.createAnnotation();
```
robertknight, line 126: This change allows the tests to check when the activity triggered by clicking "Annotate" or "Highlight" is finished.

src/annotator/guest.js
```diff
@@ -535,53 +535,46 @@ export default class Guest extends Delegator {
    * @param {Partial<AnnotationData>} annotation - Initial properties of the
    *   new annotation, other than `target`, `document` and `uri` which are
    *   set by this method.
+   * @return {Promise<AnnotationData>} - The updated annotation with selectors
+   *   and metadata populated.
```
robertknight, line 539: This method used to return an `AnnotationData` despite being async internally. It now returns a Promise as it should.

src/annotator/test/guest-test.js
```diff
@@ -29,10 +29,6 @@ class FakeTextRange {
   }
 }
 
-// A little helper which returns a promise that resolves after a timeout
-const timeoutPromise = (millis = 0) =>
-  new Promise(resolve => setTimeout(resolve, millis));
-
```
robertknight, line 35: This is no longer needed because tests can explicitly await `createAnnotation`'s results.

src/annotator/test/guest-test.js
```diff
@@ -609,29 +627,55 @@ describe('Guest', () => {
...
       const annotation = {};
 
-      guest.createAnnotation(annotation);
+      await guest.createAnnotation(annotation);
 
-      return timeoutPromise().then(() => {
-        assert.equal(annotation.uri, fakeDocumentMeta.uri());
-        assert.deepEqual(annotation.document, fakeDocumentMeta.metadata);
-      });
+      assert.equal(annotation.uri, fakeDocumentMeta.uri());
```
LMS007, line 636: This test, while valid, is testing the mutated param, but not an actual returned variable. I think the function should either return nothing so it's very clear its mutating the param, or it should test the thing it actually is returning.

`const annotation = await guest.createAnnotation({});`

src/annotator/guest.js
```diff
@@ -535,53 +535,46 @@ export default class Guest extends Delegator {
...
    */
-  createAnnotation(annotation = {}) {
-    const root = this.element;
-
+  async createAnnotation(annotation = {}) {
```
LMS007, line 541: It might be a little more clear to create a new var that gets returned rather than returning the mutated param? 

```
async createAnnotation(partialAnnotation = {}) {
   const annotation = {...partialAnnotation}
```

src/annotator/guest.js
```diff
@@ -535,53 +535,46 @@ export default class Guest extends Delegator {
...
+    const root = this.element;
+    const rangeSelectors = await Promise.all(
+      ranges.map(range => this.anchoring.describe(root, range))
+    );
+    annotation.target = rangeSelectors.map(selectors => ({
+      source: info.uri,
 
-    const selectors = Promise.all(ranges.map(getSelectors));
-    const targets = Promise.all([info, selectors]).then(setTargets);
+      // In the Hypothesis API the field containing the selectors is called
```
LMS007, line 556: 👍

LMS007 (approved):
`createAnnotation` is less confusing and better tested. My only hangup was the questionable mutation of the `annotation` param and if that is still something we can to continue doing? Otherwise looks good to go.

src/annotator/guest.js
```diff
@@ -535,53 +535,46 @@ export default class Guest extends Delegator {
...
-    const setTargets = ([info, selectors]) => {
-      // `selectors` is an array of arrays: each item is an array of selectors
-      // identifying a distinct target.
-      const source = info.uri;
-      annotation.target = selectors.map(selector => ({
-        source,
-        selector,
-      }));
-    };
+    const info = await this.getDocumentInfo();
```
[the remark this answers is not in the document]
robertknight: > If you agree, perhaps we can do another clean-up pass in the future to address that?

Yes, I agree this needs further cleanup.

src/annotator/guest.js
```diff
@@ -535,53 +535,46 @@ export default class Guest extends Delegator {
...
+    annotation.document = info.metadata;
+    annotation.uri = info.uri;
 
-    const info = this.getDocumentInfo();
-    info.then(setDocumentInfo);
+    const root = this.element;
+    const rangeSelectors = await Promise.all(
+      ranges.map(range => this.anchoring.describe(root, range))
```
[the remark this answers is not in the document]
robertknight: Good question! - Currently the behavior is the same as before - no annotation gets created and an error is logged into the browser console. A future improvement may be to show some kind of notice and log error details to Sentry.

robertknight:
> My only hangup was the questionable mutation of the annotation param and if that is still something we can to continue doing?

Thanks for the feedback. It turns out the only remaining use of the mutable annotation param was to set the `$highlight` flag when clicking on the "Highlight" button. I've replaced this param with an options object that has a single `highlight` option and added some related missing tests.

PR #3094, opened 2021-03-03
Render bucket bar immediately upon construction

robertknight:
Render the bucket bar immediately using the Guest's current set of anchors when it
is constructed.

This resolves an issue introduced in 68d44395 that the background of the
bucket bar did not appear until after the first batch of annotations had
been fetched from the API and synced from the sidebar to the annotator.
It also ensures that buckets are drawn for any anchors that already
exist in the Guest when the BucketBar is initialized.

### DIFF
--- a/src/annotator/bucket-bar.js
+++ b/src/annotator/bucket-bar.js
@@ -31,6 +31,9 @@ export default class BucketBar {
       ));
       scrollableElement?.addEventListener('scroll', this.updateFunc);
     });
+
+    // Immediately render the buckets for the current anchors.
+    this._update();
   }
 
   destroy() {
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
I can confirm that I am able to reproduce the "missing bucket background" issue on `master` locally, and that this branch fixes the glitch.

robertknight:
For posterity, a clarifying comment from Slack about why this change fixes the regression:

>  Because of the removal of the .annotator-bucket-bar element in 68d44395aae4432dd4dddf81fb30b2a5db2491e1, and .Buckets__list doesn’t get rendered until the bucket-bar is updated

PR #3097, opened 2021-03-04
Stop propagation of click and touchstart events

esanzgar:
On touch devices every touch event should eventually result on a click
event, as explained [here](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent#event_order).
However, this is not true for mobile Safari versions < 13 (2019).

Because we still rely on listening to touchstart events, we decided to
sandbox the shadow DOMs, by stopping the propagation of click and
touchstart at that level. This prevents the bubbling of these events
outside the shadow DOM and better isolates the pieces of the application
that should not interact with the host page.

### DIFF
--- a/src/annotator/util/shadow-root.js
+++ b/src/annotator/util/shadow-root.js
@@ -30,6 +30,7 @@ function loadStyles(shadowRoot) {
  */
 export function createShadowRoot(container) {
   if (!container.attachShadow) {
+    stopEventPropagation(container);
     return container;
   }
 
@@ -43,5 +44,24 @@ export function createShadowRoot(container) {
     applyFocusVisible(shadowRoot);
   }
 
+  stopEventPropagation(shadowRoot);
   return shadowRoot;
+}
+
+/**
+ * Stop bubbling up of 'click' and 'touchstart' events.
+ *
+ * This makes the host page a little bit less aware of the annotator activity.
+ * It is still possible for the host page to manipulate the events on the capturing
+ * face.
+ *
+ * Another benefit is that click and touchstart typically causes the sidebar to close.
+ * By preventing the bubble up of these events, we don't have to individually stop
+ * the propagation.
+ *
+ * @param {HTMLElement|ShadowRoot} element
+ */
+function stopEventPropagation(element) {
+  element.addEventListener('click', event => event.stopPropagation());
+  element.addEventListener('touchstart', event => event.stopPropagation());
 }
--- a/src/annotator/util/test/shadow-root-test.js
+++ b/src/annotator/util/test/shadow-root-test.js
@@ -14,6 +14,7 @@ describe('annotator/util/shadow-root', () => {
     container.remove();
     window.applyFocusVisiblePolyfill = applyFocusVisiblePolyfill;
   });
+
   describe('createShadowRoot', () => {
     it('attaches a shadow root to the container', () => {
       const shadowRoot = createShadowRoot(container);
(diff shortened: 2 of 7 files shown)


src/annotator/util/shadow-root.js
```diff
@@ -30,6 +30,7 @@ function loadStyles(shadowRoot) {
  */
 export function createShadowRoot(container) {
   if (!container.attachShadow) {
+    // stopEventPropagation(container);
```
esanzgar, line 33: @robertknight, what should we do if shadow DOMs are not available? Should we still stop the propagation of events? It all depends on the structure of the page.

esanzgar:
Related to furthering the isolation: I wonder if we should change the shadow DOM mode from `open` to `close` so we enforce that all interaction and communication between all parts of the client app is done via the devised event communication system.

src/annotator/util/shadow-root.js
robertknight: Can you see any problems caused by stopping propagation of events on the `container` element instead of the shadow root?

esanzgar: On a second thought, I don't see any problems.

Propagation of events happens parent -> child and child -> parent.

```
A host element
        |
    - Stop propagation -
        |
        | -> annotator element
```

robertknight: OK, let's do this then.

esanzgar: My only concerns was if the `container` element was ever host element.

I can see that `createShadowRoot` is only used 3 times on the following elements created by the annotator:
- `hypothesis-adder`
- `hypothesis-sidebar`
- `hypothesis-notebook`

Hence, I there is no risk of affecting the host page in anyway.

esanzgar: OK, I have uncommented this line.

src/annotator/util/shadow-root.js
```diff
@@ -43,5 +44,18 @@ export function createShadowRoot(container) {
     applyFocusVisible(shadowRoot);
   }
 
+  stopEventPropagation(shadowRoot);
   return shadowRoot;
 }
+
+/**
+ * Stop bubbling up of the 'click' and 'touchstart' events.
```
robertknight, line 52:
```suggestion
 * Stop bubbling up of 'click' and 'touchstart' events.
```

Can you add a couple of sentences after the summary about the reason for doing this, for future readers. Can you also add a blank line between the documentation and the param docs. In other words, format JSDoc as:

```js
/**
 * Short summary
 *
 * More detailed notes.
 *
 * @param ...
 */

File: src/annotator/util/shadow-root.js [hunk 3]
```diff
@@ -43,5 +44,18 @@ export function createShadowRoot(container) {
...
+ * @param {HTMLElement|ShadowRoot} element
+ */
+function stopEventPropagation(element) {
+  element.addEventListener('click', event => event.stopImmediatePropagation());
```
robertknight, line 56: Is the use of `stopImmediatePropagation` vs `stopPropagation` here important? From looking at the code in the `Guest` class it looks like it adds the event listener to `this.element` which is `document.body` (see annotator/index.js). I can see that it is needed with the way that the tests have been written.

src/annotator/util/test/shadow-root-test.js
```diff
@@ -56,5 +57,33 @@ describe('annotator/util/shadow-root', () => {
       assert.isNull(linkEl);
       link.setAttribute('rel', 'stylesheet');
     });
+
+    it('stops propagation of click events', () => {
+      const onClick = sinon.stub();
+      container.addEventListener('click', onClick);
+
+      const shadowRoot = createShadowRoot(container);
+      const innnerElement = document.createElement('div');
```
robertknight, line 66:
```suggestion
      const innerElement = document.createElement('div');
```

src/annotator/util/test/shadow-root-test.js
```diff
@@ -56,5 +57,33 @@ describe('annotator/util/shadow-root', () => {
...
+
+      assert.notCalled(onClick);
+    });
+
+    it('stops propagation of touchstart events', () => {
+      const onTouch = sinon.stub();
+      container.addEventListener('touchstart', onTouch);
+
+      const shadowRoot = createShadowRoot(container);
+      const innnerElement = document.createElement('div');
```
robertknight, line 80:
```suggestion
      const innerElement = document.createElement('div');
```

src/annotator/util/shadow-root.js
```diff
@@ -43,5 +44,18 @@ export function createShadowRoot(container) {
     applyFocusVisible(shadowRoot);
   }
 
+  stopEventPropagation(shadowRoot);
```
robertknight, line 47: I wasn't sure if this was needed given that only events that are specifically marked as `composed` propagate across shadow DOM boundaries, but [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed) this is the case for all browser-generated events:

> All UA-dispatched UI events are composed (click/touch/mouseover/copy/paste, etc.).

robertknight (approved):
I found a couple of typos and requested some additional docs for `stopEventPropagation`. Otherwise LGTM.

src/annotator/util/shadow-root.js
esanzgar: You are absolutely correct.

In one of my tests, it took me a while to emulate the browser-generate event `element.click()` which propagated through the shadow DOM, with the `new Event('click')` which didn't. The reason was the `composed` property.

esanzgar: I will add a comment in these line for future reference:
 https://github.com/hypothesis/client/blob/cb4af8934c37253059c0de71945eb0b3378ae8bb/src/annotator/util/test/shadow-root-test.js#L69

src/annotator/util/shadow-root.js
esanzgar: No, it is not important. Both will stop the propagation. `stopImmediatePropagation` in addition to stop the propagation will prevent other events attached to the same element to be called.
https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation

It is a subtle and in this context unimportant.

I will change `stopImmediatePropagation` for `stopPropagation` and fix the tests.

src/annotator/util/shadow-root.js
esanzgar: No problem, I will do.

PR #3099, opened 2021-03-04
Added utility to manage collection of listeners

esanzgar:
Following on @robertknight [suggestion](https://github.com/hypothesis/client/pull/3035#pullrequestreview-599668554), I created a tiny utility to make easier to unregister event listeners.

### DIFF
--- a/src/annotator/util/listener-collection.js
+++ b/src/annotator/util/listener-collection.js
@@ -1,0 +1,35 @@
+/**
+ * @typedef Listener
+ * @prop {EventTarget} eventTarget
+ * @prop {string} eventType
+ * @prop {(event: Event) => void} listener
+ */
+
+/**
+ * Utility that provides a way to conveniently remove a set of DOM event
+ * listeners when they are no longer needed.
+ */
+export class ListenerCollection {
+  constructor() {
+    /** @type {Listener[]} */
+    this._listeners = [];
+  }
+
+  /**
+   * @param {Listener['eventTarget']} eventTarget
+   * @param {Listener['eventType']} eventType
+   * @param {Listener['listener']} listener
+   * @param {AddEventListenerOptions} [options]
+   */
+  add(eventTarget, eventType, listener, options) {
+    eventTarget.addEventListener(eventType, listener, options);
+    this._listeners.push({ eventTarget, eventType, listener });
+  }
+
+  removeAll() {
+    this._listeners.forEach(({ eventTarget, eventType, listener }) => {
+      eventTarget.removeEventListener(eventType, listener);
+    });
+    this._listeners = [];
+  }
+}
--- a/src/annotator/util/test/listener-collection-test.js
+++ b/src/annotator/util/test/listener-collection-test.js
@@ -1,0 +1,33 @@
+import { ListenerCollection } from '../../util/listener-collection';
+
+describe('ListenerCollection', () => {
+  let listeners;
+
+  beforeEach(() => {
+    listeners = new ListenerCollection();
+  });
+
+  afterEach(() => {
+    listeners.removeAll();
+  });
+
+  it('registers and triggers event listener', () => {
+    const listener = sinon.stub();
+    listeners.add(window, 'resize', listener);
+
+    window.dispatchEvent(new Event('resize'));
+    assert.calledOnce(listener);
+  });
+
+  it('unregisters event listeners', () => {
+    const listener1 = sinon.stub();
+    const listener2 = sinon.stub();
+    listeners.add(window, 'resize', listener1);
+    listeners.add(window, 'resize', listener2);
+    listeners.removeAll();
+
+    window.dispatchEvent(new Event('resize'));
+    assert.notCalled(listener1);
+    assert.notCalled(listener2);
+  });
+});
--- a/src/annotator/bucket-bar.js
+++ b/src/annotator/bucket-bar.js
@@ -2,6 +2,7 @@ import { render } from 'preact';
 import Buckets from './components/Buckets';
 
 import { anchorBuckets } from './util/buckets';
+import { ListenerCollection } from './util/listener-collection';
 
 /**
  * @typedef BucketBarOptions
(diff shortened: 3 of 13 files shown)


src/annotator/util/listener-collection.js
```diff
@@ -0,0 +1,34 @@
...
+ * @typedef Listener
+ * @prop {Window|Element|Document} eventTarget
+ * @prop {string} eventType
+ * @prop {(event: Event) => void} listener
+ */
+
+/**
+ * Utility that prevents forgetting to remove event listener.
+ */
+export function listenerCollection() {
```
robertknight, line 11: Was there a reason for making this a function that returns an object rather than a class? I know that in some parts of the sidebar (eg. services) this pattern is common, but that's mostly because it pre-dates ES classes. In modern code we tend to use actual classes partly because it makes it more obvious to the reader what they are looking at and partly because the tooling understands it better.

src/annotator/util/test/listener-collection-test.js
```diff
@@ -0,0 +1,30 @@
+import { listenerCollection } from '../../util/listener-collection';
+
+describe('register/unregister event listener', () => {
```
robertknight, line 3: As a general rule, the top-level `describe` block in a file should either be named after the class/function being tested, if there is only one and it is a reasonably unique name, or the path relative to `src/` otherwise (`annotator/util/listener-collection`) in this context. The reason is just to make it easier to know what file to go looking at if the tests fail.

src/annotator/util/listener-collection.js
```diff
@@ -0,0 +1,34 @@
...
+  /** @type{Listener[]} */
+  let listeners = [];
+
+  /**
+   * @param {Listener['eventTarget']} eventTarget
+   * @param {Listener['eventType']} eventType
+   * @param {Listener['listener']} listener
+   * @param {boolean | AddEventListenerOptions | undefined} [options]
+   */
+  function register(eventTarget, eventType, listener, options) {
```
robertknight, line 21: I wonder about naming these `add`/`removeAll` to be more similar to `addEventListener` and `removeEventListener`. Thoughts?

robertknight (commented):
This helper looks like it will be useful. I added some thoughts on the API design.

src/annotator/util/listener-collection.js
esanzgar: It was a personal choice. I will transform the function into a class for consistency with the rest of the code.

src/annotator/util/listener-collection.js
esanzgar: No issues.

src/annotator/util/test/listener-collection-test.js
esanzgar: 👍🏼

src/annotator/util/listener-collection.js
esanzgar: There is a however an advantage in the context of react functional components, it doesn't need to be wrap on a useEffect or useMemo;

/Users/esanzgar/Development/hypothesis/client/src/sidebar/components/ThreadList.js
  55:9  error  The 'listeners' object construction makes the dependencies of useEffect Hook (at line 171) change on every render. Move it inside the useEffect callback. Alternatively, wrap the initialization of 'listeners' in its own useMemo() Hook

esanzgar:
@robertknight I have implemented your suggestions.

src/annotator/util/test/listener-collection-test.js
```diff
@@ -0,0 +1,30 @@
...
+
+  it('registers and triggers event listener', () => {
+    const listener = sinon.stub();
+    listeners.add(window, 'resize', listener);
+
+    window.dispatchEvent(new Event('resize'));
+    assert.calledOnce(listener);
+  });
+
+  it('unregister event listeners', () => {
```
robertknight, line 22:
```suggestion
  it('unregisters event listeners', () => {
```

robertknight (changes requested):
This looks good except that I think the various `listeners` fields should all have an underscore prefix to mark them as private. I know that some of the existing code in the annotator classes does not adhere to this convention, but I do think it is important for future maintenance to be clear about what are intended to be private vs public fields and methods. Fixing up the existing code to be consistent in this respect would be a good thing to do, but separate from this PR!

src/annotator/util/listener-collection.js
```diff
@@ -0,0 +1,34 @@
...
+  constructor() {
+    /** @type{Listener[]} */
+    this.listeners = [];
+  }
+
+  /**
+   * @param {Listener['eventTarget']} eventTarget
+   * @param {Listener['eventType']} eventType
+   * @param {Listener['listener']} listener
+   * @param {boolean | AddEventListenerOptions | undefined} [options]
```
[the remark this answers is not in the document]
esanzgar: Good idea

esanzgar:
Thank you for the feedback.

- I have renamed `_listeners` the private property as per convention. 
- Changed some types as suggested

src/annotator/util/listener-collection.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @typedef Listener
+ * @prop {EventTarget} eventTarget
+ * @prop {string} eventType
+ * @prop {(event: Event) => void} listener
+ */
+
+/**
+ * Utility that prevents forgetting to remove event listener.
```
robertknight, line 9: The wording is perhaps a little strong here. It only avoids failing to remove event listeners if you use it correctly. Perhaps something like "Utility that provides a way to conveniently remove a set of DOM event listeners when they are no longer needed."?

src/annotator/util/test/listener-collection-test.js
robertknight: Hi Eduardo - This was marked as resolved but the suggestion wasn't applied.

src/annotator/util/test/listener-collection-test.js
```diff
@@ -0,0 +1,30 @@
...
+    const listener = sinon.stub();
+    listeners.add(window, 'resize', listener);
+    listeners.removeAll();
```
robertknight, line 25: I think it would be good to add a couple of listeners in this test, to verify that _all_ of them are removed.

robertknight (approved):
LGTM

src/annotator/util/listener-collection.js
esanzgar: Thanks

src/annotator/util/test/listener-collection-test.js
esanzgar: Sorry, for that. I fixed that.

PR #3100, opened 2021-03-04
Avoid leaving elements from bucket-bar test

esanzgar:
These changes enforce that the bucket-bar is always destroyed and allows multiple creations of the bucket bar, if needed.

### DIFF
--- a/src/annotator/test/bucket-bar-test.js
+++ b/src/annotator/test/bucket-bar-test.js
@@ -5,15 +5,18 @@ describe('BucketBar', () => {
   const sandbox = sinon.createSandbox();
   let fakeAnnotator;
   let fakeBucketUtil;
-  let bucketBar;
+  let bucketBars;
   let bucketProps;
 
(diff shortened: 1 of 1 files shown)


LMS007 (approved):
Looks reasonable Eduardo , can you help me understand the circumstance where 2 bars might be created in a test?

esanzgar:
Multiple bucket bars could be created by using loops in a tests like this: https://github.com/hypothesis/client/blob/cebe7dbd0e98772c7ed7d92f3a7be2d8440de0cc/src/annotator/test/guest-test.js#L475

esanzgar:
The last test in this file was using a reference to a bucket bar that was created previously. 
https://github.com/hypothesis/client/blob/cebe7dbd0e98772c7ed7d92f3a7be2d8440de0cc/src/annotator/test/bucket-bar-test.js#L115

This PR enforce to create a bucket bar instance each time.

PR #3102, opened 2021-03-05
Git ignore yalc related files

esanzgar:
The @hypothesis/frontend-shared lives now in another repo. The
recommended way to test the package in the context of the client is
using yalc. Therefore, it is better to ignores yalc related files so
they are not committed by mistake.


robertknight:
👍

LMS007 (approved):
Thanks Eduardo

PR #3103, opened 2021-03-05
Remove element bucket-bar upon destruction

esanzgar:
This was never an issue because the parent elements are destroyed,
however for completeness is better to add it.

In addition, I modified the test so that all the elements created are
added to a single parent container who is deleted after each test.

### DIFF
--- a/src/annotator/test/bucket-bar-test.js
+++ b/src/annotator/test/bucket-bar-test.js
@@ -7,15 +7,16 @@ describe('BucketBar', () => {
   let fakeBucketUtil;
   let bucketBars;
   let bucketProps;
+  let container;
 
   const createBucketBar = function (options) {
-    const element = document.createElement('div');
-    const bucketBar = new BucketBar(element, fakeAnnotator, options);
+    const bucketBar = new BucketBar(container, fakeAnnotator, options);
     bucketBars.push(bucketBar);
     return bucketBar;
   };
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


src/annotator/test/bucket-bar-test.js
```diff
@@ -113,5 +113,11 @@ describe('BucketBar', () => {
       bucketBar.update();
       assert.notCalled(window.requestAnimationFrame);
     });
+
+    it('deletes the bucketbar element after destroy method is called', () => {
+      const bucketBar = createBucketBar();
+      bucketBar.destroy();
+      assert.isNull(bucketBar.element.parentElement);
```
LMS007, line 120: I would do the assert from the test node rather than peaking into the bucketBar's member `element`. If we change it, then we can treat it more like a blackbox and have less code-specific tests.

`assert.isFalse(container.hasChildNodes())`

I noticed a previous test also reference .element, but possibly could be refactored to use `container`
OR  is `element` intended to be public?

LMS007 (approved):
Looks good, I just had one suggestion on the assert to make it less code dependent, but I didn't have full context on this module, so please override my suggestion if .element is in fact public

src/annotator/test/bucket-bar-test.js
esanzgar: It is a good suggestion. Thank you very much.

PR #3107, opened 2021-03-08
Change `BucketBar` to support only a single content container element

robertknight:
In current usage of the `BucketBar` we only ever have a single scrollable container to observe. Simplify the `BucketBar` API to pass in this element directly rather than a selector and add associated tests.

This change is one of several pieces of refactoring to make it easier to encapsulate all the document-type/viewer specific logic (eg. for PDFs, HTML documents, ebooks) in the annotator into a single place, so we can more easily add customizations for different document types (eg. ebook viewers). See also [this Slack thread](https://hypothes-is.slack.com/archives/C1M8NH76X/p1614684393090400?thread_ts=1614684393.090400). Note that for new document types the `BucketBar` might have to work completely differently (eg. if the content is in a different iframe or scrolls left-to-right instead of top-to-bottom). For now the interface only has the flexibility required by the currently supported types (HTML, PDF).

### DIFF
--- a/src/annotator/bucket-bar.js
+++ b/src/annotator/bucket-bar.js
@@ -6,7 +6,8 @@ import { anchorBuckets } from './util/buckets';
 /**
  * @typedef BucketBarOptions
  * @prop {Element} [contentContainer] - The scrollable container element for the
- *   highlights that the bucket bar's buckets will point at
+ *   document content. All of the highlights that the bucket bar's buckets point
+ *   at should be contained within this element.
  */
 
 export default class BucketBar {
--- a/src/annotator/bucket-bar.js
+++ b/src/annotator/bucket-bar.js
@@ -5,7 +5,8 @@ import { anchorBuckets } from './util/buckets';
 
 /**
  * @typedef BucketBarOptions
- * @prop {string[]} [scrollables] - Selectors for the scrollable elements on the page
+ * @prop {Element} [contentContainer] - The scrollable container element for the
+ *   highlights that the bucket bar's buckets will point at
  */
 
 export default class BucketBar {
@@ -14,8 +15,8 @@ export default class BucketBar {
    * @param {Pick<import('./guest').default, 'anchors'|'selectAnnotations'>} guest
    * @param {BucketBarOptions} [options]
    */
-  constructor(container, guest, options = {}) {
-    this.options = options;
+  constructor(container, guest, { contentContainer = document.body } = {}) {
+    this._contentContainer = contentContainer;
     this.element = document.createElement('div');
 
     this.guest = guest;
@@ -25,12 +26,7 @@ export default class BucketBar {
 
     window.addEventListener('resize', this.updateFunc);
     window.addEventListener('scroll', this.updateFunc);
-    this.options.scrollables?.forEach(scrollable => {
-      const scrollableElement = /** @type {HTMLElement | null} */ (document.querySelector(
-        scrollable
-      ));
-      scrollableElement?.addEventListener('scroll', this.updateFunc);
-    });
+    contentContainer.addEventListener('scroll', this.updateFunc);
 
     // Immediately render the buckets for the current anchors.
     this._update();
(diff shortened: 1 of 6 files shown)


src/annotator/bucket-bar.js
```diff
@@ -5,7 +5,8 @@ import { anchorBuckets } from './util/buckets';
 
 /**
  * @typedef BucketBarOptions
- * @prop {string[]} [scrollables] - Selectors for the scrollable elements on the page
+ * @prop {Element} [contentContainer] - The scrollable container element for the
```
esanzgar, line 8: I would suggest `scrollableContainer`

esanzgar: or better `scrollableElement`

src/annotator/test/bucket-bar-test.js
```diff
@@ -80,30 +80,37 @@ describe('BucketBar', () => {
...
       beforeEach(() => {
-        const scrollableEls1 = document.createElement('div');
-        scrollableEls1.className = 'scrollable-1';
-        const scrollableEls2 = document.createElement('div');
-        scrollableEls2.className = 'scrollable-2';
-        scrollableEls.push(scrollableEls1, scrollableEls2);
-        document.body.appendChild(scrollableEls1);
-        document.body.appendChild(scrollableEls2);
+        container = document.createElement('div');
+        container.className = 'scrollable-1';
```
esanzgar, line 88: It is not used, so maybe it could be deleted.

src/annotator/sidebar.js
```diff
@@ -89,11 +89,9 @@ export default class Sidebar extends Delegator {
...
         this.iframeContainer.classList.add('annotator-frame--theme-clean');
       } else {
-        const bucketBar = new BucketBar(
-          this.iframeContainer,
-          guest,
-          config.BucketBar
-        );
+        const bucketBar = new BucketBar(this.iframeContainer, guest, {
+          contentContainer: config.contentContainer || document.body,
+        });
```
esanzgar, line 94: I would rather not overwrite the default value in the bucket bar constructor.
What about?
```suggestion
        const bucketBar = new BucketBar(this.iframeContainer, guest, {
          contentContainer: config.contentContainer
          });
```
Or even:
```suggestion
        const bucketBar = new BucketBar(this.iframeContainer, guest, config);
```

esanzgar (approved):
👍🏼

src/annotator/bucket-bar.js
esanzgar: `contentContainer` is ambiguous. It seems to refer to the container of bucket-bar itself.

robertknight: `scrollableContainer` is fine with me.

src/annotator/sidebar.js
robertknight: I see a problem with passing `config` here because it ends up passing all of the sidebar's configuration options to the bucket bar. That could potentially have unintended consequences.

src/annotator/bucket-bar.js
robertknight: Hmm ... I'm having second thoughts about this. `scrollableContainer` doesn't really capture what I mean. This element is really the container for the _document content_ (hence the original name) and all of the highlights. "content" is short for "document content". In future we might want to observe other kinds of events on this element - such as resizing which might alter the layout.

src/annotator/test/bucket-bar-test.js
robertknight: Indeed. I'll remove this.

src/annotator/sidebar.js
robertknight: I do agree about not overriding `BucketBar`'s defaults when no `contentContainer` is specified. That leaves the responsibility of how to observe "the whole (HTML) document scrolling" up to `BucketBar`, which makes more sense.

robertknight:
Thanks for the feedback Eduardo. I've made a few changes (see last commits) in response. Regarding the `contentContainer` name I kept it for now, see my comment above for the rationale, but revised the documentation.

PR #3118, opened 2021-03-08
Remove prop types from components

robertknight:
See individual commit messages for details. The only non-obvious change is that `mockImportedComponents` now needs a different (and not completely accurate, but Good Enough™) method to detect whether an import is a component or not.

Fixes https://github.com/hypothesis/client/issues/3030

### DIFF
--- a/src/sidebar/components/MenuKeyboardNavigation.js
+++ b/src/sidebar/components/MenuKeyboardNavigation.js
@@ -1,6 +1,5 @@
 import { normalizeKeyName } from '@hypothesis/frontend-shared';
 import { useEffect, useRef } from 'preact/hooks';
-import propTypes from 'prop-types';
 
 function isElementVisible(element) {
   return element.offsetParent !== null;
@@ -110,18 +109,3 @@ export default function MenuKeyboardNavigation({
     </div>
   );
 }
-
-MenuKeyboardNavigation.propTypes = {
-  className: propTypes.string,
-
-  // Callback when the menu is closed via keyboard input
-  closeMenu: propTypes.func,
-
-  // When  true`, sets focus on the first item in the list
-  // which has a role="menuitem" attribute.
-  visible: propTypes.bool,
-
-  // Array of nodes which may contain <MenuItems> or any nodes
-  // that have role set to 'menuitem', 'menuitemradio', or 'menuitemcheckbox'
-  children: propTypes.any.isRequired,
-};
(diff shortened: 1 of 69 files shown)


robertknight:
Lyza is out of action today, would you mind taking a look Kyle?

src/sidebar/components/MenuKeyboardNavigation.js
```diff
@@ -110,18 +109,3 @@ export default function MenuKeyboardNavigation({
...
-  className: propTypes.string,
-
-  // Callback when the menu is closed via keyboard input
-  closeMenu: propTypes.func,
-
-  // When  true`, sets focus on the first item in the list
-  // which has a role="menuitem" attribute.
-  visible: propTypes.bool,
-
-  // Array of nodes which may contain <MenuItems> or any nodes
```
LMS007, line 124: The comment for `children` does not match that in `MenuKeyboardNavigationProps`, Generally speaking, I actually don't know our source of truth in these cases, but I'm included to just go with what we have for the jsdocs.

src/sidebar/components/ShareLinks.js
```diff
@@ -23,16 +32,12 @@ function ShareLink({ label, iconName, uri, onClick }) {
...
-  /** Accessible label / tooltip for link. */
-  label: propTypes.string.isRequired,
-  /** URI for sharing this annotation on the given social service */
-  uri: propTypes.string.isRequired,
-  /** click callback (for analytics tracking) */
-  onClick: propTypes.func.isRequired,
-};
+/**
+ * @typedef ShareLinksProps
+ * @prop {object} analytics
```
LMS007, line 37: We use `{Object}` a lot more than `{object}`, it might be nice at some point to get em all the same.

src/sidebar/service-context.js
```diff
@@ -44,7 +44,7 @@ export const ServiceContext = createContext(fallbackInjector);
  * Wrap a React component to inject any services it depends upon as props.
  *
  * Components declare their service dependencies in an `injectedProps` static
- * property. These services also need to be declared in `propTypes`.
```
LMS007, line 47: 👍

LMS007 (approved):
GTG!

src/sidebar/components/MenuKeyboardNavigation.js
robertknight: The JSDoc is the only source of truth going forwards :)

src/sidebar/components/ShareLinks.js
robertknight: We discussed on Slack, but `object` is more specific (a non-primitive) than `Object` (which is an alias for `any`) and so `object` is preferable where applicable in new code.

PR #3121, opened 2021-03-09
Replace inheritance from Delegator

esanzgar:
Delegator class was a pub/sub event emitter for communication between
different parts of the application in the host page.

Instead of inherit this functionality, I have replaced by composition.

Some arguments send by this system were unnecessarily wrapped in an
array. I have unwrapped these arguments.


robertknight (commented):
Hello Eduardo,

Thanks for having a go at revising this code. The current `Delegator` name and its `config` argument is a vestige of its original use. I've had a look through the revised code and have some thoughts on the design.

----

There are two kinds of entity involved in event handling:

1. The event bus
2. Event emitters/subscribers which are connected to the event bus

In the current code in master, the event bus is state associated with the root element passed to the `Guest`/`Sidebar` etc. constructors and `Delegator` instances are emitters/subscribers connected to the bus belonging to the DOM element.

In the proposed design, the event bus is implicit global state (`GuestEmitter.emitter`) and `GuestEmitter` instances are all subscribers to this. I see a couple of problems with this:

- In order to understand `GuestEmitter` you really need to be aware of the event bus and its relation to individual subscribers. Because it is created implicitly it is easy not to really be aware of it.
- From past experience we've seen that global state is a hazard in tests because it allows one set of tests to inadvertently interfere with another apparently unrelated set.

My main suggestion here would be to make the event bus an explicit concept, to make it more obvious for the reader how the classes are interacting and so that each test suite can use a separate event bus. Here is a sketch of one possible approach to that:

```js
// In annotator/index.js
const guest = new Guest(eventBus, ...);
const sidebar = new Sidebar(eventBus, ...);

// In Guest constructor
this.eventEmitter = eventBus.createEmitter(); // or `createSubscriber`?

// Use of `this.eventEmitter` is then the same as `this.guestEmitter` in this PR,
// including calling `destroy()` when the consumer is destroyed.
```

As well as making the event bus concept explicit, this also means that in tests you don't need to access the private `eventEmitter` instance of whatever you are testing. Instead you create the event bus from outside and emit events on it. The class that you are testing can handle these events internally however it normally would. This approach also matches better what happens in the actual application.

esanzgar:
Thank you very much for the suggested design approach, Robert.

On a second commit I implemented it.

src/annotator/util/test/emitter-test.js
```diff
@@ -0,0 +1,53 @@
...
+    assert.calledOnce(callback);
+    assert.calledWith(callback, 'foo', 'bar');
+
+    emitterB.unsubscribe('someEvent', callback);
+    emitterA.publish('someEvent', 'foo', 'bar');
+
+    assert.calledOnce(callback);
+  });
+
+  it('fires events only emitters using the same eventBus', () => {
```
robertknight, line 22:
```suggestion
  it('fires events only to emitters using the same EventBus', () => {
```

src/annotator/util/test/emitter-test.js
```diff
@@ -0,0 +1,53 @@
+import EventBus from '../emitter';
+
+describe('EvenBus', () => {
```
robertknight, line 3:
```suggestion
describe('Emitter', () => {
```

Although these tests use `EventBus`, I think it would be more accurate to describe them as a set of tests for the `Emitter` class.

src/shared/discovery.js
```diff
@@ -33,15 +33,17 @@
...
+   * @param {string} [options.origin]
    */
   constructor(target, options = {}) {
     /** The window to send and listen for messages with. */
     this.target = target;
 
     /**
      * Set whether this frame acts as a server (fetches annotations from the
-     * API) or a client (contains annotatable content and displays highlights).
+     * API) or a client (contains annotable content and displays highlights).
```
robertknight, line 46: The previous spelling was actually correct here. See https://en.wiktionary.org/wiki/annotatable.

src/annotator/util/emitter.js
```diff
@@ -0,0 +1,74 @@
...
+   */
+  destroy() {
+    for (let [event, callback] of this._subscriptions) {
+      this._emitter.off(event, callback);
+    }
+    this._subscriptions = [];
+  }
+}
+
+export default class EventBus {
```
robertknight, line 66: In recent code we've started to move away from using default exports in most scenarios. The reason is that it can lead to the same type unintentionally ending up with different names in different files which can be confusing. In other words `import FooBar from './util/some-module` can silently "rename" whatever the actual name of the default export is to `FooBar`. If importing by name you end up making this explicit at least (`import { Foo as FooBar } ...`. There is an exception for Preact components to follow ecosystem conventions.

src/annotator/test/guest-test.js
```diff
@@ -48,7 +48,8 @@ describe('Guest', () => {
 
   const createGuest = (config = {}) => {
     const element = document.createElement('div');
-    return new Guest(element, { ...guestConfig, ...config });
+    const eventBus = new EventBus();
```
robertknight, line 51: Generally it is preferable for tests to avoid accessing the private API (eg. `_emitter`) for the classes they are testing. Here you could avoid that by creating the `EventBus` in the `beforeEach`, saving it as an `eventBus` variable in the scope of the top-level describe block and then calling `eventBus.emit(...)` in individual tests. That wouldn't really have any immediate practical benefits, although it might discourage future tests from using private APIs when they shouldn't be.

robertknight (approved):
LGTM, with one stylistic request which is to make `EventBus` a named rather than a default export and a few small suggestions.

There is one other piece of feedback I had that I'm ambivalent about, which is whether tests should use `someObject._emitter.publish(...)` vs capturing a reference to the `EventBus` from outside and then calling `eventBus.emit(...)` externally.

src/shared/discovery.js
esanzgar: Corrected

src/annotator/util/emitter.js
esanzgar: I fully agree. Generally, I dislike default exports.

src/annotator/test/guest-test.js
esanzgar: I followed your suggestion of no accessing private properties.

esanzgar: Furthermore, it would be confusing to expose the internal API of the private TinyEmitter class. Generally, developers should use the API of `Emitter`.

PR #3138, opened 2021-03-12
Optimize annotation fetch order in notebook

robertknight:
Reduce the perceived loading time in the notebook by fetching
annotations in, roughly, the reverse chronological order in which they
are displayed. In my testing this makes a big difference in the time-until-useful-content
in the Hypothesis Reading group (~1400 threads, ~1900 annotations) and my personal
notes group (~1500 threads, ~1700 annotations).

By default `SearchClient` fetches annotations by ascending creation
date. This makes sense in the sidebar as it typically correlates,
roughly, with the order in which the sidebar presents annotations by
default (document location order). For the notebook however this is
suboptimal.

 - Add `sortBy` and `sortOrder` options to `SearchClient` and
   `loadAnnotations` to control fetch order, with the defaults remaining
   the same as before (`created`, `asc`).
 - In `NotebookView` set `sortBy` to `updated` and `sortOrder` to desc
   to fetch annotations in reverse chronological order.

### DIFF
--- a/src/sidebar/components/NotebookView.js
+++ b/src/sidebar/components/NotebookView.js
@@ -52,6 +52,19 @@ function NotebookView({ loadAnnotationsService }) {
       loadAnnotationsService.load({
         groupId: focusedGroup.id,
         maxResults: 5000,
+
+        // Load annotations in reverse-chronological order because that is how
+        // threads are sorted in the notebook view. By aligning the fetch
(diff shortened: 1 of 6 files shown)


lyzadanger (approved):
Bravo. A really nice improvement (I can see significant differences even locally) and I can't find a _single nit_ to complain about in code or commenting. SHIP IT.

PR #3162, opened 2021-03-16
Use term "page" rather than "chunk" or "batch" in SearchClient

robertknight:
Consistently use the term "page" to refer to a subset of results fetched
from the search API instead of "batch" or "chunk". This is more
consistent within the rest of the client code and also consistent with
the API documentation.

 - Rename `chunkSize` constructor argument to `pageSize`
 - Change internal references to "batch" or "chunk" in SearchClient to
   "page"


lyzadanger (approved):
I really like this change. It had sort of partially occurred to me yesterday when reviewing other related changes... yay!

PR #3170, opened 2021-03-18
Remove Google Analytics from the client

### LINKED ISSUE #3027: Remove Google Analytics from the client
Google Analytics has been turned off for the Hypothesis client in production for over a year, and we have no plans to re-enable it any time soon.

We do have some analytics use cases around finding out a) which sites embed Hypothesis, b) how users interact with various features of the client and c) which browsers are being used. We currently think these are likely to be better served by other tools. In particular for c) we have https://github.com/hypothesis/user-agent-analysis.

On this basis, we've got the go-ahead to remove this code from the client.

For more details, see https://hypothes-is.slack.com/archives/C4K6M7P5E/p1613564158208800

robertknight:
We have had Google Analytics disabled for the Hypothesis client for over
a year and do not plan to re-enable it. We may need analytics for other
purposes in future (for example, to help inform product design). In that
case we will likely implement a solution that is a better fit for
whatever those needs end up being.

This commit removes the existing analytics code in the meantime so that
we don't spend time maintaining something that is not used.

- Remove the code that loads the Google Analytics script (`src/sidebar/ga.js`)
- Remove the `analytics` service which was a wrapper around GA
- Remove all usage of the `analytics` service across various components

Fixes #3027


LMS007 (approved):
👍

PR #3172, opened 2021-03-19
Reload iframe on every `openNotebook` event

esanzgar:
While there is no mechanism to sync new annotations in the notebook, we
force re-rendering of the iframe on every 'openNotebook' event, so we
fetch the new annotations.

This is a temporary fix that is intended to be removed.

### DIFF
--- a/src/annotator/components/NotebookModal.js
+++ b/src/annotator/components/NotebookModal.js
@@ -26,7 +26,6 @@ function NotebookIframe({ config, groupId }) {
 
   return (
     <iframe
-      key={groupId} // force re-rendering on group change
       title={'Hypothesis annotation notebook'}
       className="Notebook__iframe"
       // Enable media in annotations to be shown fullscreen
@@ -48,6 +47,11 @@ function NotebookIframe({ config, groupId }) {
  * @param {NotebookModalProps} props
  */
 export default function NotebookModal({ eventBus, config }) {
+  // Temporary solution: while there is no mechanism to sync new annotations in
+  // the notebook, we force re-rendering of the iframe on every 'openNotebook'
+  // event, so that the new annotations are displayed.
+  // https://github.com/hypothesis/client/issues/3182
+  const [iframeKey, setIframeKey] = useState(0);
   const [isHidden, setIsHidden] = useState(true);
   const [groupId, setGroupId] = useState(/** @type {string|null} */ (null));
   const originalDocumentOverflowStyle = useRef('');
@@ -81,6 +85,7 @@ export default function NotebookModal({ eventBus, config }) {
       /** @type {string} */ groupId
     ) => {
       setIsHidden(false);
+      setIframeKey(iframeKey => iframeKey + 1);
       setGroupId(groupId);
     });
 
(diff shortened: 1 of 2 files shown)


src/annotator/components/NotebookModal.js
```diff
@@ -48,6 +48,10 @@ function NotebookIframe({ config, groupId }) {
  * @param {NotebookModalProps} props
  */
 export default function NotebookModal({ eventBus, config }) {
+  // Temporal solution: while there is no mechanism to sync new annotations in
+  // the notebook, we force re-rendering of the iframe on every 'openNotebook'
+  // event so we fetch the new annotations.
+  const [iframeKey, setIframeKey] = useState(Math.random());
```
LMS007, line 54: I think we can just make this a super simple counter and just increment on change

`useState(0);`

src/annotator/components/NotebookModal.js
```diff
@@ -81,6 +85,7 @@ export default function NotebookModal({ eventBus, config }) {
       /** @type {string} */ groupId
     ) => {
       setIsHidden(false);
+      setIframeKey(Math.random());
```
LMS007, line 88: `useState(iframeKey + 1);`

src/annotator/components/test/NotebookModal-test.js
```diff
@@ -58,6 +58,9 @@ describe('NotebookModal', () => {
     );
   });
 
+  // Test disabled temporarily until a sync mechanism is implemented in the
```
LMS007, line 61: I think it's still valid to leave this test even though it's now not possible to change the groupid and _not_ the iframeKey.  

**OR** I would suggesting getting rid of the `key` on the iframe too and just removing all the code bits and this test too.

src/annotator/components/NotebookModal.js
```diff
@@ -48,6 +48,10 @@ function NotebookIframe({ config, groupId }) {
  * @param {NotebookModalProps} props
  */
 export default function NotebookModal({ eventBus, config }) {
+  // Temporal solution: while there is no mechanism to sync new annotations in
```
LMS007, line 51: Is there a ticket to track this sync improvement? It might be good to add a link in the comments here if there is one.

LMS007 (changes requested):
Thanks Eduardo. The solution here works just fine, but I had just a few suggestions on the counter and maybe we can remove all the code associated with the iframe key or just leave that one test?

robertknight:
I concur with @LMS007's feedback here. One small note I'll add is that I think you mean "temporary" (meaning short-lived) rather than "temporal" (relating to time or the time aspect of something).

src/annotator/components/test/NotebookModal-test.js
esanzgar: I removed the `key` from the iframe and re-enable the test adjusting to the new behaviour.

esanzgar:
Thanks for your comments. This is ready for a second review.

src/annotator/components/NotebookModal.js
esanzgar: I added a ticket.

LMS007 (approved):
Thanks Eduardo, looks good!

Issue #3131, opened 2021-03-10
Client not fully being "destroyed" with destroy event dispatched (unload-client.js)


matt-e-king:
I am using the Hypothesis Client embedded within a single page application (VueJS) with client side routing (Browser History API). I've been informed in previous discussions that Hypothesis doesn't directly support SPA implementation because of the dynamic nature of client side routing. The annotation selection targeting gets all screwy :)

However, @robertknight has recommended I use the [browser-extension `unload-client`](https://github.com/hypothesis/browser-extension/blob/master/src/unload-client.js) approach to reload the embedded client on demand. So in my VueJS application, when a route changes I run the code found in the [unload-client](https://github.com/hypothesis/browser-extension/blob/master/src/unload-client.js) script. The `addEventListener` for the `destroy` event can be found [here](https://github.com/hypothesis/client/blob/master/src/annotator/index.js#L60).

**This was working beautifully!** However, as of late this appears to be broken - when I navigate pages the selection highlights become "stacked" on top of each other, the annotation counts double, and the Hypothesis menu when making a selection no longer works.

**UPDATE:** This bug however will most likely appear once the browser extension updates to the newest build. See the first set of steps to reproduce.

### Steps to reproduce
If you have a local copy of the `client` and `h` running:
1. Navigate to localhost:3000
2. Click on "Open sidebar"
3. Click on the "Page Note" icon in the Hypothesis sidebar, this should open up a new annotation input
4. Click "Unload Client"
5. Click "Load Client"
6. You can no longer make a Page Note annotation or a text selection annotation. Here is a GIF (it loops so it might be confusing)
![HypothesisClientBug](https://user-images.githubusercontent.com/3094245/111839549-32a58380-88b8-11eb-8e92-3d0de03479d2.gif)

Another way to recreate this "unload client" bug. You don't need to spin up an SPA or have Hypothesis locally to recreate this issue, the following steps should suffice:
1. **Turn off Hypothesis browser extension first - this seems to only be happening when embedding the client**
2. Navigate to https://web.hypothes.is/
3. Open up the dev console and paste in the following to embed the client:
```
var hypothScript = document.createElement('script')
hypothScript.setAttribute('src', 'https://hypothes.is/embed.js')
hypothScript.setAttribute('async', true)
document.head.appendChild(hypothScript)
```
4. This creates the Hypothesis sidebar on the right (make sure your annotation group is set to Public)
5. Now paste the following code in the dev console:
```
[...]

PR #3174, opened 2021-03-19
Destroy crossframe instance when guest is destroyed

matt-e-king:
Referencing issue https://github.com/hypothesis/client/issues/3131

It appears [`frame-sync.js`](https://github.com/hypothesis/client/blob/master/src/sidebar/services/frame-sync.js) is responsible for setting up a subscription to `beforeCreateAnnotation` which then calls `annotationsService.create`. This subscription is never destroyed so when the client is "unloaded/loaded" it creates duplicate subscriptions, makings things buggy.

From what I'm seeing, [`cross-frame.js`](https://github.com/hypothesis/client/blob/master/src/annotator/plugin/cross-frame.js#L60) is the only place responsible for dispatching `destroyFrame` to the `bridge` instance in `frame-sync.js`, but the `destroy` method is never invoked when the `Guest` instance is destroyed.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -342,6 +342,7 @@ export default class Guest {
 
     this.pdfIntegration?.destroy();
     this._emitter.destroy();
+    this.crossframe.destroy();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


robertknight (approved):
This fix looks correct. Thanks @matt-e-king - we'll get this merged and shipped early next week. It looks like there is an issue running GitHub Actions-based CI from third-party pull requests. We'll look into that as well.

robertknight:
nb. GitHub Actions checks didn't run on this PR but I did run them locally on this PR. I will resolve that in a separate PR.

PR #3175, opened 2021-03-22
TagEditor now removes repeated tags without changing filter() order.

peterdm:
The design of the tag-suggestions feature intends to defer ordering criteria to the tag-service, 
however this was not respected in the component.

The current tag service (`filter`) orders matching tag suggestions by descending frequency.
The TagEditor component (since #1558 - 15 months back) would then reorder tag suggestions alphabetically.

This change modifies the TagEditor to display tag-suggestions in the order returned by the service.

- Adds test "preserves order of tag-suggestions returned from filter()"
- Updates several tests which presumed the component would alphabetize tags.
- Reimplements `TagEditor.updateSuggestions` to not alphabetize or otherwise reorder suggestions.
- Closes https://github.com/hypothesis/product-backlog/issues/1196

**Impact**:  Users may notice a change in the ordering criteria from alpha back to frequecy-desc (current tag-service impl).

### DIFF
--- a/src/sidebar/components/TagEditor.js
+++ b/src/sidebar/components/TagEditor.js
@@ -72,13 +72,11 @@ function TagEditor({
    * @return {string[]}
    */
   const removeDuplicates = (suggestions, duplicates) => {
-    const suggestionsSet = [];
-    suggestions.forEach(suggestion => {
-      if (duplicates.indexOf(suggestion) < 0) {
-        suggestionsSet.push(suggestion);
-      }
-    });
-    return suggestionsSet.sort();
+    // suggestions already sorted the way the service wants
+
+    // only remove duplicates
+    // (array.filter returns a new array and preserves original input order)
+    return suggestions.filter(suggestion => duplicates.indexOf(suggestion) < 0);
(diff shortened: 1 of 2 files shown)


src/sidebar/components/TagEditor.js
```diff
@@ -72,13 +72,11 @@ function TagEditor({
...
-      if (duplicates.indexOf(suggestion) < 0) {
-        suggestionsSet.push(suggestion);
-      }
-    });
-    return suggestionsSet.sort();
+    // suggestions already sorted the way the service wants
+
+    // only remove duplicates
+    // (array.filter returns a new array and preserves original input order)
+    return suggestions.filter(suggestion => duplicates.indexOf(suggestion) < 0);
```
esanzgar, line 79:
```suggestion
    return suggestions.filter(suggestion => !duplicates.includes(suggestion));
```

esanzgar (commented):
If the assumption it's correct that the service returns the suggestions in frequency order and that we want to display these in that order (not alphabetical), I agree this could be a good change.

PR #3181, opened 2021-03-22
Run GitHub Actions checks on `pull_request` rather than `push` event

robertknight:
This ensures that GitHub Actions checks run for PRs submitted from forks,
which includes all PRs submitted by people outside the Hypothesis
organization.

Note that this change which commit the checks run on. See [1] for details.

This change also matches how GitHub CI is set up in other projects (eg. lms, h).

[1] https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

This fixes an issue discovered when reviewing https://github.com/hypothesis/client/pull/3174

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,5 +1,5 @@
 name: Continuous integration
-on: [push]
+on: pull_request
 jobs:
   ci:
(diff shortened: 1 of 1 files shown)


LMS007:
Yep, I think we do this for all frontend projects

LMS007 (approved):

PR #3187, opened 2021-03-23
Move creation of the outer container element to `Adder`

esanzgar:
`Notebook` and `Sidebar` components create `<hypothesis-...>` elements
that attach shadow DOMs. This PR makes the `Adder` class responsible of
creating its own outer element.

### DIFF
--- a/src/annotator/adder.js
+++ b/src/annotator/adder.js
@@ -83,24 +83,23 @@ export class Adder {
    *
    * The adder is initially hidden.
    *
-   * @param {HTMLElement} container - The DOM element into which the adder will be created
+   * @param {HTMLElement} element - The DOM element into which the adder will be created
    * @param {AdderOptions} options - Options object specifying `onAnnotate` and `onHighlight`
    *        event handlers.
    */
-  constructor(container, options) {
-    this._container = container;
-    this._shadowRoot = createShadowRoot(container);
+  constructor(element, options) {
+    this._outerContainer = document.createElement('hypothesis-adder');
+    element.appendChild(this._outerContainer);
+    this._shadowRoot = createShadowRoot(this._outerContainer);
 
     // Set initial style
-    Object.assign(container.style, {
-      display: 'block',
-
+    Object.assign(this._outerContainer.style, {
       // take position out of layout flow initially
       position: 'absolute',
       top: 0,
     });
 
-    this._view = /** @type {Window} */ (container.ownerDocument.defaultView);
+    this._view = /** @type {Window} */ (element.ownerDocument.defaultView);
 
     this._width = () => {
       const firstChild = /** @type {Element} */ (this._shadowRoot.firstChild);
@@ -137,6 +136,11 @@ export class Adder {
   hide() {
     this._isVisible = false;
     this._render();
+  }
+
+  destroy() {
+    render(null, this._shadowRoot); // First, unload the Preact component
+    this._outerContainer.remove();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/annotator/guest.js
```diff
@@ -116,11 +116,8 @@ export default class Guest {
     this.element = element;
     this._emitter = eventBus.createEmitter();
     this.visibleHighlights = false;
-    this.adderToolbar = document.createElement('hypothesis-adder');
-    this.adderToolbar.style.display = 'none';
-    this.element.appendChild(this.adderToolbar);
 
-    this.adderCtrl = new Adder(this.adderToolbar, {
+    this.adderCtrl = new Adder(this.element, {
```
robertknight, line 120: I suggest renaming `this.adderCtrl` to `this.adder` since there is no longer a need to differentiate between two different things related to the adder.

src/annotator/adder.js
```diff
@@ -83,24 +83,23 @@ export class Adder {
...
    * @param {AdderOptions} options - Options object specifying `onAnnotate` and `onHighlight`
    *        event handlers.
    */
-  constructor(container, options) {
-    this._container = container;
-    this._shadowRoot = createShadowRoot(container);
+  constructor(element, options) {
+    this._outerContainer = document.createElement('hypothesis-adder');
+    element.appendChild(this._outerContainer);
+    this._shadowRoot = createShadowRoot(this._outerContainer);
```
robertknight, line 93: There is a pattern that could potentially be extracted in future for setting up an element with a custom `hypothesis-` name, creating a shadow DOM, rendering a Preact component into it and then cleaning up later.

API sketch:

```js
// In util/preact-container.js
class PreactContainer {
  constructor(element, name) { /* create element and shadow root */ }
  render(vnode) { /* render vnode into shadow root */ }
  remove() { /* unmount and remove container */ }
}

// Usage example
this._container = new PreactContainer(element, 'hypothesis-adder');
this._container.render(<Adder .../>);
this._container.remove();
```

src/annotator/adder.js
```diff
@@ -83,24 +83,23 @@ export class Adder {
...
-   * @param {HTMLElement} container - The DOM element into which the adder will be created
+   * @param {HTMLElement} element - The DOM element into which the adder will be created
    * @param {AdderOptions} options - Options object specifying `onAnnotate` and `onHighlight`
    *        event handlers.
    */
-  constructor(container, options) {
-    this._container = container;
-    this._shadowRoot = createShadowRoot(container);
+  constructor(element, options) {
+    this._outerContainer = document.createElement('hypothesis-adder');
```
robertknight, line 91: I see that the naming of the `element` argument here and `this._outerContainer` is consistent with the `Notebook` class. I think it would be good to change the sidebar in future to also be consistent. One way to do this might be to extract a pattern (see below).

robertknight (approved):
This is a nice piece of cleanup 👍 . I left one suggestion for this PR about renaming `this.adderCtrl` in `Guest` and a suggestion for the future about extracting the common logic for setting up an element with a shadow DOM and rendering a Preact component into it.

src/annotator/adder.js
esanzgar: Good idea. I like that.

src/annotator/guest.js
esanzgar: I thought the same. Thanks, I will do it.

esanzgar: I also renamed `adderCtrl` to `adder` in `adder-test.js`.

Issue #3142, opened 2021-03-13
Editing annotations causes threads to change position in notebook


robertknight:
Threads in the notebook are currently sorted by the last-updated time of the root annotation. If the root annotation is edited directly from the notebook this time changes, causing the thread to jump to the top of the first page. This is IMO disorienting particularly when you're not on the first page. Sorting by the original creation date of the root annotation would prevent this from happening.

PR #3189, opened 2021-03-24
Sort root threads by date created in notebook

LMS007:
- Add `Created` sorter method to sort by date created as opposed to date modified.

fixes https://github.com/hypothesis/client/issues/3142

### DIFF
--- a/src/sidebar/helpers/test/thread-sorters-test.js
+++ b/src/sidebar/helpers/test/thread-sorters-test.js
@@ -25,22 +25,22 @@ describe('sidebar/util/thread-sorters', () => {
   describe('sorting by newest annotation thread first', () => {
     [
       {
-        a: [{ updated: 40 }, { updated: 5 }],
-        b: [{ updated: 20 }, { updated: 3 }],
+        a: [{ created: 40 }, { created: 5 }],
+        b: [{ created: 20 }, { created: 3 }],
         expected: -1,
       },
       {
-        a: [{ updated: 20 }, { updated: 3 }],
-        b: [{ updated: 20 }, { updated: 3 }],
+        a: [{ created: 20 }, { created: 3 }],
+        b: [{ created: 20 }, { created: 3 }],
         expected: 0,
       },
       {
-        a: [{ updated: 20 }, { updated: 3 }],
-        b: [{ updated: 40 }, { updated: 5 }],
+        a: [{ created: 20 }, { created: 3 }],
+        b: [{ created: 40 }, { created: 5 }],
         expected: 1,
       },
     ].forEach(testCase => {
-      it('sorts by newest updated root annotation', () => {
+      it('sorts by newest created root annotation', () => {
         // Disable eslint: `sorters` properties start with capital letters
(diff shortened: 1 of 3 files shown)


robertknight (commented):
Reading this PR makes me think that perhaps we should change the definition of the existing "Newest" and "Oldest" sort methods to work by creation date rather than updated date.

The problem that this PR solves in the notebook also applies in the sidebar when sorting by Newest or Oldest. If sorting by Newest and you make a tiny edit to an "old" annotation it will jump to the top of the list.

lyzadanger (changes requested):
Thanks for this work! It’s cleanly implemented. Alas, the more I look at this problem set, the more nuanced it becomes. I think there’s some baggage resulting from not explicitly defining (or at least, not documenting) the desired sorting behavior in the past. “This has grown organically,” etc., etc.

First, I think there’s an issue here with conflation of two things:

* **Sort by** (I always want to call this “column” because I think of it DB-wise)
* **Sort order** (e.g DESC, ASC)

Each of the pre-existing “sort keys” correspond to an implied pairing of sort-by and sort-order (implied, but not super clear, which means these weren’t exactly squeaky clean to begin with):

* “Newest” = sort by updated date, DESC
* “Oldest” = sort by updated date, ASC
* “Location” = sort by location in document, ASC. (DESC is logically nonsensical, so it’s sort of “OK” that there is no sort key corresponding to that variant)

These changes add:

* “Created” — This specifies a “sort by” but is unclear about “sort order.” And its existence alongside “Newest” and “Oldest” is confusing.

Rob’s note about the jumpiness of working with threads sorted by updated date, even in the sidebar, is interesting. **It sounds like we might want to consider,** in places where default sort order should be chronological (cf. sidebar Annotations tab, where default sort order is LOCATION ASC and so this doesn’t apply), **chrono-sorting by created instead of updated.** In this case, the current set of “sort keys” are still valid (i.e. no need for a new sort key) but they always refer to `created`, not `updated`. This is actually making this sort do what it _sounds_ like it would do.

This leaves the question: **Should we also provide the ability to users to sort by most/least recently updated root threads?** My hunch is that, yes, this is a quite useful sort option—I know I make use of it, e.g., in GitHub frequently. Here’s how GitHub handles it (ignore the most/least commented in our case, though, heck, the ability to sort by most/least replies _could_ be useful in the future. But let’s ignore it for the moment):

image

Given the above, I think we might be able to break this work down into chunks:

1. **Determine if it’s appropriate to make “Oldest” and “Newest” always operate** on `created` **instead** of `updated` (FWIW, replies are always sorted by `created DESC`; this happens, counterintuitively, within the `build-thread` helper). That could be a surgical change using the approach you’ve already used here.
2. (This could be done later/prioritized/not done, whatever): Determine if we want to add the ability for users to sort by most-recent and least-recently updated top-level thread.

LMS007:
Updated to PR to change the meaning of newest/oldest to use date created.

lyzadanger (approved):
This looks good and works as expected in my local.

Issue #3194, opened 2021-03-24
Adder does not adjust its position when the window is resized


esanzgar:
Adder does not adjust its position when the window is resized.

### Steps to reproduce
1. Make a selection to display the adder
2. Resize the window

### Expected behaviour
The position of the adder should be re-adjusted on window resize or pan events

### Actual behaviour
The adder maintains the original position when it was 'created'.

### Browser/system information
All.

PR #3197, opened 2021-03-24
Adjust the position of the adder on window.resize

esanzgar:
First commit is a refactor to use the `ListenerCollection` utility.

Second commit is the actual fix for #3194.

Applying this PR, this is how the `adder` repositions on `window.resize`:

https://user-images.githubusercontent.com/8555781/112584454-2b223500-8df8-11eb-94c1-3d72f65f42bf.mov

### DIFF
--- a/src/annotator/test/guest-test.js
+++ b/src/annotator/test/guest-test.js
@@ -45,14 +45,18 @@ describe('Guest', () => {
 
   let PDFIntegration;
   let fakePDFIntegration;
+  let guests;
 
   const createGuest = (config = {}) => {
     const element = document.createElement('div');
     const eventBus = new EventBus();
-    return new Guest(element, eventBus, { ...guestConfig, ...config });
+    const guest = new Guest(element, eventBus, { ...guestConfig, ...config });
+    guests.push(guest);
+    return guest;
   };
 
   beforeEach(() => {
+    guests = [];
     FakeAdder.instance = null;
     guestConfig = {};
     highlighter = {
@@ -125,6 +129,7 @@ describe('Guest', () => {
   });
 
   afterEach(() => {
+    guests.forEach(guest => guest.destroy());
     sandbox.restore();
     $imports.$restore();
   });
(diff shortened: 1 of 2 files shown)


src/annotator/test/guest-test.js
```diff
@@ -487,6 +492,22 @@ describe('Guest', () => {
         assert.notCalled(guest.crossframe.call);
       }
     });
+
+    it('reposition the adder on window "resize" event', () => {
+      sandbox.stub(guest, '_repositionAdder').callThrough();
+      sandbox.stub(guest, '_onSelection');
+      window.dispatchEvent(new Event('resize'));
+
+      // If the adder is not displayed on 'window.resize' event does nothing
```
LMS007, line 501: Maybe reword this a little...

"Does not call _onSelection() if the adder is hidden"

src/annotator/test/guest-test.js
```diff
@@ -487,6 +492,22 @@ describe('Guest', () => {
...
+      assert.called(guest._repositionAdder);
+      assert.notCalled(guest._onSelection);
+
+      // If the adder is displayed then it calls 'guest._onSelection'
```
LMS007, line 505: Maybe reword this a little...

"Calls _onSelection() if the adder is visible"

And I would actually just make this a second test since its testing a different code path.

LMS007 (changes requested):
Code changes look good. I just had some minor suggestions to break the test up into 2 tests.  And there seems to be a missing condition that codecov is barking at. 

This is a noticeable improvement for HTML documents. 👍

src/annotator/test/guest-test.js
esanzgar: Good idea

src/annotator/test/guest-test.js
esanzgar: I have split the tests in two. I name the test more in terms of what it does, more than what function calls.

esanzgar:
On a third commit, I will add tests for some of the lines that are missing code coverage: 
image

esanzgar:
@LMS007, would you mind if we merge this PR and I improve the code coverage in another PR?

I have created #3203 to keep track.

LMS007 (approved):
Thanks for waiting Eduardo. Yep, just follow up with the coverage additions

Issue #3128, opened 2021-03-09
Create a standard interface for Guest/Sidebar to invoke document-type specific functionality


robertknight:
The logic in the annotator that is specific to a particular document type is currently spread out in various places and has slightly different interfaces. This logic includes:

- Document URI and metadata extraction
- Generating selectors from the current selection and resolving ("anchoring") them to DOM ranges
- Implementing side-by-side mode (currently only implemented for PDFs)
- Responding to content changes (eg. scrolling, change of loaded document)

The lack of a common interface that encompasses this functionality means that there is no obvious way to plug in support for a new document type. We're planning to expand ebook support and so I think this would be a good time to solve this. As well as making it easier to customize the logic for various ebook viewers, this will also make it easier to make some of the functionality that is currently PDF-only also work for other document types (eg. side-by-side mode, re-anchoring annotations when document content changes).

I have a sketch of what this could look like in https://github.com/hypothesis/client/tree/z-doc-type-integration-api. Below is a list of subtasks I identified along the way:

- [x] Simplify BucketBar interface to support only a single container element (https://github.com/hypothesis/client/pull/3107)
- [x] Finish extracting WarningBanner into its own module (https://github.com/hypothesis/client/pull/3065)
- [x] Move non-document-type specific side-by-side logic from PdfSidebar class to Sidebar (https://github.com/hypothesis/client/pull/3190)
- [ ] Move non-PDF-specific annotation refresh logic from PDF plugin to Guest (punted for now. See https://github.com/hypothesis/client/issues/3128#issuecomment-809214461)
- [x] Define an interface that includes all the necessary methods for the Guest/Sidebar to interact with the document-type specific functionality, and convert the existing HTML/PDF integration to use it (https://github.com/hypothesis/client/pull/3199)

See also Slack notes on a [proposed module structure](https://hypothes-is.slack.com/archives/C1M8NH76X/p1614684393090400)

robertknight:
This is now done except for two aspects of document type integration:

- The logic for iterating over the current anchors and re-anchoring any which are no longer valid lives in the PDF integration, although this is probably useful in a similar form for other document types 
- The logic for creating highlights in PDF.js still lives together with the HTML highlighting logic in `highlighter.js`

I'm not certain what the ideal abstraction will look like for both of these aspects, so I'm inclined to punt on it until we have a concrete use cases to work with.

PR #3198, opened 2021-03-25
Rename `PDF` => `PDFIntegration` and move to `src/annotator/integrations` (2/n)

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/3190**~~

Create an `integrations` directory within `src/annotator` which will
house the `*Integration` classes that encapsulate document
type/viewer-specific functionality, and create the first version of the
PDF integration by moving/renaming the existing `PDF` class. The
supporting `PDFMetadata` class has also been moved.

Per recently agreed code conventions [1], the `PDFIntegration` and
`PDFMetadata` classes are no longer default exports.

[1] https://github.com/hypothesis/frontend-toolkit/blob/master/docs/js-guide.md#default-exports
[Related Slack thread](https://hypothes-is.slack.com/archives/C1M8NH76X/p1615303150029900?thread_ts=1615302509.028600)

### DIFF
--- a/src/annotator/test/guest-test.js
+++ b/src/annotator/test/guest-test.js
@@ -44,8 +44,8 @@ describe('Guest', () => {
   let DocumentMeta;
   let fakeDocumentMeta;
 
-  let PdfIntegration;
-  let fakePdfIntegration;
+  let PDFIntegration;
+  let fakePDFIntegration;
 
   const createGuest = (config = {}) => {
     const element = document.createElement('div');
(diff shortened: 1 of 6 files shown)


src/annotator/test/guest-test.js
```diff
@@ -44,8 +44,8 @@ describe('Guest', () => {
   let DocumentMeta;
   let fakeDocumentMeta;
 
-  let PdfIntegration;
-  let fakePdfIntegration;
+  let PDFIntegration;
+  let fakePDFIntegration;
```
robertknight, line 48: The capitalization here was corrected to match the real class. See related Slack discussion: https://hypothes-is.slack.com/archives/C1M8NH76X/p1616742301079700.

esanzgar (approved):

PR #3199, opened 2021-03-25
Unify HTML and PDF code paths in `Guest` (3/n)

robertknight:
**Depends on https://github.com/hypothesis/client/pull/3198**

Unify the code paths in the `Guest` class that invoke document
type/viewer-specific code such as getting the document URI and metadata
and anchoring and generating selectors.

 - Replace `pdfIntegration` and `documentMeta` fields in `Guest` with an
   `integration` field which provides the interface to document-type
   specific functionality
 - Replace the roundabout method of `PDFIntegration` customizing the
   anchoring logic with `anchor` and `describe` methods which the
   `Guest` calls as needed
 - Add documentation and tests for `uri` and `getMetadata` methods in
   `PDFIntegration`

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -111,9 +111,8 @@ export default class Guest {
    * @param {EventBus} eventBus -
    *   Enables communication between components sharing the same eventBus
    * @param {Record<string, any>} [config]
-   * @param {typeof htmlAnchoring} [anchoring] - Anchoring implementation
    */
-  constructor(element, eventBus, config = {}, anchoring = htmlAnchoring) {
+  constructor(element, eventBus, config = {}) {
     this.element = element;
     this._emitter = eventBus.createEmitter();
     this.visibleHighlights = false;
@@ -146,16 +145,19 @@ export default class Guest {
 
     // Setup the document type-specific integration consisting of metadata extraction,
     // anchoring module and logic to respond to activity (eg. scrolling) in the page.
-    //
-    // In future the plan is to put these behind a common interface so that the
-    // `Guest` class doesn't need different code paths depending on the document type.
-
-    // nb. The `anchoring` field defaults to HTML anchoring and in PDFs is replaced
-    // by `PDFIntegration` below.
-    this.anchoring = anchoring;
-    this.documentMeta = new DocumentMeta();
     if (config.documentType === 'pdf') {
-      this.pdfIntegration = new PDFIntegration(this);
+      this.integration = new PDFIntegration(this);
+    } else {
+      const documentMeta = new DocumentMeta();
+      this.integration = {
+        anchor: htmlAnchoring.anchor,
+        contentContainer: () => this.element,
+        describe: htmlAnchoring.describe,
+        destroy: () => {},
+        fitSideBySide: () => false,
+        getMetadata: () => Promise.resolve(documentMeta.getDocumentMetadata()),
+        uri: () => Promise.resolve(documentMeta.uri()),
+      };
     }
 
     // Set the frame identifier if it's available.
@@ -241,21 +243,10 @@ export default class Guest {
    * Retrieve metadata for the current document.
    */
   getDocumentInfo() {
-    let metadataPromise;
-    let uriPromise;
-
-    if (this.pdfIntegration) {
-      metadataPromise = this.pdfIntegration.getMetadata();
-      uriPromise = this.pdfIntegration.uri();
-    } else {
-      uriPromise = Promise.resolve(this.documentMeta.uri());
-      metadataPromise = Promise.resolve(this.documentMeta.metadata);
-    }
-
-    uriPromise = uriPromise.catch(() =>
-      decodeURIComponent(window.location.href)
-    );
-    metadataPromise = metadataPromise.catch(() => ({
+    const uriPromise = this.integration
+      .uri()
+      .catch(() => decodeURIComponent(window.location.href));
+    const metadataPromise = this.integration.getMetadata().catch(() => ({
       title: document.title,
       link: [{ href: decodeURIComponent(window.location.href) }],
     }));
@@ -338,7 +329,7 @@ export default class Guest {
 
     removeAllHighlights(this.element);
 
-    this.pdfIntegration?.destroy();
+    this.integration.destroy();
     this._emitter.destroy();
     this.crossframe.destroy();
   }
@@ -394,7 +385,7 @@ export default class Guest {
       }
 
       // Find a target using the anchoring module.
-      return this.anchoring
+      return this.integration
         .anchor(root, target.selector)
         .then(range => ({
           annotation,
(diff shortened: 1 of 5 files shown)


src/annotator/guest.js
```diff
@@ -149,16 +148,19 @@ export default class Guest {
...
+      const documentMeta = new DocumentMeta();
+      this.integration = {
+        anchor: htmlAnchoring.anchor,
+        contentContainer: () => this.element,
+        describe: htmlAnchoring.describe,
+        destroy: () => {},
+        fitSideBySide: () => false,
+        getMetadata: () => Promise.resolve(documentMeta.getDocumentMetadata()),
+        uri: () => Promise.resolve(documentMeta.uri()),
+      };
```
robertknight, line 163: The HTML implementation of `integration` is currently defined inline in the `Guest` class. I plan to extract it into a separate `HTMLIntegration` class in future for consistency with `PDFIntegration`.

esanzgar: 👍🏼

esanzgar: Please, would it be possible for you to create the `Integration` interface and use `@implements {Integration}`on the `PDFIntegration` class and temporary `@type {Integration}` on `this.integration` declaration?

src/annotator/guest.js
```diff
@@ -241,21 +243,10 @@ export default class Guest {
...
-      metadataPromise = Promise.resolve(this.documentMeta.metadata);
-    }
-
-    uriPromise = uriPromise.catch(() =>
-      decodeURIComponent(window.location.href)
-    );
-    metadataPromise = metadataPromise.catch(() => ({
+    const uriPromise = this.integration
+      .uri()
+      .catch(() => decodeURIComponent(window.location.href));
```
esanzgar, line 248: Do we want to silently default to `window.location.href` in case of errors? If the correctness of this piece of information is essential for the application it maybe better to not catch the error and make the application fail.

src/annotator/guest.js
```diff
@@ -241,21 +243,10 @@ export default class Guest {
...
+    const metadataPromise = this.integration.getMetadata().catch(() => ({
```
esanzgar, line 249: Same as above.

src/annotator/guest.js
```diff
@@ -146,16 +145,19 @@ export default class Guest {
...
-    // by `PDFIntegration` below.
-    this.anchoring = anchoring;
-    this.documentMeta = new DocumentMeta();
     if (config.documentType === 'pdf') {
-      this.pdfIntegration = new PDFIntegration(this);
+      this.integration = new PDFIntegration(this);
+    } else {
+      const documentMeta = new DocumentMeta();
+      this.integration = {
+        anchor: htmlAnchoring.anchor,
```
esanzgar, line 153: You could `import { anchor, describe } from './anchoring/html';` and then use 
```suggestion
        anchor,
```
Same for `describe`. 

The import above would be more consistent with the way you import from `anchoring/pdf` below.

src/annotator/integrations/pdf.js
```diff
@@ -89,14 +89,46 @@ export class PDFIntegration {
...
+  /**
+   * Resolve serialized `selectors` from an annotation to a range.
+   *
+   * @param {HTMLElement} root
+   * @param {Selector[]} selectors
+   * @return {Promise<Range>}
+   */
+  anchor(root, selectors) {
+    // nb. The `root` argument is not really used by `anchor`. It existed for
+    // consistency between HTML and PDF anchoring and could be removed.
```
esanzgar, line 115: 👍🏼 , good note

src/annotator/integrations/pdf.js
```diff
@@ -115,7 +147,7 @@ export class PDFIntegration {
     }
 
     try {
-      const hasText = await pdfAnchoring.documentHasText();
+      const hasText = await documentHasText();
       this._showNoSelectableTextWarning(!hasText);
```
esanzgar, line 151: Unrelated with this PR, but this is unusual. Why not?
```suggestion
      if (!hasText) {
          this._showNoSelectableTextWarning();
      }
```

Why do we want to create the `<hypothesis-banner>` with its shadow DOM for nothing?

src/annotator/integrations/pdf.js
```diff
@@ -89,14 +89,46 @@ export class PDFIntegration {
...
+  }
+
+  /**
+   * Generate selectors for the text in `range`.
+   *
+   * @param {HTMLElement} root
+   * @param {Range} range
+   * @return {Promise<Selector[]>}
+   */
+  describe(root, range) {
```
esanzgar, line 126: For me (newbie), `describe` is a very vague name that doesn't tell me much about the aim of the property.

src/annotator/integrations/test/pdf-test.js
```diff
@@ -54,11 +54,15 @@ describe('PDFIntegration', () => {
     };
 
     fakePdfAnchoring = {
```
esanzgar, line 56: I think it should be `fakePDFAnchoring`
```suggestion
    fakePDFAnchoring = {
```

esanzgar (approved):
I left a few comments.

src/annotator/guest.js
robertknight: I agree with this change, but I think it will make more sense once the HTML integration has been extracted in a follow-up PR, where it will be obvious that `anchor` and `describe` are about HTML anchoring.

src/annotator/guest.js
robertknight: Note for posterity: This change was implemented in the follow-up PR https://github.com/hypothesis/client/pull/3202

src/annotator/guest.js
robertknight: I agree. I think changing this is out of scope for this PR but we definitely don't want `window.location.href` being used in PDF documents for example! I filed https://github.com/hypothesis/client/issues/3204 to track this issue.

src/annotator/guest.js
robertknight: Agreed. Also covered by https://github.com/hypothesis/client/issues/3204.

src/annotator/integrations/pdf.js
robertknight: Ah - the method name is confusing here. The way it behaves is more like `this._toggleSelectableTextWarning(<visible>)`. If the value is `false` it just removes the banner if it already exists.

src/annotator/integrations/pdf.js
robertknight: I'm open to changing the term, but I think we'd need to change it everywhere for consistency. Another option is that we could treat it as one of a modest number of domain-specific terms (like "anchor" or "selector") which a newbie will need to look up when they first see it, but will then know what it means the next time they encounter it.

robertknight:
I pushed a couple of small commits with refactorings based on the comments and filed an issue about changing the fallback behavior if document URI or metadata extraction fails. See https://github.com/hypothesis/client/issues/3204.

PR #3201, opened 2021-03-25
Hide sidebar on 'mousedown' event

### LINKED ISSUE #463: Sidebar obscures annotate popup
When the sidebar partially covers the page content, then when your selection includes some of the covered content the sidebar can end up partially or entirely obscuring the `Annotate | Highlight` popup:

peek 2017-06-23 17-50

esanzgar:
When the selection of text finishes on the sidebar's iframe, the click
event is not fired. Hence, the sidebar is not collapsed and the adder is
not displayed. This is because events don't bubble up through the
iframe.

This PR makes the sidebar to collapse on 'mousedown' events on the host
page. This is similar in nature to the 'touchstart' event that `Guest`
already listens (although for totally different reason).

Closes #463

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -180,32 +180,35 @@ export default class Guest {
 
     // Hide the sidebar in response to a document click or tap, so it doesn't obscure
     // the document content.
-    const maybeCloseSidebar = event => {
-      if (!this.closeSidebarOnDocumentClick || this.isEventInAnnotator(event)) {
-        // Don't hide the sidebar if event occurred inside Hypothesis UI, or
-        // the user is making a selection, or the behavior was disabled because
-        // the sidebar doesn't overlap the content.
+    const maybeCloseSidebar = element => {
+      if (!this.closeSidebarOnDocumentClick) {
+        // Don't hide the sidebar if event was disabled because the sidebar
+        // doesn't overlap the content.
+        return;
+      }
+      if (annotationsAt(element).length) {
+        // Don't hide the sidebar if the event comes from an element that contains a highlight
         return;
       }
       this.crossframe.call('closeSidebar');
     };
 
-    addListener('click', event => {
+    addListener('mouseup', event => {
       const annotations = annotationsAt(event.target);
       if (annotations.length && this.visibleHighlights) {
         const toggle = event.metaKey || event.ctrlKey;
         this.selectAnnotations(annotations, toggle);
-      } else {
-        maybeCloseSidebar(event);
       }
+    });
+
+    addListener('mousedown', event => {
+      maybeCloseSidebar(event.target);
     });
 
     // Allow taps on the document to hide the sidebar as well as clicks.
     // On iOS < 13 (2019), elements like h2 or div don't emit 'click' events.
     addListener('touchstart', event => {
-      if (!annotationsAt(event.target).length) {
-        maybeCloseSidebar(event);
-      }
+      maybeCloseSidebar(event.target);
     });
 
     addListener('mouseover', event => {
@@ -626,18 +629,6 @@ export default class Guest {
     } else {
       this.showAnnotations(annotations);
     }
-  }
-
-  /**
-   * Did an event originate from an element in the sidebar UI? (eg. the sidebar
-   * iframe, or its toolbar)
-   *
-   * @param {Event} event
-   */
-  isEventInAnnotator(event) {
-    return (
-      /** @type {Element} */ (event.target).closest('.annotator-frame') !== null
-    );
   }
 
   /**
--- a/src/annotator/test/guest-test.js
+++ b/src/annotator/test/guest-test.js
@@ -462,29 +462,20 @@ describe('Guest', () => {
       fakeSidebarFrame?.remove();
     });
 
-    it('hides sidebar when the user taps or clicks in the page', () => {
-      for (let event of ['click', 'touchstart']) {
+    it('hides sidebar on user "mousedown" or "touchstart" events in the document', () => {
+      for (let event of ['mousedown', 'touchstart']) {
         rootElement.dispatchEvent(new Event(event));
         assert.calledWith(guest.crossframe.call, 'closeSidebar');
+        guest.crossframe.call.resetHistory();
       }
     });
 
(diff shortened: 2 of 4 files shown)


esanzgar:
@LMS007, today, I have been reviewing a few PRs from Robert that change `guest.js` and `guest-test.js` substantially.

Would you mind if I left the increase of code coverage for later, once Robert's PRs are merged.

I have created a issue to track down this improvement: #3203 3203

LMS007:
I actually think the biggest problem with this issue is that as you drag text from behind the sidebar, it just jumps to the end of the page and highlights everything below. In light of that, and the fact that we do close the sidebar on click, we could probably just close it on mousedown. I will note I discovered this also closes it on right mousedown, but that's probably ideally as well.

src/annotator/guest.js
```diff
@@ -208,6 +208,12 @@ export default class Guest {
       }
     });
 
+    addListener('mousedown', event => {
+      if (!annotationsAt(event.target).length) {
+        maybeCloseSidebar(event);
```
LMS007, line 213: Because we close this on mouse down, I think we can get rid of the `else` case line 207 in the `click` listener and any corresponding test. The new `mousedown` listener should now take care of this.

src/annotator/util/test/shadow-root-test.js
```diff
@@ -74,7 +74,23 @@ describe('annotator/util/shadow-root', () => {
       assert.notCalled(onClick);
     });
 
-    it('stops propagation of touchstart events', () => {
+    it('stops propagation of "mousedown" events', () => {
```
LMS007, line 77: Probably a good test to add, looks like no code changes needed to shadow-root as it already does this.

LMS007 (changes requested):
Eduardo this looks good, and a worthwhile improvement. I think we can now simplify the click handler and then its good to go.

src/annotator/guest.js
esanzgar: You are correct, thanks.

Another change we could make it is to change the `click` event for `mouseup`. In that way, it would be symmetrical. What do you think?

src/annotator/util/test/shadow-root-test.js
esanzgar: I added the code in the shadow-root to stop the bubbling up of the `mousedown`, otherwise all clicks in `<hypothesis-....>` will cause the sidebar to close.

src/annotator/test/guest-test.js
```diff
@@ -462,29 +462,20 @@ describe('Guest', () => {
       fakeSidebarFrame?.remove();
     });
 
-    it('hides sidebar when the user taps or clicks in the page', () => {
-      for (let event of ['click', 'touchstart']) {
+    it('hides sidebar on user "mousedown" or "touchstart" events in the document', () => {
+      for (let event of ['mousedown', 'touchstart']) {
         rootElement.dispatchEvent(new Event(event));
         assert.calledWith(guest.crossframe.call, 'closeSidebar');
+        guest.crossframe.call.resetHistory();
```
esanzgar, line 469: Important to add this, otherwise it could succeed on the first loop and fail on the second and still pass the test.

src/annotator/test/guest-test.js
```diff
@@ -462,29 +462,20 @@ describe('Guest', () => {
...
-    });
-
-    it('does not hide sidebar if event occurs inside annotator UI', () => {
-      fakeSidebarFrame = document.createElement('div');
-      fakeSidebarFrame.className = 'annotator-frame';
-      rootElement.appendChild(fakeSidebarFrame);
-
-      for (let event of ['click', 'touchstart']) {
-        fakeSidebarFrame.dispatchEvent(new Event(event, { bubbles: true }));
-        assert.notCalled(guest.crossframe.call);
```
esanzgar, line 487: This code was actually no testing correctly this behaviour, which is quite difficult to test.

esanzgar:
Thanks for your review. 

I have changed the `click` event and divided into two `mouseup` and `mousedown`. `mousedown` could close sidebar and `mouseup` could open it if the target is a hypothesis highlight.

Could you let me know what you think?

src/annotator/test/guest-test.js
LMS007: Good catch

esanzgar:
There is still one case where this fix won't solve: If the selection starts on a highlight, the sidebar is not collapsed.

Overall I think this is an improvement from the previous behaviour.

src/annotator/guest.js
```diff
@@ -180,32 +180,34 @@ export default class Guest {
...
-        // Don't hide the sidebar if event occurred inside Hypothesis UI, or
-        // the user is making a selection, or the behavior was disabled because
-        // the sidebar doesn't overlap the content.
+    const maybeCloseSidebar = element => {
+      if (!this.closeSidebarOnDocumentClick) {
+        // Don't hide the sidebar if event was disabled because the sidebar
+        // doesn't overlap the content.
+        return;
+      }
+      if (annotationsAt(element).length) {
```
LMS007, line 189: perhaps a small comment here

`// Don't hide if highlight clicked`

src/annotator/guest.js
```diff
@@ -180,32 +180,34 @@ export default class Guest {
 
     // Hide the sidebar in response to a document click or tap, so it doesn't obscure
     // the document content.
-    const maybeCloseSidebar = event => {
-      if (!this.closeSidebarOnDocumentClick || this.isEventInAnnotator(event)) {
-        // Don't hide the sidebar if event occurred inside Hypothesis UI, or
-        // the user is making a selection, or the behavior was disabled because
-        // the sidebar doesn't overlap the content.
+    const maybeCloseSidebar = element => {
+      if (!this.closeSidebarOnDocumentClick) {
```
LMS007, line 184: I guess `isEventInAnnotator` ever never returning `true`. Good catch.

src/annotator/test/guest-test.js
LMS007: 👍

LMS007 (approved):
Looks good Eduardo!
thank you

src/annotator/guest.js
esanzgar: Good idea.

PR #3202, opened 2021-03-26
Extract HTMLIntegration class (4/n)

robertknight:
**Depends on https://github.com/hypothesis/client/pull/3199**

This PR extracts the integration that handles HTML documents into an `HTMLIntegration` class, to match the `PDFIntegration` class added in https://github.com/hypothesis/client/pull/3198. It also renames the existing `DocumentMeta` class to `HTMLMetadata` to match the `PDFMetadata` class and moves it from `src/annotator/plugins` to `src/annotator/integrations`.

The `HTMLIntegration` and `PDFIntegration` class implement a common `Integration` interface which is also added in this PR in `src/types/annotator.js`.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -1,11 +1,10 @@
 import scrollIntoView from 'scroll-into-view';
 
 import { Adder } from './adder';
+import { HTMLIntegration } from './integrations/html';
 import { PDFIntegration } from './integrations/pdf';
 import CrossFrame from './plugin/cross-frame';
-import DocumentMeta from './plugin/document';
 
-import * as htmlAnchoring from './anchoring/html';
 import { TextRange } from './anchoring/text-range';
 import {
   getHighlightsContainingNode,
@@ -148,16 +147,7 @@ export default class Guest {
     if (config.documentType === 'pdf') {
       this.integration = new PDFIntegration(this);
     } else {
-      const documentMeta = new DocumentMeta();
-      this.integration = {
-        anchor: htmlAnchoring.anchor,
-        contentContainer: () => this.element,
-        describe: htmlAnchoring.describe,
-        destroy: () => {},
-        fitSideBySide: () => false,
-        getMetadata: () => Promise.resolve(documentMeta.getDocumentMetadata()),
-        uri: () => Promise.resolve(documentMeta.uri()),
-      };
+      this.integration = new HTMLIntegration(this.element);
     }
 
     // Set the frame identifier if it's available.
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -22,8 +22,9 @@ import { normalizeURI } from './util/url';
  * @typedef {import('./util/emitter').EventBus} EventBus
  * @typedef {import('../types/annotator').AnnotationData} AnnotationData
  * @typedef {import('../types/annotator').Anchor} Anchor
+ * @typedef {import('../types/annotator').Integration} Integration
+ * @typedef {import('../types/annotator').SidebarLayout} SidebarLayout
  * @typedef {import('../types/api').Target} Target
- * @typedef {import('./sidebar').LayoutState} LayoutState
  */
 
 /**
@@ -142,13 +143,11 @@ export default class Guest {
     /** @type {Anchor[]} */
     this.anchors = [];
 
-    // Setup the document type-specific integration consisting of metadata extraction,
-    // anchoring module and logic to respond to activity (eg. scrolling) in the page.
-    if (config.documentType === 'pdf') {
-      this.integration = new PDFIntegration(this);
-    } else {
-      this.integration = new HTMLIntegration(this.element);
-    }
+    /** @type {Integration} */
+    this.integration =
+      config.documentType === 'pdf'
+        ? new PDFIntegration(this)
+        : new HTMLIntegration(this.element);
 
     // Set the frame identifier if it's available.
     // The "top" guest instance will have this as null since it's in a top frame not a sub frame
@@ -664,7 +663,7 @@ export default class Guest {
   /**
    * Attempt to fit the document content alongside the sidebar.
    *
-   * @param {LayoutState} sidebarLayout
+   * @param {SidebarLayout} sidebarLayout
    */
   fitSideBySide(sidebarLayout) {
     const active = this.integration.fitSideBySide(sidebarLayout);
--- a/src/annotator/integrations/html.js
+++ b/src/annotator/integrations/html.js
@@ -1,0 +1,34 @@
+import { anchor, describe } from '../anchoring/html';
+
+import { HTMLMetadata } from './html-metadata';
+
+export class HTMLIntegration {
+  constructor(container = document.body) {
+    this.container = container;
+    this.anchor = anchor;
+    this.describe = describe;
+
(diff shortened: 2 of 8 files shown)


src/annotator/integrations/test/html-test.js
```diff
@@ -0,0 +1,69 @@
...
+      const integration = new HTMLIntegration();
+      assert.equal(integration.contentContainer(), document.body);
+    });
+  });
+
+  describe('#destroy', () => {
+    it('cleans up integration', () => {
+      new HTMLIntegration().destroy();
+    });
+  });
```
robertknight, line 46: This test and the one below are stubs just to check that the methods don't result in any errors. There are no expected effects and hence no asserts.

src/annotator/integrations/html.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  fitSideBySide() {
+    // Not yet implemented.
+    return false;
+  }
+
+  async getMetadata() {
+    return this._htmlMeta.getDocumentMetadata();
+  }
```
robertknight, line 41: `getMetadata` and `uri` don't need to be async in this implementation, but they are for consistency with `PDFIntegration`.

src/annotator/guest.js
```diff
@@ -143,22 +143,11 @@ export default class Guest {
...
-        fitSideBySide: () => false,
-        getMetadata: () => Promise.resolve(documentMeta.getDocumentMetadata()),
-        uri: () => Promise.resolve(documentMeta.uri()),
-      };
-    }
+    /** @type {Integration} */
+    this.integration =
+      config.documentType === 'pdf'
+        ? new PDFIntegration(this)
+        : new HTMLIntegration(this.element);
```
robertknight, line 150: In future it will probably sense to pass in the integration implementation rather than have `Guest` know how to construct it.

src/annotator/guest.js
```diff
@@ -23,8 +22,9 @@ import { normalizeURI } from './util/url';
  * @typedef {import('./util/emitter').EventBus} EventBus
  * @typedef {import('../types/annotator').AnnotationData} AnnotationData
  * @typedef {import('../types/annotator').Anchor} Anchor
+ * @typedef {import('../types/annotator').Integration} Integration
```
esanzgar, line 25: 👍🏼

src/annotator/integrations/html.js
```diff
@@ -0,0 +1,46 @@
...
+ * @typedef {import('../../types/annotator').Integration} Integration
+ */
+
+/**
+ * Document type integration for ordinary web pages.
+ *
+ * This integration is used for web pages and applications that are not handled
+ * by a more specific integration (eg. for PDFs).
+ *
+ * @implements {Integration}
```
esanzgar, line 15: If you are not using `Integration` type only once, maybe you could do this:

```suggestion
 * @implements {import('../../types/annotator').Integration}
```

src/annotator/integrations/test/html-test.js
```diff
@@ -0,0 +1,69 @@
+import { HTMLIntegration, $imports } from '../html';
+
+describe('HTMLIntegration', () => {
+  let fakeHtmlAnchoring;
```
esanzgar, line 4:
```suggestion
  let fakeHTMLAnchoring;
```

src/annotator/integrations/test/html-test.js
```diff
@@ -0,0 +1,69 @@
...
+  let fakeHtmlMetadata;
```
esanzgar, line 5:
```suggestion
  let fakeHTMLMetadata;
```

src/types/annotator.js
```diff
@@ -72,6 +72,36 @@
  * @prop {(ann: AnnotationData) => Promise<Anchor[]>} anchor
  */
 
+/**
+ * @typedef SidebarLayout
+ * @prop {boolean} expanded
```
esanzgar, line 77: Could you add a description? I guess this is if the sidebar is open or closed, correct?

And the width of the sidebar in pixels correct?

src/types/annotator.js
```diff
@@ -72,6 +72,36 @@
...
+ * @prop {number} width
+ */
+
+/**
+ * Interface for document type/viewer integrations that handle all the details
+ * of supporting a specific document type (web page, PDF, ebook etc.).
```
esanzgar, line 83:
```suggestion
 * of supporting a specific document type (web page, PDF, ebook, etc.).
```

src/types/annotator.js
```diff
@@ -72,6 +72,36 @@
...
+ *   Attempt to resolve a set of serialized selectors to the corresponding content in the
+ *   current document.
+ * @prop {(root: HTMLElement, range: Range) => Selector[]|Promise<Selector[]>} describe -
+ *   Generate a list of serializable selectors which represent the content in
+ *   `range`.
+ * @prop {() => HTMLElement} contentContainer -
+ *   Return the main element that contains the document content. This is used
+ *   by controls such as the bucket bar to know when the content might have scrolled.
+ * @prop {() => void} destroy -
+ *   Clean up the integration and remove any event listeners, caches etc.
```
esanzgar, line 96:
```suggestion
 *   Clean up the integration and remove any event listeners, caches, etc.
```

src/types/annotator.js
```diff
@@ -72,6 +72,36 @@
...
+ * @prop {(layout: SidebarLayout) => boolean} fitSideBySide -
+ *   Attempt to resize the content so that it is visible alongside the sidebar.
+ *   Returns `true` if the sidebar and content are displayed side-by-side or
+ *   false otherwise.
+ * @prop {() => Promise<object>} getMetadata
```
esanzgar, line 101: Please, could you add a description here. The other definitions are super helpful. I better understood the anchor <-> describe relation. By the way, `anchorSelectors` and `generateSelectors` could be more self documented variations.

src/types/annotator.js
```diff
@@ -72,6 +72,36 @@
...
+ * @prop {() => Promise<string>} uri - Return the URL of the currently loaded document
```
esanzgar, line 102:
```suggestion
 * @prop {() => Promise<string>} uri - Return the encoded URL of the currently loaded document
```

I don't know why it needs to be encoded.

src/annotator/integrations/test/html-test.js
```diff
@@ -0,0 +1,69 @@
...
+
+  describe('#contentContainer', () => {
+    it('returns body by default', () => {
+      const integration = new HTMLIntegration();
+      assert.equal(integration.contentContainer(), document.body);
+    });
+  });
+
+  describe('#destroy', () => {
+    it('cleans up integration', () => {
```
esanzgar, line 43:
```suggestion
    it('does nothing', () => {
```

esanzgar (approved):
Some of my suggestions in parts 3/n are outdated, because you implemented them in this PR.

I left some comments.

src/types/annotator.js
robertknight: Yes, I'll update this.

robertknight:
Thanks for the feedback. I've pushed various corrections as additional commits. One thing I've punted on for now is renaming `anchor` and `describe`. This is because we need to decide whether to update references to this term in other places for consistency.

PR #3205, opened 2021-03-29
Remove `plugin` directory from src/annotator/

robertknight:
The term "plugin" is vestigial. The remaining cross-frame.js module in this
directory has been moved up to the top level of `src/annotator/` since
there isn't a more specific home that is suitable. Also change the
`CrossFrame` class to be a named rather than default export per current
conventions.

In the longer term I think this class will probably need a complete overhaul and maybe before that we might group it in some other subdirectory, but for this PR I'm just moving it out of the `plugin` directory, since that's clearly inappropriate.

### DIFF
--- a/src/annotator/test/integration/anchoring-test.js
+++ b/src/annotator/test/integration/anchoring-test.js
@@ -53,12 +53,12 @@ describe('anchoring', function () {
 
   before(() => {
     guestImports.$mock({
-      './plugin/cross-frame': FakeCrossFrame,
+      './cross-frame': { CrossFrame: FakeCrossFrame },
     });
   });
 
   after(() => {
-    guestImports.$restore({ './plugins/cross-frame': true });
+    guestImports.$restore({ './cross-frame': true });
   });
 
   beforeEach(() => {
--- a/src/annotator/test/integration/anchoring-test.js
+++ b/src/annotator/test/integration/anchoring-test.js
@@ -58,7 +58,7 @@ describe('anchoring', function () {
   });
 
   after(() => {
-    guestImports.$restore({ './cross-frame': true });
+    guestImports.$restore();
   });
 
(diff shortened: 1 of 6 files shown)


esanzgar (approved):
Alternatively, maybe `cross-frame.js` could have been moved to the `util` directory.

src/annotator/test/integration/anchoring-test.js
```diff
@@ -53,12 +53,12 @@ describe('anchoring', function () {
...
   before(() => {
     guestImports.$mock({
-      './plugin/cross-frame': FakeCrossFrame,
+      './cross-frame': { CrossFrame: FakeCrossFrame },
     });
   });
 
   after(() => {
-    guestImports.$restore({ './plugins/cross-frame': true });
+    guestImports.$restore({ './cross-frame': true });
```
esanzgar, line 61: Why not to restore all the imports like this `guestImports.$restore()`?

robertknight: Good catch. I'm guessing there was a reason for this originally but it no longer applies.

robertknight:
> Alternatively, maybe cross-frame.js could have been moved to the util directory.

The `util/` directory is intended for code which doesn't have dependencies on other parts of the application. In the sidebar application we also have a distinction between code which is deeply application/domain-specific in `src/sidebar/helpers` (eg modules like `build-thread.js`) and code which is not (eg. modules like `src/sidebar/util/copy-to-clipboard.js`).

In the LMS frontend we also have a `utils/` directory which follows the "no dependencies on code in other directories" rule but not the "not domain-specific" guideline.

`cross-frame.js` is both highly application-specific and also has dependencies on non-utility code. We don't have a place for this really in `src/annotator/` at present but we might create one in future.

PR #3253, opened 2021-04-06
Improve JSX code formatting in dev server

robertknight:
Improve the formatting of code in `jsxToString`.

 - Indent children by two spaces

 - Handle boolean props closer to how we normally set them in JSX by omitting
   them if false and including them without explicit values if true.
   This assumes that boolean props default to false. If there is a
   scenario in future where this is not the case, this behavior may need
   to be configurable.

 - Handle `bigint` numeric values. We have no use case for this, but
   TypeScript pointed out that it wasn't handled.

 - Add JSDoc comments

----

New output:

JSX Menu after

JSX buttons after

### DIFF
--- a/dev-server/ui-playground/util/jsx-to-string.js
+++ b/dev-server/ui-playground/util/jsx-to-string.js
@@ -85,7 +85,7 @@ export function jsxToString(vnode) {
       return `<${name}${propStr}>\n${childrenStr}\n</${name}>`;
     } else {
       // No children - use a self-closing tag.
-      return `<${name}${propStr}/>`;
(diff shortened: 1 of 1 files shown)


lyzadanger (approved):
This looks nice! Thanks for picking this up.

I notice one small detail, which is that the last prop for self-closing tags doesn't have a space after it, which looks a little cramped:

`<IconButton icon="trash" title="Delete annotation" expanded/>`

Not having a space looks fine for tags that aren't self-closing:

```
<LabeledButton expanded>
  Edit
</LabeledButton>
```

PR #3262, opened 2021-04-07
Refactor `anchor` method in `Guest`

robertknight:
Rewrite the `Guest#anchor` method to simplify the control flow and generally make it easier to understand and change. There are no functional changes for annotations with one target (associated document location), which includes all existing Hypothesis annotations.

There is a functional change to handling of annotations with multiple targets [2]. Previously `anchor` would try to be smart about only re-anchoring targets which were not already anchored. In the new implementation all targets for an annotation are re-anchored. This will have no effect in practice because the Hypothesis client and server only support creating and storing one target per annotation. The anchoring logic has always supported multiple targets in theory though. Maybe in future we could just drop all support for multiple targets in the Hypothesis client. I demurred from going that far in this PR however.

 - Convert `anchor` method to async and replace Promise chains with async/await
 - Replace the logic that removes existing anchors and highlights for an annotation with a call to the `detach` method. This required adding an internal parameter to `detach` to control whether `anchorsChanged` is emitted, so that `anchor` only emits `anchorsChanged` once.
 - Add an explicit error to `Promise.reject` calls in tests so that the tests are easier to debug if they fail
 - Add some notes to the docs for the `Target` type and `annotation.target` field about usage in Hypothesis

[1] `Guest#createAnnotation` creates one target per `Range` in the `Selection` returned by `window.getSelection()`. The [Selection API documentation](https://w3c.github.io/selection-api/#definition) states that a selection "can be associated with a single range" (ie. there are zero or one ranges in the selection).
[2] _Multiple targets_ means that an annotation refers to more than one location in a document. Not supported on the Hypothesis client or server, but the representation of annotations in the API allows for it.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -335,83 +335,64 @@ export default class Guest {
   }
 
   /**
-   * Anchor (locate) an annotation's selectors in the document.
+   * Anchor an annotation's selectors in the document.
+   *
+   * _Anchoring_ resolves a set of selectors to a concrete region of the document
+   * which is then highlighted. The results of anchoring are broadcast to the
+   * rest of the application via `CrossFrame#sync`.
+   *
+   * Any existing anchors associated with `annotation` will be removed before
+   * re-anchoring the annotation.
    *
    * @param {AnnotationData} annotation
    * @return {Promise<Anchor[]>}
    */
-  anchor(annotation) {
-    let anchor;
-    const root = this.element;
-
-    // Anchors for all annotations are in the `anchors` instance property. These
-    // are anchors for this annotation only. After all the targets have been
-    // processed these will be appended to the list of anchors known to the
-    // instance. Anchors hold an annotation, a target of that annotation, a
-    // document range for that target and an Array of highlights.
-    const anchors = [];
-
-    // The targets that are already anchored. This function consults this to
-    // determine which targets can be left alone.
-    const anchoredTargets = [];
-
-    // These are the highlights for existing anchors of this annotation with
-    // targets that have since been removed from the annotation. These will
-    // be removed by this function.
-    let deadHighlights = [];
-
-    // Initialize the target array.
-    if (!annotation.target) {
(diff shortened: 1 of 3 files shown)


src/annotator/guest.js
```diff
@@ -421,98 +402,62 @@ export default class Guest {
...
     }
 
-    // Remove all the highlights that have no corresponding target anymore.
-    removeHighlights(deadHighlights);
+    // Set flag indicating whether anchoring succeeded. For each target,
+    // anchoring is successful either if there are no selectors (ie. this is a
+    // Page Note) or we successfully resolved the selectors to a range.
+    annotation.$orphan =
+      anchors.length > 0 &&
+      anchors.every(anchor => anchor.target.selector && !anchor.range);
```
LMS007, line 424: nice use of `.every`this is certainly easier to read than before.

src/annotator/guest.js
```diff
@@ -421,98 +402,62 @@ export default class Guest {
...
    */
-  detach(annotation) {
+  detach(annotation, notify = true) {
     const anchors = [];
-    let unhighlight = [];
     for (let anchor of this.anchors) {
-      if (anchor.annotation === annotation) {
-        unhighlight.push(...(anchor.highlights ?? []));
-      } else {
+      if (anchor.annotation !== annotation) {
```
LMS007, line 443: I don't see a test case where the annotation references are not equal.
e.g. we create 2 annotations, but don't push one to anchors, and then try and delete it.

LMS007 (approved):
Look much more simple Rob, thank you. my only bit of feedback was I noticed one missing test case and other then that, I can't see any red flags. It's still a complex bit of code, but its less complex now . If you can add that one test, then GTG

src/annotator/guest.js
robertknight: Indeed! - I've added a test to cover this and also refactored the existing `destroy` tests a little.

PR #3264, opened 2021-04-07
Add PDF version of The Great Gatsby to test documents

robertknight:
Add a full novel as a test document for testing performance and
behaviors related to long PDF documents, such as re-anchoring when pages
are rendered and un-rendered while scrolling through the document.

We try to keep example PDFs small to avoid bloating the git repository, but
this one is on the larger side (~700KB) due to its length.

This PDF was produced by downloading the EPUB version of the novel [1]
from Project Gutenberg and then converting it to a PDF using Calibre [2]

[1] https://www.gutenberg.org/ebooks/64317
[2] https://calibre-ebook.com


lyzadanger (approved):
Nice to have, thanks.

PR #3270, opened 2021-04-08
Show clearer error if opening login popup window fails

robertknight:
If opening the login popup window using `window.open` fails or succeeds but fails to return a `Window` reference to the caller (i.e. returns `null`) then show a more helpful error message.

There are various rare scenarios (browser bugs, browser extensions) which might cause this to happen. We have to use a popup window for the login so we're limited in what we can do to mitigate the problem, but at least we can show a better error message.

For context, we're trying to track down an issue that a new Hypothesis team member is encountering in Firefox: https://hypothes-is.slack.com/archives/C2BLQDKHA/p1617899428389800 (_Edit: The problem fixed itself for reasons unknown, but hey, we might as well continue to improve the error handling here_)

### DIFF
--- a/src/sidebar/util/oauth-client.js
+++ b/src/sidebar/util/oauth-client.js
@@ -255,10 +255,16 @@ export default class OAuthClient {
       })
       .replace(/&/g, ',');
 
-    return /** @type {Window} */ ($window.open(
+    const authWindow = $window.open(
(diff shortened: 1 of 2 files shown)


LMS007 (approved):
Yes, I think this is more helpful

PR #3275, opened 2021-04-12
Refactor side-by-side state in Guest

robertknight:
In issue discussions, tests and various parts of the code there are
mentions of side-by-side mode being active or not. The `Guest` class
didn't store this state directly but instead had a
`closeSidebarOnDocumentClick` state which is always the inverse of
"is side-by-side active?".

Make the code easier to follow by storing the "is side-by-side active?" state
directly. This will also avoid a misnamed variable if in future the
state has other effects besides changing whether clicking on the
document closes the sidebar.

 - Replace `closeSidebarOnDocumentClick` with a private
   `_sideBySideActive` property, exposed via a read-only
   `sideBySideActive` getter

 - Refactor tests to only use the public API of Guest (calling
   `Guest#fitSideBySide` or reading `sideBySideActive`)

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -162,8 +162,7 @@ export default class Guest {
     });
     this.crossframe.onConnect(() => this._setupInitialState(config));
 
-    // Whether clicks on non-highlighted text should close the sidebar
-    this.closeSidebarOnDocumentClick = true;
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
Thanks for this clarity improvement!

PR #3277, opened 2021-04-12
Improve readability of focused PDF highlights

robertknight:
The visible part of PDF highlights are created by SVG elements which are multiply-blended with the PDF content to enhance legibility of highlighted text.  The blueish background for focused highlights was still being applied to the transparent text layer on top of the PDF which contains the `<hypothesis-highlight>` elements. This impacted the legibility of focused highlights.

Improve the readability of PDF highlights by instead modifying the fill color of the SVG element.

**Before:**

PDF highlight before

**After:**

PDF highlight after

### DIFF
--- a/src/annotator/highlighter.js
+++ b/src/annotator/highlighter.js
@@ -327,6 +327,9 @@ export function removeHighlights(highlights) {
  */
 export function setHighlightsFocused(highlights, focused) {
   highlights.forEach(h => {
+    // In PDFs the visible highlight is created by an SVG element, so the focused
+    // effect is applied to that. In other documents the effect is applied to the
+    // `<hypothesis-highlight>` element.
     if (h.svgHighlight) {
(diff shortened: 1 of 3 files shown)


src/annotator/highlighter.js
```diff
@@ -326,9 +326,13 @@ export function removeHighlights(highlights) {
  * @param {boolean} focused
  */
 export function setHighlightsFocused(highlights, focused) {
-  highlights.forEach(h =>
-    h.classList.toggle('hypothesis-highlight-focused', focused)
-  );
+  highlights.forEach(h => {
+    if (h.svgHighlight) {
```
LMS007, line 330: If this only applies to PDFs, then it might helpful to point that out in comment here, and likewise for the else case for HTML.

LMS007 (approved):
Looks good Rob, might need a comment if that makes sense?

Issue #3236, opened 2021-04-01
Improve: config settings in the annotator 


LMS007:
## 1. Current implementation issues

Currently, the configuration settings in the annotator are gathered from `settings.js` and pulled into an object in `config/index.js`. There are several issues with the current implementation that are confusing to read and could be slightly improved

### 1a. Default Value

All values are set to a default value of null if they are missing / undefined. But there is also a `defaultValue` setting that can further override this null value.  Can we simply only use `defaultValue` and anything not defined can be just omitted from the config? At the very least, don't bake in null values and defined a `defaultValue` set to null for each item that needs to be null. At least this way its explicit from looking at a configuration definition. 

### 1b. Coerce methods

Currently, only one value has a coerce method (several more in the sidebar--unrelated to this ticket). We do need to maintain a way of augmenting values because incoming config may have older standards we need to maintain for backwards compatibility. Secondly, we need to support the more limited types that via can produce. e.g. It only supports strings. 

`showhighlights` also needs the be changed (depending on logic) and its currently done in `settings.js`. It seems that much of that logic is suitable for a coerce method similar to `openSidebar`. It's confusing that logic to coerce values is in multiple places.

### 1c. Browser extension context

All values are subject to being removed/nulled if in the browser extension context. This is separate from an application context because both can be true or worth orthogonally. In other words, we still need to honor `allowInBrowserExt` in the notebook context or sidebar, etc. It makes sense that each setting gets an `allowInBrowserExt` flag. Currently, this is not the case. All settings that come from `hostPageSetting` do honor the `allowInBrowserExt` flag (defaults to `false`), but the getter methods in `settings.js` no not honor the flag. In fact, `showHighlights` even calls down into `hostPageSetting()` which has the ability to check the flag, but `showHighlights` does not actually pass an options object down so it misses the opportunity. While it may be true that all the config settings derived from such getters should be allowed in the browser extension context, it's unclear from the code that they should be allowed in the first place. In other words, it should be expressed clearly that they are allowed or not.

getters:

```
annotations()
clientUrl()
group()
notebookAppUrl()
showHighlights()
sidebarAppUrl() 
query()
```

## 2. Resulting configs need to be different distinct applications.  (New feature)

[...]

PR #3291, opened 2021-04-13
Refactor annotator/config/index.js

LMS007:
Continuation of https://github.com/hypothesis/client/pull/3232 _Since the branch was force-pushed, I can't reopen the PR._

The changes here are mostly structural. I removed the class API and replaced it with just a single getConfig export. Additionally, I have used a separate list of keys with simple filters to defined the application contexts.

A follow up would be an audit of sidebar and annotator to see if any other keys could be omitted from those contexts. And, additionally, do we need to default all these values to null?

fixes https://github.com/hypothesis/client/issues/3236

### Commits

Expose `isBrowserExtension` flag from settingsFrom() …

The `isBrowserExtension` is needed external to the settings to be able to filter config values. As a result, settings-test.js needs a minor reconfiguration so that the link is mocked before each test.

---------

Refactor config/index.js to add app contexts …

Refactor config/index.js in an effort to filter the config based on the application context (sidebar, notebook, annotator, This refactor moves all the rules to a single configuration definition that makes it clear 1. where a setting comes from and 2. what rules it follows before getting passed along to the rest of the app.

Additionally, it was unclear which settings were allowed in the browser extension context because only the hostPageSetting settings made use of the boolean flag but some of the getters did not.

Other changes:

- When in browser context (isBrowserExtension), values which are prohibited are now omitted rather than set to null
- When requeuing a config using an app context, values outside of that context are omitted.
- Move isBrowserExtension flag from settings to index.js configDefinitions

--------

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -2,52 +2,218 @@ import settingsFrom from './settings';
 import { toBoolean } from '../../shared/type-coercions';
 
 /**
- * Reads the Hypothesis configuration from the environment.
- *
- * @param {Window} window_ - The Window object to read config from.
+ * @typedef ConfigDefinition
+ * @prop {boolean} allowInBrowserExt - Allow this name to be available in the browser extension
+ * @prop {any} [defaultValue] - Sets a default if `valueFn` returns undefined
+ * @prop {([configName]: string) => any} valueFn -
+ *   Method to retrieve the value from the incoming source
+ * @prop {(value: any) => any} [coerce] - Transform a value's type value or value
  */
-export default function configFrom(window_) {
-  const settings = settingsFrom(window_);
+
+/**
+ * @typedef {Record<string, ConfigDefinition>} ConfigDefinitionMap
+ */
+
+/**
+ * Total list of configuration keys.
+ */
+const configurationKeys = [
+  'annotations',
+  'assetRoot',
+  'branding',
+  'clientUrl',
+  'enableExperimentalNewNoteButton',
+  'experimental',
+  'group',
+  'focus',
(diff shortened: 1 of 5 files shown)


src/annotator/config/index.js
```diff
@@ -2,52 +2,218 @@ import settingsFrom from './settings';
...
+ * @prop {any} [defaultValue] - Sets a default if `valueFn` returns undefined
+ * @prop {([configName]: string) => any} valueFn -
+ *   Method to retrieve the value from the incoming source
+ * @prop {(value: any) => any} [coerce] - Transform a value's type value or value
  */
-export default function configFrom(window_) {
-  const settings = settingsFrom(window_);
+
+/**
+ * @typedef {Record<string, ConfigDefinition>} ConfigDefinitionMap
```
LMS007, line 14: Per suggestion, I changed this to be a Record rather than a defined key: value pair of every setting. This is clearly better.

src/annotator/config/index.js
```diff
@@ -2,52 +2,218 @@ import settingsFrom from './settings';
...
+export default function getConfig(appContext = 'annotator', window_ = window) {
+  const config = {};
+
+  // Filter the config based on the application context as some config values
+  // may be inappropriate or erroneous for some applications.
+  let partialConfigDef = {};
+  const settings = settingsFrom(window_);
+  const configDefs = configDefinitions(settings);
+
+  // One of 'annotator', 'sidebar', or 'notebook' or 'all'
```
LMS007, line 192:
```suggestion
  // One of 'annotator', 'sidebar', 'notebook' or 'all'
```

PR #3296, opened 2021-04-15
Convert `router` service to a native ES class

robertknight:
Services in `src/sidebar/services` are effectively classes. They are
currently implemented with a convention that was more common pre-ES6 where
the constructor is a function that creates the fields as local variables
and the methods using closures and returns an object that references the
closures.

This pattern has advantages, especially pre-ES6, as it avoids issues with
incorrect use of `this` and hides internal state from consumers. However
it also has downsides:

 - It is less obvious to readers that they are looking at something that
   is logically a class

 - This is not an idiom we use elsewhere in the codebase, where we use
   native classes instead

 - Static analysis tools don't support this pattern for creating a class
   as well as they support a native class. For example `TS` creates a
   named type for native classes, which is convenient to reference
   in JSDoc comments

This commit starts a process of refactoring service classes to ES
classes which are named `<Thing>Service`, using the router service as a
first example. Per recently agreed conventions, the classes are named
rather than default exports.

Related Slack discussion: https://hypothes-is.slack.com/archives/C1M8NH76X/p1618312556291000?thread_ts=1618307159.290300&cid=C1M8NH76X

### DIFF
--- a/src/sidebar/components/StreamSearchInput.js
+++ b/src/sidebar/components/StreamSearchInput.js
@@ -5,7 +5,7 @@ import SearchInput from './SearchInput';
 
 /**
  * @typedef StreamSearchInputProps
- * @prop {Object} router - Injected service
+ * @prop {import('../services/router').RouterService} router
  */
 
 /**
--- a/src/sidebar/services/router.js
+++ b/src/sidebar/services/router.js
@@ -5,14 +5,24 @@ import * as queryString from 'query-string';
  * implied by the URL and the corresponding route state in the store.
  */
 // @inject
-export default function router($window, store) {
+export class RouterService {
+  /**
+   * @param {Window} $window
+   * @param {import('../store').SidebarStore} store
+   */
+  constructor($window, store) {
+    this._window = $window;
+    this._store = store;
+    this._didRegisterPopstateListener = false;
+  }
+
   /**
    * Return the name and parameters of the current route.
    */
-  function currentRoute() {
-    const path = $window.location.pathname;
+  currentRoute() {
+    const path = this._window.location.pathname;
     const pathSegments = path.slice(1).split('/');
-    const params = queryString.parse($window.location.search);
+    const params = queryString.parse(this._window.location.search);
 
     // The extension puts client resources under `/client/` to separate them
     // from extension-specific resources. Ignore this part.
@@ -49,8 +59,11 @@ export default function router($window, store) {
 
   /**
    * Generate a URL for a given route.
+   *
+   * @param {string} name
+   * @param {Object.<string,string>} params
    */
-  function routeUrl(name, params = {}) {
+  routeUrl(name, params = {}) {
     let url;
     const queryParams = { ...params };
 
@@ -81,8 +94,6 @@ export default function router($window, store) {
     return url;
   }
 
-  let didRegisterPopstateListener = false;
-
   /**
    * Synchronize the route name and parameters in the store with the current
    * URL.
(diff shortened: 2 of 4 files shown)


src/sidebar/components/StreamSearchInput.js
```diff
@@ -5,7 +5,7 @@ import SearchInput from './SearchInput';
 
 /**
  * @typedef StreamSearchInputProps
- * @prop {Object} router - Injected service
+ * @prop {import('../services/router').RouterService} router
```
esanzgar, line 8: 👍🏼

src/sidebar/services/router.js
```diff
@@ -81,43 +94,39 @@ export default function router($window, store) {
...
+      this._didRegisterPopstateListener = true;
     }
   }
 
   /**
    * Navigate to a given route.
    *
    * @param {string} name
-   * @param {Object} params
+   * @param {Object.<string,string>} params
```
esanzgar, line 126:
```suggestion
   * @param {Record<string, string>} params
```

src/sidebar/services/router.js
```diff
@@ -49,8 +59,11 @@ export default function router($window, store) {
 
   /**
    * Generate a URL for a given route.
+   *
+   * @param {string} name
```
esanzgar, line 63:
```suggestion
   * @param {'annotation'|'notebook'|'stream'} route
```

src/sidebar/services/router.js
```diff
@@ -49,8 +59,11 @@ export default function router($window, store) {
...
+   * @param {Object.<string,string>} params
```
esanzgar, line 64:
```suggestion
   * @param {Record<'id', string>} [params]
```

src/sidebar/services/router.js
```diff
@@ -81,43 +94,39 @@ export default function router($window, store) {
...
       });
-      didRegisterPopstateListener = true;
+      this._didRegisterPopstateListener = true;
     }
   }
 
   /**
    * Navigate to a given route.
    *
    * @param {string} name
```
esanzgar, line 125:
```suggestion
   * @param {'annotation'|'notebook'|'stream'} route
```

src/sidebar/services/router.js
esanzgar: Or maybe the type suggested above.

src/sidebar/services/router.js
```diff
@@ -49,8 +59,11 @@ export default function router($window, store) {
...
   /**
    * Generate a URL for a given route.
+   *
+   * @param {string} name
+   * @param {Object.<string,string>} params
    */
-  function routeUrl(name, params = {}) {
+  routeUrl(name, params = {}) {
     let url;
     const queryParams = { ...params };
```
esanzgar, line 68:
```suggestion
    const {id, queryParams} =  params;
```
And then 3 lines (73-75) can be deleted

src/sidebar/services/router.js
robertknight: We're currently using a mix of `Record<Key, Value>` and `Object.<Key,Value>` in the code. We had a discussion about this a while back and you mentioned that [`Record` is stricter](https://hypothes-is.slack.com/archives/C1M8NH76X/p1611230893009700?thread_ts=1611226352.008200&cid=C1M8NH76X) which could help to avoid some errors, so I think I'm good with this change going forwards.

src/sidebar/services/router.js
robertknight: On this line `id` may or may not be a URL parameter for the route. Its only within the switch branches that we know what the valid route params are.

I expect in future as our routing needs get more sophisticated we'll introduce a better abstraction or possibly re-use some library to handles the route name <-> URL conversion process.

esanzgar (approved):
👍🏼 

I wish JSDoc would allow declare function/method overloading:
https://github.com/microsoft/TypeScript/issues/25590 

Otherwise, we could improve create a more constrained type fo routeUrl method.

src/sidebar/services/router.js
esanzgar: If the other two routes don't use `id`  then it is safe.

PR #3297, opened 2021-04-15
Convert `apiRoutes` service to ES class

robertknight:
Follow the pattern introduced in https://github.com/hypothesis/client/pull/3296 to convert the API routes service to a class and improve various types along the way.

### DIFF
--- a/src/sidebar/services/oauth-auth.js
+++ b/src/sidebar/services/oauth-auth.js
@@ -27,6 +27,7 @@ import { resolve } from '../util/url';
  *
  * @param {Window} $window
  * @param {import('./toast-messenger').ToastMessengerService} toastMessenger
+ * @param {import('./api-routes').APIRoutesService} apiRoutes
  * @inject
  */
 export default function auth(
--- a/src/types/api.js
+++ b/src/types/api.js
@@ -6,6 +6,22 @@
  */
 
 /**
+ * An entry in the API index response (`/api`) describing an API route.
+ *
+ * @typedef RouteMetadata
+ * @prop {string} method - HTTP method
+ * @prop {string} url - URL template
+ * @prop {string} desc - Description of API route
+ */
+
+/**
+ * Structure of the `links` field of the API index response (`/api`) describing
+ * available API routes.
+ *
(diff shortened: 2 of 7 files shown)


src/types/api.js
```diff
@@ -5,6 +5,22 @@
...
+ * @prop {string} method - HTTP method
+ * @prop {string} url - URL template
+ * @prop {string} desc - Description of API route
+ */
+
+/**
+ * Structure of the `links` field of the API index response (`/api`) describing
+ * available API routes.
+ *
+ * @typedef {{ [key: string]: RouteMap|RouteMetadata }} RouteMap
```
robertknight, line 21: I tried using `Record<string, RouteMap|RouteMetadata>` here but that turns out not to support the recursive nature of this type.

robertknight:
Code coverage found that one of the tests wasn't executing the intended code paths.

src/types/api.js
esanzgar: Understand

esanzgar (changes requested):
All looks good.

Once you make the slow test to not wait I will approve it.

esanzgar:
If you can make the fix for the slow test in a follow up commit, it will help me to review the PR.

src/sidebar/services/oauth-auth.js
```diff
@@ -25,6 +25,9 @@ import { resolve } from '../util/url';
  * Interaction with OAuth endpoints in the annotation service is delegated to
  * the `OAuthClient` class.
  *
+ * @param {Window} $window
+ * @param {import('./api-routes').APIRoutesService} apiRoutes
```
esanzgar, line 29: Thank you for adding these missing types!

robertknight:
I've pushed a commit to fix the slow test.

esanzgar (approved):
Another options to fix the slow test would have been to use `clock = sinon.useFakeTimers();`.

robertknight:
> Another options to fix the slow test would have been to use clock = sinon.useFakeTimers();.

Yes, that could have worked too. An upside is less code. A downside is that it means the `APIRoutesService` test would have knowledge of implementation details of `retryPromiseOperation`. In this instance I think either approach would be OK, as long as the risks of each are understood.

Issue #3298, opened 2021-04-15
Convert services to ES classes


robertknight:
As explained in https://github.com/hypothesis/client/pull/3296, sidebar services currently implemented classes in a way that makes it non-obvious to a reader that they are classes and isn't as well supported by our tooling (eg. TypeScript, ESLint) as native classes.

This issue covers converting the existing services in the client to ES classes, as well as improving the types of fields and methods within the class and ensuring that references to the class are correctly typed. This will help make future refactorings safer.

- [x] './services/annotations'
- [x] './services/api'
- [x] './services/api-routes' (https://github.com/hypothesis/client/pull/3297)
- [x] './services/oauth-auth' (https://github.com/hypothesis/client/pull/3303)
- [x] './services/autosave'
- [x] './services/features'
- [x] './services/frame-sync'
- [x] './services/groups' (https://github.com/hypothesis/client/pull/3370)
- [x] './services/load-annotations'
- [x] './services/local-storage'
- [x] './services/persisted-defaults'
- [x] './services/router' (https://github.com/hypothesis/client/pull/3296)
- [x] './services/service-url'
- [x] './services/session' (https://github.com/hypothesis/client/pull/3322)
- [x] './services/stream-filter' (https://github.com/hypothesis/client/pull/3302)
- [x] './services/streamer (https://github.com/hypothesis/client/pull/3368)
- [x] './services/tags'
- [x] './services/threads'
- [x] './services/toast-messenger' (https://github.com/hypothesis/client/pull/3301)

robertknight:
This is almost done now. https://github.com/hypothesis/client/pull/3405 is the last PR that needs to land to close this out.

robertknight:
https://github.com/hypothesis/client/pull/3405 is merged so this is done now.

PR #3301, opened 2021-04-16
Convert `toastMessenger` service to ES class

robertknight:
- Convert toastMessenger service to an ES class using a named rather than default export
 - Specify type of `toastMessenger` references to this service

This is part of https://github.com/hypothesis/client/issues/3298.


esanzgar (approved):
👍🏼

PR #3303, opened 2021-04-16
Convert `auth` service to ES class

robertknight:
This is part of https://github.com/hypothesis/client/issues/3298.

This converts the `auth` service to an ES class (`AuthService`) and renames the module from `oauth-auth.js` to `auth.js`. The previous name is a vestige from when there were multiple authentication services during a transition from cookie-based auth to OAuth-based auth.

The `AuthService` is not yet an idiomatic ES class because the fields and methods are still created using local variables and closures as before. I took the contents of the `auth` constructor function moved them into the `constructor` of the new class. Refactoring to an idiomatic class is non-trivial because there are quite a lot of fields and helpers. Hence I've left that as a follow-up exercise.

- The first commit renames the module and converts the `auth` function to the constructor of an `AuthService` class
- The second commit updates references to this service elsewhere in the code to be typed
- The last commit fixes some incorrect types I found in the groups service while doing this refactoring


src/sidebar/services/api.js
```diff
@@ -92,18 +89,13 @@ function get(object, path) {
  * @param links - Object or promise for an object mapping named API routes to
  *                URL templates and methods
  * @param route - The dotted path of the named API route (eg. `annotation.create`)
- * @param [APIMethodOptions] - Configuration for the API method
+ * @param {APIMethodOptions} options - Configuration for the API method
```
esanzgar, line 92: Shouldn't `options` be optional?

robertknight: > Shouldn't options be optional?

You'd expect so, but the naming here was bad. This is really a set of callbacks invoked at various points during the API request. We _could_ make some of them optional, but there is no benefit because this is an internal helper in the module and in practice we're always going to specify all of them.

esanzgar:
```suggestion
 * @param {APIMethodCallbacks} callbacks - Callbacks for the API method
```

esanzgar (approved):
You could convert the methods in `auth.js` to ordinary class methods, here in another commit or on a follow up PR.

src/sidebar/services/api.js
```diff
@@ -63,10 +67,10 @@ function stripInternalProperties(obj) {
  */
 
 /**
- * Configuration for an API method.
+ * Callbacks invoked at various points during an API call to get an access token etc.
  *
- * @typedef {Object} APIMethodOptions
- * @prop {() => Promise<string>} getAccessToken -
+ * @typedef {Object} APIMethodCallbacks
+ * @prop {() => Promise<string|null>} getAccessToken -
```
robertknight, line 73: A `null` result here indicates that the user is not authenticated.

src/sidebar/services/api.js
```diff
@@ -75,9 +79,6 @@ function stripInternalProperties(obj) {
  * @prop {() => any} onRequestFinished - Callback invoked when the API request finishes.
  */
 
-// istanbul ignore next
-const noop = () => null;
```
robertknight, line 79: The need for this was removed by making all fields of the `callbacks` argument to `createAPICall` mandatory.

src/sidebar/services/auth.js
```diff
@@ -0,0 +1,311 @@
...
+  /**
+   * @param {Window} $window
+   * @param {import('./api-routes').APIRoutesService} apiRoutes
+   * @param {import('./local-storage').LocalStorageService} localStorage
+   * @param {import('./toast-messenger').ToastMessengerService} toastMessenger
+   */
+  constructor($window, apiRoutes, localStorage, settings, toastMessenger) {
+    super();
+
+    /**
```
robertknight, line 40: The code from here until the end of the constructor is taken from the body of the old `auth` function without changes.

robertknight:
> You could convert the methods in auth.js to ordinary class methods, here in another commit or on a follow up PR.

From a previous experimentation that was a reasonably large change so I think it will be best to do that in a follow-up PR.

PR #3309, opened 2021-04-19
Convert `localStorage` service to an ES class

robertknight:
Convert `localStorage` service to an ES class and add types to various
places where it is referenced.

Part of https://github.com/hypothesis/client/issues/3298.


esanzgar (approved):
All good!

PR #3312, opened 2021-04-19
Convert `tags` service to ES class

robertknight:
This conversion revealed some incorrect and inconsistent types.

- The `filter` method was documented to return a `Tag` array but actually
  returned a string array
- The `store` method required a list of `Tag`s but only used the `text`
  property from it. This method has been simplified to just accept a
  list of tag strings instead.


esanzgar (approved):
Good catch!

PR #3313, opened 2021-04-19
Focused group notebook

esanzgar:


LMS007:
Looks like this solutions works exactly as you would expect. I fear this may complicate our settings structure even more, and it might be a good time to take a look at how things work.

I think the tricky part is going be how do we reconcile the use of the hostConfig after app launches with the more statically defined `settings` (merged config) at the start of the app launch.

I think this is the only example of hostConfig being used outside of fetchConfig (I think) and not at the start of the app. So this may be a valid use case and we might need to adopt a way or standard of getting updated/dynamic config settings, but then I don't know how we differentiate that with the static `settings` object. I think at the very least we need to make sure it's clear that settings are at load time, and are static. Then there may need to be runtime settings that may change during the course of the application. 

Or, perhaps there is a way we can move the event bits, such as `hashchange` into the host-config and/or the settings service so that if the group's source of truth is updated (e.g, the hash itself), then the value is updated too in the `settings` service object. This approach would likely want to morph the thing into more of a store, rather than a static object.

I'd like to hear more feedback, and I don't totally know if I'm making a mountain out of a mole hill.

esanzgar:
Thanks @LMS007 for your comments and for testing this PR! It is appreciated.

Did you have the chance to test in the situation you described [here](https://github.com/hypothesis/client/pull/3295#issue-615571639)? I think it should work.

I agree with you, I feel this solution could lead to more confusion around the configuration. If there is a cleaner solution to solve this, I would be happy to implement. Maybe @robertknight  can give an additional perspective.

src/sidebar/components/NotebookView.js
```diff
@@ -45,7 +46,21 @@ function NotebookView({ loadAnnotationsService, streamer }) {
   // likely to be focused (eg. because the notebook has been configured to
   // display a particular group when launched), we can optimistically fetch
   // annotations from that group.
-  const groupId = focusedGroup?.id || store.directLinkedGroupId();
+  const [groupId, setGroupId] = useState(
+    focusedGroup?.id || store.directLinkedGroupId()
+  );
```
robertknight, line 51: Since you are setting the focused group below in response to a fragment change, and this component is re-rendered whenever the focused group changes, you don't need to `groupId` to be a separate piece of state here.

robertknight (commented):
This is a simple solution that hadn't occurred to me, and it does perceptibly speed up opening the notebook for a second time by avoiding the need to re-fetch the profile and groups. I think the main recommendation I have is to abstract the details of how the notebook app gets focused-group updates into a service, so that it can be more easily extended and changed in future without having to modify notebook UI components. This way we could for example change it to use some `postMessage`-based communication system.

Regarding the current `hashchange`-based mechanism, there may be some complications with future changes when we introduce more proper routing into the notebook app, such that the URL ends up reflecting the current group or filter status. The notebook's URL is currently https://hypothes.is/notebook, but if we support opening the notebook in a new tab in future then I expect we'll want to have the URL incorporate the major filtering state (eg. https://hypothes.is/notebook?group={xyz}&user={userid} etc.).

Another complication I see is if a user joins a new group after launching the notebook for the first time, and then switches to it before opening the notebook. I would guess in that scenario that attempting to focus the group will fail currently.

A few comments about the code structure:

- The `groupId` in `NotebookView` does not need to become a piece of state. Whenever the focused group changes this component re-renders anyway and `groupId` will always refer to the `focusedGroup`. I would want to avoid having a notion of a focused group ID in multiple places, since there is code elsewhere in the app that will assume `store.focusedGroup()` is the current group.
- As I mentioned above, I think it would be wise to extract the logic that knows how to listen for notebook configuration changes into a separate service, so the `NotebookView` component isn't cluttered with these details and associated dependencies. A question that arises then is whether that service should also be responsible for updating the focused group.
- I understand Kyle's point about potential confusion surrounding `settings` which is generally static and re-using the hash to dynamically update app configuration. One answer to that would be to say explicitly that we are not introducing a mechanism to support dynamic configuration updates here, but are understand just providing a way for the annotator to ask the already-running Notebook app to focus a particular group. If we make that decision, then we don't need to worry about how to update settings and making all the things that currently use settings reactive to dynamic updates.
- If we are removing the `iframeKey` logic in `NotebookModal` I think it is important to add some comments there about how changes in the `NotebookIframe` props are communicated to the running application

esanzgar:
We have decided to investigate further to possibility of communication between the `sidebar` and `notebook` iframes.

PR #3314, opened 2021-04-20
Convert `autosave` and `persistedDefaults` services to ES classes

robertknight:
This is a straightforward conversion of the `autosave` and `persistedDefaults` services to ES classes.

There are no significant changes to the APIs or internals.

### DIFF
--- a/src/sidebar/services/autosave.js
+++ b/src/sidebar/services/autosave.js
@@ -1,66 +1,70 @@
-/**
- * A service for automatically saving new highlights.
- */
 import { retryPromiseOperation } from '../util/retry';
 
 /**
- * @param {import('./annotations').AnnotationsService} annotationsService
- * @param {import('../store').SidebarStore} store
+ * A service for automatically saving new highlights.
+ *
+ * @inject
  */
-// @inject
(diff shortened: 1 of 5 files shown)


src/sidebar/services/autosave.js
```diff
@@ -1,66 +1,70 @@
...
+     * @return {boolean}
+     */
+    const shouldSaveHighlight = htag => {
+      return !this._saving.has(htag) && !this._failed.has(htag);
+    };
+
+    /**
+     * Store-subscribed call back. Automatically save new highlights.
+     */
+    this._autosaveNewHighlights = () => {
```
esanzgar, line 36: I would suggest to either (1) move this function to the `init` method or (2) make it its own method. 

The `shouldSaveHighlight` could be a helper function inside `init` or make it its own method.

esanzgar (changes requested):
I feel it would be better to move the majority of the logic in the `constructor` to somewhere else.

robertknight:
> I feel it would be better to move the majority of the logic in the constructor to somewhere else.

Yes, I agree. I've pushed a commit that makes this change.

esanzgar (approved):

PR #3315, opened 2021-04-20
Modernize and simplify `features` service

robertknight:
**Depends on https://github.com/hypothesis/client/pull/3314**

 - Convert `features` service to a class
 - Remove the unused `flagEnabled` method. Calls to this method have
   already been replaced by `store.isFeatureEnabled(...)`.
 - Move the initialization of the background sidebar => annotator
   synchronization of feature flag information into an `init` method
   which is then called from the sidebar's entry point. This pattern is
   consistent with several other services. It is also necessary for services
   that are not depended upon by anything else because services are
   lazily instantiated.

Part of https://github.com/hypothesis/client/issues/3298.

### DIFF
--- a/src/sidebar/services/features.js
+++ b/src/sidebar/services/features.js
@@ -1,60 +1,43 @@
-/**
- * Provides access to feature flag states for the current
- * Hypothesis user.
- *
- * This service is a thin wrapper around the feature flag data in
- * the session state.
- *
- * Users of this service should assume that the value of any given flag can
- * change at any time and should write code accordingly. Feature flags should
- * not be cached, and should not be interrogated only at setup time.
- */
-
 import bridgeEvents from '../../shared/bridge-events';
-import warnOnce from '../../shared/warn-once';
 import { watch } from '../util/watch';
 
-// @inject
-export default function features(bridge, session, store) {
-  const currentFlags = () => store.profile().features;
-  const sendFeatureFlags = () => {
-    bridge.call(bridgeEvents.FEATURE_FLAGS_UPDATED, currentFlags() || {});
-  };
-
-  // Re-send feature flags to connected frames when flags change or a new
-  // frame connects.
-  watch(store.subscribe, [currentFlags, () => store.frames()], () =>
-    sendFeatureFlags()
-  );
-
+/**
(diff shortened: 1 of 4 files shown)


src/sidebar/services/features.js
```diff
@@ -1,60 +1,43 @@
...
+      );
+    };
+
+    // Re-send feature flags to connected frames when flags change or a new
+    // frame connects.
+    watch(
+      this._store.subscribe,
+      [currentFlags, () => this._store.frames()],
+      () => sendFeatureFlags()
+    );
```
esanzgar, line 41:
```suggestion
      [currentFlags, this._store.frames],
      sendFeatureFlags
    );
```

src/sidebar/services/features.js
```diff
@@ -1,60 +1,43 @@
...
 
-  return {
-    flagEnabled: flagEnabled,
-  };
+  init() {
+    const currentFlags = () => this._store.profile().features;
+    const sendFeatureFlags = () => {
+      this._bridge.call(
+        bridgeEvents.FEATURE_FLAGS_UPDATED,
+        currentFlags() || {}
```
esanzgar, line 31: Would this render the same results?

```suggestion
    const currentFlags = () => this._store.profile().features || {};
    const sendFeatureFlags = () => {
      this._bridge.call(
        bridgeEvents.FEATURE_FLAGS_UPDATED,
        currentFlags
```

esanzgar (approved):

src/sidebar/services/features.js
robertknight: Practically yes, although in the first case it makes an assumption that `frames` is a bound method. That's currently true and I don't see it changing. In general its a bit of an unfortunate refactoring hazard though.

src/sidebar/services/features.js
robertknight: This would result in a behavior change because the result of `currentFlags` would change its identity on each call which would affect the `watch` function below. It could however be changed to:

```js
const emptyFlags = {};  // Ensure same empty object on each call.
const currentFlags = () => this._store.profile().features || emptyFlags;
...
```

PR #3316, opened 2021-04-20
Refactor AuthService internals and rename `tokenGetter` to `getAccessToken`

robertknight:
This PR renames AuthService's weirdly-named `tokenGetter` method [1] to the more obvious `getAccessToken` and simplifies some internals of AuthService by using modern syntax (Promise chains => async/await) and extracting some helper functions out of the long `getAccessToken` method (`getGrantToken`, `exchangeAuthCodeForToken`). There are no functional changes.

There is still a TODO at the bottom of the AuthService constructor about converting closures into normal class methods. I plan to address that separately. It turned out to be easier to do this cleanup first.

[1] From memory I think the name `tokenGetter` came from some Angular 1.x-related feature originally.

### DIFF
--- a/src/sidebar/services/auth.js
+++ b/src/sidebar/services/auth.js
@@ -117,24 +117,6 @@ export class AuthService extends TinyEmitter {
     }
 
     /**
-     * Exchange the refresh token for a new access token and refresh token pair.
-     *
-     * @param {string} refreshToken
-     * @param {RefreshOptions} options
-     * @return {Promise<TokenInfo>} Promise for the new access token
-     */
-    function refreshAccessToken(refreshToken, options) {
-      return oauthClient()
-        .then(client => client.refreshToken(refreshToken))
-        .then(tokenInfo => {
-          if (options.persist) {
-            saveToken(tokenInfo);
-          }
-          return tokenInfo;
-        });
-    }
-
-    /**
      * Listen for updated access & refresh tokens saved by other instances of the
      * client.
      */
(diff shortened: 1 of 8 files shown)


src/sidebar/services/auth.js
```diff
@@ -116,24 +116,6 @@ export class AuthService extends TinyEmitter {
...
-    function refreshAccessToken(refreshToken, options) {
-      return oauthClient()
-        .then(client => client.refreshToken(refreshToken))
-        .then(tokenInfo => {
-          if (options.persist) {
-            saveToken(tokenInfo);
-          }
-          return tokenInfo;
-        });
-    }
```
robertknight, line 135: This function was moved lower in the file.

src/sidebar/services/auth.js
```diff
@@ -142,45 +124,91 @@ export class AuthService extends TinyEmitter {
...
-          // on the next call to `tokenGetter()`.
+          // on the next call to `getAccessToken()`.
           tokenInfoPromise = null;
           this.emit('oauthTokensChanged');
         }
       });
     };
 
-    function oauthClient() {
+    const oauthClient = async () => {
```
robertknight, line 134: Using an arrow function here will avoid a hazard with `this` in a later refactoring.

esanzgar: I noticed that on the `login` method below you chose not to use arrow function.

src/sidebar/services/auth.js
```diff
@@ -245,19 +251,14 @@ export class AuthService extends TinyEmitter {
...
-        const usingGrantToken =
-          typeof serviceConfig(settings)?.grantToken === 'string';
-
+        const grantToken = getGrantToken();
         tokenInfoPromise = refreshAccessToken(token.refreshToken, {
-          // If we are using automatic login via a grant token, do not persist the
-          // initial access token or refreshed tokens.
-          persist: !usingGrantToken,
+          // Only persist tokens if automatic login is not being used.
+          persist: typeof grantToken === 'undefined',
```
esanzgar, line 257: Why not simply?

```suggestion
          persist: grantToken === undefined,
```

I understand `undefined` can be redefined (before ES5) but this doesn't affect our application.

esanzgar (approved):
I found a little bit more challenging to review this refactor because it also included some other improvements.

src/sidebar/services/auth.js
robertknight: It's more a case of "I didn't get to refactoring it in this PR".

src/sidebar/services/auth.js
```diff
@@ -142,45 +124,96 @@ export class AuthService extends TinyEmitter {
...
-        return Promise.resolve(client);
-      }
-      return apiRoutes.links().then(links => {
-        client = new OAuthClient({
-          clientId: settings.oauthClientId,
-          authorizationEndpoint: links['oauth.authorize'],
-          revokeEndpoint: links['oauth.revoke'],
-          tokenEndpoint: tokenUrl,
-        });
         return client;
```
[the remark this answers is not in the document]
robertknight: This is an async function now so any non-Promise return value is automatically wrapped in a Promise.

src/sidebar/services/auth.js
robertknight: > I understand undefined can be redefined (before ES5) but this doesn't affect our application.

Good point. I think I did this out of habit. I will simplify it.

robertknight: Elsewhere in the codebase we have similar checks that could be simplified, as well as some where `typeof` is still needed to avoid triggering a `ReferenceError` if a variable does not exist.

PR #3318, opened 2021-04-21
Make listener passive

esanzgar:
Chrome 89 printed the following warnings on the console:

```
shadow-root.js:67 [Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive
```

See also: https://www.chromestatus.com/feature/5745543795965952

I have removed a listener that it isn't no longer needed, because it is wrapped in
a shadow DOM that already stop the bubbling up of the event.

Another 'touchstart' listener had the potential to slow down scrolling
on the host page on mobile devices. To avoid that we have made listener
passive.

These changes remove the warnings without affecting the current
behaviour of the application.

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -241,9 +241,6 @@ export default class Sidebar {
   _setupGestures() {
     const toggleButton = this.toolbar.sidebarToggleButton;
     if (toggleButton) {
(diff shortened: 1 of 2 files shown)


robertknight (approved):
LGTM

PR #3319, opened 2021-04-22
Convert `api` service to an ES class

robertknight:
This is a pretty straightforward conversion with no significant refactoring other than the function => class conversion.

### DIFF
--- a/src/sidebar/services/api.js
+++ b/src/sidebar/services/api.js
@@ -14,7 +14,7 @@ import { replaceURLParams } from '../util/url';
  */
 function translateResponseToError(response, data) {
   let message = response.status + ' ' + response.statusText;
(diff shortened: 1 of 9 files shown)


src/sidebar/services/api.js
```diff
@@ -193,70 +193,77 @@ function createAPICall(
...
       update: apiCall('profile.update'),
-    },
+    };
+  }
 
-    // The `links` endpoint is not included here. Clients should fetch these
-    // from the `apiRoutes` service.
-  };
+  /**
+   * Set the "client ID" sent with API requests.
```
esanzgar, line 258: Thanks for the explanation!

esanzgar (approved):

PR #3320, opened 2021-04-22
Convert `frameSync` service to ES class

robertknight:
Convert the `frameSync` service to an ES class and rewrite the
documentation that explains what the service does.

Some internal `frameSync` methods are still set up in the constructor
with closures. These can be converted to ES class methods in later
refactoring.

Some of the syntax of the `FrameSyncService` internals has been modernized but there is no
change to the logic in this PR, although I saw a few things that could be simplified in future.

### DIFF
--- a/src/sidebar/services/frame-sync.js
+++ b/src/sidebar/services/frame-sync.js
@@ -2,7 +2,7 @@ import debounce from 'lodash.debounce';
 
 import bridgeEvents from '../../shared/bridge-events';
 import Discovery from '../../shared/discovery';
-import * as metadata from '../helpers/annotation-metadata';
+import { isReply, isPublic } from '../helpers/annotation-metadata';
 import { watch } from '../util/watch';
 
 /**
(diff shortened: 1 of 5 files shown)


src/sidebar/services/frame-sync.js
```diff
@@ -25,205 +25,223 @@ export function formatAnnot(ann) {
...
 
 /**
- * This service runs in the sidebar and is responsible for keeping the set of
- * annotations displayed in connected frames in sync with the set shown in the
- * sidebar.
+ * Service that synchronizes annotations between the sidebar and host page.
  *
- * @param {import('./annotations').AnnotationsService} annotationsService
- * @param {import('../store').SidebarStore} store
+ * Annotations are synced in both directions. New annotations created in the host
```
esanzgar, line 30: Thanks for adding additional description to the class. It helps a lot!

esanzgar (approved):

PR #3321, opened 2021-04-22
Retry `/api/links` fetch if it fails

robertknight:
The `APIRoutesService` service had logic to automatically retry fetching of the `/api` endpoint used to get the metadata for various endpoints, but not the endpoint used to fetch URL templates for URLs in the H service. This means that if the request failed due to a temporary issue, all links in the sidebar/notebook pointing into the H service would remain blank for the rest of the session.

Move the logic for retrying API metadata fetching into the `getJSON` helper function so that it is used both when fetching the API index (`/api`) and the service links (`/api/links`).

### DIFF
--- a/src/sidebar/services/api-routes.js
+++ b/src/sidebar/services/api-routes.js
@@ -1,17 +1,23 @@
 import { retryPromiseOperation } from '../util/retry';
 
-/** @param {string} url */
+/**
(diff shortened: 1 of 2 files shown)


esanzgar (approved):
Good improvement! Thanks!

PR #3322, opened 2021-04-23
Convert `session` service to an ES class

robertknight:
This PR converts the `session` service to an ES class as part of https://github.com/hypothesis/client/issues/3298. In the process two unnecessary public fields (`state`, `profileFetchRetryOpts`) were removed.

In the process I also improved the types and docs for the `retryPromiseOperation` utility which we use in a few places to automatically retry requests. See the last commit for details.

Changes to the `session` service:

 - Convert the `session` service to an ES class
 - Update references to the `session` service elsewhere in the code to be typed

Changes to the tests:

 - Added a new test to check what happens if fetching the profile fails after all the retries are exhausted
 - Remove unnecessary use of custom Sinon sandbox
 - Refactor the way the service instance is created in tests to allow individual tests to do custom setup before the service is constructed
 - Change the way that the delay between retries is removed in tests to match how this is done in the `APIRoutesService` tests, using a simple `retryPromiseOperation` mock

### DIFF
--- a/src/sidebar/components/HelpPanel.js
+++ b/src/sidebar/components/HelpPanel.js
@@ -43,7 +43,7 @@ function HelpPanelTab({ linkText, url }) {
 /**
  * @typedef HelpPanelProps
  * @prop {AuthState} auth - Object with auth and user information
- * @prop {Object} session - Injected service
+ * @prop {import('../services/session').SessionService} session
  */
 
 /**
@@ -86,17 +86,16 @@ function HelpPanel({ auth, session }) {
     setActiveSubPanel(panelName);
   };
 
-  const dismissFn = session.dismissSidebarTutorial; // Reference for useCallback dependency
   const onActiveChanged = useCallback(
     active => {
       if (!active && hasAutoDisplayPreference) {
         // If the tutorial is currently being auto-displayed, update the user
         // preference to disable the auto-display from happening on subsequent
         // app launches
-        dismissFn();
+        session.dismissSidebarTutorial();
       }
     },
-    [dismissFn, hasAutoDisplayPreference]
+    [session, hasAutoDisplayPreference]
   );
 
   return (
--- a/src/sidebar/services/session.js
+++ b/src/sidebar/services/session.js
@@ -1,5 +1,5 @@
 import serviceConfig from '../config/service-config';
-import * as retryUtil from '../util/retry';
+import { retryPromiseOperation } from '../util/retry';
 import * as sentry from '../util/sentry';
 
 const CACHE_TTL = 5 * 60 * 1000; // 5 minutes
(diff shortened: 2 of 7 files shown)


src/sidebar/components/HelpPanel.js
```diff
@@ -86,17 +86,16 @@ function HelpPanel({ auth, session }) {
     setActiveSubPanel(panelName);
   };
 
-  const dismissFn = session.dismissSidebarTutorial; // Reference for useCallback dependency
```
robertknight, line 89: I'm not sure why a reference to this method was extracted in this way previously. Perhaps a mistaken assumption about how callback dependencies work.

src/sidebar/services/test/session-test.js
```diff
@@ -31,47 +23,59 @@ describe('sidebar/services/session', function () {
...
+    const retryPromiseOperation = async callback => {
+      // eslint-disable-next-line no-constant-condition
+      while (true) {
+        try {
+          return await callback();
+        } catch (err) {
+          // Try again
+        }
+      }
+    };
```
robertknight, line 53: This mock matches how `retryPromiseOperation` delays are removed in `APIRoutesService` tests.

src/sidebar/util/retry.js
```diff
@@ -1,29 +1,41 @@
 import retry from 'retry';
 
+/**
+ * Options passed that control how the operation is retried.
+ *
+ * See https://github.com/tim-kos/node-retry#retrytimeoutsoptions
+ *
+ * @typedef RetryOptions
+ * @prop {number} minTimeout
+ */
```
robertknight, line 10: This type is incomplete. I've only included the properties that we've actually found a use for so far.

src/sidebar/services/test/session-test.js
```diff
@@ -148,6 +150,22 @@ describe('SessionService', () => {
         });
       });
 
+      it('should reject if the profile fetch repeatedly fails', async () => {
+        const fetchError = new Error('Server error');
+        fakeApi.profile.read.rejects(fetchError);
+
+        const session = createService();
+        let error;
+        try {
```
LMS007, line 159: I think we can just make this a 1-liner

`await assert.rejects(session.load(), fetchError.message);`

src/sidebar/services/session.js
```diff
@@ -10,31 +10,32 @@ const CACHE_TTL = 5 * 60 * 1000; // 5 minutes
...
-      return null;
-    }
-    return service.authority;
-  }
+export class SessionService {
+  /**
+   * @param {import('../store').SidebarStore} store
+   * @param {import('./api').APIService} api
+   * @param {import('./auth').AuthService} auth
+   * @param {import('./toast-messenger').ToastMessengerService} toastMessenger
```
LMS007, line 20: looks like `settings` is missing from the param types

LMS007 (approved):
My knowledge of this module is limited, but appear to work the same as before. I just had 2 small suggestions below. If you agree, gtg.

src/sidebar/services/session.js
robertknight: That's currently true of all services. My plan was to go around and address that in all services separately.

src/sidebar/services/test/session-test.js
robertknight: Done! - I did forget to use `await` when I first made this change to I had to push a follow-up commit. This issue (an atypical usage pattern) makes me thing that we probably shouldn't be monkey-patching this method onto `assert` and instead just import the helper explicit from some module.

PR #3324, opened 2021-04-23
Rephrased comment referring to Delegator

esanzgar:
Delegator does no longer exits.

The new comment explains the source of the events and its final
destination.

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -208,8 +208,8 @@ export default class Sidebar {
     this.guest.crossframe.on('openSidebar', () => this.open());
     this.guest.crossframe.on('closeSidebar', () => this.close());
 
-    // Re-publish the crossframe event so that anything extending Delegator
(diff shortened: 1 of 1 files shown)


src/annotator/sidebar.js
```diff
@@ -208,8 +208,8 @@ export default class Sidebar {
     this.guest.crossframe.on('openSidebar', () => this.open());
     this.guest.crossframe.on('closeSidebar', () => this.close());
 
-    // Re-publish the crossframe event so that anything extending Delegator
-    // can subscribe to it (without need for crossframe)
+    // Sidebar listens to the `openNotebook` event coming from the sidebar's
+    // iframe and re-publish it via the emitter to the Notebook
```
robertknight, line 212:
```suggestion
    // iframe and re-publishes it via the emitter to the Notebook
```

robertknight (approved):
👍

PR #3325, opened 2021-04-23
Simplification of code

esanzgar:
A few lines have been removed as a result of the code simplification.

### DIFF
--- a/src/annotator/cross-frame.js
+++ b/src/annotator/cross-frame.js
@@ -42,8 +42,6 @@ export class CrossFrame {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


src/annotator/cross-frame.js
```diff
@@ -42,8 +42,6 @@ export class CrossFrame {
         return;
       }
 
-      const { clientUrl } = config;
```
esanzgar, line 45: Moved closer to where it is used.

LMS007 (approved):
👍

PR #3344, opened 2021-04-26
Fix scrolling to highlights in un-rendered PDF pages

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/3343**~~

When scrolling to an anchor in a page that has not been rendered by
PDF.js, the anchor actually references a placeholder element in the
middle of the page. In order to scroll to the correct location, scrolling
for such anchors needs to happen in three phases:

 1. Scroll to the approximate location of the final anchor, given by the
    placeholder anchor. This will trigger PDF.js to re-render the target
    page.
 2. Wait for PDF.js to finish re-rendering the page and for the client
    to finish re-anchoring the annotation.
 3. Scroll to the real/non-placeholder anchor.

Changes in detail:

- Add a `scrollElement` utility in `annotator/util/scroll.js` which animates an element's scroll to a target offset and returns a promise for when the scroll is complete
- Change `PDFIntegration`'s `scrollToAnchor` method to make use of `scrollElement` to implement the three-phase scrolling process described above

Known issues:

- When scrolling to an annotation in an un-rendered page, the highlight should appear focused, but does not. I will address this separately (see https://github.com/hypothesis/client/pull/3345).
- Depending on system performance, PDF complexity etc. scrolling can feel too slow. I'm planning to tweak this once the basic functionality is working properly. In this PR the total scrolling time should be about the same as before

Fixes https://github.com/hypothesis/client/issues/3269

### DIFF
--- a/src/annotator/integrations/pdf.js
+++ b/src/annotator/integrations/pdf.js
@@ -353,6 +353,9 @@ export class PDFIntegration {
       return;
     }
 
+    // nb. We only compute the scroll offset once at the start of scrolling.
+    // This is important as the highlight may be removed from the document during
+    // the scroll due to a page transitioning from rendered <-> un-rendered.
     await scrollElement(this.contentContainer(), offset);
 
     if (inPlaceholder) {
--- a/src/annotator/integrations/pdf.js
+++ b/src/annotator/integrations/pdf.js
@@ -54,7 +54,7 @@ export class PDFIntegration {
   /**
    * @param {Annotator} annotator
    * @param {object} options
-   *   @param {number} [options.reanchoringWait] - Max time to wait for
+   *   @param {number} [options.reanchoringMaxWait] - Max time to wait for
    *     re-anchoring to complete when scrolling to an un-rendered page.
    */
   constructor(annotator, options = {}) {
@@ -82,7 +82,7 @@ export class PDFIntegration {
      * Amount of time to wait for re-anchoring to complete when scrolling to
      * an anchor in a not-yet-rendered page.
      */
-    this._reanchoringWait = options.reanchoringWait ?? 3000;
+    this._reanchoringMaxWait = options.reanchoringMaxWait ?? 3000;
 
     /**
      * A banner shown at the top of the PDF viewer warning the user if the PDF
@@ -361,7 +361,7 @@ export class PDFIntegration {
     if (inPlaceholder) {
       const anchor = await this._waitForAnnotationToBeAnchored(
         annotation,
-        this._reanchoringWait
+        this._reanchoringMaxWait
       );
       if (!anchor) {
         return;
@@ -390,6 +390,9 @@ export class PDFIntegration {
       anchor = this.annotator.anchors.find(a => a.annotation === annotation);
       if (!anchor || anchorIsInPlaceholder(anchor)) {
         anchor = null;
+
+        // If no anchor was found, wait a bit longer and check again to see if
+        // re-anchoring completed.
         await delay(20);
       }
     } while (!anchor && Date.now() - start < maxWait);
--- a/src/annotator/integrations/pdf.js
+++ b/src/annotator/integrations/pdf.js
@@ -35,7 +35,7 @@ const MIN_PDF_WIDTH = 680;
  *
  * @param {Anchor} anchor
  */
-function anchorIsPlaceholder(anchor) {
+function anchorIsInPlaceholder(anchor) {
   const highlight = anchor.highlights?.[0];
   return highlight && isInPlaceholder(highlight);
 }
@@ -347,7 +347,7 @@ export class PDFIntegration {
    */
   async scrollToAnchor(anchor) {
     const annotation = anchor.annotation;
-    const isPlaceholder = anchorIsPlaceholder(anchor);
+    const inPlaceholder = anchorIsInPlaceholder(anchor);
     const offset = this._anchorOffset(anchor);
     if (offset === null) {
       return;
@@ -355,7 +355,7 @@ export class PDFIntegration {
 
     await scrollElement(this.contentContainer(), offset);
 
-    if (isPlaceholder) {
+    if (inPlaceholder) {
       const anchor = await this._waitForAnnotationToBeAnchored(
         annotation,
         this._reanchoringWait
@@ -385,7 +385,7 @@ export class PDFIntegration {
       // nb. Re-anchoring might result in a different anchor object for the
       // same annotation.
       anchor = this.annotator.anchors.find(a => a.annotation === annotation);
-      if (!anchor || anchorIsPlaceholder(anchor)) {
+      if (!anchor || anchorIsInPlaceholder(anchor)) {
         anchor = null;
         await delay(20);
       }
--- a/src/annotator/integrations/pdf.js
+++ b/src/annotator/integrations/pdf.js
@@ -1,6 +1,5 @@
 import debounce from 'lodash.debounce';
 import { render } from 'preact';
-import scrollIntoView from 'scroll-into-view';
 
 import {
   RenderingStates,
@@ -8,15 +7,17 @@ import {
   describe,
   documentHasText,
 } from '../anchoring/pdf';
-import { removePlaceholder } from '../anchoring/placeholder';
+import { isInPlaceholder, removePlaceholder } from '../anchoring/placeholder';
 import WarningBanner from '../components/WarningBanner';
 import { createShadowRoot } from '../util/shadow-root';
 import { ListenerCollection } from '../util/listener-collection';
+import { offsetRelativeTo, scrollElement } from '../util/scroll';
 
 import { PDFMetadata } from './pdf-metadata';
 
 /**
  * @typedef {import('../../types/annotator').Anchor} Anchor
+ * @typedef {import('../../types/annotator').AnnotationData} AnnotationData
  * @typedef {import('../../types/annotator').Annotator} Annotator
  * @typedef {import('../../types/annotator').HypothesisWindow} HypothesisWindow
  * @typedef {import('../../types/annotator').Integration} Integration
@@ -30,6 +31,21 @@ import { PDFMetadata } from './pdf-metadata';
 const MIN_PDF_WIDTH = 680;
 
 /**
+ * Return true if `anchor` is in an un-rendered page.
+ *
+ * @param {Anchor} anchor
+ */
+function anchorIsPlaceholder(anchor) {
+  const highlight = anchor.highlights?.[0];
+  return highlight && isInPlaceholder(highlight);
+}
+
+/** @param {number} ms */
+function delay(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+/**
  * Integration that works with PDF.js
  *
  * @implements {Integration}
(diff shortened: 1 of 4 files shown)


src/annotator/integrations/pdf.js
```diff
@@ -309,14 +334,81 @@ export class PDFIntegration {
...
+    const inPlaceholder = anchorIsInPlaceholder(anchor);
+    const offset = this._anchorOffset(anchor);
+    if (offset === null) {
+      return;
+    }
+
+    // nb. We only compute the scroll offset once at the start of scrolling.
+    // This is important as the highlight may be removed from the document during
+    // the scroll due to a page transitioning from rendered <-> un-rendered.
+    await scrollElement(this.contentContainer(), offset);
```
robertknight, line 359: https://github.com/hypothesis/client/issues/3269 notes a sporadic problem where the PDF would sometimes jump right to the top after clicking on an annotation card:

> Additionally if you repeat steps 3 and 4 a few times you may also encounter times when the viewer starts scrolling and then reverts to scrolling back up to the top.

The cause of this, prior to this PR, is when the highlight element got removed from the DOM mid-scroll, causing `scrollIntoView` to compute its offset as 0.

src/annotator/integrations/pdf.js
```diff
@@ -29,6 +30,21 @@ import { PDFMetadata } from './pdf-metadata';
 // is enough room. Otherwise, allow sidebar to overlap PDF
 const MIN_PDF_WIDTH = 680;
 
+/**
+ * Return true if `anchor` is in an un-rendered page.
+ *
+ * @param {Anchor} anchor
+ */
+function anchorIsPlaceholder(anchor) {
```
lyzadanger: I worry again about the conflation of _placeholder_ to mean both the single placeholder-per-page in which anchors are quarantined until after render _and_ a single anchor-highlight that is within said one-per-page placeholder...

src/annotator/integrations/pdf.js
```diff
@@ -59,6 +78,12 @@ export class PDFIntegration {
       subtree: true,
     });
 
+    /**
+     * Amount of time to wait for re-anchoring to complete when scrolling to
+     * an anchor in a not-yet-rendered page.
+     */
+    this._reanchoringWait = options.reanchoringWait ?? 3000;
```
lyzadanger, line 85: I had to go look at the implementation of the `_waitForAnnotationToBeAnchored` method to understand that this is a max wait/timeout value. Perhaps the comment or the attribute name could be adjusted to make this more clear? I initially read this as a "wait exactly this amount of time."

src/annotator/integrations/pdf.js
```diff
@@ -309,14 +334,78 @@ export class PDFIntegration {
...
+
+    if (isPlaceholder) {
+      const anchor = await this._waitForAnnotationToBeAnchored(
+        annotation,
+        this._reanchoringWait
+      );
+      if (!anchor) {
+        return;
+      }
+      const offset = this._anchorOffset(anchor);
```
lyzadanger, line 369: Is there any value in making sure that this `offset` is different from the estimated/placeholder offset (and not re-scrolling if they are equivalent), or is that so unlikely as to be not worth the check?

src/annotator/integrations/pdf.js
```diff
@@ -309,14 +334,78 @@ export class PDFIntegration {
...
+   * Wait for an annotation to be anchored in a rendered page.
+   *
+   * @param {AnnotationData} annotation
+   * @param {number} maxWait
+   * @return {Promise<Anchor|null>}
+   */
+  async _waitForAnnotationToBeAnchored(annotation, maxWait) {
+    const start = Date.now();
+    let anchor;
+    do {
```
lyzadanger, line 387: Man, ace use of `do/while`. 5 points.

src/annotator/integrations/pdf.js
```diff
@@ -309,14 +334,78 @@ export class PDFIntegration {
...
+      // nb. Re-anchoring might result in a different anchor object for the
```
lyzadanger, line 388: What's the impact if that happens?

_Update_: I'm working through this by following a test case. It believe it means that you may well get a different `anchor` object result when asking the same question (`this.annotator.anchors.find(an anchor for this annotation)`) on different loop iterations. Correct?

src/annotator/integrations/pdf.js
```diff
@@ -309,14 +334,78 @@ export class PDFIntegration {
...
+      // same annotation.
+      anchor = this.annotator.anchors.find(a => a.annotation === annotation);
```
lyzadanger, line 390: This makes me wonder if a method should exist on `this.annotator` for finding an `anchor`. Is it kosher to reach that far down in? Minor note, just curious. See also questions on the test for this method.

src/annotator/integrations/pdf.js
```diff
@@ -309,14 +334,78 @@ export class PDFIntegration {
...
+      if (!anchor || anchorIsPlaceholder(anchor)) {
+        anchor = null;
+        await delay(20);
```
lyzadanger, line 393: Can you add a comment on this `delay` (what it's for?)?

src/annotator/integrations/test/pdf-test.js
```diff
@@ -369,15 +373,119 @@ describe('PDFIntegration', () => {
...
+      );
+
+      // Simulate a delay while rendering of the text layer for the page happens
+      // and re-anchoring completes.
+      await delay(5);
+
+      // Create a new anchor for the annotation created by re-anchoring.
+      const finalHighlight = document.createElement('div');
+      fakeScrollUtils.scrollElement.resetHistory();
+      fakeAnnotator.anchors.push({
```
lyzadanger, line 439: It's opaque here what happens when you push an `anchor` object to the `annotator`'s `anchors` collection that pertains to an `annotation` already referenced by a (or multiple) `anchor` objects in that collection. Given that `Array.prototype.find` (used in the method under test) will only return the first match, it matters. Can you help shed some light on this?

src/annotator/integrations/test/pdf-test.js
```diff
@@ -369,15 +373,119 @@ describe('PDFIntegration', () => {
...
+      const highlight = createPlaceholderHighlight();
+      const integration = createPDFIntegration({ reanchoringWait: 10 });
+      const annotation = { $tag: 'tag1' };
+      const anchor = { annotation, highlights: [highlight] };
+
+      const scrollDone = integration.scrollToAnchor(anchor);
+      await delay(5); // Simulate delay in re-anchoring
+      fakeScrollUtils.scrollElement.resetHistory();
+
+      // Wait until the re-anchoring timeout expires.
```
lyzadanger, line 467: I understand that the `delay(5)` is not actually related to the timeout here... The timeout happens because there isn't information/elements present to allow scrolling within `10ms` (non-placeholder anchor highlights are _never_ available, `10ms` or `10000ms` from now). I presume the `10ms` value is just because it's fairly quick and lets tests run quickly...Correct?

src/annotator/integrations/pdf.js
lyzadanger: Golly. Good catch. I could see this causing a hard-to-track-down bug, yeah.

lyzadanger (approved):
Neat! I can see the improvement locally on comments on _The Great Gatsby_.

I encountered some questions about multiple anchors referencing the same `annotation` that I'd like to understand better...

src/annotator/util/scroll.js
```diff
@@ -0,0 +1,74 @@
...
+ * @return {Promise<void>} - A promise that resolves once the scroll animation
+ *   is complete
+ */
+export async function scrollElement(
+  element,
+  offset,
+  { maxDuration = 500 } = {}
+) {
+  const initialOffset = element.scrollTop;
+  const targetOffset = offset;
```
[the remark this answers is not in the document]
robertknight: Maybe `startOffset` and `endOffset`?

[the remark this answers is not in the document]
lyzadanger: That seems good!

src/annotator/util/scroll.js
```diff
@@ -0,0 +1,74 @@
...
+  const scrollStart = Date.now();
+
+  // Choose a scroll duration proportional to the scroll distance, but capped
+  // to avoid it being too slow.
+  const pixelsPerMs = 3;
```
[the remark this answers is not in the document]
robertknight: I think it makes sense since `Ms` is an abbreviation rather than an acryonym. An alternative would be to use "millis" instead, but we do use `ms` quite widely as a shorthand.

src/annotator/util/test/scroll-test.js
```diff
@@ -0,0 +1,71 @@
...
+      const child = document.createElement('div');
+      child.style.position = 'absolute';
+      child.style.top = '100px';
+
+      assert.equal(offsetRelativeTo(child, parent), 0);
+    });
+  });
+
+  describe('scrollElement', () => {
+    it("animates the element's `scrollTop` offset to the target position", async () => {
```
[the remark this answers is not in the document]
robertknight: No, there is no testing of the animation specifics here, just that the end value reached is correct. We _could_ do that, but we'd have to be careful that it wasn't sensitive to test execution timing.

src/annotator/integrations/pdf.js
robertknight: I renamed this in a later commit to `anchorIsInPlaceholder`.

src/annotator/integrations/pdf.js
robertknight: The scroll delay is proportional to the difference between current and final offset, so if the two are identical it will simply be instant.

src/annotator/integrations/pdf.js
robertknight: Yes. Most importantly, the anchor associated with an annotation's tag in `this.annotator.anchor` will be different when the anchor is in a placeholder vs when it is in the real text layer.

src/annotator/integrations/pdf.js
robertknight: Good question - this is dubious but we're already traversing the `anchors` list elsewhere in this module, so we're not exposing new internals of the `Guest` class to `PDFIntegration` here.

src/annotator/integrations/pdf.js
robertknight: Will do. This is just a sleep to let the Guest/PDF.js/PDF anchoring code do its thing. Sleeping is a slightly lazy solution here - a better one would be to have some concrete "annotation with tag X is anchored" promise that could be waited upon.

src/annotator/integrations/test/pdf-test.js
robertknight: In practice all annotations created with Hypothesis or stored in h will only ever have a single target and thus one anchor. If an annotation happened to have multiple targets (eg. because it had been stored in some non-h backend that supported it) then only the first would be used.

src/annotator/integrations/test/pdf-test.js
robertknight: Right, the 10ms value is just to shorten test execution. Importantly it is less than the 2000ms timeout for tests.

robertknight:
Thanks for the feedback Lyza. I've updated variable/field names and added a few comments accordingly.

PR #3350, opened 2021-04-28
Rename `BucketBar.element`

robertknight:
The name was vestigial. Rename it to `_bucketsContainer` to be more
meaningful and indicate that it is private. Several other private fields
have also been prefixed accordingly.


esanzgar (approved):
Thanks for adding the comment on top of the class.

PR #3353, opened 2021-04-28
Remove unnecessary check and simplify code

esanzgar:
- Observe changes in `enable-annotation` attribute. Currently, if an
  existing iframe adds the attribute it won't be detected.

- `sidebar` or `notebook` iframes are shadow DOMed so they are not
  returned by `querySelector...` or `getElement...` (from elements
  outside the shadow root). Therefore, it is not necessary to filter
  these iframes out.

### DIFF
--- a/src/annotator/frame-observer.js
+++ b/src/annotator/frame-observer.js
@@ -29,6 +29,7 @@ export default class FrameObserver {
     this._mutationObserver.observe(this._target, {
       childList: true,
       subtree: true,
+      attributeFilter: ['enable-annotation'],
     });
   }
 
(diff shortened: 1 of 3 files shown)


robertknight:
> sidebar or notebook iframes are shadow DOMed so they are not returned by querySelector... or getElement... (from elements outside the shadow root).

We currently still support Edge v17/v18 which do not support shadow DOM. Do we need to retain this check for those browsers?

robertknight (approved):
LGTM. I agree that the check to filter out the Hypothesis sidebar was unnecessary since we are creating that iframe ourselves and control the attributes on it.

PR #3356, opened 2021-04-29
Modernize RPC client/server class.

robertknight:
Modernize the `RPC` class in `src/shared/frame-rpc` that serves as both
the client and server for RPC requests between frames.

This code was originally imported from
https://github.com/substack/frame-rpc but now that we've adopted it, it
makes sense to modernize it to match the conventions and requirements of
the rest of our code (eg. in terms of linting and style).

nb. We're missing unit tests for this module specifically but we do have indirect tests via other modules for much of it.

 - Convert `frame-rpc.js` to modern syntax
 - Update JSDoc documentation for methods and class
 - Convert `RPC` class to a named rather than default export
 - Inline the `apply` method into the `call` method, since only `call`
   was used externally
 - Remove unused option for `methods` constructor argument to be a
   function
 - Rename `src` and `dest` public fields to the more explicit
   `sourceFrame` and `destFrame`

Depending on what we decide to do with regards to fixing sidebar <-> notebook communication this may end up being obsolete. We'll see 🤷

### DIFF
--- a/src/shared/frame-rpc.js
+++ b/src/shared/frame-rpc.js
@@ -1,6 +1,7 @@
-/* eslint-disable */
+/*
+  This module was adapted from `index.js` in https://github.com/substack/frame-rpc.
 
-/** This software is released under the MIT license:
+  This software is released under the MIT license:
 
   Permission is hereby granted, free of charge, to any person obtaining a copy
   of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


src/shared/frame-rpc.js
```diff
@@ -21,106 +22,144 @@
...
+    this.sourceFrame = sourceFrame;
+    this.destFrame = destFrame;
 
-  this._sequence = 0;
-  this._callbacks = {};
+    if (origin === '*') {
+      this.origin = '*';
+    } else {
+      const uorigin = new URL(origin);
+      this.origin = uorigin.protocol + '//' + uorigin.host;
```
esanzgar, line 68:
```suggestion
      const url = new URL(origin);
      this.origin = `${url.protocol}//${url.host}`;
```

src/shared/frame-rpc.js
```diff
@@ -21,106 +22,144 @@
...
+  /**
+   * Send an RPC request to the destination frame.
+   *
+   * If the final argument in `args` is a function, it is treated as a callback
+   * which is invoked with the response.
+   *
+   * @param {string} method
+   * @param {any[]} args
+   */
+  call(method, ...args) {
```
esanzgar, line 112: A curiosity, typescript allows `...rest` parameters at the begging or middle of an assignment:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-2-beta/#leading-middle-rest-elements-in-tuple-types

esanzgar (approved):

src/shared/frame-rpc.js
robertknight: Huh, that's interesting. Good to know!

PR #3368, opened 2021-05-03
Convert `streamer` service to an ES class (1/n)

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/3371**~~

Convert the `streamer` service that handles WebSocket connections to an ES class as part of https://github.com/hypothesis/client/issues/3298. Other than changing `StreamerService` from a default to a named export there are no API changes. I did add some missing tests for existing error paths in the code.

Since this class has quite a few fields and helper methods the conversion is being done in two stages. This first PR converts the function to a class but all the logic remains in the constructor and the exported fields/methods are assigned as properties at the end of the constructor. A later PR will move code out of the constructor and into other methods.

### DIFF
--- a/src/sidebar/components/TopBar.js
+++ b/src/sidebar/components/TopBar.js
@@ -28,7 +28,7 @@ import UserMenu from './UserMenu';
  * @prop {() => any} onLogout - Callback invoked when user clicks "Logout" action in account menu.
  * @prop {() => any} onSignUp - Callback invoked when user clicks "Sign up" button.
  * @prop {MergedConfig} settings
- * @prop {import('../services/streamer').default} streamer
+ * @prop {import('../services/streamer').StreamerService} streamer
  */
 
 /**
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -36,7 +36,11 @@ if (process.env.NODE_ENV !== 'production') {
 // Install Preact renderer options to work around browser quirks
 rendererOptions.setupBrowserFixes();
 
-// @inject
+/**
+ * @param {import('./services/api').APIService} api
+ * @param {import('./services/streamer').StreamerService} streamer
+ * @inject
+ */
 function setupApi(api, streamer) {
   api.setClientId(streamer.clientId);
 }
@@ -83,8 +87,8 @@ function initServices(
 /**
  * @param {import('./services/frame-sync').FrameSyncService} frameSync
  * @param {import('./store').SidebarStore} store
+ * @inject
  */
-// @inject
 function setupFrameSync(frameSync, store) {
   if (store.route() === 'sidebar') {
     frameSync.connect();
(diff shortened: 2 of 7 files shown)


src/sidebar/components/TopBar.js
```diff
@@ -21,19 +21,21 @@ import UserMenu from './UserMenu';
...
- * @prop {Object} [streamer]
+ * @prop {AuthState} auth
+ * @prop {Bridge} [bridge]
+ * @prop {boolean} isSidebar - Flag indicating whether the app is the sidebar or a top-level page.
+ * @prop {() => any} onLogin - Callback invoked when user clicks "Login" button.
+ * @prop {() => any} onLogout - Callback invoked when user clicks "Logout" action in account menu.
+ * @prop {() => any} onSignUp - Callback invoked when user clicks "Sign up" button.
+ * @prop {MergedConfig} settings
+ * @prop {import('../services/streamer').StreamerService} streamer
  */
```
robertknight, line 32: Update the type of `streamer` and correct the optionality of various other properties. It was incorrect but not noticed because the `TopBarProps` type was not actually used.

src/sidebar/components/TopBar.js
```diff
@@ -21,19 +21,21 @@ import UserMenu from './UserMenu';
...
 
 /**
  * The toolbar which appears at the top of the sidebar providing actions
  * to switch groups, view account information, sort/filter annotations etc.
+ *
+ * @param {TopBarProps} props
```
robertknight, line 38: Fix a mistake where the `TopBarProps` type was not actually used. This revealed various mistakes in the types.

src/sidebar/components/HypothesisApp.js
```diff
@@ -173,6 +173,7 @@ function HypothesisApp({ auth, bridge, settings, session, toastMessenger }) {
       {route !== 'notebook' && (
         <TopBar
           auth={authState}
+          bridge={bridge}
```
esanzgar, line 176: ?

esanzgar (changes requested):
Robert, would you mind rebasing and resolving the conflicts? Thanks.

Regarding the bridge prop, I thought it was injected.

robertknight:
Hi @esanzgar - I have rebased this PR. The changes related to `TopBar` and its use of `bridge` were part of an earlier version of another PR that are no longer in this branch.

src/sidebar/index.js
```diff
@@ -36,7 +36,11 @@ if (process.env.NODE_ENV !== 'production') {
 // Install Preact renderer options to work around browser quirks
 rendererOptions.setupBrowserFixes();
 
-// @inject
+/**
+ * @param {import('./services/api').APIService} api
+ * @param {import('./services/streamer').StreamerService} streamer
+ * @inject
+ */
```
esanzgar, line 43: 👍🏼

Issue #3375, opened 2021-05-05
Upgrade from Python 3.6


jon-betts:
For: https://github.com/hypothesis/product-backlog/issues/1199

It looks like there actually is still some Python 3.6 usage in the client which we should get off of.

At the same time it might be a good idea to look about and see if there's any other standard stuff we should be doing in the Python `tox` realm.

jon-betts:
Done

PR #3376, opened 2021-05-05
Upgrade to Python 3.9 and add some of the newer Python magic we have

jon-betts:
For: https://github.com/hypothesis/client/issues/3375

 * This adds pinned versions of the docs dependencies, so they don't drift between local and CI
 * This is managed with `hdev requirements` so support for that has been added
 * Using the pyenv integration we have elsewhere so tox tracks the versions in the version file

## Review notes

This brings the Python portion of the client much more inline with our other projects and removes some possible cases where versions can drift.

## Test notes

 * With luck everything required here should automatically install itself upon running any `tox` command
 * If you want to try compiling the requirements you will need to install `hdev` as well - https://github.com/hypothesis/hdev
    * Then run `hdev requirements`
 * Other than that, everything should work as before:
    * `make docs`
    * `make checkdocs` - This fails with an indentation warning for me... but it also does in `master`


Makefile
```diff
@@ -35,11 +35,11 @@ lint: node_modules/.uptodate
 
 .PHONY: docs
 docs: python
-	tox -e py36-docs
+	tox -e docs
```
jon-betts, line 38: We shouldn't need to specify the version here as we now use `tox-pyenv` will now find the version specified in `.python-version` when we don't specify.

tox.ini
```diff
@@ -1,18 +1,19 @@
 [tox]
-envlist = py3-docs
+envlist = docs
 skipsdist = true
+requires =
+  tox-pip-sync
+  tox-pyenv
+  tox-run-command
```
jon-betts, line 7: In order this adds:

 * `tox-pip-sync` - Our new project to update your tox envs if the dependencies change
 * `tox-pyenv` - To make the version be determined by the `.python-version` file
 * `tox-run-command` - Required by `hdev` to work, but also quite handy as it lets you do `tox -e docs --run-command <ANYTHING YOU LIKE>`

tox.ini
```diff
@@ -1,18 +1,19 @@
...
 
 [testenv]
 skip_install = true
 deps =
-    {docs,checkdocs}: recommonmark
-    {docs,checkdocs}: sphinx-autobuild
-    {docs,checkdocs}: sphinx
-    {docs,checkdocs}: sphinx_rtd_theme
+    {docs,checkdocs}: -r requirements/docs.txt
```
jon-betts, line 12: We now pin all our Python requirements for a few reasons:

 * So you can't get weirdness where it passes in CI but not locally or vice versa
 * So dependabot can poke us

tox.ini
```diff
@@ -1,18 +1,19 @@
...
+    {env:EXTRA_DEPS:}
 passenv =
     HOME
-changedir =
-    {docs,checkdocs}: docs
```
jon-betts, line 15: Changing dirs isn't necessary, and makes `make requirements` write out local paths to the compiled files.

requirements/docs.txt
```diff
@@ -0,0 +1,78 @@
+#
```
jon-betts, line 1: The file above is manually created, but this one is created by running `hdev requirements`

LMS007 (commented):
Thanks Jon. Everything worked for me just fine

requirements/docs.txt
robertknight: It would be great if the instructions at the top of this file could be updated to reflect the current process.

robertknight (approved):
LGTM. I can reproduce the issue with `make checkdocs` but as you said, that's unrelated to the changes on this branch.

requirements/docs.txt
jon-betts: We actually can't do that unfortunately without creating problems with Dependabot. We both use `pip-compile` to create this file I think. So if we change the header Dependabot will change it back. This means a lot of PRs would have noise adding and removing a custom header.

PR #3377, opened 2021-05-06
Fix documentation generation errors

robertknight:
This PR fixes various errors when building the documentation using `make docs` and checking it using `make checkdocs`:

- The existing Markdown parsing setup triggered an "Unexpected indentation" on a code block in `adr-003.md`. I changed the markdown parsing to follow the [current instructions in the Sphinx docs](https://www.sphinx-doc.org/en/master/usage/markdown.html) and this resolved the issue.
- Add a missing `notebookAppUrl` documentation entry that was referenced from `envvars.rst`

We only have a single Markdown documentation file currently, so we could just have converted it to reStructuredText instead. However all of the frontend developers are far more familiar with Markdown, so I think it makes sense to keep the option to write future documentation in this format.

I briefly looked into getting `make checkdocs` running as part of CI, but that's a little more involved as we need to get Python installed and set up correctly, so I'll leave that task for a separate PR.

**Testing:**

- Run `make checkdocs` locally. It should complete without errors.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,8 +15,6 @@
 import os
 # import sys
 
-from recommonmark.parser import CommonMarkParser
-
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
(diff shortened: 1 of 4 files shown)


jon-betts (approved):
Not particularly familiar with sphinx docs, but this absolutely seems to build fine and emits no warnings, so I think this looks good!

PR #3378, opened 2021-05-06
Replace `make clean` with `hdev clean`

jon-betts:
For: https://github.com/hypothesis/hdev/issues/35

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,6 @@ help:
 	@echo "make sure              Make sure that the formatter, linter, tests, etc all pass"
 	@echo "make docs              Build docs website and serve it locally"
 	@echo "make checkdocs         Crash if building the docs website fails"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[tool.hdev.clean]
+# File and directory patterns relative to the root to remove.
+# These support shell expansions and patterns
+dirs=["build"]
```
jon-betts, line 4: Unfortunately this wasn't included in the default set... maybe it should be?

jon-betts: Done in https://github.com/hypothesis/hdev/pull/36

seanh (approved):

robertknight:
This unfortunately broke the deployment process: https://jenkins.hypothes.is/job/client/job/master/2891/console. I'm going to revert this to unblock deploys and then we can revisit this later.

PR #3381, opened 2021-05-07
Remove ts-expect-error and improve types

esanzgar:
I removed a `@ts-expect-error` and tight some types.

I removed the returned `provider` property from the
`authStateFromProfile` function as it was not used.


src/sidebar/components/UserMenu.js
```diff
@@ -16,16 +16,18 @@ import MenuSection from './MenuSection';
...
 
 /**
- * @typedef AuthState
- * @prop {'logged-in'|'logged-out'|'unknown'} status
+ * @typedef AuthStateLoggedIn
+ * @prop {'logged-in'} status
  * @prop {string} displayName
  * @prop {string} userid
  * @prop {string} username
+ * @prop {string} provider
```
robertknight, line 24: From what I could tell using grep and TS's "Find References" tool, nothing reads the provider field. Therefore we could just remove it from the type entirely.

src/sidebar/components/HypothesisApp.js
```diff
@@ -162,6 +165,8 @@ function HypothesisApp({ auth, bridge, settings, session, toastMessenger }) {
     session.logout();
   };
 
+  // eslint-disable-next-line no-unused-vars
+  const { status, ...authHelpPanel } = authState;
```
robertknight, line 169: It wasn't clear to me at first why this was done. I see that if trying to pass `authState` directly to `HelpPanel` you get an error, even though the types are compatible, due to the way TS's compatibility heuristics work. After a little more experimentation I found this also works, and I think is more obvious:

```diff
-        <HelpPanel auth={authHelpPanel} />
+        <HelpPanel auth={authState.status === 'logged-in' ? authState : {}} />
```

robertknight (changes requested):
I think in future we could do away with the `authState` field in `HypothesisApp` entirely and instead compute the necessary information in individual components. I'm OK with this as an intermediate step though. I did have a couple of suggestions to simplify it though.

src/sidebar/components/HypothesisApp.js
esanzgar: 👍🏼

src/sidebar/components/UserMenu.js
esanzgar: I thought about removing this unused property, but it leads to changing some tests. I will do that.

esanzgar:
I agree this is an intermediate step and simplifying the state will be better.

I incorporated your suggestion and removed the unused of `provider` property.

src/sidebar/components/HypothesisApp.js
```diff
@@ -24,16 +24,18 @@ import TopBar from './TopBar';
...
 /**
  * Return the user's authentication status from their profile.
  *
  * @param {Profile} profile - The profile object from the API.
+ * @return {AuthState}
  */
 function authStateFromProfile(profile) {
   const parsed = parseAccountID(profile.userid);
-  if (parsed) {
+  if (parsed && profile.userid) {
```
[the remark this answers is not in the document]
esanzgar: You are correct.

I could have cast it the `profile.userid` below, but I have preferred to do the actual check. As you sed if `parsed` is non-null implies `parsed.userid` is non-null.

PR #3382, opened 2021-05-07
Add fallback for `PDFViewerApplication.appConfig` for older PDF.js releases

robertknight:
Specify the type of `window.PDFViewerApplication` in src/types/pdfjs.js and add various missing properties required by `PDFIntegration`. This turned up some missing checks for potentially null/undefined properties, one of which caused https://github.com/hypothesis/product-backlog/issues/1205. These are fixed in this PR as well.

 - Add a fallback for older versions of PDF.js where `PDFViewerApplication`
   did not have an `appConfig` property [1]
 - Add a missing check for `getPageView` returning `null`

Fixes https://github.com/hypothesis/product-backlog/issues/1205

[1] The property was introduced in https://github.com/mozilla/pdf.js/commit/b6c74f20561b04573cd00dd924484be7f80a09ce

----

**Testing:**

1. Check out https://github.com/hypothesis/pdf.js-hypothes.is and then check out commit 3d8c5eebc922804f6ce83489a3898348a51f1112, which is the last commit that used an old version of PDF.js
2. Change the URL pointing to `hypothes.is/embed.js` at the bottom of `viewer/web/viewer.html` in the pdf.js-hypothes.is repo to `localhost:5000/embed.js` so that it loads the local client
3. Run `python -m http.server 6006` and browse to http://localhost:6006. You should see PDF.js + Hypothesis. The browser console should indicate that PDF.js is v1.1.114. There will be an error in the console about `this.pdfContainer` in the `fitSideBySide` method
4. Check out this branch in the client repo and reload http://localhost:6006. You should see that the error has disappeared and side-by-side mode should be working. When the sidebar is closed and you click on a highlight or create a new annotation, the sidebar should open.

### DIFF
--- a/src/annotator/integrations/pdf.js
+++ b/src/annotator/integrations/pdf.js
@@ -23,6 +23,7 @@ import { PDFMetadata } from './pdf-metadata';
  * @typedef {import('../../types/annotator').Integration} Integration
  * @typedef {import('../../types/annotator').SidebarLayout} SidebarLayout
  * @typedef {import('../../types/api').Selector} Selector
+ * @typedef {import('../../types/pdfjs').PDFViewerApplication} PDFViewerApplication
  */
 
 // The viewport and controls for PDF.js start breaking down below about 670px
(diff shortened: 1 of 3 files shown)


src/annotator/integrations/pdf.js
```diff
@@ -61,14 +62,18 @@ export class PDFIntegration {
...
+    // Assume this class is only used if we're in the PDF.js viewer.
+    const pdfViewerApp = /** @type {PDFViewerApplication} */ (window_.PDFViewerApplication);
+
+    this.pdfViewer = pdfViewerApp.pdfViewer;
     this.pdfViewer.viewer.classList.add('has-transparent-text-layer');
 
     // Get the element that contains all of the PDF.js UI. This is typically
     // `document.body`.
-    this.pdfContainer = window_.PDFViewerApplication?.appConfig?.appContainer;
+    this.pdfContainer = pdfViewerApp.appConfig?.appContainer ?? document.body;
```
robertknight, line 74: The null-ness check for `window_.PDFViewerApplication` here was redundant. The `document.body` fallback is added for PDF.js versions before v1.5.188.

robertknight:
Hi @LMS007 - I added some testing steps to the PR that explains how to test using an old version of https://github.com/hypothesis/pdf.js-hypothes.is.

LMS007:
Confirmed the fix works in both the older version of PDF js and the current.

Screen Shot 2021-05-13 at 10 01 23 AMCon

Screen Shot 2021-05-13 at 9 59 20 AM

LMS007 (approved):

PR #3405, opened 2021-05-11
Refactor streamer service

robertknight:
This PR is a follow up to https://github.com/hypothesis/client/pull/3368 which converts the `StreamerService` class to a more conventional class, by which I mean using instance properties for fields and methods rather than closures assigned in the constructor. Some unused public methods and fields have also been removed. There are no functional changes.

Changes in this PR:

- Internal state of the `StreamerService` has been moved from closed-over variables in the constructor to instance properties
- Methods have been converted from closures in the constructor to regular instance methods
- The internal `_connect` helper has been renamed to `_reconnect` to differentiate it from the public `connect` method
- The public `reconnect` method and its tests have been removed as it was unused
- The public `configMessages` field has been made private
- Some of the tests have been modernized slightly (functions => arrow functions) but are otherwise unchanged.

This is the final PR needed to close out https://github.com/hypothesis/client/issues/3298.

### DIFF
--- a/src/sidebar/services/streamer.js
+++ b/src/sidebar/services/streamer.js
@@ -80,31 +80,6 @@ export class StreamerService {
     this._store.clearPendingUpdates();
   }
 
-  _handleAnnotationNotification(message) {
-    const action = message.options.action;
-    const annotations = message.payload;
-
-    switch (action) {
-      case 'create':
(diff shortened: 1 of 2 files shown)


esanzgar (commented):
👍🏼

PR #3406, opened 2021-05-12
Update export style and documentation for sidebar store

robertknight:
_nb. This PR contains some non-urgent refactoring/documentation updates. Please review at your leisure._

Convert the `store/index.js` and `store/create-store.js` modules to use
named rather than default exports per our current conventions, and
revise the documentation.

The goal of the revised documentation is to more clearly describe what the
store is and how it is used within the sidebar/notebook app, assuming
that the reader is likely have at least some familiarity with Redux (or
can read the "Introduction" section of the linked website if not). In
particular I have tried to convey:

 - The separation between the "base" `createStore` function which is not
   application-specific and the `createSidebarStore` function and the
   modules it uses which are
 - How the store in the sidebar app differs from a standard/base Redux store
 - Best practices around using the store from other parts of the app
   (use `useStoreProxy` in UI components, use selector and action
   methods rather than `getState` and `dispatch`)

### DIFF
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -131,7 +131,7 @@ import { ThreadsService } from './services/threads';
 import { ToastMessengerService } from './services/toast-messenger';
 
 // Redux store.
-import store from './store';
+import { createSidebarStore } from './store';
 
 // Utilities.
 import { Injector } from '../shared/injector';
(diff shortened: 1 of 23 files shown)


esanzgar (approved):
I appreciated the documentation, thanks.

src/sidebar/index.js
```diff
@@ -131,7 +131,7 @@ import { ThreadsService } from './services/threads';
 import { ToastMessengerService } from './services/toast-messenger';
 
 // Redux store.
-import store from './store';
+import { createSidebarStore } from './store';
```
esanzgar, line 134: This is also used by the `Notebook`, right?

robertknight: Yes. Currently the sidebar and the notebook share the same store. In future we might change this so that the sidebar and notebook app uses stores with different sets of modules.

PR #3408, opened 2021-05-13
Refactor annotator/config/ (1/N)

LMS007:
Refactor `annotator/config/index.js` to expose a slightly different method for getting the config called `getConfig`. This now takes a app name as a string which may return a subset of config values appropriate for a given application such as the "sidebar" or "notebook".

Previously, part of this change was in this longer PR: https://github.com/hypothesis/client/pull/3291
The plan is still to move forward with a similar (if not same approach) but broken up in to smaller steps.

Partially fixes https://github.com/hypothesis/client/issues/3236

The next steps at a high level are

- [ ] Audit the existing keys to see if which keys may be able to be omitted from the app contexts (notebook, sidebar, guest)
- [x] Add a config def map that has a valueFn and gets the setting from setting.js [3411](https://github.com/hypothesis/client/pull/3411)
- [x] Move the browser extension logic out of settings and add flags into the config defs as needed [3421](https://github.com/hypothesis/client/pull/3421)
- [x] Consolidate the default value(s) and move them into the config defs. Currently some undefined configs become null and its not very explicit. (https://github.com/hypothesis/client/pull/3421)
- [x] Move the existing coerce methods or logic out into a coerce fn's in the config defs. Some of that logic is buried in settings [3421](https://github.com/hypothesis/client/pull/3421)
- [ ] Audit the settings.js file and see if any existing gettings need custom coerce methods
- [ ] Add validation methods as needed to pass/fail incoming config values. Note this is different than "coerce". e.g. "true" may be `true`, but "foo" would be an error.

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -61,7 +61,14 @@ function configurationKeys(appContext) {
       return contexts.notebook;
     case 'all':
       // Complete list of configuration keys used for testing.
-      return [...contexts.annotator, ...contexts.sidebar, ...contexts.notebook];
+      return [
+        ...contexts.annotator,
+        ...contexts.sidebar,
+        ...contexts.notebook,
+        // "experimental" currently has no uses in any app contexts, but is included
+        // for test coverage
+        'experimental',
+      ];
     default:
       throw new Error(`Invalid application context used: "${appContext}"`);
   }
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -2,17 +2,74 @@ import settingsFrom from './settings';
 import { toBoolean } from '../../shared/type-coercions';
 
 /**
+ * @typedef {'sidebar'|'notebook'|'annotator'|'all'} AppContext
+ */
+
+/**
+ * List of allowed configuration keys per application context. Keys omitted
+ * in a given context will be removed from the relative configs when calling
+ * getConfig.
+ *
+ * @param {AppContext} [appContext] - The name of the app.
+ */
+function configurationKeys(appContext) {
+  const contexts = {
+    annotator: ['clientUrl', 'showHighlights', 'subFrameIdentifier'],
+    sidebar: [
+      'appType',
+      'annotations',
+      'branding',
+      'enableExperimentalNewNoteButton',
+      'externalContainerSelector',
+      'focus',
+      'group',
+      'onLayoutChange',
+      'openSidebar',
+      'query',
+      'requestConfigFromFrame',
+      'services',
+      'showHighlights',
+      'sidebarAppUrl',
+      'theme',
+      'usernameUrl',
+    ],
+    notebook: [
+      'branding',
+      'group',
+      'notebookAppUrl',
+      'requestConfigFromFrame',
+      'services',
+      'theme',
+      'usernameUrl',
+    ],
+  };
+
+  switch (appContext) {
+    case 'annotator':
+      return contexts.annotator;
+    case 'sidebar':
+      return contexts.sidebar;
+    case 'notebook':
+      return contexts.notebook;
+    case 'all':
+      // Complete list of configuration keys used for testing.
+      return [...contexts.annotator, ...contexts.sidebar, ...contexts.notebook];
+    default:
+      throw new Error(`Invalid application context used: "${appContext}"`);
+  }
+}
+
+/**
  * Reads the Hypothesis configuration from the environment.
  *
+ * @param {string[]} settingsKeys - List of settings that should be returned.
  * @param {Window} window_ - The Window object to read config from.
  */
-export default function configFrom(window_) {
+function configFrom(settingsKeys, window_) {
   const settings = settingsFrom(window_);
-  return {
+  const allConfigSettings = {
     annotations: settings.annotations,
-    // URL where client assets are served from. Used when injecting the client
-    // into child iframes.
-    assetRoot: settings.hostPageSetting('assetRoot', {
+    appType: settings.hostPageSetting('appType', {
       allowInBrowserExt: true,
     }),
     branding: settings.hostPageSetting('branding'),
@@ -50,4 +107,24 @@ export default function configFrom(window_) {
       'externalContainerSelector'
     ),
   };
+
+  // Only return what we asked for
+  const resultConfig = {};
+  settingsKeys.forEach(key => {
+    resultConfig[key] = allConfigSettings[key];
+  });
+  return resultConfig;
+}
+
+/**
+ * Return the configuration for a given application context.
+ *
+ * @param {AppContext} [appContext] - The name of the app.
+ */
+export function getConfig(appContext = 'annotator', window_ = window) {
+  // Filter the config based on the application context as some config values
+  // may be inappropriate or erroneous for some applications.
+  const filteredKeys = configurationKeys(appContext);
+  const config = configFrom(filteredKeys, window_);
+  return config;
 }
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -3,6 +3,14 @@ import { toBoolean } from '../../shared/type-coercions';
 
 /**
  * @typedef {'sidebar'|'notebook'|'annotator'|'all'} AppContext
+ *
+ * @typedef ConfigDefinition
+ * @prop {() => any} valueFn -
+ *   Method to retrieve the value from the incoming source
+ */
+
+/**
+ * @typedef {Record<string, ConfigDefinition>} ConfigDefinitionMap
  */
 
 /**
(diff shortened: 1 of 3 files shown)


src/annotator/config/index.js
```diff
@@ -1,12 +1,57 @@
 import settingsFrom from './settings';
 import { toBoolean } from '../../shared/type-coercions';
 
+/**
+ * Total list of configuration keys.
```
robertknight, line 5:
```suggestion
 * Complete list of configuration keys.
```

"Total" to me implies a sum, whereas here you mean the "complete set" or "full set".

src/annotator/config/index.js
```diff
@@ -1,12 +1,57 @@
...
+ * in a given context will be removed from the relative configs when calling
+ * getConfig.
+ */
+const configurationContexts = {
+  // For testing
+  all: [...configurationKeys],
+
+  annotator: [...configurationKeys],
+
+  sidebar: [...configurationKeys],
```
robertknight, line 41: Is there a need to clone the list? Since we're not mutating it you could just do `sidebar: configurationKeys`.

src/annotator/config/index.js
```diff
@@ -51,3 +96,23 @@ export default function configFrom(window_) {
...
+/**
+ * Return the configuration for a given application context.
+ *
+ * @param {'sidebar'|'notebook'|'annotator'|'all'} [appContext] - The name of the app.
+ */
+export default function getConfig(appContext = 'annotator', window_ = window) {
+  // Filter the config based on the application context as some config values
+  // may be inappropriate or erroneous for some applications.
+  let partialConfig = {};
+  const fullConfig = configFrom(window_);
```
robertknight, line 109: `configFrom` presumably ends up doing some redundant work fetching values for keys which are never used. Perhaps it could accept a whitelist of keys as an argument?

src/annotator/config/index.js
```diff
@@ -51,3 +96,23 @@ export default function configFrom(window_) {
...
+
+/**
+ * Return the configuration for a given application context.
+ *
+ * @param {'sidebar'|'notebook'|'annotator'|'all'} [appContext] - The name of the app.
+ */
+export default function getConfig(appContext = 'annotator', window_ = window) {
+  // Filter the config based on the application context as some config values
+  // may be inappropriate or erroneous for some applications.
+  let partialConfig = {};
```
robertknight, line 108: `partialConfig` and `filteredKeys` are never re-assigned (`partialConfig` is modified, but it still refers to the same object), so both of these bindings can use `const` rather than `let`.

src/annotator/config/index.js
```diff
@@ -51,3 +96,23 @@ export default function configFrom(window_) {
     ),
   };
 }
+
+/**
+ * Return the configuration for a given application context.
+ *
+ * @param {'sidebar'|'notebook'|'annotator'|'all'} [appContext] - The name of the app.
+ */
+export default function getConfig(appContext = 'annotator', window_ = window) {
```
robertknight, line 105: Can you use a named rather than default export here per our recently adopted conventions.

src/annotator/config/index.js
```diff
@@ -1,12 +1,57 @@
...
+/**
+ * List of allowed configuration keys per application context. Keys omitted
+ * in a given context will be removed from the relative configs when calling
+ * getConfig.
+ */
+const configurationContexts = {
+  // For testing
+  all: [...configurationKeys],
+
+  annotator: [...configurationKeys],
```
robertknight, line 39: It might not be that difficult to list just the relevant keys for each context here rather than using the complete list for all contexts and then just removing one of the items in the `notebook` context.

src/annotator/index.js
```diff
@@ -33,34 +33,29 @@ const appLinkEl = /** @type {Element} */ (
   )
 );
 
-const config = configFrom(window);
+const annotatorConfig = getConfig('annotator');
```
robertknight, line 36: It looks like the `annotatorConfig` variable could be pushed down into `init`.

src/annotator/index.js
```diff
@@ -33,34 +33,29 @@ const appLinkEl = /** @type {Element} */ (
...
 
   // Load the PDF anchoring/metadata integration.
-  config.documentType = isPDF ? 'pdf' : 'html';
+  annotatorConfig.documentType = isPDF ? 'pdf' : 'html';
 
   const eventBus = new EventBus();
-  const guest = new Guest(document.body, eventBus, config);
-  const sidebar = !config.subFrameIdentifier
-    ? new Sidebar(document.body, eventBus, guest, config)
+  const guest = new Guest(document.body, eventBus, annotatorConfig);
```
robertknight, line 52: There is an existing complication here that the `Guest` passes its `config` argument through to the `CrossFrame` constructor which in turn passes it through to the `injectHypothesis` utility that injects configuration into child frames. I think it should be OK to pass just the guest-related configuration to the `Guest` constructor because in child frames only the `Guest` part of the annotator is instantiated and not the `Sidebar` or `Notebook` parts.

robertknight (changes requested):
The overall structure looks good. I left a few change requests.

src/annotator/config/index.js
LMS007: Thats the long term plan, but if it's okay with you, id like to land this unchanged from how it works now. I still want to   fix the things outlined in the ticket and then circle back to this and come up with an audited list of config settings that we agree the annotator needs.  This is also a risker change than just the structure of how the config gets created.

src/annotator/index.js
LMS007: Understood and I think we need to strongly consider that for the audit which is why I want to continue to pass the total config to the guess for now even though it's called "annotatorConfig"

src/annotator/config/index.js
LMS007: Sure, but in the next PR or two, I will propose this function goes away. So I hope not to get too attached to details with configFrom(). Part of what has been hard was landing intermittent changes that make sense.

robertknight (commented):
Hi Kyle,

I've had an initial look over this. The two main items of feedback I have are:

- The current `getConfig` API leads to the same configuration being reparsed three times during annotator initialization. This is not super expensive in the grand scheme of things but it would be nicer if this didn't happen.
- We currently return (almost) every key for every context. I think one of the most valuable things we could do as part of this work is figure out what keys are actually needed in the three different contexts. Later we could also look at organizing them into groups so it is easier to see all the settings that pertain to some aspect of the client (eg. keys that affect what is initially focused/selected in the sidebar, or keys that affect appearance or UI behavior). Right now there is a big list of keys without an obvious grouping.

src/annotator/index.js
```diff
@@ -33,34 +33,28 @@ const appLinkEl = /** @type {Element} */ (
   )
 );
 
-const config = configFrom(window);
-
 function init() {
+  const annotatorConfig = getConfig('annotator');
```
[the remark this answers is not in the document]
LMS007: The original solution I proposed was a class instance, but  I got feedback that steered this in another direction
The major push back was making the API seem like the config would change the lifetime of the instance and perhaps not be deterministic.  More discussion here.

https://github.com/hypothesis/client/pull/3232#pullrequestreview-626217791

src/annotator/config/index.js
```diff
@@ -1,14 +1,60 @@
...
+  'onLayoutChange',
+  'openSidebar',
+  'query',
+  'requestConfigFromFrame',
+  'services',
+  'showHighlights',
+  'notebookAppUrl',
+  'sidebarAppUrl',
+  'subFrameIdentifier',
+  'externalContainerSelector',
```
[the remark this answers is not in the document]
LMS007: I think I may do a crude sort first, I don't have a good enough sense to group this a more sensible way.

src/annotator/config/index.js
```diff
@@ -1,14 +1,60 @@
...
+
+  annotator: configurationKeys,
+
+  sidebar: configurationKeys,
+
+  // Notebook must omit the `annotations` setting
+  notebook: configurationKeys.filter(
+    configProperty => configProperty !== 'annotations'
+  ),
+};
```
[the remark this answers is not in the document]
LMS007: I did a preliminary pass at this a while back, I can do a more in depth audit to see which keys are actually not even used in each context and we can perhaps move forward with that. I added a todo in the PR description for that

src/annotator/config/index.js
```diff
@@ -1,14 +1,84 @@
...
+        'group',
+        'notebookAppUrl',
+        'requestConfigFromFrame',
+        'services',
+        'showHighlights',
+        'theme',
+        'usernameUrl',
+      ];
+    case 'all':
+    default:
```
LMS007, line 46: Maybe add 'boot' eventually too, but that would likely mean we move these files outside of /annotator

src/annotator/config/index.js
```diff
@@ -1,14 +1,84 @@
...
+        'services',
+        'showHighlights',
+        'sidebarAppUrl',
+        'theme',
+        'usernameUrl',
+      ];
+    case 'notebook':
+      return [
+        'branding',
+        'enableExperimentalNewNoteButton',
```
robertknight, line 35: The `enableExperimentalNewNoteButton` setting only makes sense in the sidebar because it is used in the "Page Notes" tab to create new notes. We don't currently support creating new annotations in the Notebook.

src/annotator/config/index.js
```diff
@@ -1,14 +1,84 @@
...
+        'focus',
```
robertknight, line 36: I'm not sure if we want to support the `focus` setting in the notebook yet.

src/annotator/config/index.js
```diff
@@ -1,14 +1,84 @@
...
+        'group',
+        'notebookAppUrl',
+        'requestConfigFromFrame',
+        'services',
+        'showHighlights',
```
robertknight, line 41: `showHighlights` isn't needed for the notebook.

src/annotator/index.js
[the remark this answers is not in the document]
robertknight: I think Lyza's feedback in https://github.com/hypothesis/client/pull/3232#pullrequestreview-626217791 was quite reasonable. However I don't interpret it as a hard request not to use a class but instead a request to 1) justify its value and 2) clarify when parsing occurs. I think I would respond to point (2) by just adjusting the API design/docs/implementation to make it clear when parsing happens and regarding (1) I think it's clear from this PR that it results in doing parsing / validation once rather than thrice.

I think making any further changes in response to this comment can be the subject of a later PR though.

```js
class Config {
  constructor(allSettings) { ... }
  
  configFor(context) {
    // Return subset of `allSettings` for current context
  }
}

// `parseConfig` name makes it clear that configuration extraction/parsing happens
// when this function is called.
export function parseConfig(window_) {
  const allSettings = readAllTheSettings();
  return new Config(allSettings);
}
```

robertknight (changes requested):
Hi Kyle,

I did some testing of this in various scenarios and it looks like the configuration is being passed correctly except for a small number of keys that I called out in comments. I requested a couple of other minor changes. Once those are addressed I suggest we land this without any major API changes. We can revisit changing the API to avoid re-parsing multiple times in a follow-up.

src/annotator/config/index.js
```diff
@@ -1,14 +1,84 @@
...
+      return [
+        'annotations',
+        'branding',
+        'enableExperimentalNewNoteButton',
+        'externalContainerSelector',
+        'focus',
+        'group',
+        'onLayoutChange',
+        'openSidebar',
+        'requestConfigFromFrame',
```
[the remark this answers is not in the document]
LMS007: bah! must have got lost in the audit work here, good catch.

src/annotator/config/index.js
```diff
@@ -1,14 +1,84 @@
...
+ *
+ * @param {'sidebar'|'notebook'|'annotator'|'all'} [appContext] - The name of the app.
+ */
+function configurationContexts(appContext) {
+  switch (appContext) {
+    case 'annotator':
+      return ['clientUrl', 'showHighlights', 'subFrameIdentifier'];
+    case 'sidebar':
+      return [
+        'annotations',
```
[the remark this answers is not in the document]
LMS007: yep, looks like something the sidebar uses so let's add it to that context

src/annotator/config/index.js
```diff
@@ -1,14 +1,84 @@
...
+        'openSidebar',
+        'query',
+        'requestConfigFromFrame',
+        'services',
+        'showHighlights',
+        'sidebarAppUrl',
+        'subFrameIdentifier',
+        'theme',
+        'usernameUrl',
+      ];
```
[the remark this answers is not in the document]
LMS007: Sounds good. I did this a bit different since I tried to keep it all in one function. I also don't see the need for Set() since the overlapping keys will cancel out, but please let me know if i missed something obvious or you intended for Set to improve readability.

LMS007:
@robertknight I noticed that `assetRoot` is not used in any of the contexts. Its use in boot only at near as I can tell. I created a ticket to potentially make a boot context https://github.com/hypothesis/client/issues/3474

But since its not used in sidebar or annotator or notebook, and all is now a combination of those 3, assetRoot is now omitted entirely.  I left it in configFrom() for now because it's likely we will re-used this module for the boot context (per ticket 3474). Or, we could pull it completely out for now?

src/annotator/config/index.js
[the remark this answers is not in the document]
robertknight: Hi Kyle - I used a `Set` to ensure that the keys were unique and avoid potential issues with upstream code that didn't handle duplicates properly. If it isn't a problem to have duplicates with the current upstream code then you can get away without it, but you probably want to note this assumption explicitly.

robertknight:
> @robertknight I noticed that assetRoot is not used in any of the contexts. Its use in boot only at near as I can tell. I created a ticket to potentially make a boot context #3474

That's correct. `assetRoot` is only used by the boot script. Omitting it from this code is fine with me.

> I left it in configFrom() for now because it's likely we will re-used this module for the boot context (per ticket 3474)

I'm not certain we'll want to do that. The boot script has some additional constraints which become harder to enforce if the amount of code in it grows:

- It targets older browsers than the rest of the Hypothesis client code, so that it can at least show some kind of notice if you try to use the client in the browser that is too old. This notice is currently just a console warning, but I had planned that we might add a crude UI in future.
- It is designed to be a very small script that just bootstraps the execution of the rest of the code

src/annotator/config/index.js
```diff
@@ -1,14 +1,69 @@
...
+  switch (appContext) {
+    case 'annotator':
+      return contexts.annotator;
+    case 'sidebar':
+      return contexts.sidebar;
+    case 'notebook':
+      return contexts.notebook;
+    case 'all':
+      // Complete list of configuration keys used for testing.
+      return [...contexts.annotator, ...contexts.sidebar, ...contexts.notebook];
```
robertknight, line 52: 👍 - I prefer this to my suggestion.

src/annotator/config/index.js
```diff
@@ -50,4 +105,24 @@ export default function configFrom(window_) {
...
+  settingsKeys.forEach(key => {
+    resultConfig[key] = allConfigSettings[key];
+  });
+  return resultConfig;
+}
+
+/**
+ * Return the configuration for a given application context.
+ *
+ * @param {'sidebar'|'notebook'|'annotator'|'all'} [appContext] - The name of the app.
```
robertknight, line 120: I suggest defining an alias for this union with typedef at the top of the file rather than duplicating it.

src/annotator/config/test/index-test.js
```diff
@@ -95,28 +95,78 @@ describe('annotator.config.index', function () {
...
-        return settings[settingName];
-      };
-
-      const settingValue = configFrom('WINDOW')[settingName];
-
-      assert.equal(settingValue, settings[settingName]);
+  ['branding', 'openSidebar', 'requestConfigFromFrame', 'services'].forEach(
+    function (settingName) {
+      it(
+        'returns the ' + settingName + ' value from the host page',
```
robertknight, line 101: This could could be modernized a little here to use string interpolation and arrow functions.

robertknight (approved):
The proposed changes look good to me. When this lands on QA, please make sure to check that everything still works as expected in the Via and LMS contexts, as those are the most important "customers" that use custom settings.

LMS007:
> * very small script that just bootstraps the execution of the rest of the code

okay, then maybe we don't proceed with a 'boot' context for this config module in the future and boot config will just remain a light-weight ad-hoc implementation

LMS007:
@robertknight , i have make some changes here to add `documentType` and `appType` to allConfigSettings. See slack discussion https://hypothes-is.slack.com/archives/C1M8NH76X/p1622739037190400

I think this is the right thing to do, but we can discuss further.  I have made the other suggestions changed you requests.

robertknight:
> @robertknight , i have make some changes here to add documentType and appType to allConfigSettings. See slack discussion https://hypothes-is.slack.com/archives/C1M8NH76X/p1622739037190400

Hi @LMS007 - Adding `appType` is correct. `documentType` is a purely internal property though which is not currently intended to be configurable by the user. Logically the `config` argument to `Guest` is a combination of the user-specified configuration properties and the `documentType` option.

PR #3411, opened 2021-05-14
Refactor annotator/config/ (2/N)

LMS007:
depends on https://github.com/hypothesis/client/pull/3408

-------

Change configFrom() to configDefinitions()

The main difference is that configFrom returned config key:value pairs, but now configDefinitions returns key:object pairs. Where the object may hold additional flags and methods for each specific config value. Currently the only method included is valueFn() which gets the value from its source. Others will follow.

-------

The end goal here is to have the definitions look more or less like this

```...
    openSidebar: {
      allowInBrowserExt: true,
      defaultValue: false,
      coerce: toBoolean,
      valueFn: settings.hostPageSetting,
    },
    ...

```

Where getConfig does some of the logic to set a default, coerce, restrict based on allowInBrowserExt, and define a value getter as valueFn. Much of this logic is currently inside settings and that will be pulled out so settings can be focused on just retrieving the value.

_This PR has no external changes_

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -61,7 +61,14 @@ function configurationKeys(appContext) {
       return contexts.notebook;
     case 'all':
       // Complete list of configuration keys used for testing.
-      return [...contexts.annotator, ...contexts.sidebar, ...contexts.notebook];
+      return [
+        ...contexts.annotator,
+        ...contexts.sidebar,
+        ...contexts.notebook,
+        // "experimental" currently has no uses in any app contexts, but is included
+        // for test coverage
+        'experimental',
+      ];
     default:
       throw new Error(`Invalid application context used: "${appContext}"`);
   }
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -3,6 +3,14 @@ import { toBoolean } from '../../shared/type-coercions';
 
 /**
  * @typedef {'sidebar'|'notebook'|'annotator'|'all'} AppContext
+ *
+ * @typedef ConfigDefinition
+ * @prop {() => any} valueFn -
+ *   Method to retrieve the value from the incoming source
+ */
+
+/**
+ * @typedef {Record<string, ConfigDefinition>} ConfigDefinitionMap
  */
 
 /**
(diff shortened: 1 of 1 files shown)


src/annotator/config/index.js
```diff
@@ -111,11 +151,19 @@ function configFrom(settingsKeys, window_) {
  * @param {'sidebar'|'notebook'|'annotator'|'all'} [appContext] - The name of the app.
  */
 export function getConfig(appContext = 'annotator', window_ = window) {
+  const settings = settingsFrom(window_);
```
LMS007, line 154: I left settings here instead of putting it inside of configDefinitions because later on we need to see if  `settings.isBrowserExtension` is true as that logic will move out of settings into this method. However, I may revisit that and see if there is a more straightforward way to get that without using settings.js file.  If that's the case, this var would be better place inside of configDefinitions() rather than being passed in.

src/annotator/config/index.js
```diff
@@ -56,62 +66,98 @@ function configurationKeys(appContext) {
...
+          allowInBrowserExt: true,
+        }),
+    },
     // URL where client assets are served from. Used when injecting the client
     // into child iframes.
-    assetRoot: settings.hostPageSetting('assetRoot', {
-      allowInBrowserExt: true,
-    }),
-    branding: settings.hostPageSetting('branding'),
+    assetRoot: {
```
LMS007, line 85: This is untested because `assetRoot` does not belong to any app context because its part of the 'boot' context which is still TBD. Issue filed here. I could use some suggestions about what to do now.

https://github.com/hypothesis/client/issues/3474

src/annotator/config/index.js
```diff
@@ -56,62 +66,98 @@ function configurationKeys(appContext) {
...
+    clientUrl: {
+      valueFn: () => settings.clientUrl,
+    },
+    enableExperimentalNewNoteButton: {
+      valueFn: () =>
+        settings.hostPageSetting('enableExperimentalNewNoteButton'),
+    },
+    experimental: {
+      valueFn: () =>
+        settings.hostPageSetting('experimental', {
```
LMS007, line 105: Again, this has no app context and is not used anywhere in our client as near as I can tell. Should we add this to the sidebar context?

LMS007:
Rob, I rebased this, but the patch is not 100% anymore, I had some questions about that.

Also, do we want to move forward with these changes and 3/N before refactoring the config to call into settings only once vs each time a different context is called? Or should we try to sort that out now? Maybe take a look at 3/N PR to get a better sense of where I'm going with this work.

src/annotator/config/index.js
robertknight: `assetRoot` is only used by the boot script, so you can ignore it here.

src/annotator/config/test/index-test.js
```diff
@@ -51,7 +51,10 @@ describe('annotator/config/index', function () {
     });
   });
 
-  ['assetRoot', 'subFrameIdentifier'].forEach(function (settingName) {
+  ['appType', 'subFrameIdentifier'].forEach(function (settingName) {
+    // Note, 'assetRoot' should be tested here, but is currently not used in any context
+    // so it needs to be omitted for now.
+    // See https://github.com/hypothesis/client/issues/3474
```
robertknight, line 57: To reiterate, `assetRoot` doesn't need to be handled in this context.

LMS007: Roger, thanks.

PR #3423, opened 2021-05-18
Display the annotation header on the top-level annotation

esanzgar:
This PR ensures that the annotation header is always shown, even if the
top-level annotation is hidden by a filter. The annotation header
contains the document target. This information is specially important in
the context of the notebook, where annotations from different documents
are displayed together.

I created a wrapper around the `AnnotationHeader` component. I couldn't use it directly because I needed a couple of methods from the store.

On the notebook context:

Before  | After
------- | -------
image | image

On the sidebar context:

Before  | After
------- | -------
image | image

Closes # 3250

### DIFF
--- a/src/sidebar/components/Thread.js
+++ b/src/sidebar/components/Thread.js
@@ -7,6 +7,7 @@ import { withServices } from '../service-context';
 import { countHidden, countVisible } from '../helpers/thread';
 
 import Annotation from './Annotation';
+import AnnotationHeader from './Annotation/AnnotationHeader';
 import ModerationBanner from './ModerationBanner';
 
 /** @typedef {import('../helpers/build-thread').Thread} Thread */
@@ -108,11 +109,22 @@ function Thread({ thread, threadsService }) {
         {annotationContent}
 
         {showHiddenToggle && (
-          <div className="Thread__hidden-toggle-button-container">
-            <LabeledButton onClick={() => threadsService.forceVisible(thread)}>
-              Show {countHidden(thread)} more in conversation
-            </LabeledButton>
-          </div>
+          <>
+            {!thread.parent && (
+              <ThreadHeader
+                annotation={thread.annotation}
+                replyCount={thread.replyCount}
+                threadIsCollapsed={thread.collapsed}
+              />
+            )}
+            <div className="Thread__hidden-toggle-button-container">
+              <LabeledButton
+                onClick={() => threadsService.forceVisible(thread)}
+              >
+                Show {countHidden(thread)} more in conversation
+              </LabeledButton>
+            </div>
+          </>
         )}
 
         {showChildren && (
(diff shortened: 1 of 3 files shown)


src/sidebar/components/Thread.js
```diff
@@ -129,4 +141,31 @@ function Thread({ thread, threadsService }) {
...
 }
 
+/**
+ * Wrapper around AnnotationHeader
+ * @param {object} props
+ *   @param {Thread['annotation']} props.annotation
+ *   @param {number} props.replyCount
+ *   @param {boolean} props.threadIsCollapsed
+ */
+function AnnotationHeaderWrapper({ annotation, ...restProps }) {
```
esanzgar, line 151: This is the wrapper component I created to emulate the same behaviour than when `AnnotationHeader` is called through `Annotation` component.

src/sidebar/components/Thread.js
```diff
@@ -129,4 +141,31 @@ function Thread({ thread, threadsService }) {
...
+  const store = useStoreProxy();
+  const isSaving = annotation && store.isSavingAnnotation(annotation);
+  const isEditing = annotation && !!store.getDraft(annotation) && !isSaving;
```
esanzgar, line 154: Copied from `Annotation` component.

lyzadanger:
This is a screen shot from a slightly unusual state in which there are filters applied, and the top-level annotation has been deleted (in the Notebook here):

image

I think this is fine behavior for now. I still think there is a broader need to examine what belongs to a "thread" and what belongs to an "annotation," but I think this degrades gracefully enough for the moment!

src/sidebar/components/Thread.js
```diff
@@ -108,11 +109,22 @@ function Thread({ thread, threadsService }) {
...
 
         {showHiddenToggle && (
-          <div className="Thread__hidden-toggle-button-container">
-            <LabeledButton onClick={() => threadsService.forceVisible(thread)}>
-              Show {countHidden(thread)} more in conversation
-            </LabeledButton>
-          </div>
+          <>
+            {!thread.parent && (
+              <AnnotationHeaderWrapper
```
lyzadanger, line 114: Might I suggest choosing a name that indicates a little more clearly what this component is responsible for? Perhaps `HiddenAnnotationHeader` or `ThreadHeader`? The latter (`ThreadHeader`) might be helpful to highlight that the metadata rendered within pertains to the entire thread...

src/sidebar/components/Thread.js
lyzadanger: Any chance this logic could be hoisted up to `Thread` level and we can make `Annotation`, `AnnotationHeader` and this component a little "dumber" and DRY?

src/sidebar/components/Thread.js
```diff
@@ -129,4 +141,31 @@ function Thread({ thread, threadsService }) {
   );
 }
 
+/**
+ * Wrapper around AnnotationHeader
```
lyzadanger, line 145: Let's get a bit fuller of an explanation of what this component does here...

lyzadanger (commented):
I think this approach—wrapping the `AnnotationHeader` in some capacity—is generally fine. At some point, we'll want to make the `AnnotationHeader` a dumber component, and do more unpicking of what belongs to a single annotation and what is something that pertains to a thread.

I haven't gone into detail in my review about styling (though in my quick testing it _looks_ good) or tests; I'll wait until we move out of draft here.

Thanks for being thoughtful about this!

src/sidebar/components/Thread.js
esanzgar: I agree, thanks.

esanzgar:
> Any chance this logic could be hoisted up to Thread level and we can make Annotation, AnnotationHeader and this component a little "dumber" and DRY?

We could. However, I am afraid that because it uses the `store` I could trigger unnecessary re-rendering of some of the components of the thread if the annotation is saved or get in draft mode. I would need your advice here.

esanzgar:
The `AnnotationHeader` _is_ a dumb component. It depends on a smart component to drive it. 

In the closed #3423 PR, I chose to leave `Annotation` as the smart component, to  not duplicate some of the logic.

On the other hand, in this PR, I made a thin smart wrapper (`ThreadHeader`) to drive the `AnnotationHeader`.

lyzadanger (approved):
This looks good. I briefly re-tested this updated branch locally, but just for clarity, I have not tested in multiple browsers. Thanks!

PR #3424, opened 2021-05-20
Remove flag icon from annotation cards for LMS users

esanzgar:
1. Added an option (`allowFlagging`) in the service configuration to
   enable/disable flagging.

2. Added logic in `AnnotationActionBar` component to hide the flag icon
   if the service configuration has the option `allowFlagging`.

Closes https://github.com/hypothesis/product-backlog/issues/1126
(together with  https://github.com/hypothesis/lms/pull/2713)

### DIFF
--- a/src/sidebar/components/Annotation/AnnotationActionBar.js
+++ b/src/sidebar/components/Annotation/AnnotationActionBar.js
@@ -1,6 +1,7 @@
 import { IconButton } from '@hypothesis/frontend-shared';
 
 import { confirm } from '../../../shared/prompts';
+import serviceConfig from '../../config/service-config';
 import {
   sharingEnabled,
   annotationSharingLink,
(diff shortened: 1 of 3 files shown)


robertknight:
> Added an option (disableFlagging) in the service configuration to
disable flagging.

The existing configuration options on the same object have an opposite verb (`enableShareLinks`, `allowLeavingGroups`). I suggest changing this to `enableFlagging` with a default of `true` for consistency.

esanzgar:
I changed `disableFlagging` for `enableFlagging` for consistency. Thanks, @robertknight.

robertknight (approved):
This works as expected. I added a note about using an existing `serviceConfig` helper to get the main/active service from the `settings` object.

esanzgar:
I adopted the typical `serviceConfig` usage, thanks again.

PR #3428, opened 2021-05-20
Refactor of `serviceConfig`

esanzgar:
Simplified the logic. It doesn't change the behaviour.

### DIFF
--- a/src/sidebar/config/service-config.js
+++ b/src/sidebar/config/service-config.js
@@ -11,9 +11,9 @@
  * @return {Service|null}
  */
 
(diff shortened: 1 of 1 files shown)


robertknight:
@esanzgar - This needs a rebase on master.

esanzgar:
Rebase done.

LMS007 (approved):
I approved this on the other PR. I suspect it may be overkill to check for an array because the type should be implied, but since this is coming from another source, we likely need a validation step which we don't have. I think it's safe to leave that check for now.

PR #3429, opened 2021-05-20
Refactor of `isThirdPartyService`

esanzgar:
Simplified the logic but keep the same behaviour.


src/sidebar/config/service-config.js
```diff
@@ -11,9 +11,9 @@
  * @return {Service|null}
```
LMS007, line 11: My guess is we can do away with the return value here

LMS007 (approved):
Tested on LMS, Looks good. I think we might be able to do away with the return values since TS can figure those out from code. I think this may need to be rebased before merging.

src/sidebar/config/service-config.js
esanzgar: You are right, it could be inferred.

PR #3432, opened 2021-05-20
Add a `Destroyable` interface

esanzgar:
The purpose of this interface is to highlight developers that in order
to cleanly get ride of the classes that implement the interface, they
need to call the `destroy` method.

This interface doesn't actually enforce calling `destroy`, but hopefully
makes the omission more notorious.


esanzgar:
If this is consider not helpful I am fine closing the PR.

src/types/annotator.js
```diff
@@ -151,3 +151,11 @@
 
 // Make TypeScript treat this file as a module.
 export const unused = {};
+
+/**
+ * Destroyable classes implements the `destroy` method to properly remove all
```
robertknight, line 156:
```suggestion
 * Destroyable classes implement the `destroy` method to properly remove all
```

src/annotator/integrations/html.js
```diff
@@ -16,6 +17,7 @@ import { HTMLMetadata } from './html-metadata';
  * by a more specific integration (eg. for PDFs).
  *
  * @implements {Integration}
+ * @implements {Destroyable}
```
robertknight, line 20: I wonder if `Integration` could imply/extend `Destroyable`? I'm not sure if this is supported when using JSDoc to define types.

src/annotator/guest.js
```diff
@@ -24,6 +24,7 @@ import { ListenerCollection } from './util/listener-collection';
  * @typedef {import('../types/annotator').Integration} Integration
  * @typedef {import('../types/annotator').SidebarLayout} SidebarLayout
  * @typedef {import('../types/api').Target} Target
+ * @typedef {import('../types/annotator').Destroyable} Destroyable
```
robertknight, line 27: Can you sort these typedefs lines.

src/annotator/integrations/pdf.js
```diff
@@ -24,6 +24,7 @@ import { PDFMetadata } from './pdf-metadata';
  * @typedef {import('../../types/annotator').SidebarLayout} SidebarLayout
  * @typedef {import('../../types/api').Selector} Selector
  * @typedef {import('../../types/pdfjs').PDFViewerApplication} PDFViewerApplication
+ * @typedef {import('../../types/annotator').Destroyable} Destroyable
```
robertknight, line 27: Can you sort these typedef lines.

robertknight (approved):
Introducing an interface to document the `destroy` method in one place seems like a good idea to me.

src/annotator/guest.js
esanzgar: Sure

src/annotator/integrations/html.js
esanzgar: Indeed. I merged both types in this PR: https://github.com/hypothesis/client/pull/3454

PR #3425, opened 2021-05-20
Update GitHub API authentication and add logging to release scripts

robertknight:
Recent releases of the client don't have the expected release notes in https://github.com/hypothesis/client/releases. The release notes should be a list of titles of PRs that were merged along with a link to the PR. Instead the release notes are just the commit description from the commit that was tagged. When I run the `generate-change-list.js` script locally I get the expected output, so I suspect the issue may relate to the Node version that Jenkins is using.

This PR adds logging to the `create-github-release.js` script to show the changelog that has been generated locally before making the request to GitHub to create the release. It also replaces a deprecated approach to authenticating with the GitHub API.

PR #3434, opened 2021-05-21
Fix GitHub release notes

robertknight:
This is a follow up to https://github.com/hypothesis/client/pull/3425#issuecomment-845884247.

`create-github-release.js` had an implicit assumption that it was run
before the new version was tagged. It would generate the change log
based on PRs merged since the most recent tag. However Jenkins now runs
this script after the tag is created. As a result the change log was
always empty.

Fix the issue by explicitly passing the previous version as an argument
to the script so that it doesn't matter whether the new version has been
tagged or not.


lyzadanger (approved):
Seems like an improvement over a blank changelog :)

PR #3436, opened 2021-05-21
Small visual improvements for the notebook close button

esanzgar:
Because the `notebook` and `sidebar` are not able to communicate
directly (hopefully this will change soon), it is not possible to place
the close button on the `notebook` app itself.

While waiting for the above to be fixed, I propose a few minor style changes:

- Changed the `LabeledButton` to `IconButton` and only show the `X` (cancel
  icon). This is a widely understood and accepted pattern for modals

- Increase the size of the cancel icon

- Re-enable the cursor when hovering on the button (it seems that the
  `cursor: pointer` rule was not longer functional)

- Made the margin bigger to avoid the browser scrolling bar to hidden by the
  button.

These changes are meant to be a temporary improvement.

On a second commit, I addressed a behaviour only observed in Firefox, where the close button was displayed briefly at the bottom of the page when reloading a document with the hypothesis client.

The close button doesn't hide the scrolling bar (tested in latest version of
Chrome, Firefox and Safari):

Screenshot 2021-06-30 at 14 32 31

The close button still looks awkward when floating on small screen devices, but that's the only way to not take top space from the top part of the notebook app:

image

### DIFF
--- a/src/annotator/components/NotebookModal.js
+++ b/src/annotator/components/NotebookModal.js
@@ -98,6 +98,10 @@ export default function NotebookModal({ eventBus, config }) {
     emitter.current.publish('closeNotebook');
   };
 
+  if (groupId === null) {
+    return null;
+  }
+
   return (
     <div
       className={classnames('NotebookModal__outer', { 'is-hidden': isHidden })}
@@ -111,9 +115,7 @@ export default function NotebookModal({ eventBus, config }) {
             variant="dark"
           />
         </div>
-        {groupId !== null && (
-          <NotebookIframe key={iframeKey} config={config} groupId={groupId} />
-        )}
+        <NotebookIframe key={iframeKey} config={config} groupId={groupId} />
       </div>
     </div>
   );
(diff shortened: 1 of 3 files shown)


lyzadanger:
@esanzgar I noticed a couple of commits on this branch that look like they might be unrelated to the button work (and I think they might have been part of another PR?). Can you take a look and let me know if they belong with these changes?

esanzgar:
@lyzadanger, I fixed the commits.

lyzadanger:
@esanzgar Can you take a look at https://github.com/hypothesis/frontend-shared/pull/79 and see if the dark `IconButton` variant would do the trick (versus customizing the button styling)?

esanzgar:
Yes, no problem. I will take a look.

esanzgar:
@lyzadanger, indeed, the dark variant for the `IconButton` is perfect. Thanks!

esanzgar:
@lyzadanger, I am encountering a centering issue with the `IconButton`. It seems that `justify-content: flex-start;` is the default. However, maybe for IconButton it should be `justify-content: center;`

image

lyzadanger:
@esanzgar Please see https://github.com/hypothesis/frontend-shared/pull/109

lyzadanger (approved):
Thanks for your patience on this. Looks good!

PR #3446, opened 2021-05-24
Replace * and / with math.div() in sass files

LMS007:
See https://sass-lang.com/documentation/breaking-changes/slash-div for details

After upgrading to sass to v1.32, a new deprecation warning was appearing in the console,  but not actually failing the tests so CI was passing.

https://github.com/hypothesis/client/runs/2655035699


robertknight (approved):
👍

LMS007:
@robertknight manually

PR #3448, opened 2021-05-25
Replace `<Fragment>` for `<>`

esanzgar:
Generally, we prefer `<></>` JSX syntax to  `<Fragment></Fragment>`.

I reorder the imports to match our conventions.


esanzgar:
That's true, thanks.

PR #3455, opened 2021-05-26
Fix an error when dismissing toast messages

robertknight:
Services were recently converted to ES classes. One of the side effects / downsides of this is that service methods now depend on being called with the correct receiver (ie. `this` property). This was not the case for a call made to dismiss toast messages when clicking them.


lyzadanger (approved):
Let's get this one deployed!

PR #3457, opened 2021-05-27
Fix frame discoverability

esanzgar:
After one failed attempt to limit the frame discoverability (#3352), I
suggest here another temporary fix which involves:

* Disable the top-down, breadth-first traversal of frames to enable
  discoverability.

* Replace it by a targeted discoverability: identify the sender of the
  initial postMessage message and direct it to the right frame:

     - if the sender is the `sidebar` iframe (server frame), the
       postMessage is sent *only* to the `host` frame using
       `window.parent`.

     - if the sender is an annotatable iframe(s), send the postMessage
       to the `sidebar` iframe.

Pros:

- resolves the broken ePub example

- resolves the hyper-connectivity of frames

Cons:

- the `notebook` iframe is still not able to be discovered

- introduce an artificial delay on the discoverablity of annotatable
  iframes that could potentially brake

## Background

The client relies on an inter-frame communication system between the
`host` frame, where the client is initially loaded, and a number of
children iframes. For the communication to work, every frame needs to be
able to discover the iframe that acts as a server by sending a
`frame.postMessage`. Currently, the `sidebar` iframe is the server.

The following frames must establish communication with the server
`sidebar` iframe:

- `host` frame (where the client is initially loaded)
- `notebook` iframe
- additional annotatable iframe(s) (each have an `enable-annotation`
  attribute) where the another client instance is injected.

This layout represents the current arrangement of frames:

```
host frame (client)
|-> (generally, shadow DOMed) sidebar iframe (server)
|-> (generally, shadow DOMed) notebook iframe (client)
|-> [annotatable iframe/s] (client)
     |-> [annotatable iframe/s] (client)

```

There are two problems with the current discoverability algorithm:

1. It relies on `window.frames` to list all the other frames. Because
   `sidebar` and `notebook` iframes are generally wrapped on a shadow
   DOM they are not listed on `window.frames`.

2. It is very generic: the algorithm starts from the top-most frame in
   the hierarchy (`window.top`) and send messages to all the frame
   children *recursively*. If there are several clients initialised on
   individual frames, this algorithm causes *all* the `host` frames to
   be connected to all the `sidebar` iframes.

### DIFF
--- a/dev-server/documents/html/sidebar-external-container.mustache
+++ b/dev-server/documents/html/sidebar-external-container.mustache
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Annotatable iframe test</title>
+    <style>
+      body {
+        font-family: sans-serif;
+      }
+      .container {
+        display: flex;
+      }
+      #my-sidebar {
+        flex: 0 0 20%;
+      }
+      iframe {
+        flex: 1;
+        height: 400px;
+        resize: both;
+        overflow: auto;
+      }
+    </style>
+  </head>
+  <body>
+    <h1>Sidebar on external container test</h1>
+    <p>
+      The sidebar is placed inside the element with `#my-sidebar` id. The iframe
+      below has an `enable-annotation` attribute. The client detects and injects
+      the annotator into the iframe, which makes the iframe content annotatable.
+    </p>
+    <div class="container">
+      <div id="my-sidebar"></div>
+      <iframe src="/document/injectable-frame" enable-annotation></iframe>
+    </div>
+    {{{hypothesisScript}}}
+    <script>
+      window.hypothesisConfig = function () {
+        return {
+          externalContainerSelector: '#my-sidebar',
+        };
+      };
+    </script>
+  </body>
+</html>
--- a/dev-server/templates/index.mustache
+++ b/dev-server/templates/index.mustache
@@ -22,6 +22,7 @@
     <li><a href="/document/z-index">z-index test</a></li>
     <li><a href="/document/shadow-dom">Shadow DOM test</a></li>
     <li><a href="/document/parent-frame">Annotatable iframe test</a></li>
+    <li><a href="/document/sidebar-external-container">Sidebar on external container test</a></li>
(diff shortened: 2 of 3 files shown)


robertknight (commented):
Hi Eduardo,

The solution here relies on the guest and host frames being same-origin, so that the guest can directly query the DOM of the host. That's true in the common case of the guest/host being the same frame. I think it might also be true for all of the epub demos that we've built so far but I'm not certain about that. It might not be true for other use cases in future - especially when the content is put in an iframe for security reasons. I'm going to add some comments on how we might "fix up" the approach sketched in this PR, and then I'll say something about alternative communication mechanisms that avoid the guest <-> host same-origin constraint.

Regarding this PR:

1. I think it is important to avoid having knowledge about the DOM structure of the sidebar, notebook etc. spread across multiple places in the code. Otherwise someone else might change the code in the `Sidebar` class in future and not realize they are breaking something here.

2. It is also important to preserve the layering of the various modules that make up the client. `src/shared/` is intended to be at the root of the dependency graph so it can be used by `src/{annotator, sidebar}` but not the other way around. Cyclic dependencies can cause various problems down the line.

3. Dealing with the different orderings in which cross-origin frames can load is one of the trickier aspects of setting up communication. I think there are some things you can rely on (eg. the host will always load before the sidebar), others that you can't (whether guest frames load before the sidebar) and some guarantees which we can probably make at present but may not always hold (whether we'll always be able to guarantee that the host code runs before any guest frames have loaded). In the case where we can't guarantee the order in which two frames load, having one party send repeated "hello" messages at an interval to the other until communication is established or a timeout elapses may actually be a reasonable thing to do.

If we were to rethink the frame communication mechanism to avoid the guest <-> host same-origin constraint then it could work something like:

   1. The sidebar frame notifies the host when it is ready to communicate with guests
   2. Guest frames notify the host when they are ready to communicate with the sidebar (in the common case where the host also contains a guest, this can happen without any messages being sent)
   3. When each partner in a (sidebar, guest) pair are ready to communicate, the host frame creates a `MessageChannel` and sends one port to the guest and one port to the sidebar. Each side then calls `start` on the channel after receiving it to start receiving messages.
   
 Here step (1) and (2) can happen in either order. Step (2) assumes that code will run in the host frame before guest frames. If we can't make that assumption then step (2) might have to involve some polling by the guest until the host responds.

esanzgar:
Thank you very much for the advise and I agree with that.

I will close this draft PR and work on an alternative solution.

PR #3476, opened 2021-06-07
Improve HTMLMetadata test descriptions and remove unused `uris` method

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/3475/**~~

This PR improves the descriptions for the tests of `HTMLMetadata`'s `getDocumentMetadata` method to make it more obvious what is being tested. I also made some other minor cleanups in the process:

 - Fix a test case ("should have twitter card metadata") that was not executed because it was declared inside another test
 - Remove a test case ("uri() returns the canonical uri'") that was in the wrong `describe` block and duplicated another test
 - Convert function expressions => arrow functions
 - Remove an unused `uris` method

### DIFF
--- a/src/annotator/integrations/html-metadata.js
+++ b/src/annotator/integrations/html-metadata.js
@@ -78,22 +78,6 @@ export class HTMLMetadata {
   }
 
   /**
-   * Returns all uris for the document being annotated
-   *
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
👍🏻

PR #3491, opened 2021-06-07
Refactor annotator/config/ (2/N)

LMS007:
Change configFrom() to configDefinitions() 

The main difference is that configFrom returned config key:value pairs, but now configDefinitions returns key:object pairs. Where the object may hold additional flags and methods for each specific config value. Currently the only method included is valueFn() which gets the value from its source. Others will follow.

Additionally:
- Remove `experimental` from configDefinitions/configFrom as it is not used

-------

The end goal here is to have the definitions look more or less like this

```...
    openSidebar: {
      allowInBrowserExt: true,
      defaultValue: false,
      coerce: toBoolean,
      getValue: settings => settings.hostPageSetting,
    },
    ...

```

Where getConfig will do the logic to get a value from getValue, set a optional default, coerce the value if needed, and/or restrict based on allowInBrowserExt. Much of this logic is currently inside settings.js and that will be pulled out so settings can be reduced to only retrieving the value from source.

_This PR has no external changes_

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -3,6 +3,14 @@ import { toBoolean } from '../../shared/type-coercions';
 
 /**
  * @typedef {'sidebar'|'notebook'|'annotator'|'all'} AppContext
+ *
+ * @typedef {import('./settings').SettingsGetters} SettingsGetters
+ *
+ * @typedef ConfigDefinition
+ * @prop {(settings: SettingsGetters) => any} getValue -
+ *   Method to retrieve the value from the incoming source
+ *
+ * @typedef {Record<string, ConfigDefinition>} ConfigDefinitionMap
  */
 
 /**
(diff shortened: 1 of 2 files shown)


src/annotator/config/index.js
```diff
@@ -53,67 +61,104 @@ function configurationKeys(appContext) {
       return contexts.notebook;
     case 'all':
       // Complete list of configuration keys used for testing.
-      return [...contexts.annotator, ...contexts.sidebar, ...contexts.notebook];
+      return [
+        ...contexts.annotator,
+        ...contexts.sidebar,
+        ...contexts.notebook,
+        // "experimental" currently has no uses in any app contexts, but is included
+        // for test coverage
```
robertknight, line 69: If the `experimental` field is not actively used, and isn't mentioned in the [public documentation](https://h.readthedocs.io/projects/client/en/latest/publishers/config/) is there anything stopping us from just removing it?

src/annotator/config/index.js
```diff
@@ -3,6 +3,14 @@ import { toBoolean } from '../../shared/type-coercions';
 
 /**
  * @typedef {'sidebar'|'notebook'|'annotator'|'all'} AppContext
+ *
+ * @typedef ConfigDefinition
+ * @prop {() => any} valueFn -
```
robertknight, line 8: I think something like `getValue` might be a slightly more obvious name here.

src/annotator/config/index.js
```diff
@@ -53,67 +61,104 @@ function configurationKeys(appContext) {
...
  */
-function configFrom(settingsKeys, window_) {
-  const settings = settingsFrom(window_);
-  const allConfigSettings = {
-    annotations: settings.annotations,
-    appType: settings.hostPageSetting('appType', {
-      allowInBrowserExt: true,
-    }),
-    branding: settings.hostPageSetting('branding'),
+function configDefinitions(settings) {
```
robertknight, line 81: Logically `configDefinitions` returns a static description of all supported configuration properties. Currently this is wrapped inside a function so that you can pass in `settings`. If however the `valueFn` took `settings` as an argument then you could make `configDefinitions` a static variable:

```js
const configDefinitions = {
  annotations: {
    valueFn: settings => settings.annotations,
  },
  ...
}
```

The advantage of this would be primarily to make it obvious that the shape of `configDefinitions` is static and only the part involving reading the value is dynamic. A side benefit is that you could potentially reference the shape of `configDefinitions` in types more easily.

robertknight (commented):
Hi Kyle,

I had a couple of pieces of feedback here. One is around the name of the `valueFn` field. The other is that the general trend here is to make the definitions of available settings more declarative, which I think is good, and I suggested a way to push this a bit further and a rationale for doing so.

src/annotator/config/index.js
LMS007: Negative, lets remove it.

src/annotator/config/index.js
LMS007: I think that's fine and it makes total sense. In the next set of commits in the third PR, I had been passing the key name in as well to reduce some redundancy, but maybe there is a way we can still accomplish that and still pass settings too.

LMS007:
Updated the PR.
- configDefinitions is no longer a function
- valueFn is now getValue
- removed `experimental`  config key
- added a new type `SettingsGetters` (terrible name) but it asserts what the settings() will return. That object is then passed to the getValue() functions in index.js for fetching each value. We could instead just declare this to be an generic `Object` and reduce the overhead of managing this type--what settings returns. I could use feedback here.

src/annotator/config/settings.js
```diff
@@ -3,6 +3,21 @@ import { parseJsonConfig } from '../../boot/parse-json-config';
...
+/**
+ * @typedef SettingsGetters
+ * @prop {string|null} annotations
+ * @prop {string} clientUrl
+ * @prop {string|null} group
+ * @prop {string} notebookAppUrl
+ * @prop {string} showHighlights
+ * @prop {string} sidebarAppUrl
+ * @prop {string|null} query
+ * @prop {(name: string, options?: Object) => (string|null)} hostPageSetting
```
robertknight, line 15: I think this type would be easier to skim if the related properties/methods were grouped together (eg. `annotations` with `query` and `group` and `notebookAppUrl` with `sidebarAppUrl`).

Writing this down makes it more obvious that this object contains a mix of generic methods (`hostPageSetting`) and a bunch of special-case getters. In future I think it would be good to try and have fewer special cases.

src/annotator/config/index.js
```diff
@@ -60,71 +68,100 @@ function configurationKeys(appContext) {
 }
 
 /**
- * Reads the Hypothesis configuration from the environment.
- *
- * @param {string[]} settingsKeys - List of settings that should be returned.
- * @param {Window} window_ - The Window object to read config from.
+ * The definition for configuration sources, their default values
```
robertknight, line 71:
```suggestion
 * Definitions of configuration keys
```

The entries in this may are configuration _keys_ rather than _sources_ (which I take to mean JSON script tags, link elements etc.). I'm not completely sure about the term "definition" vs say "metadata", but lets roll with it.

src/annotator/config/index.js
```diff
@@ -60,71 +68,100 @@ function configurationKeys(appContext) {
...
+  const settings = settingsFrom(window_);
+  const config = {};
   // Filter the config based on the application context as some config values
   // may be inappropriate or erroneous for some applications.
-  const filteredKeys = configurationKeys(appContext);
-  const config = configFrom(filteredKeys, window_);
+  let filteredKeys = configurationKeys(appContext);
+  filteredKeys.forEach(name => {
+    const configDef = configDefinitions[name];
+    // Get the value from the configuration source and run through an optional coerce method
```
robertknight, line 162:
```suggestion
    // Get the value from the configuration source
```

There is no coercion happening currently. Presumably this will come in future.

src/annotator/config/settings.js
```diff
@@ -3,6 +3,21 @@ import { parseJsonConfig } from '../../boot/parse-json-config';
 import configFuncSettingsFrom from './config-func-settings-from';
 import isBrowserExtension from './is-browser-extension';
 
+/**
+ * @typedef SettingsGetters
```
robertknight, line 7: As you said, this name is a little awkward, but it makes sense.

robertknight (approved):
I left some minor corrections, but otherwise this looks good to me.

src/annotator/config/index.js
LMS007: I believe some of these comments were merge errors dude to upstream changes before. This might be one of them, good catch.

LMS007:
fixed typos in comments and re-ordered SettingsGetters to

```
 * @typedef SettingsGetters
 * @prop {string|null} annotations
 * @prop {string|null} query
 * @prop {string|null} group
 * @prop {string} showHighlights
 * @prop {string} clientUrl
 * @prop {string} sidebarAppUrl
 * @prop {string} notebookAppUrl
 ```

PR #3492, opened 2021-06-07
Always use current `document.baseURI` in `HTMLMetadata`

robertknight:
`HTMLMetadata` read `document.baseURI` only once in the constructor,
risking issues due to the cached value becoming stale. `document.baseURI` is
available in all supported browsers and is easily mocked where needed in
tests, so just always read the current value when needed.

Also remove a `normalizeURI` test seam that was unused. If we did want
to mock this in future we'd use the `$imports` object as we do in other
tests.


LMS007 (approved):
Looks good Rob.

PR #3495, opened 2021-06-09
Refactor annotator/config/ (3/N) 

LMS007:
### Move coerce() into configDefinitions

Previously, coerce was passed to the hostPageSetting. Moving the logic up into getConfig allows any setting to have a coerce method, not just settings pulled from hostPageSetting. Eventually, other settings may need coerce logic such as showHighlights.

------

relates to 

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -9,7 +9,10 @@ import { toBoolean } from '../../shared/type-coercions';
  * @typedef ConfigDefinition
  * @prop {(settings: SettingsGetters) => any} getValue -
  *   Method to retrieve the value from the incoming source
- *
+ * @prop {(value: any) => any} [coerce] - Transform a value's type, value or both
+ */
+
+/**
  * @typedef {Record<string, ConfigDefinition>} ConfigDefinitionMap
  */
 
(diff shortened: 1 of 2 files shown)


src/annotator/config/index.js
```diff
@@ -159,8 +162,9 @@ export function getConfig(appContext = 'annotator', window_ = window) {
   let filteredKeys = configurationKeys(appContext);
   filteredKeys.forEach(name => {
     const configDef = configDefinitions[name];
-    // Get the value from the configuration source
-    config[name] = configDef.getValue(settings);
+    const coerceFn = configDef.coerce ? configDef.coerce : name => name; // use no-op if omitted
+    // Get the value from the configuration source and run through an optional coerce method
+    config[name] = coerceFn(configDef.getValue(settings));
```
robertknight, line 167: I think I would probably have written this without the no-op function as:

```js
const value = configDef.getValue(settings);
config[name] = configDef.coerce ? configDef.coerce(value) : value;
```

I find this very slightly easier to follow.

robertknight (approved):
I added a minor suggestion, but otherwise LGTM.

PR #3496, opened 2021-06-10
Refactor annotator/config/ (4/N)

LMS007:
### Move urlFromLinkTag to its own module

- Add a new parameter to urlFromLinkTag called "type" to allow urlFromLinkTag to get the client url as well as sidebar and notebook urls. This eliminates the need to have a custom clientUrl method in settings.js

- urlFromLinkTag() will also soon be needed outside of settings.js as part of a future refactor to move logic into getConfig() which determines if the browser extension context is true.

- Remove the deprecated "coerce" option from hostPageSetting as this is now moved to configDefinitions

---------

relates to 

### DIFF
--- a/src/annotator/config/settings.js
+++ b/src/annotator/config/settings.js
@@ -3,6 +3,7 @@ import { toBoolean } from '../../shared/type-coercions';
 
 import configFuncSettingsFrom from './config-func-settings-from';
 import isBrowserExtension from './is-browser-extension';
+import { urlFromLinkTag } from './url-from-link-tag';
 
 /**
  * @typedef SettingsGetters
@@ -31,75 +32,6 @@ export default function settingsFrom(window_) {
     jsonConfigs = {};
   } else {
     jsonConfigs = parseJsonConfig(window_.document);
-  }
-
-  /**
-   * Return the href of the first annotator link in the given
-   * document with this `rel` attribute.
-   *
-   * Return the value of the href attribute of the first
-   * `<link type="application/annotator+html" rel="${rel}">`
-   * element in the given document. This URL is used as the `src` for sidebar
-   * or notebook iframes.
-   *
-   * @param {string} rel - The `rel` attribute to match
-   * @return {string} - The URL to use for the iframe
-   * @throws {Error} - If there's no link with the `rel` indicated, or the first
-   *   matching link has no `href`
-   */
-  function urlFromLinkTag(rel) {
-    const link = window_.document.querySelector(
-      `link[type="application/annotator+html"][rel="${rel}"]`
-    );
-
-    if (!link) {
-      throw new Error(
-        `No application/annotator+html (rel="${rel}") link in the document`
-      );
-    }
-
-    if (!link.href) {
-      throw new Error(
-        `application/annotator+html (rel="${rel}") link has no href`
-      );
-    }
-
-    return link.href;
-  }
-
-  /**
-   * Return the href URL of the first annotator client link in the given document.
-   *
-   * Return the value of the href attribute of the first
-   * `<link type="application/annotator+javascript" rel="hypothesis-client">`
-   * element in the given document.
-   *
-   * This URL is used to identify where the client is from and what url should
-   * be used inside of subframes.
-   *
-   * @return {string} - The URL that the client is hosted from
-   * @throws {Error} - If there's no annotator link or the first annotator has
-   *   no href.
-   *
-   */
-  function clientUrl() {
-    const link = window_.document.querySelector(
-      'link[type="application/annotator+javascript"][rel="hypothesis-client"]'
-    );
-
-    if (!link) {
-      throw new Error(
-        'No application/annotator+javascript (rel="hypothesis-client") link in the document'
-      );
-    }
-
-    if (!link.href) {
-      throw new Error(
-        'application/annotator+javascript (rel="hypothesis-client") link has no href'
-      );
-    }
-
-    return link.href;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/annotator/config/test/url-from-link-tag-test.js
```diff
@@ -0,0 +1,67 @@
...
+      'javascript'
+    );
+    assert.equal(
+      urlFromLinkTag(window, 'hypothesis-client', 'javascript'),
+      'http://example.com/app.html'
+    );
+    link.remove();
+  });
+
+  context('when there are multiple annotator+html notebook links', () => {
```
robertknight, line 37: Given that we're testing a generic utility here, I would suggest that the tests also need to be described in a similarly generic way, though they may use parameters/data from a real use case. In this case, the `context` block's description could be "when there are multiple matching links" and then the inner `it` test's description would be "returns the href from the first matching link".

src/annotator/config/test/url-from-link-tag-test.js
```diff
@@ -0,0 +1,67 @@
...
+      assert.equal(
+        urlFromLinkTag(window, 'notebook', 'html'),
+        'http://example.com/app1'
+      );
+      link1.remove();
+      link2.remove();
+    });
+  });
+
+  context('incorrect link tag', () => {
```
robertknight, line 50:
```suggestion
  context('missing or invalid link tag', () => {
```

The contents of this block encompass both missing `<link>` tags and ones that are invalid because they have no `href`.

src/annotator/config/test/url-from-link-tag-test.js
```diff
@@ -0,0 +1,67 @@
...
+    ['html', 'javascript'].forEach(type => {
+      it('throws an error if link tag is missing', async () => {
```
robertknight, line 52: This test contains no `await` statements, so it doesn't need to be `async`. Same with the test below.

src/annotator/config/test/url-from-link-tag-test.js
```diff
@@ -0,0 +1,67 @@
+import { urlFromLinkTag } from '../url-from-link-tag';
+
+describe('annotator/config/url-from-link-tag', () => {
```
robertknight, line 3: When a test module contains only a single thing (function, class etc.) being tested and the item's name is reasonably unique, the convention is to name the `describe` block after the item. If a module contains multiple functions/classes etc. being tested, then you'd structure it as:

```js
describe('path/to/module', () => {
  describe('functionOne', () => {
     ...
  });

  describe('functionTwo', () => {
     ...
  });
});
```

This convention just makes it easier to know what a test refers to if it fails.

src/annotator/config/url-from-link-tag.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * * `<link type="application/annotator+javascript" rel="hypothesis-client">`
+ *
+ * element in the given document. This URL is used as the `src` for sidebar
+ * or notebook iframes, or to identify where the client is from and what url
+ * should be used inside of subframes.
+ *
+ * @param {Window} window_
+ * @param {string} rel - The `rel` attribute to match
+ * @param {'javascript'|'html'} type - document type
```
robertknight, line 18:
```suggestion
 * @param {'javascript'|'html'} type - Type of resource that the link refers to
```

src/annotator/config/url-from-link-tag.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Return the href of the first annotator link in the given
```
robertknight, line 2: I think the wording here could be more succinct. Something along the lines of:

```
Return the URL of a resource related to the Hypothesis client that has been stored in
the page via a `<link type="application/annotator+{type}">` tag.
```

Below that I think it would be useful to give some additional context about why this mechanism exists.

```
These link tags are generally written to the page by the boot script, which may be executed
in a separate JavaScript realm (eg. in the browser extension), and thus can share information
with the annotator code via the DOM but not JS globals.
```

I suggest the rest of the comment, which refers to specific uses of this function, can be removed as it is going to get out of date if new uses of this code are added elsewhere.

robertknight (approved):
I requested some small changes to the tests and description of `urlFromLinkTag`. Otherwise the changes look good.

src/annotator/config/test/settings-test.js
```diff
@@ -22,245 +27,42 @@ describe('annotator/config/settingsFrom', () => {
...
-          it('returns the href from the notebook link', () => {
-            assert.equal(
-              settingsFrom(window).notebookAppUrl,
-              'http://example.com/app.html'
-            );
-          });
-        }
+  describe('#notebookAppUrl', () => {
+    it('calls urlFromLinkTag with appropriate params', () => {
+      assert.exists(settingsFrom(window).notebookAppUrl);
```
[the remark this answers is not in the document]
LMS007: I'm not sure it needs to because it's moved to a different module. From the standpoint of this module, my assumption was it just needs to make sure it calls that notebookAppUrl with the appropriate params. url-from-link-tag-test.js is where the specific values are tested. Does that seem reasonable?

src/annotator/config/settings.js
```diff
@@ -189,20 +121,20 @@ export default function settingsFrom(window_) {
   function hostPageSetting(name, options = {}) {
     const allowInBrowserExt = options.allowInBrowserExt || false;
     const hasDefaultValue = typeof options.defaultValue !== 'undefined';
-    // Optional coerce method, or a no-op.
-    const coerceValue =
-      typeof options.coerce === 'function' ? options.coerce : name => name;
```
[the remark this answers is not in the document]
LMS007: I think I intended it to be part of this because of all the test changes too. What I'll do is update the git commit message to explain this a bit better.

src/annotator/config/test/settings-test.js
```diff
@@ -22,245 +27,42 @@ describe('annotator/config/settingsFrom', () => {
...
+      assert.calledWith(
+        fakeUrlFromLinkTag,
+        // ignore window object
```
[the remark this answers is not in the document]
LMS007: I assumed it might be expensive, but testing this shows its not. Adding it back in.

src/annotator/config/test/settings-test.js
```diff
@@ -22,245 +27,42 @@ describe('annotator/config/settingsFrom', () => {
...
-      context("when there's no annotator+html sidebar link", () => {
-        it('throws an error', () => {
-          assert.throws(() => {
-            settingsFrom(window).sidebarAppUrl; // eslint-disable-line no-unused-expressions
-          }, 'No application/annotator+html (rel="sidebar") link in the document');
-        });
-      });
+  describe('#sidebarAppUrl', () => {
+    it('calls urlFromLinkTag with appropriate params', () => {
+      assert.exists(settingsFrom(window).sidebarAppUrl);
```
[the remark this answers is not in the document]
LMS007: Same situation, I strongly feel we should not have redundant tests. If we start to test things in a deeper mannar, than our unit tests look less like unit tests and more like end to end tests.  The reason I called .exists is so lint would pass. I really don't need need any tests here, but since I have to have one, it could simply test the value, however this makes it somewhat redundant.

src/annotator/config/test/url-from-link-tag-test.js
LMS007: good catch!

src/annotator/config/test/url-from-link-tag-test.js
LMS007: okay makes sense

PR #3497, opened 2021-06-10
Refactor annotator/config/ (5/N)

LMS007:
### Add allowInBrowserExt and defaultValue to configDefinitions 

- Logic for `allowInBrowserExt` and `defaultValue` are moved up into getConfig() from various places inside settings. There are no external changes here, but there are a few subtle changes to how a config values work.

  1. Previously if in a browser ext context and  `allowInBrowserEx` was false (default), then the value was was changed to null. This bit of logic was not entirely explicit, but now the logic is that the value will be set to the `defaultValue` if provided or otherwise the value will be omitted entirely. Currently, all values have a `defaultValue` and most of the defaults are null. This ensures things work the same, but also helps bring clarity to what values will be transformed under this condition. Later we can revisit these config values and determine if any don't need a default or simply leave them as is.

  2. The `defaultValue` was previously only used for several fields, but now its set for every field for clarity. It should also be noted that the `defaultValue` will now take precedent over the `coerce` method if settings returns an undefined value for that config key. This is in contrast to how it worked before, but was problematic under some possible circumstances. For example, if a settings value was not found or undefined, and a `defaultValue` was true and the coerce was `toBoolean`, then the value would become false rather than true and that is confusing because the default would not work due to the `coerce` taking precedent. This is now fixed.

- Remove default export from is-browser-extension

--------

relates to 

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -1,5 +1,7 @@
+import { isURLFromBrowserExtension } from './is-browser-extension';
 import settingsFrom from './settings';
 import { toBoolean } from '../../shared/type-coercions';
+import { urlFromLinkTag } from './url-from-link-tag';
 
 /**
  * @typedef {'sidebar'|'notebook'|'annotator'|'all'} AppContext
@@ -8,7 +10,15 @@ import { toBoolean } from '../../shared/type-coercions';
  *
  * @typedef ConfigDefinition
  * @prop {(settings: SettingsGetters) => any} getValue -
- *   Method to retrieve the value from the incoming source
+ *  Method to retrieve the value from the incoming source
+ * @prop {boolean} allowInBrowserExt -
+ *  Allow this setting to be read in the browser extension. If this is false
+ *  and browser extension context is true, use `defaultValue` if provided otherwise
+ *  ignore the config key
+ * @prop {any} [defaultValue] - Sets a default if `getValue` returns undefined
+ *   nb. Currently all setting have a defaultValue. An improvement will be to
+ *   audit each setting and see if we can remove the defaultValue which then allows
+ *   the setting to be omitted from the config when a value is undefined.
  * @prop {(value: any) => any} [coerce] - Transform a value's type, value or both
  */
 
@@ -76,75 +86,104 @@ function configurationKeys(appContext) {
  */
 const configDefinitions = {
   annotations: {
+    allowInBrowserExt: true,
+    defaultValue: null,
     getValue: settings => settings.annotations,
   },
   appType: {
-    getValue: settings =>
-      settings.hostPageSetting('appType', {
-        allowInBrowserExt: true,
-      }),
+    allowInBrowserExt: true,
+    defaultValue: null,
+    getValue: settings => settings.hostPageSetting('appType'),
   },
   branding: {
+    defaultValue: null,
+    allowInBrowserExt: false,
     getValue: settings => settings.hostPageSetting('branding'),
   },
   // URL of the client's boot script. Used when injecting the client into
   // child iframes.
   clientUrl: {
+    allowInBrowserExt: true,
+    defaultValue: null,
     getValue: settings => settings.clientUrl,
   },
   enableExperimentalNewNoteButton: {
+    allowInBrowserExt: false,
+    defaultValue: false,
     getValue: settings =>
       settings.hostPageSetting('enableExperimentalNewNoteButton'),
   },
   group: {
+    allowInBrowserExt: true,
+    defaultValue: null,
     getValue: settings => settings.group,
   },
   focus: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.hostPageSetting('focus'),
   },
   theme: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.hostPageSetting('theme'),
   },
   usernameUrl: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.hostPageSetting('usernameUrl'),
   },
   onLayoutChange: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.hostPageSetting('onLayoutChange'),
   },
   openSidebar: {
+    allowInBrowserExt: true,
+    defaultValue: false,
     coerce: toBoolean,
-    getValue: settings =>
-      settings.hostPageSetting('openSidebar', {
-        allowInBrowserExt: true,
-      }),
+    getValue: settings => settings.hostPageSetting('openSidebar'),
   },
   query: {
+    allowInBrowserExt: true,
+    defaultValue: null,
     getValue: settings => settings.query,
   },
   requestConfigFromFrame: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.hostPageSetting('requestConfigFromFrame'),
   },
   services: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.hostPageSetting('services'),
   },
   showHighlights: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.showHighlights,
   },
   notebookAppUrl: {
+    allowInBrowserExt: true,
+    defaultValue: null,
     getValue: settings => settings.notebookAppUrl,
   },
   sidebarAppUrl: {
+    allowInBrowserExt: true,
+    defaultValue: null,
     getValue: settings => settings.sidebarAppUrl,
   },
   // Sub-frame identifier given when a frame is being embedded into
   // by a top level client
   subFrameIdentifier: {
-    getValue: settings =>
-      settings.hostPageSetting('subFrameIdentifier', {
-        allowInBrowserExt: true,
-      }),
+    allowInBrowserExt: true,
+    defaultValue: null,
+    getValue: settings => settings.hostPageSetting('subFrameIdentifier'),
   },
   externalContainerSelector: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.hostPageSetting('externalContainerSelector'),
   },
 };
@@ -162,8 +201,29 @@ export function getConfig(appContext = 'annotator', window_ = window) {
   let filteredKeys = configurationKeys(appContext);
   filteredKeys.forEach(name => {
     const configDef = configDefinitions[name];
+    const hasDefault = configDef.defaultValue !== undefined; // A default could be null
+    const isURLFromBrowserExtensionTrue = isURLFromBrowserExtension(
+      urlFromLinkTag(window_, 'sidebar', 'html')
+    );
+
+    // Only allow certain values in the browser extension context
+    if (!configDef.allowInBrowserExt && isURLFromBrowserExtensionTrue) {
+      if (hasDefault) {
+        config[name] = configDef.defaultValue;
+      }
+      return;
+    }
+
+    // Get the value from the configuration source
     const value = configDef.getValue(settings);
-    // Get the value from the configuration source and run through an optional coerce method
+    if (value === undefined) {
+      if (hasDefault) {
+        config[name] = configDef.defaultValue;
+      }
+      return;
+    }
+
+    // Finally, run the value through an optional coerce method
     config[name] = configDef.coerce ? configDef.coerce(value) : value;
   });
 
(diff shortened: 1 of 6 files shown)


src/annotator/config/index.js
```diff
@@ -8,7 +10,12 @@ import { toBoolean } from '../../shared/type-coercions';
  *
  * @typedef ConfigDefinition
  * @prop {(settings: SettingsGetters) => any} getValue -
- *   Method to retrieve the value from the incoming source
+ *  Method to retrieve the value from the incoming source
+ * @prop {boolean} allowInBrowserExt -
+ *  Allow this name to be available in the browser extension. If this is false
```
robertknight, line 15:
```suggestion
 *  Allow this setting to be read in the browser extension. If this is false
```

I changed the terms here for clarity.

src/annotator/config/index.js
```diff
@@ -8,7 +10,12 @@ import { toBoolean } from '../../shared/type-coercions';
...
+ *  and browser extension context is true, use `defaultValue` if provided otherwise
+ *  ignore the config key
+ * @prop {any} [defaultValue] - Sets a default if `valueFn` returns undefined
```
robertknight, line 18:
```suggestion
 * @prop {any} [defaultValue] - Sets a default if `getValue` returns undefined
```

The `valueFn` property was renamed after an earlier discussion. This needs to be updated accordingly.

src/annotator/config/index.js
```diff
@@ -76,75 +83,104 @@ function configurationKeys(appContext) {
...
     getValue: settings => settings.clientUrl,
   },
   enableExperimentalNewNoteButton: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings =>
       settings.hostPageSetting('enableExperimentalNewNoteButton'),
   },
   group: {
+    allowInBrowserExt: false,
```
robertknight, line 114: We do want to allow this setting in the browser extension actually. It is used for direct links to groups.

src/annotator/config/index.js
```diff
@@ -162,9 +198,35 @@ export function getConfig(appContext = 'annotator', window_ = window) {
   let filteredKeys = configurationKeys(appContext);
   filteredKeys.forEach(name => {
     const configDef = configDefinitions[name];
+    const hasDefault = configDef.defaultValue !== undefined; // A default could be null
+    const browserExtensionTrue = isBrowserExtension(
```
robertknight, line 202: The name `browserExtensionTrue` is a little awkward. I see this was done to avoid a conflict with the `isBrowserExtension` function. I think it might have been better to rename that function (eg. `isURLFromBrowserExtension`)

src/annotator/config/is-browser-extension.js
```diff
@@ -6,6 +6,6 @@
  *   website.
  *
  */
-export default function isBrowserExtension(app) {
+export function isBrowserExtension(app) {
```
robertknight, line 9: This would be a good time to rename `app` to something better (perhaps just `url`?) and add a `@param` entry to document its type.

src/annotator/config/test/index-test.js
```diff
@@ -51,24 +67,117 @@ describe('annotator/config/index', function () {
...
-  ) {
-    it(`reads ${settingName} from the host page, even when in a browser extension`, function () {
-      getConfig('all', 'WINDOW');
-      assert.calledWith(
-        fakeSettingsFrom().hostPageSetting,
-        settingName,
-        sinon.match({ allowInBrowserExt: true })
-      );
+  describe('browser extension', () => {
+    context('browser extension set to true', function () {
```
robertknight, line 71:
```suggestion
    context('when client is loaded from the browser extension', () => {
```

The previous naming suggested there was a flag somewhere that is set to a boolean value indicating if this is the browser extension, but that's not how it works.

I rephrased this in a way that better describes the situation rather than how we check for that situation.

src/annotator/config/test/index-test.js
```diff
@@ -51,24 +67,117 @@ describe('annotator/config/index', function () {
...
+            subFrameIdentifier: 'fakeValue',
+            theme: null,
+            usernameUrl: null,
+          },
+          config
+        );
+      });
+    });
+
+    context('browser extension set to false', function () {
```
robertknight, line 102:
```suggestion
    context('when client is not loaded from browser extension', () => {
```

See previous comments.

src/annotator/config/test/index-test.js
```diff
@@ -51,24 +67,117 @@ describe('annotator/config/index', function () {
...
+          group: undefined,
+          notebookAppUrl: undefined,
+          showHighlights: undefined,
+          sidebarAppUrl: undefined,
+          query: undefined,
+        }),
+      });
+    });
+
+    this.afterEach(() => {
```
robertknight, line 151:
```suggestion
    afterEach(() => {
```

I assume the use of `this` here was not intended? In this case I think `this` will actually refer to the containing `context` block rather than the `describe` block due the use of arrow functions for the `describe` but function expressions for `context`.

We generally avoid the use of `this` to refer to the Mocha context in tests unless absolutely necessary - and I'm not even sure if we have ever needed it in the client.

LMS007: Correct, not intended and typo!

src/annotator/config/index.js
```diff
@@ -8,7 +10,12 @@ import { toBoolean } from '../../shared/type-coercions';
  *
  * @typedef ConfigDefinition
  * @prop {(settings: SettingsGetters) => any} getValue -
- *   Method to retrieve the value from the incoming source
+ *  Method to retrieve the value from the incoming source
+ * @prop {boolean} allowInBrowserExt -
+ *  Allow this setting to be read in the browser extension. If this is false
+ *  and browser extension context is true, use `defaultValue` if provided otherwise
+ *  ignore the config key
+ * @prop {any} [defaultValue] - Sets a default if `getValue` returns undefined
```
robertknight, line 18: `defaultValue` is optional here but you're setting it for every key in the `configDefinitions` map below. _Should_ it be optional?

src/annotator/config/index.js
```diff
@@ -76,75 +83,104 @@ function configurationKeys(appContext) {
...
   // URL of the client's boot script. Used when injecting the client into
   // child iframes.
   clientUrl: {
+    allowInBrowserExt: true,
+    defaultValue: null,
     getValue: settings => settings.clientUrl,
   },
   enableExperimentalNewNoteButton: {
+    allowInBrowserExt: false,
+    defaultValue: null,
```
robertknight, line 109: Since this is a boolean, I think it would make sense for the default value to be `false`.

src/annotator/config/index.js
```diff
@@ -76,75 +83,104 @@ function configurationKeys(appContext) {
...
     getValue: settings => settings.hostPageSetting('requestConfigFromFrame'),
   },
   services: {
+    allowInBrowserExt: false,
+    defaultValue: null,
     getValue: settings => settings.hostPageSetting('services'),
   },
   showHighlights: {
+    allowInBrowserExt: false,
+    defaultValue: null,
```
robertknight, line 161: The default value here should be `always`. This is resolved in https://github.com/hypothesis/client/pull/3500.

src/annotator/config/index.js
```diff
@@ -162,8 +198,33 @@ export function getConfig(appContext = 'annotator', window_ = window) {
   let filteredKeys = configurationKeys(appContext);
   filteredKeys.forEach(name => {
     const configDef = configDefinitions[name];
+    const hasDefault = configDef.defaultValue !== undefined; // A default could be null
+    const isURLFromBrowserExtension = isBrowserExtension(
```
robertknight, line 202: I was thinking to rename the `isBrowserExtension` function rather than the local variable here, so as to use the shorter name for the most local symbol. Also `isURLFromBrowserExtension` makes more sense as the function name because it tests the provided URL, whereas the local variable here means "is the client being loaded from the browser extension".

src/annotator/config/index.js
```diff
@@ -162,8 +198,33 @@ export function getConfig(appContext = 'annotator', window_ = window) {
...
+      urlFromLinkTag(window_, 'sidebar', 'html')
+    );
+
+    // Only allow certain values in the browser extension context
+    if (!configDef.allowInBrowserExt && isURLFromBrowserExtension) {
+      // If the value is not allowed here, then set to the default if provided, otherwise ignore
+      // the key:value pair
```
robertknight, line 209: IMO this comment is a bit redundant, as the code is very straightforward and the naming makes it clear what is going on.

src/annotator/config/index.js
```diff
@@ -162,8 +198,33 @@ export function getConfig(appContext = 'annotator', window_ = window) {
...
+        config[name] = configDef.defaultValue;
+      }
+      return;
+    }
+
+    // Get the value from the configuration source
     const value = configDef.getValue(settings);
-    // Get the value from the configuration source and run through an optional coerce method
+    if (value === undefined) {
+      // If there is no value (e.g. undefined), then set to the default if provided,
```
robertknight, line 219: As above, I think this comment is a little redundant as it repeats the code.

src/annotator/config/index.js
```diff
@@ -162,8 +198,33 @@ export function getConfig(appContext = 'annotator', window_ = window) {
   let filteredKeys = configurationKeys(appContext);
   filteredKeys.forEach(name => {
     const configDef = configDefinitions[name];
+    const hasDefault = configDef.defaultValue !== undefined; // A default could be null
+    const isURLFromBrowserExtension = isBrowserExtension(
+      urlFromLinkTag(window_, 'sidebar', 'html')
```
robertknight, line 203: When initially reading this I wondered what would happen if the link tag was missing. This triggers an exception and the client will fail to boot.

src/annotator/config/test/index-test.js
```diff
@@ -51,24 +67,117 @@ describe('annotator/config/index', function () {
...
-    it(`reads ${settingName} from the host page, even when in a browser extension`, function () {
-      getConfig('all', 'WINDOW');
-      assert.calledWith(
-        fakeSettingsFrom().hostPageSetting,
-        settingName,
-        sinon.match({ allowInBrowserExt: true })
-      );
+  describe('browser extension', () => {
+    context('when client is loaded from the browser extension', function () {
+      it('returns only config values where `allowInBrowserExt` is true or the defaultValue if provided', () => {
```
robertknight, line 72: Since a `defaultValue` is set for every setting, the case where a config value is not returned is not exercised here. A more accurate description of what is tested currently would be "reads only config values where `allowInBrowserExt` is true, otherwise returns `defaultValue`".

robertknight (commented):
The `defaultValue` field is marked as optional and there are code paths to handle the case where a `defaultValue` is not provided. However a `defaultValue` is set for every setting, so those code paths don't get used. What's the intent here?

Other than that I had a small comment regarding the `isBrowserExtension` local variable renaming and a couple of comments that I think could be omitted. The rest of the changes looked good.

src/annotator/config/index.js
LMS007: Good Q. Previous to any change changes here any missing value got a "null". I'd like to change this to allow any missing values to be missing (e.g. omitted)

However, I don't feel 100% confident we can do that just yet and it needs a bit more investigation. So the stop gap was to simply fixup the code here to allow missing values to be omitted, but then add a `defaultValue: null` to all the values that don't have some other specialized  default.  In other words, its defined for every setting.

The final step would be to back and see how many of the null defaults we could remove.  Perhaps to make the code more clear, we can make `defaultValue` not be optional for now, and when we do that final audit, it can be reverted to optional?

src/annotator/config/index.js
LMS007: Awe, I totally misread that :)
**will change**

src/annotator/config/index.js
LMS007: Sure, but that was always the case, or are you saying this is a difference with this PR?

src/annotator/config/index.js
robertknight: Hi Kyle - This sounds reasonable to me. Can you add a comment in the code that the `defaultValue: null` entries are temporary until you've validated that it can be removed entirely.

src/annotator/config/index.js
robertknight: Oh, I didn't mean there was a change in this PR. I'm just making a note of a query that popped into my head while reading this.

src/annotator/config/index.js
LMS007: I'll leave this for now since these will land all at the same time.

PR #3498, opened 2021-06-10
Refactor annotator/config/ (6/N) 

LMS007:
### Pass config name directly to getValue()

This change fixes the usage of redundant names being expressed in the configDefinitions map.

--------------

relates to 

### DIFF
--- a/src/annotator/config/index.js
+++ b/src/annotator/config/index.js
@@ -5,21 +5,18 @@ import { urlFromLinkTag } from './url-from-link-tag';
 
 /**
  * @typedef {'sidebar'|'notebook'|'annotator'|'all'} AppContext
- *
  * @typedef {import('./settings').SettingsGetters} SettingsGetters
+ * @typedef {(settings: SettingsGetters, name: string) => any} ValueGetter
  *
  * @typedef ConfigDefinition
- * @prop {(settings: SettingsGetters) => any} getValue -
- *  Method to retrieve the value from the incoming source
+ * @prop {ValueGetter} getValue - Method to retrieve the value from the incoming source
  * @prop {boolean} allowInBrowserExt -
  *  Allow this setting to be read in the browser extension. If this is false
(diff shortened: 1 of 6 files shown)


src/annotator/config/index.js
```diff
@@ -90,12 +90,12 @@ const configDefinitions = {
   appType: {
     allowInBrowserExt: true,
     defaultValue: null,
-    getValue: settings => settings.hostPageSetting('appType'),
+    getValue: (settings, name) => settings.hostPageSetting(name),
```
robertknight, line 93: The `(settings, name) => settings.hostPageSetting(name)` closure is duplicated several times. Those could be combined into a single closure, or alternatively you could make `getValue` optional and default to calling `settings.hostPageSetting(name)` if no `getValue` function is provided.

robertknight (changes requested):
I had a suggestion regarding the duplicated `(settings, name) => settings.hostPageSetting(name)` closures.

src/annotator/config/index.js
LMS007: I'd rather not get too cute with implicit options so I'll pass a function to de-duplicate those calls but still make getValue requried.

src/annotator/config/index.js
```diff
@@ -8,9 +8,10 @@ import { urlFromLinkTag } from './url-from-link-tag';
  *
  * @typedef {import('./settings').SettingsGetters} SettingsGetters
  *
+ * @typedef {(settings: SettingsGetters, name: string) => any} ValueGetter
```
robertknight, line 11: A small stylistic note: When there are a group of single-line typedefs I tend to group them together without blank lines in between, just to let a little more useful code fit on a screen.

robertknight (approved):
LGTM with a tiny stylistic request.

src/annotator/config/index.js
LMS007: I'll do that!

PR #3501, opened 2021-06-10
Refactor annotator/config/ (9/N) 

LMS007:
**Remove options.defaultValue from hostPageSetting**

- defaultValue is now handled inside of getConfig so this code is unused.

- hostPageSetting now returns undefined, rather than null, if a value is not found in the config. The undefined value will trigger the defaultValue to take effect in getConfig if a defaultValue is provided.

------

relates to 

### DIFF
--- a/src/annotator/config/settings.js
+++ b/src/annotator/config/settings.js
@@ -84,7 +84,7 @@ export default function settingsFrom(window_) {
   function showHighlights() {
     let showHighlights_ = hostPageSetting('showHighlights');
 
-    if (showHighlights_ === null) {
+    if (showHighlights_ === undefined) {
       showHighlights_ = 'always'; // The default value is 'always'.
     }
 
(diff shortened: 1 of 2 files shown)


src/annotator/config/settings.js
```diff
@@ -129,11 +130,7 @@ export default function settingsFrom(window_) {
       return jsonConfigs[name];
     }
 
-    if (hasDefaultValue) {
-      return options.defaultValue;
-    }
-
-    return null;
+    return undefined;
```
robertknight, line 133: `undefined` is the implicit return value, but returning it explicitly here is good for clarity I think.

src/annotator/config/settings.js
```diff
@@ -118,9 +118,10 @@ export default function settingsFrom(window_) {
     return jsonConfigs.query || queryFromURL();
   }
 
-  function hostPageSetting(name, options = {}) {
-    const hasDefaultValue = typeof options.defaultValue !== 'undefined';
-
+  /**
+   * @param {string} name - Unique name of the setting
```
robertknight, line 122: It would be useful to have a short summary of what sources this function consults and what it returns (the setting value if found or `undefined` otherwise).

robertknight (approved):
I requested some additional JSDoc info. Otherwise LGTM.

robertknight:
I have manually tested this PR which incorporates all the changes in the preceding PRs (https://github.com/hypothesis/client/pull/3495, https://github.com/hypothesis/client/pull/3496, https://github.com/hypothesis/client/pull/3497, https://github.com/hypothesis/client/pull/3498, https://github.com/hypothesis/client/pull/3499, https://github.com/hypothesis/client/pull/3500). I have not tested the earlier PRs in the series manually. I checked the following:

- Test pages in the client's dev server which customize various settings
- Manually changing various settings in `dev-server/templates/client-config.js.mustache` and then checking everything behaved as expected
- That assignments in the LMS app, where the client also has some custom configuration, still work

Things I still need to check:

- The browser extension with a build of the client from this branch

src/annotator/config/settings.js
LMS007: my thoughts too

LMS007:
> * browser extension with a build of the client from this branch

PR #3502, opened 2021-06-11
Add support for `ignoreOtherConfiguration` setting option

esanzgar:
As part of https://github.com/hypothesis/viahtml/pull/192 we decided to introduce a new `ignoreOtherConfiguration` option in the hypothesis settings.

On the first commit, I added support on non-JSON configuration (`window.hypothesisConfig()`)

On the second commit, I added a test case to  check the effect on this new option.

### DIFF
--- a/dev-server/templates/index.mustache
+++ b/dev-server/templates/index.mustache
@@ -24,6 +24,7 @@
     <li><a href="/document/parent-frame">Annotatable iframe test</a></li>
     <li><a href="/document/sidebar-external-container">Sidebar in external container test</a></li>
     <li><a href="/document/multi-frames">Multi-frame test</a></li>
+    <li><a href="/document/ignore-other-configuration"><code>ignoreOtherConfiguration</code> test</a></li>
   </ul>
 
   <h2>Test PDF documents</h2>
--- a/src/annotator/config/settings.js
+++ b/src/annotator/config/settings.js
@@ -1,4 +1,5 @@
 import { parseJsonConfig } from '../../boot/parse-json-config';
+import { toBoolean } from '../../shared/type-coercions';
 
 import configFuncSettingsFrom from './config-func-settings-from';
 import isBrowserExtension from './is-browser-extension';
@@ -20,8 +21,17 @@ import isBrowserExtension from './is-browser-extension';
  * @return {SettingsGetters}
  */
 export default function settingsFrom(window_) {
-  const jsonConfigs = parseJsonConfig(window_.document);
+  // Prioritize the `window.hypothesisConfig` function over the JSON format
+  // Via uses `window.hypothesisConfig` and makes it non-configurable and non-writable.
+  // In addition, Via sets the `ignoreOtherConfiguration` option to prevent configuration merging.
   const configFuncSettings = configFuncSettingsFrom(window_);
+
+  let jsonConfigs;
+  if (toBoolean(configFuncSettings.ignoreOtherConfiguration)) {
+    jsonConfigs = {};
+  } else {
+    jsonConfigs = parseJsonConfig(window_.document);
+  }
 
   /**
    * Return the href of the first annotator link in the given
--- a/dev-server/documents/html/ignore-other-configuration.mustache
+++ b/dev-server/documents/html/ignore-other-configuration.mustache
@@ -1,0 +1,55 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Ignore configuration test</title>
+    <style>
+      body {
+        font-family: sans-serif;
+      }
+    </style>
+
+    <!-- This configuration is ignored -->
+    <script type="application/json" class="js-hypothesis-config">
+      {
+        "theme": "clean",
+        "openSidebar": true
+      }
+    </script>
+
+    <!-- This configuration wins -->
+    <script>
+      Object.defineProperty(window, 'hypothesisConfig', {
+        value: function () {
+          return {
+            openSidebar: false,
+            ignoreOtherConfiguration: true,
+          };
+        },
+      });
+    </script>
+  </head>
+  <body>
+    <h1>Ignore configuration test</h1>
+    <p>
+      <em>viahtml</em> sets the <code>ignoreOtherConfiguration</code> option in
+      the result of <code>window.hypothesisConfig()</code>. This option
+      disregards other sources of Hypothesis configuration that the host page
+      could include.
+    </p>
+    <p>
+      In addition, <em>viahtml</em> makes
+      <code>window.hypothesisConfig</code> non-configurable and non-writable.
+      These two settings together makes the client consistently behave on
+      <em>via</em> (PDF or html).
+    </p>
+    <p>
+      If this test works, it should make the sidebar remain closed and use the
+      classic theme. The options
+      <code>openSidebar : true</code> and <code>theme: "clean"</code>
+      should be ignored.
+    </p>
+    {{{hypothesisScript}}}
+  </body>
+</html>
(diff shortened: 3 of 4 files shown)


src/annotator/config/settings.js
```diff
@@ -20,9 +21,21 @@ import isBrowserExtension from './is-browser-extension';
  * @return {SettingsGetters}
  */
 export default function settingsFrom(window_) {
-  const jsonConfigs = parseJsonConfig(window_.document);
+  // Prioritize the `window.hypothesisConfig` function over the JSON format
+  // Via uses `window.hypothesisConfig` and makes it non-configurable and non-writable.
+  // In addition, `via` sets the `ignoreOtherConfiguration` option to prevent configuration merging.
```
robertknight, line 26:
```suggestion
  // In addition, Via sets the `ignoreOtherConfiguration` option to prevent configuration merging.
```

robertknight (commented):
The changes related to `window.hypothesisConfig` look good. Regarding JSON script tags, I think we should leave that out for the moment. There are multiple ways that we might want the flag to work in that context (apply only the configuration from the JSON tag containing `ignoreOtherConfiguration`, apply the cumulative configuration up until the `ignoreOtherConfiguration` flag is encountered as you've done here) and I think it would be wise to wait for a concrete use case to inform which of them to implement.

esanzgar:
> Regarding JSON script tags, I think we should leave that out for the moment.

I agree with you. I dropped the second commit and instead I added a test scenario to check this new option.

dev-server/documents/html/configurations.mustache
```diff
@@ -0,0 +1,48 @@
...
+          };
+        },
+      });
+    </script>
+  </head>
+  <body>
+    <h1>Ignore configuration test</h1>
+    <p>
+      `viahtml` sets the `ignoreOtherConfiguration` option on
+      `window.hypotesysConfig`. This option disregards other JSON hypothesis
```
robertknight, line 33:
```suggestion
      `window.hypothesisConfig`. This option disregards other sources of Hypothesis
```

I changed "JSON" to "sources" here since the point is that it doesn't matter what kind of configuration is used elsewhere, it just ignores them all.

dev-server/documents/html/configurations.mustache
```diff
@@ -0,0 +1,48 @@
...
+      configurations that the host page could include.
```
robertknight, line 34:
```suggestion
      configuration that the host page could include.
```

dev-server/documents/html/configurations.mustache
```diff
@@ -0,0 +1,48 @@
...
+    </p>
+    <p>
+      In addition, `viahtml` makes `window.hypotesysConfig` non-configurable and
```
robertknight, line 37:
```suggestion
      In addition, `viahtml` makes `window.hypothesisConfig` non-configurable and
```

dev-server/documents/html/configurations.mustache
```diff
@@ -0,0 +1,48 @@
...
+      non-writable. These two settings together makes the client consistenly
+      behave on `via` (PDF or html).
+    </p>
+    <p>
+      If this test works, it should make the sidebar to remain closed and to
+      have the classic theme. The configurations `"openSidebar" : true` and
```
robertknight, line 43:
```suggestion
      If this test works, it should make the sidebar remain closed and
      use the classic theme. The options `"openSidebar" : true` and
```

src/annotator/config/settings.js
```diff
@@ -20,9 +21,21 @@ import isBrowserExtension from './is-browser-extension';
...
-  const jsonConfigs = parseJsonConfig(window_.document);
+  // Prioritize the `window.hypothesisConfig` function over the JSON format
+  // Via uses `window.hypothesisConfig` and makes it non-configurable and non-writable.
+  // In addition, Via sets the `ignoreOtherConfiguration` option to prevent configuration merging.
   const configFuncSettings = configFuncSettingsFrom(window_);
 
+  let jsonConfigs;
+  if (
+    configFuncSettings.hasOwnProperty('ignoreOtherConfiguration') &&
+    toBoolean(configFuncSettings.ignoreOtherConfiguration) === true
```
robertknight, line 32: Assuming that `toBoolean` always returns a boolean as the name implies the `=== true` part is not strictly needed. Also it it casts undefined => boolean then the `hasOwnProperty` part is not strictly needed either.

dev-server/documents/html/configurations.mustache
```diff
@@ -0,0 +1,48 @@
...
+            ignoreOtherConfiguration: true,
+          };
+        },
+      });
+    </script>
+  </head>
+  <body>
+    <h1>Ignore configuration test</h1>
+    <p>
+      `viahtml` sets the `ignoreOtherConfiguration` option on
```
robertknight, line 32: The dev server doesn't process markdown in in templates so you'll need to use `<code>someSymbol</code` rather than backticks.

dev-server/documents/html/configurations.mustache
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html>
```
robertknight, line 1: This template has a rather generic name (`configurations.mustache`). I suggest something like `ignore-other-configuration.mustache` would leave room for other configuration-related tests in future.

dev-server/templates/index.mustache
```diff
@@ -24,6 +24,7 @@
     <li><a href="/document/parent-frame">Annotatable iframe test</a></li>
     <li><a href="/document/sidebar-external-container">Sidebar in external container test</a></li>
     <li><a href="/document/multi-frames">Multi-frame test</a></li>
+    <li><a href="/document/configurations">Ignore configuration test</a></li>
```
robertknight, line 27:
```suggestion
    <li><a href="/document/configurations"><code>ignoreOtherConfiguration</code> test</a></li>
```

I think it might be hard for a new developer to understand what "ignore configuration" means. Naming it the way I've suggested here would hopefully make it more obvious that it refers to a specific configuration option.

dev-server/documents/html/configurations.mustache
robertknight, line 32: Including this context about the purpose and usage of the option is very helpful.

dev-server/documents/html/configurations.mustache
```diff
@@ -0,0 +1,48 @@
...
+  <body>
+    <h1>Ignore configuration test</h1>
+    <p>
+      `viahtml` sets the `ignoreOtherConfiguration` option on
+      `window.hypotesysConfig`. This option disregards other JSON hypothesis
+      configurations that the host page could include.
+    </p>
+    <p>
+      In addition, `viahtml` makes `window.hypotesysConfig` non-configurable and
+      non-writable. These two settings together makes the client consistenly
```
robertknight, line 38:
```suggestion
      non-writable. These two settings together makes the client consistently
```

dev-server/documents/html/configurations.mustache
robertknight, line 33:
```suggestion
      `viahtml` sets the `ignoreOtherConfiguration` option in the result of
      `window.hypothesisConfig()`. This option disregards other JSON hypothesis
```

To be precise, `hypothesisConfig` is a function that returns an object, rather than an object itself as the previous wording might have suggested.

robertknight (changes requested):
This looks good from a functional point of view. I requested some corrections on the wording and formatting of the test page.

dev-server/documents/html/configurations.mustache
esanzgar: Sure.

src/annotator/config/settings.js
esanzgar: That's true. The only reason to use `hasOwnProperty` would be to avoid prototype pollution.

esanzgar:
@robertknight Thanks for your suggestions.

I have incorporated all of them.

robertknight (approved):
LGTM.

PR #3532, opened 2021-06-21
Modernize WebSocket wrapper

robertknight:
Modernize our auto-reconnecting wrapper for `WebSocket` to match current
conventions in the client and improve the documentation.

 - Use named rather than default exports
 - Convert functions to arrow functions and remove unneeded `self` alias.
 - Add JSDoc comments and types
 
_Added since first review:_

- Add three additional tests
- Improve documentation for internal `operation` variable used for automatic reconnections
- Remove a code path that could not be triggered related to calling `connect` while already connected

A circular reference between the `connect` and `onAbnormalClose` functions was broken by adding a second argument to `onAbnormalClose`.

### DIFF
--- a/src/sidebar/services/streamer.js
+++ b/src/sidebar/services/streamer.js
@@ -2,7 +2,7 @@ import * as queryString from 'query-string';
 
 import warnOnce from '../../shared/warn-once';
 import { generateHexString } from '../util/random';
-import Socket from '../websocket';
+import { Socket } from '../websocket';
 import { watch } from '../util/watch';
 
 /**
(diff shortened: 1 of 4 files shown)


robertknight:
This PR highlighted a few missing tests in this code. I'll address that.

LMS007:
Rob, changes here seem straightforward, I think to fix the coverage, we can send multiple  `fakeSocket.onclose({ code: CLOSE_ABNORMAL });` events in a row and it should break into the condition where `operation` is not null in onAbnormalClose().

I suppose we could also add a test for `socket.onerror` and get that line covered too.

robertknight:
I have updated this PR to add missing tests for the `Socket` class, as well as improve some internal documentation and remove a code path that could not be triggered given the current way `Socket` works. See the three most recent commits for details.

LMS007 (approved):
Thanks for adding the tests! This looks ready to merge.

PR #3537, opened 2021-06-23
Enable MessageChannel for host <--> sidebar communication

esanzgar:
- [ ] remove `Channel` suffix from the channel types
- [ ] modify `cross-frame` service to enable `MessageChannel` communication in the `host` frame
- [ ] modify `frame-sync` service to enable `MessageChannel` communication in the `sidebar` frame
- [ ] other trivial refactors

### DIFF
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -25,7 +25,7 @@ import Notebook from './notebook';
 import Sidebar from './sidebar';
 import { EventBus } from './util/emitter';
 
-const window_ = /** @type {HypothesisWindow} */ (window);
+const _window = /** @type {HypothesisWindow} */ (window);
 
 // Look up the URL of the sidebar. This element is added to the page by the
 // boot script before the "annotator" bundle loads.
@@ -51,11 +51,11 @@ function init() {
   if (annotatorConfig.subFrameIdentifier) {
     // Other modules use this to detect if this frame context belongs to hypothesis.
     // Needs to be a global property.
-    Object.defineProperty(window_, '__hypothesis_frame', { value: true });
+    Object.defineProperty(_window, '__hypothesis_frame', { value: true });
     hostPort = null;
   }
 
-  const isPDF = typeof window_.PDFViewerApplication !== 'undefined';
+  const isPDF = typeof _window.PDFViewerApplication !== 'undefined';
 
   let sidebar = /** @type {Sidebar?} */ (null);
   let notebook = /** @type {Notebook?} */ (null);
@@ -68,7 +68,7 @@ function init() {
     documentType: isPDF ? 'pdf' : 'html',
   });
 
-  if (!window_.__hypothesis_frame) {
+  if (!_window.__hypothesis_frame) {
     frameConnector.listen();
 
     sidebar = new Sidebar(document.body, eventBus, guest, sidebarConfig);
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -17,6 +17,8 @@ import { registerIcons } from '@hypothesis/frontend-shared';
 import iconSet from './icons';
 registerIcons(iconSet);
 
+import { FrameConnector } from '../shared/communicator';
+
 import { getConfig } from './config/index';
 import Guest from './guest';
 import Notebook from './notebook';
@@ -37,12 +39,9 @@ function init() {
   const annotatorConfig = getConfig('annotator');
   const isPDF = typeof window_.PDFViewerApplication !== 'undefined';
 
-  if (annotatorConfig.subFrameIdentifier) {
-    // Other modules use this to detect if this
-    // frame context belongs to hypothesis.
-    // Needs to be a global property that's set.
-    window_.__hypothesis_frame = true;
-  }
+  let frameConnector = /** @type {FrameConnector?} */ (null);
+  let sidebar = /** @type {Sidebar?} */ (null);
+  let notebook = /** @type {Notebook?} */ (null);
 
   const eventBus = new EventBus();
   const guest = new Guest(document.body, eventBus, {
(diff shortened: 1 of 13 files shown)


src/sidebar/components/SidebarView.js
```diff
@@ -91,9 +91,20 @@ function SidebarView({
...
     const framePort = new PortFinder();
+
+    // Enables communication between `host` and `sidebar` frames
+    framePort
+      .discover({
+        channel: 'hostToSidebar',
+        hostFrame: window.parent,
+        port: 'sidebar',
+      })
+      .then(port => frameSync.connect(port));
```
robertknight, line 102: Might make sense to move the `framePort` here into the `FrameSync` service to keep the pattern that all communication from parts of the sidebar to the host page and guests goes via that service. Alternatively we could have separate services for the sidebar to communicate with guests vs. communicating with non-guests.

src/sidebar/services/frame-sync.js
```diff
@@ -223,7 +225,7 @@ export class FrameSyncService {
     const addFrame = channel => {
       channel.call('getDocumentInfo', (err, info) => {
         if (err) {
-          channel.destroy();
+          channel.destroy(); // TODO: I am not sure if this works with `MessagePort`
```
robertknight, line 228: I think the idea here is that if the sidebar is unable to get the URL and metadata for a guest then it stops communicating with it and doesn't try to fetch annotations etc. I think that makes sense. For a `MessagePort` all we'd need to do is call `close()` on it.

src/annotator/index.js
```diff
@@ -36,32 +36,38 @@ const appLinkEl = /** @type {Element} */ (
...
     // Load the PDF anchoring/metadata integration.
     // nb. documentType is an internal config property only
     documentType: isPDF ? 'pdf' : 'html',
   });
 
-  if (annotatorConfig.subFrameIdentifier) {
-    // Other modules use this to detect if this
-    // frame context belongs to hypothesis.
-    // Needs to be a global property that's set.
-    window_.__hypothesis_frame = true;
```
robertknight, line 57: I would really like to get rid of this global. From what I can see it is used to detect whether the client has already been loaded into a frame. There are other ways we could do that, such as checking for the `<link>` tags that the boot script adds to the page.

robertknight (commented):
I left a few comments on parts of this. As mentioned [in Slack](https://hypothes-is.slack.com/archives/C1M8NH76X/p1624447473075900?thread_ts=1624442390.071000&cid=C1M8NH76X) after looking at this I think perhaps a first step that might help is to split up or replace `CrossFrame`. Currently there is a single object that is used to communicate between different parties (guest <-> host, guest <-> sidebar, sidebar <-> host) so you can't easily change the communication between two of those parties without affecting the other. There is also the ugliness where the `Guest` ends up creating an object that is used for host <-> sidebar communication. If we could tease apart this communication channels then they could be changed individually.

src/annotator/index.js
```diff
@@ -36,32 +36,38 @@ const appLinkEl = /** @type {Element} */ (
 
 function init() {
   const annotatorConfig = getConfig('annotator');
-  const isPDF = typeof window_.PDFViewerApplication !== 'undefined';
+  const sidebarConfig = getConfig('sidebar');
+  const { sidebarAppUrl } = sidebarConfig;
+
+  const frameConnector = new FrameConnector({
+    sidebarAppUrl,
+  });
```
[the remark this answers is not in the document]
esanzgar: Yes, I noticed too. Well spotted. However, because only the `frameConnector.listen()` is called in the `host` frame the extra ports created are not been used in the `guest` frames.

However, I will address this more cleanly later.

src/annotator/index.js
```diff
@@ -36,32 +36,38 @@ const appLinkEl = /** @type {Element} */ (
...
+  let hostPort = frameConnector.getPort({
+    channel: 'hostToSidebar',
+    port: 'host',
+  });
```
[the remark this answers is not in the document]
esanzgar: Thanks, I will rework this code.

src/annotator/index.js
```diff
@@ -36,32 +36,38 @@ const appLinkEl = /** @type {Element} */ (
...
+
+  const isPDF = typeof _window.PDFViewerApplication !== 'undefined';
 
-  let frameConnector = /** @type {FrameConnector?} */ (null);
   let sidebar = /** @type {Sidebar?} */ (null);
   let notebook = /** @type {Notebook?} */ (null);
 
   const eventBus = new EventBus();
-  const guest = new Guest(document.body, eventBus, {
+  const guest = new Guest(document.body, eventBus, hostPort, {
```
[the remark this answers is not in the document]
esanzgar: Correct. However the `hostPort` is used later in the annotator `index.js` to create the channel from the port.

src/annotator/index.js
esanzgar: Sure.

src/annotator/index.js
```diff
@@ -71,6 +77,11 @@ function init() {
     notebook = new Notebook(document.body, eventBus, getConfig('notebook'));
   }
 
+  // This enables the communication channel between `host` <-> `sidebar or
+  // `guest` <-> `sidebar` frames. It needs to happen after the `Sidebar` is
+  // created because `Sidebar` registers event listeners using `guest.crossframe`.
```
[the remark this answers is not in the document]
esanzgar: The annotator functionality is very similar if it is embedded in the `host` or `guest` frames.

src/sidebar/components/SidebarView.js
esanzgar: I totally agree.

src/sidebar/services/frame-sync.js
esanzgar: Thanks for the clarification.

esanzgar (commented):
Thanks @robertknight  your comments are very helpful and I will implement them.

PR #3552, opened 2021-07-01
Remove showHighlights from settings

LMS007:
_Updated_

Add new `toShowHighlights` coerce method and apply that to the showHighlights config definition in config.index.js

Note. There is a small change to how the new coerce method works and the older settings.showHighlights getter worked. The old getter allowed any type to be passed through if it was not a boolean, whereas the new coerce method is much more strict and only returns the allowed string values or undefined

Additionally, there is added logic to getConfig after the coerce method runs to check for undefined and set a default is necessary. Currently, only the showHighlights coerce method will return and undefined. A future change could replace this with validation

------

This is one of the last clean-up details with the previous config refactor. This change simply removes the showHighlights getter and replaces it with a coerce method similar to how `openSidebar` is defined.

The rationale for making the new coerce more strict (e.g., i won't freely pass any value back) is that we don't have any cases were we need to pass a non-string and it only made the tests confusing. I feel like being a bit more strict with our values will be a benefit in the long term.


src/shared/type-coercions.js
```diff
@@ -67,3 +67,19 @@ export function toString(value) {
...
 }
+
+/**
+ * Coerce incoming value to appropriate `showHighlights` string values.
+ *   - true => 'always' OR
+ *   - false => 'never' OR
+ *   - the passed in value as a string
+ *
+ * @param {any} value - initial value
+ */
```
robertknight, line 78: An `@return {...}` would be useful here to specify the allowed results.

src/shared/type-coercions.js
```diff
@@ -67,3 +67,19 @@ export function toString(value) {
...
+export function toShowHighlights(value) {
+  // Convert legacy keys/values to corresponding current configuration.
+  if (typeof value === 'boolean') {
+    return value ? 'always' : 'never';
+  }
+  return toString(value);
```
robertknight, line 84: This is consistent with the previous behavior of the `showHighlights` getter, but not ideal IMO as the type of the `showHighlights` config setting is logically `'always'|'never'|'whenSidebarOpen'`. From an end user's point of view, I think the most useful behavior would be to:

1. Convert booleans to 'always'/'never'
2. Return valid enum values ('always', 'never', 'whenSidebarOpen') unchanged
3. Return 'always' if the config value is not specified (I think this is handled elsewhere via `defaultValue`)?
4. Return 'always' for any other value, but log a warning to inform the user that the value is not valid

One possible approach might be to implement this in a generic way by having the coerce methods throw a specific for invalid values (`InvalidValueError`?) which are then caught in higher level code, causing a fallback to the default value.

If you agree, then this could perhaps be split into two phases: 1) In this PR, just make `toShowHighlights` silently fall back to `always` if the value is not valid. 2) In a follow-up, introduce a mechanism to handle coercion errors more generically.

robertknight (commented):
The change in this PR makes sense in that it moves coercion logic to the correct place and matches the previous behavior. However I think this is a good time to revisit the previous behavior as IMO it wasn't useful in the event that the value was invalid. Please have a look at my comment and let me know how you want to move forwards.

src/shared/type-coercions.js
LMS007: I think we're trending towards validation which admittedly is needed.  I have an older PR with proposed validation methods but we never got around to evaluating it. It was which was originality for the sidebar, but could work for both here and the sidebar eventually.

https://github.com/hypothesis/client/commit/7537d3aba56ab9f297074631b9141eeb8e61aa4d

(take a look at the tests)

It's a bit too robust just a simple string, but it could be used to implement a thorough validation schema for our config. If I recall correctly, it can do multiple types with various conditions, even  objects and arrays. So in this case case, we could pass a boolean or a limited set of strings, and report a console error otherwise.

I think we may need to take a step back and think how we might include validation on all the config values because I think there is a lot of upside to reporting config problems to users as you just pointed out above. A simple console warn or error might be sufficient and perhaps if a config value fails validation, we could remove its value safely from the config

The big question is at what point to we validate vs coerce? I suspect doing the validation first might be the best way as we can report directly on the value that the user provides.

For now, I can do exactly what you suggested, we'll make the coerge silently mutate to "always", but then we should follow up with a plan for full config validation.  I think there especially may be value in report errors when config values need to be structured objects.

robertknight: > The big question is at what point to we validate vs coerce? I suspect doing the validation first might be the best way as we can report directly on the value that the user provides.

My general thought is that being too liberal with coercing values can be problematic because it is hard to know how much of that liberalness users are relying on and what might break if it becomes stricter in future. For that reason I favor being quite strict ie. doing mostly validation with implicit coercion limited to very specific cases, if any.

> For now, I can do exactly what you suggested, we'll make the coerge silently mutate to "always", but then we should follow up with a plan for full config validation. I think there especially may be value in report errors when config values need to be structured objects.

In this case any invalid values will have the same effect as not supplying the value at all, since "always" is the default value. This is what we would do if we only had validation and a policy of ignoring invalid values. This sounds good to me, as does opening an issue to implement more complete validation in future.

LMS007: gotcha, I follow

LMS007:
- Updated toShowHighlights to work in more strict fashion. It will return undefined when the value does match what is expected, from there it will not be set to the config and a default will be used per config definition.
 
- Updated getConfig to honor undefined values from coerce methods and set a default if needed. Note we need to check for the undefined type after getting the 1. original value and then again after the 2. coerce method.  We can't simply do both and run a single check because for example, an invalid value may mutate to a blank string and then it won't correctly set the default because "" is not undefined.  

- Added additional comments in the tests to explain that.

robertknight (commented):
I've had a look through this. We have got to a situation where the way that different functions handle invalid values in `type-coercions.js` is different. Most of the existing functions don't support indicating that the provided value was invalid in any way, whereas `toShowHighlights` now does. `toInteger` can return `NaN` which effectively means undefined, but this is different than `toShowHighlights`.

The preference that I'm leaning towards is to stop doing any coercion in the client, except for backwards compatibility (eg. boolean => enum value for `showHighlights`) and instead require all callers to just provide data of the expected type. This will require us to make changes to h-vialib and roll it out in via and viahtml first.

We're going to need some interim step to resolve this PR though. I'll think on this a little more and get back to you.

src/annotator/config/index.js
```diff
@@ -204,13 +205,20 @@ export function getConfig(appContext = 'annotator', window_ = window) {
       urlFromLinkTag(window_, 'sidebar', 'html')
     );
 
+    /**
+     * Closure helper to set an optional default
+     */
+    function setDefault(name) {
+      if (hasDefault) {
+        config[name] = configDef.defaultValue;
```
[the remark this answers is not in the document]
LMS007: Yep, that makes more sense!

PR #3553, opened 2021-07-02
Add `Escape` shortcut to hide adder

LMS007:
The adder now hides when the escape key is pressed. (WCAG 2.1 criterion 1.4.13)

---------

Note: Low priority

fixes https://github.com/hypothesis/product-backlog/issues/1128

relates to https://github.com/hypothesis/product-backlog/issues/1114

### DIFF
--- a/src/annotator/test/adder-test.js
+++ b/src/annotator/test/adder-test.js
@@ -75,7 +75,7 @@ describe('Adder', () => {
     assert.exists(shadowRoot.querySelector('.AdderToolbar'));
   });
 
-  describe('button handling', () => {
+  describe('button and shortcut handling', () => {
     const getButton = label =>
       getContent(adder).querySelector(`button[title^="${label}"]`);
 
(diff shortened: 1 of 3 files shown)


src/annotator/test/adder-test.js
```diff
@@ -154,6 +154,15 @@ describe('Adder', () => {
       assert.called(adderCallbacks.onShowAnnotations);
     });
 
+    it('hides the adder when `Escape` shortcut is pressed', () => {
+      adder.annotationsForSelection = ['ann1'];
+      showAdder();
+      triggerShortcut('Escape');
```
robertknight, line 160: We're not terribly consistent about this in some test suites, but as a rule of thumb once tests exceed 4 or 5 lines I find it helps readability to separate the setup/action being tested/assertions (aka Arrange/Act/Assert), assuming there is a clear separation between these phases. In this case, that means adding new blank lines before and after the `triggerShortcut` call.

robertknight (approved):
LGTM. I had a look around for examples of similar behavior in other apps and found that Google Docs has similar behavior in the popover it displays when text with a spelling error or link is selected.

src/annotator/test/adder-test.js
LMS007: roger that

PR #3554, opened 2021-07-02
Remove aria-label from TagEditor and AnnotationPublishControl button

LMS007:
Per WCAG 2.1 criterion: 2.5.3 Label in Name--we should not differentiate between the aria-label and the visual name in an input element. In the case of the TagEditor, the `placeholder` and `aria-label` were different. The simple fix is to remove `aria-label`.

The second problem is with AnnotationPublishControl's button where the `aria-label` and button's text did not match. Again, the simple fix is to remove `aria-label`.

-----

In both cases I have tested with Mac voiceover and the button's text and input's placeholder are read out as expected.
fixes https://github.com/hypothesis/product-backlog/issues/1127

relates to https://github.com/hypothesis/product-backlog/issues/1114

### DIFF
--- a/src/sidebar/components/Annotation/AnnotationPublishControl.js
+++ b/src/sidebar/components/Annotation/AnnotationPublishControl.js
@@ -85,7 +85,6 @@ function AnnotationPublishControl({
           style={buttonStyle}
           onClick={onSave}
           disabled={isDisabled}
-          title={`Publish this annotation to ${publishDestination}`}
           size="large"
           variant="primary"
         >
--- a/src/sidebar/components/Annotation/test/AnnotationPublishControl-test.js
+++ b/src/sidebar/components/Annotation/test/AnnotationPublishControl-test.js
@@ -78,7 +78,7 @@ describe('AnnotationPublishControl', () => {
   });
 
   const getPublishButton = wrapper =>
-    wrapper.find('LabeledButton[title^="Publish this annotation"]');
+    wrapper.find('LabeledButton.PublishControlButton');
 
   describe('theming', () => {
     it('should apply theme styles', () => {
@@ -102,10 +102,7 @@ describe('AnnotationPublishControl', () => {
         const wrapper = createAnnotationPublishControl();
 
         const btn = getPublishButton(wrapper);
-        assert.equal(
-          btn.prop('title'),
-          `Publish this annotation to ${fakeGroup.name}`
-        );
+        assert.equal(btn.text(), `Post to ${fakeGroup.name}`);
       });
     });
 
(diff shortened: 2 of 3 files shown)


src/sidebar/components/Annotation/AnnotationPublishControl.js
```diff
@@ -85,7 +85,6 @@ function AnnotationPublishControl({
           style={buttonStyle}
           onClick={onSave}
           disabled={isDisabled}
-          title={`Publish this annotation to ${publishDestination}`}
```
robertknight, line 88: I think the origin of the problem here was that the `title` attribute was intended to set the tooltip originally, not the accessible label (`aria-label`). In that situation having a tooltip which was more verbose than the button label was OK, eg. to provide an extended description of the action that couldn't fit in the button text. All of the `*Button` components in `@hypothesis/frontend-shared` currently always set `aria-label` to match `title`. 

In this particular scenario, the tooltip doesn't provide much additional value over the button text IMO so I think it's OK to just drop it. I wonder however if there are any scenarios where an extended tooltip does provide significant value over the button text and which we'll want to expose some other way.

I think we might have had this discussion in the past, but if we have I've forgotten the conclusion. CC-ing @lyzadanger in case she knows more.

robertknight (commented):
The specific changes in this PR look reasonable to me. The change to the "Post to <group>" tooltip does make me wonder about the generic situation when a button with a short text label (eg. due to space constraints) is paired with a tooltip that provides an extended description of the action that may provide helpful, though not essential, context to users. I can't remember if we looked at this in the past and came to a conclusion already. I've pinged Lyza in case we have.

src/sidebar/components/Annotation/AnnotationPublishControl.js
LMS007: Yep, I had this thought too, lets wait to see what Lyza thinks. Because as the spec is written it sorta means that title can't be used with button text--or thats how I at least read it.

More details here:

https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html

lyzadanger (approved):
I've given the referenced success criterion documentation a read. The gist being: your accessible label (they call it _name_ but I find that a confusing term) should match your visible label. That seems like a reasonable assertion.

I think it's OK to proceed here with this change to this one component, without too much fuss. I think the issue for us in the future is more around the general UI for this component—I think it could be clearer for all users, but we don't need to open that particular can of worms just now. That is, the problem is that the current visible "Publish..." label and drop-down "privacy" selection might not be as clear as it could be, so we felt it necessary in the past to beef it up with `title` text.

PR #3568, opened 2021-07-08
Convert remaining anonymous callbacks to arrow functions

robertknight:
We've been gradually converting anonymous functions to arrow functions
since adopting ES6. This commit uses ESLint to finish the process by
converting the remaining non-named, non-`this`-using callbacks to arrow
functions.

 - Enable `prefer-arrow-callbacks` lint rule. The `allowNamedFunctions`
   option is enabled because we have a few instances (eg. in
   gulpfile.js) of explicitly naming functions for use in debugging /
   logging etc.
 - Run `eslint --fix .` and `yarn format` to automatically fix up the
   code

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,8 @@
 {
   "extends": ["hypothesis", "plugin:jsx-a11y/recommended"],
   "rules": {
+    "prefer-arrow-callback": ["error", { "allowNamedFunctions": true }],
+
     // Suppressed to make ESLint v6 migration easier.
     "no-prototype-builtins": "off",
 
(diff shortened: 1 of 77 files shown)


lyzadanger (approved):
> Run eslint --fix . and yarn format to automatically fix up the
code

Smart. Assuming we can trust the machines, and the codecov shortfall is meaningless in this situation (I believe it is), this looks GTG.

esanzgar:
It would be nice to move this rule to the `eslint-config-hypothesis` package.

robertknight:
> It would be nice to move this rule to the eslint-config-hypothesis package.

I agree. My plan was to make the client and lms codebases conform to these rules first, just to make sure it didn't cause any unexpected problems, and then add the rule to the shared package.

PR #3569, opened 2021-07-08
Always use object shorthand for properties

robertknight:
Enforce the use of object shorthand syntax everywhere for consistency.
eg. `{ foo }` instead of `{ foo: foo }`. We had already converted most
occurrences manually.

For the moment this only applies to properties and not methods, to avoid a conflict with https://github.com/hypothesis/client/pull/3568. (ie. `{ foo: function () { ... } }` is not converted to `{ foo() { ... } }`.)


lyzadanger (approved):
LGTM, especially considering there was an automated tool involved.

esanzgar:
It would be nice to move this rule to the `eslint-config-hypothesis` package.

Issue #2798, opened 2020-12-07
Add validation to incoming config values


LMS007:
Many of the values from the host page or from a via proxy need validation to produce relatvent warnings for the site owner which may otherwise prevent some features from working. The validated values must support a variety of types including objects and arrays. Our current validation/translation checking process is incomplete and inconsistent. 

**Currently**

- All values are nulled if they are not set.
- Some values are translated that come from via because it only outputs strings.
- Some values are just validated, but others are not.

![96653176-59519500-12ed-11eb-932f-4bffd46c7675](https://user-images.githubusercontent.com/3939074/101392164-403f7c00-387a-11eb-8ac8-224a748bed2a.png)

**Specifically we need to** 

- [ ]  Build a validation module that can take a schema and produce errors from a given data set against that schema
- [ ]  Don't set a value to null if its missing, just don't include it. (explore if this can be done)
- [ ]  Build a schema and validate values coming into the host page (adder)
- [ ]  Build a schema and validate values again coming into the sidebar iframe
- [ ]  Design a process for versioning the config so we don't have to support too many different versions of values
- [ ]  Translate validated values from via that might need to be move to another type for client processing. e.g. `"3"` to `3`

Relates to https://github.com/hypothesis/client/issues/1968

PR #3570, opened 2021-07-08
Proposal for validation for annotator config

LMS007:
This is a branch based off https://github.com/hypothesis/client/pull/3552 and attempts to solve the issues there in a different way. This is an alternative to the previous PR which instead of creating a complex coerce method for `showHighlights` which attempted to account for invalid values, this instead  injects a validator into the config pipeline so that the coerce methods need only deal with valid values.

_This is a huge PR, so I'll try and just highlight the bits that we need to focus on. Don't worry, most of this is irrelevant for now!_

I included the entire validation file I wrote a while back for the sidebar but only as a convenience for this draft.  Currently only a handful of methods in that file are used in this PR,  and the rest is dead code (for now).  I didn't trim down that file because that would be have been _more_ work and this is just a draft to explore the idea. However, I see enormous value in implementing more validators both in the annotator/config.index.js file and again in the sidebar and possibility in other apps too (notebook , etc). For now, don't spent too much time on the type-validation.js file and just focus on the config index.js file. The main goal of this draft is to see if things are easy to read, write and logically make sense in `getConfig()` and the `configDefinitions`

Relates to https://github.com/hypothesis/client/issues/2798


src/annotator/config/index.js
```diff
@@ -160,7 +164,12 @@ const configDefinitions = {
...
     allowInBrowserExt: false,
     defaultValue: 'always',
-    getValue: settings => settings.showHighlights,
+    validate: validators.oneOfType([
+      validators.isString.oneOf(['always', 'never', 'whenSidebarOpen']),
+      validators.isBoolean, // allow booleans for backwards compatibility
+    ]),
+    coerce: toShowHighlights, // transforms booleans to strings
+    getValue: getHostPageSetting,
   },
```
LMS007, line 173: The above code is the important thing for this draft. Do we want to add an optional "validate" method. If we did , it could look something like this for `showHighlights`. I'm also sure we can also immediately find other uses for it for just about every single config value.

The nice thing about doing this is it really separates the logic for invalid values from the user from coerced values that the app needs internally. In other words, the user does not care about showHighlights being a string internally, all they need to know is it the value they provided is a problem or not.  This gives them that feedback.

This also alleviates the messiness of the coerce methods also acting as pseudo validators.  Follow up work would be to further simply the existing coerce methods because if we had validation in front of them, then we can assume very strict incoming param values.

src/annotator/config/index.js
```diff
@@ -204,29 +213,36 @@ export function getConfig(appContext = 'annotator', window_ = window) {
...
+    // If there is a validator, run that next
+    const validationResult = configDef.validate
+      ? configDef.validate(value)
+      : true;
+    if (validationResult === true) {
+      // Finally, run the value through an optional coerce method
+      config[name] = configDef.coerce ? configDef.coerce(value) : value;
+    } else {
+      // Report the error to the user
+      console.error(
```
LMS007, line 241: So now we have a means of informing uses when bunk values are passed. In these cases, either the config will be omitted from the result, or if a default is set.

src/annotator/config/test/index-test.js
```diff
@@ -4,10 +4,23 @@ import { $imports } from '../index';
...
   beforeEach(() => {
+    sinon.stub(console, 'error'); // Validators report to console
+
+    fakeHostPageSettings = sinon.stub();
+    // default hostPageSetting to return "fakeValue"
+    fakeHostPageSettings
+      .returns('fakeValue')
+      // showHighlights has a validator so set its return value to a valid value
+      .withArgs('showHighlights')
+      .returns('always');
```
LMS007, line 18: Now with a validator, "fakeValue" won't work anymore!

src/shared/type-validation.js
```diff
@@ -0,0 +1,327 @@
+/**
```
LMS007, line 1: ignore file for now

src/shared/test/type-validation-test.js
```diff
@@ -0,0 +1,570 @@
+import { validators } from '../type-validation';
```
LMS007, line 1: ignore file for now

PR #3576, opened 2021-07-12
Fix tag service to work with non-English words

LMS007:
This was filed as a ticket relating to auto complete not working for the word "ინგლისური" (which means "English", in Georgian) 

fixes https://github.com/hypothesis/support/issues/211

**Previous behavior**
Jul-12-2021 14-09-25

### Update:

I have reworked this PR per Rob's suggestion. First split the tag by words `/W+`, then check if the query matchs any of the the start of those words. 

tag: `foo bar baz`
results:

"fo" => YES
"ba"  => YES
"baz"  => YES
"az" => NO
 
 That will ensure everything that previously worked also still works. The changed bit, is there is now a second check to see if the query matches the start of the entire tag without splitting words.  This ensures that for ASCII chars, we don't have a false negative which was exactly what the ticket demonstrated.

tag: `ინგლისური`
results:

"ინ" => YES // this case fails on master
"ნგ" => NO

However One thing that this change won't cover is cases there we have non [a-z] chars which have a delimiter such as a space.

So for example, if this was the tag
`ფუ ბარი`

And this was the query "ბარი"

It would not return that result.  I view this as a minor issue as a user could still search the first part of the tag "ფუ" and it would work. The only solution I can come up for this specific edge case is to actually split on specific chars rather than `/W+` and we would have to come up with that set of chars. I think perhaps this is a better vs best scenario were we can close out this ticket with this solution and maybe add a todo in the code for the improvement. Thoughts?

### DIFF
--- a/src/sidebar/services/tags.js
+++ b/src/sidebar/services/tags.js
@@ -41,12 +41,15 @@ export class TagsService {
     //   (e.g. tag "pink banana" matches query "ban"), OR
     // * tag has substring query occurring after a non-word character
     //   (e.g. tag "pink!banana" matches query "ban")
-    let regex = new RegExp('(\\W|\\b)' + query, 'i');
     return savedTags.filter(tag => {
-      if (tag.match(regex)) {
-        if (limit === null || resultCount < limit) {
-          // limit allows a subset of the results
-          // See https://github.com/hypothesis/client/issues/1606
+      const words = tag.split(/\W+/);
+      let regex = new RegExp(`^${query}`, 'i'); // Only match the start of the string
+      // limit allows a subset of the results
+      // See https://github.com/hypothesis/client/issues/1606
+      if (limit === null || resultCount < limit) {
+        const matches =
+          words.some(word => word.match(regex)) || tag.match(regex);
+        if (matches) {
           ++resultCount;
           return true;
         }
--- a/src/sidebar/services/tags.js
+++ b/src/sidebar/services/tags.js
@@ -1,3 +1,6 @@
+// @ts-expect-error - Ignore error about default-importing a CommonJS module.
+import escapeStringRegexp from 'escape-string-regexp';
+
 /**
  * @typedef Tag
  * @property {string} text - The label of the tag
@@ -42,17 +45,20 @@ export class TagsService {
     // * tag has substring query occurring after a non-word character
     //   (e.g. tag "pink!banana" matches query "ban")
     return savedTags.filter(tag => {
+      if (limit !== null && resultCount >= limit) {
+        // limit allows a subset of the results
+        // See https://github.com/hypothesis/client/issues/1606
+        return false;
+      }
+      // Split the string on words. An improvement would be to use a unicode word boundary
+      // algorithm implemented by the browser (when available).
+      // https://unicode.org/reports/tr29/#Word_Boundaries
       const words = tag.split(/\W+/);
-      let regex = new RegExp(`^${query}`, 'i'); // Only match the start of the string
-      // limit allows a subset of the results
-      // See https://github.com/hypothesis/client/issues/1606
-      if (limit === null || resultCount < limit) {
-        const matches =
-          words.some(word => word.match(regex)) || tag.match(regex);
-        if (matches) {
-          ++resultCount;
-          return true;
-        }
+      const regex = new RegExp(`^${escapeStringRegexp(query)}`, 'i'); // Only match the start of the string
+      const matches = words.some(word => word.match(regex)) || tag.match(regex);
+      if (matches) {
+        ++resultCount;
+        return true;
       }
       return false;
     });
--- a/src/sidebar/services/tags.js
+++ b/src/sidebar/services/tags.js
@@ -38,12 +38,7 @@ export class TagsService {
   filter(query, limit = null) {
     const savedTags = this._storage.getObject(TAGS_LIST_KEY) || [];
     let resultCount = 0;
-    // query will match tag if:
-    // * tag starts with query (e.g. tag "banana" matches query "ban"), OR
-    // * any word in the tag starts with query
-    //   (e.g. tag "pink banana" matches query "ban"), OR
-    // * tag has substring query occurring after a non-word character
-    //   (e.g. tag "pink!banana" matches query "ban")
+    // Match any tag where the query is a prefix of the tag or a word within the tag.
     return savedTags.filter(tag => {
       if (limit !== null && resultCount >= limit) {
         // limit allows a subset of the results
--- a/src/sidebar/services/test/tags-test.js
+++ b/src/sidebar/services/test/tags-test.js
@@ -57,7 +57,7 @@ describe('TagsService', () => {
         updated: stamp,
       },
       ინგლისური: {
-        // "English" in Georgian
+        // Non-ASCII characters. This is "English" in Georgian
         text: 'ინგლისური',
         count: 1,
         updated: stamp,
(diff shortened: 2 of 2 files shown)


src/sidebar/services/test/tags-test.js
```diff
@@ -56,6 +56,11 @@ describe('TagsService', () => {
         count: 1,
         updated: stamp,
       },
+      ინგლისური: { // "English" in Georgian
+        text: 'ინგლისური',
+        count: 1,
+        updated: stamp,
+      },
```
robertknight, line 63: Aside from this new test case, all of the tags consist only of letters and spaces. This misses out many scenarios involving even just ASCII characters.

robertknight (commented):
I know that Lyza was added as the reviewer here but I hope you don't mind that I had a look through out of curiosity. I added one comment clarifying my understanding of why the issue was happening and another discussing a potentially alternative way that the query/tag matching could work in future (not essential for this PR) that would be easier to align with how the backend tokenizes tags.

The fix proposed here is actually quite pragmatic, given that JS regexes don't have a convenient way to match word boundaries in a Unicode aware way (which is [quite complicated](https://www.unicode.org/reports/tr29/#Word_Boundaries)). It just adds a fallback case which allows any tags which share a prefix with the query to match. The `\b` part now appears to be redundant to me though - no tests fail if it is removed!

I did spot one significant existing problem when looking at the code which is that the `query` is not escaped when constructing the regexp. I gave an example of how this could affect even matching of ASCII queries against ASCII tags. We have an existing `escape-string-regexp` dependency in the client which can do this escaping. I also noted that we could do with some more test cases, because the pre-existing ones only involve tags with spaces and lower-case letters - no numbers, upper-case letters or punctuation.

src/sidebar/services/tags.js
```diff
@@ -41,7 +41,7 @@ export class TagsService {
     //   (e.g. tag "pink banana" matches query "ban"), OR
     // * tag has substring query occurring after a non-word character
     //   (e.g. tag "pink!banana" matches query "ban")
```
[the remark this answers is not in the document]
LMS007: Yep, I think this makes sense, and I'll go ahead and convert it over and add more tests. 

But `words.any` should be `words.some` correct?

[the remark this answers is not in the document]
robertknight: > But words.any should be words.some correct?

Gah, yes. 🤦‍♂️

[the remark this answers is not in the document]
LMS007: I actually think "any" makes more sense, I think we should tell the js gods they are wrong :)

src/sidebar/services/tags.js
```diff
@@ -41,12 +41,15 @@ export class TagsService {
...
-    let regex = new RegExp('(\\W|\\b)' + query, 'i');
     return savedTags.filter(tag => {
-      if (tag.match(regex)) {
-        if (limit === null || resultCount < limit) {
-          // limit allows a subset of the results
-          // See https://github.com/hypothesis/client/issues/1606
+      const words = tag.split(/\W+/);
+      let regex = new RegExp(`^${query}`, 'i'); // Only match the start of the string
```
robertknight, line 46: You need to escape the query when used within the regex. See `src/sidebar/helpers/groups.js` for an example of how to do this using the escape-string-regexp package. Also the regex should be `const` if not re-assigned.

src/sidebar/services/test/tags-test.js
```diff
@@ -56,6 +56,12 @@ describe('TagsService', () => {
         count: 1,
         updated: stamp,
       },
+      'žŸ¡¢£¤¥¦§': {
+        // ASCII chars
```
robertknight, line 60: All the other test cases are technically ASCII as well. Something like "Non-word ASCII chars" would be better.

src/sidebar/services/test/tags-test.js
```diff
@@ -56,6 +56,12 @@ describe('TagsService', () => {
         count: 1,
         updated: stamp,
       },
+      'žŸ¡¢£¤¥¦§': {
```
robertknight, line 59: Can you re-add the Unicode test case from the earlier version of this commit, based on the bug report. We want both since they exercise different behavior.

src/sidebar/services/tags.js
```diff
@@ -41,12 +41,15 @@ export class TagsService {
...
+      // limit allows a subset of the results
+      // See https://github.com/hypothesis/client/issues/1606
+      if (limit === null || resultCount < limit) {
```
robertknight, line 49: It would be a little more efficient, and I think easier to follow, to invert this test and hoist it to the top of the `filter` callback.

```js
savedTags.filter(tag => {
  if (limit !== null && resultCount >= limit) {
    return false;
  }

  // Test for match and increment `resultCount`
});
```

src/sidebar/services/tags.js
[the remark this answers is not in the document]
robertknight: I don't know the history of this specific method, but sometimes the standards committee are constrained by the need to avoid breaking existing websites. For example `array.includes` was originally going to be called `contains` but that turned out [to break existing code](https://github.com/tc39/proposal-Array.prototype.includes).

robertknight (changes requested):
Hi Kyle - The `query` still needs to be escaped in the regex. Also I think the test case added in the original version of this PR, based on the bug report, was valuable and should be re-added.

src/sidebar/services/tags.js
LMS007: yep, makes sense

src/sidebar/services/test/tags-test.js
```diff
@@ -83,6 +95,14 @@ describe('TagsService', () => {
       assert.deepEqual(tags.filter('b', 2), ['bar', 'bar argon']);
       assert.deepEqual(tags.filter('b', 3), ['bar', 'bar argon', 'banana']);
     });
+
+    it('returns non-word ASCII tags that start with the query string', () => {
+      assert.deepEqual(tags.filter('^žŸ¡\\'), ['^žŸ¡\\¢£¤']);
+    });
+
+    it('returns non-English tags that start with the query string', () => {
```
robertknight, line 103:
```suggestion
    it('returns non-ASCII tags that start with the query string', () => {
```

Although this example is indeed non-English, it is the use of non-ASCII characters that matters.

src/sidebar/services/tags.js
```diff
@@ -41,15 +44,21 @@ export class TagsService {
     //   (e.g. tag "pink banana" matches query "ban"), OR
     // * tag has substring query occurring after a non-word character
     //   (e.g. tag "pink!banana" matches query "ban")
```
robertknight, line 46: Given the changes in this branch, I think it would be helpful to replace the multi-line comment above the `let regex...` line with one which describes the overall intent of the code in a language-agnostic way, rather than the specific matching criteria. For example:

```js
// Match any tag where the query is a prefix of the tag or a word within the tag.
```

src/sidebar/services/test/tags-test.js
```diff
@@ -56,6 +56,18 @@ describe('TagsService', () => {
         count: 1,
         updated: stamp,
       },
+      ინგლისური: {
+        // "English" in Georgian
```
robertknight, line 60:
```suggestion
        // Non-ASCII characters. This is "English" in Georgian
```

Highlight the important aspect of this test case. Most readers will probably guess that the test case matters because of the use of Unicode characters rather than the Georgian language specifically, but I think there is no harm in being explicit.

robertknight (approved):
I had some suggestions about code comments. Otherwise LGTM.

PR #3577, opened 2021-07-14
Remove `off` method on `Bridge` class

esanzgar:
This method was not used anywhere. In addition for this method to have
the intended effect, it needed to be called before `createChannel`,
because at that point the methods are registered with the `RPC` class
and there is no way to remove the listeners.


robertknight (approved):
LGTM.

PR #3578, opened 2021-07-14
Rename `on` to `register` in the `Bridge` and `CrossFrame` classes

esanzgar:
I believe `bridge.register('method', ....)` indicates better the fact
that there is a registration phase before the `method` can be invoked.

`bridge.on` feels like the addition of the listener can happen
at anytime. That's not true, all methods must be register before the
channel is created.

### DIFF
--- a/src/annotator/annotation-counts.js
+++ b/src/annotator/annotation-counts.js
@@ -11,7 +11,7 @@ const ANNOTATION_COUNT_ATTR = 'data-hypothesis-annotation-count';
  */
 
 export default function annotationCounts(rootEl, crossframe) {
-  crossframe.on(
+  crossframe.register(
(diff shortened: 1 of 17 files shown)


esanzgar:
From @robertknight: 

This happens because the RPC class design requires all the handlers to be provided up-front. Nothing in the RPC implementation really requires that though, so we could potentially remove this limitation in future (eg. add an RPC#on method to register a handler after initial construction). On the other hand, that might be unwise because it could lead to non-obvious race conditions where a bridge listener is added after the other end tries to make an RPC call. At least with the current implementation the Bridge class could detect this misuse (calling on after createChannel).

I think it is difficult to convey the issue here with a name change alone. My thought would be to instead:
1. Document the restriction in the JSDoc for Bridge#on
2. Detect misuse of the API (calling on after createChannel) and throw an Error

PR #3583, opened 2021-07-15
General improvement on `shared/bridge.js`

### LINKED ISSUE #3580: Modernization/simplification and JSDoc improvements in `shared/bridge.js` and `annotator/cross-frame.js`
Prior to adding support for  `MessageChannel` in these classes, it might be a good idea to split these changes off into a separate smaller PR.

esanzgar:
The following improvements don't modify the functionality of `shared/bridge.js`:

* Added types and comment about a callback
* Remove unnecessary `Array.from(...)`
* Simplify promise
* Simplify `this.link` property

Closes #3580

### DIFF
--- a/src/shared/bridge.js
+++ b/src/shared/bridge.js
@@ -104,25 +104,24 @@ export default class Bridge {
     };
 
     const promises = this.links.map(l => {
-      const p = new Promise((resolve, reject) => {
+      const promise = new Promise((resolve, reject) => {
         const timeout = setTimeout(() => resolve(null), 1000);
         try {
-          return l.channel.call(method, ...args, (err, result) => {
+          l.channel.call(method, ...args, (err, result) => {
             clearTimeout(timeout);
             if (err) {
-              return reject(err);
+              reject(err);
             } else {
-              return resolve(result);
+              resolve(result);
             }
           });
         } catch (error) {
-          const err = error;
-          return reject(err);
+          reject(error);
         }
       });
 
       // Don't assign here. The disconnect is handled asynchronously.
-      return p.catch(_makeDestroyFn(l.channel));
+      return promise.catch(_makeDestroyFn(l.channel));
     });
 
     let resultPromise = Promise.all(promises);
--- a/src/shared/bridge.js
+++ b/src/shared/bridge.js
@@ -10,7 +10,7 @@ import { RPC } from './frame-rpc';
  */
 export default class Bridge {
   constructor() {
-    /** @type {Array<{channel: RPC, window: Window}>} */
+    /** @type {RPC[]} */
     this.links = [];
     /** @type {Record<string, (...args: any[]) => void>} */
     this.channelListeners = {};
@@ -24,7 +24,7 @@ export default class Bridge {
    * This removes the event listeners for messages arriving from other windows.
    */
   destroy() {
-    this.links.map(link => link.channel.destroy());
+    this.links.forEach(channel => channel.destroy());
   }
 
   /**
@@ -66,10 +66,7 @@ export default class Bridge {
     channel.call('connect', token, ready);
 
     // Store the newly created channel in our collection
-    this.links.push({
-      channel,
-      window: source,
-    });
+    this.links.push(channel);
 
     return channel;
   }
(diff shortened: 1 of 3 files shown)


src/shared/bridge.js
```diff
@@ -97,9 +95,7 @@ export default class Bridge {
     const _makeDestroyFn = c => {
       return error => {
         c.destroy();
-        this.links = Array.from(this.links)
-          .filter(l => l.channel !== c)
-          .map(l => l);
```
esanzgar, line 96: It was unnecessary.

src/shared/bridge.js
```diff
@@ -65,10 +66,7 @@ export default class Bridge {
     channel.call('connect', token, ready);
 
     // Store the newly created channel in our collection
-    this.links.push({
-      channel,
-      window: source,
```
robertknight, line 70: Was `window` not used?

src/shared/bridge.js
```diff
@@ -78,46 +76,54 @@ export default class Bridge {
...
     };
 
-    const promises = this.links.map(l => {
-      const p = new Promise((resolve, reject) => {
+    const promises = this.links.map(channel => {
+      const promise = new Promise((resolve, reject) => {
         const timeout = setTimeout(() => resolve(null), 1000);
+        // This try/catch is only used on tests, to exercise the `_makeDestroyFn`
+        // function. In real life, the execution of the channel.call's callback will
+        // never throws an exception. The try/catch can be removed when we support
```
robertknight, line 107: This assumes that the `window.postMessage` or `messagePort.postMessage` call never throws. Are you sure  that's true?

src/shared/bridge.js
```diff
@@ -78,46 +76,54 @@ export default class Bridge {
...
       };
     };
 
-    const promises = this.links.map(l => {
-      const p = new Promise((resolve, reject) => {
+    const promises = this.links.map(channel => {
+      const promise = new Promise((resolve, reject) => {
         const timeout = setTimeout(() => resolve(null), 1000);
+        // This try/catch is only used on tests, to exercise the `_makeDestroyFn`
+        // function. In real life, the execution of the channel.call's callback will
```
robertknight, line 106: The `catch` block is not invoked if `call`'s _callback_ throws but if `call` itself throws. Looking at the current implementation I think that is only possible if `postMessage` throws.

robertknight (commented):
I had a query about the comment added about the try/catch block around `RPC#call`. Otherwise LGTM.

src/shared/bridge.js
esanzgar: It was not used.

I am happy to make use of it, if you wish.

src/shared/bridge.js
esanzgar: It is possible to throws but this try/catch will never catch it. The reason is that the callback is executive somewhere else in `RPC`.

src/shared/bridge.js
robertknight: If we're not using it then it's fine to remove it.

src/shared/bridge.js
esanzgar: Yes, you are correct.

src/shared/bridge.js
robertknight: > I only found one way to make the postMessage to fail:

From looking at [the `postMessage` specification](https://html.spec.whatwg.org/multipage/web-messaging.html#message-port-post-message-steps) it looks to me like one thing that will cause an immediate `postMessage` failure is if the arguments to `postMessage` cannot be serialized. This would most likely indicate a bug in the calling code (eg. trying to invoke `call` with a non-serializable type of data.). I wonder if circular references can also cause a problem here?

esanzgar: I found a bunch at least a couple more scenarios where `postMessage` could raise exception. For example:

```
window.postMessage([],'e.com', [])
Uncaught DOMException: Failed to execute 'postMessage' on 'Window': Invalid target origin 'e.com' in a call to 'postMessage'.

window.postMessage([],'http://e.com', [])
Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('http://e.com') does not match the recipient window's origin ('https://html.spec.whatwg.org').
```

I know understand that this is more common/plausible that I thought, so I will remove the comment.

robertknight: @esanzgar - Do these exceptions occur synchronously? ie. If you surround the `postMessage` with a try..catch does it get invoked. Sometimes in the browser console you can trigger an error which appears to have been reported synchronously but has in fact been reported via some async mechanism (eg. an 'error' event).

esanzgar: The exception happens synchronously and it is intercepted by a surrounded `try/catch`.

robertknight (approved):
Changes LGTM.

esanzgar:
Thanks for the approval, @robertknight.

I made a minor, last minute change in the first commit:

```diff
80,83c80
<    *   callback with this type: `(error: string|Error|null, ...result: any[]) => void`.
<    *   This callback, if any, will be triggered once a responds (via `postMessage`)
<    *   comes back from the other frame/s. If the first argument (error) is `null`
<    *   it means successful execution of the whole remote procedure call.
---
>    *   callback with this type: `(error: string|Error, ...result: any[]) => void`
143,145c140
<    *   callback of the type: `(error: string|Error|null, ...result: any[]) => void`.
<    *   This callback must be invoked in order to respond (via `postMessage`)
<    *   to the other frame/s with a result or an error.
---
>    *   callback of the type: `(error: string|Error, ...result: any[]) => void`
158c153
<    * @param {(error: null, channel: RPC, window: Window) => void} callback
---
>    * @param {(...args: any[]) => void} callback
```

I realised about a mistake in the type for the  `onConnect` callback.

I also added a little bit more description about the purpose of the callbacks.

PR #3584, opened 2021-07-15
Revert "Replace query-string dependency with wrapper around `URLSearchParams`"

LMS007:
Reverts hypothesis/client#3582

This was causing a breaking issue with displaying annotations on PDFs


robertknight:
The issue here was that https://github.com/hypothesis/client/pull/3582 introduced a change in how arrays in query param values were handled.

```js
// Before
stringify({ uri: ['uri-1', 'uri'2] }) // "uri=uri-1&uri=uri-2"

// After
stringify({ uri: ['uri-1', 'uri'2] }) // "uri=uri-1%2Curi-2", where "%2C" is a comma
```

I'm looking into this more closely now and I've found a bunch of other subtle issues to do with repeated query parameters.

PR #3587, opened 2021-07-16
Throw an error when adding a handler if a channel has already been created

esanzgar:
First commit, 'error when adding a handler if a channel has been created':

The RPC class design requires all the handlers to be provided up-front.
Currently, adding a new handler after a channel is created has no
effect. To make more obvious this misuse, the method now throws an
error.

Second commit, 'rename `callback` to `listener`':

The word `listener` matches better the internal property names of
`this.channelListeners` and `this.onConnectListeners`. In addition, it
helps to differentiate from the `callback` function which is the last
(optional) argument of the `listener`. The `callback` is used to send
back a `postMessage` response to the other frame/s.

### DIFF
--- a/src/shared/test/bridge-test.js
+++ b/src/shared/test/bridge-test.js
@@ -168,14 +168,39 @@ describe('shared/bridge', () => {
 
   describe('#on', () => {
     it('adds a method to the method registry', () => {
-      createChannel();
       bridge.on('message1', sandbox.spy());
+      createChannel();
       assert.isFunction(bridge.channelListeners.message1);
     });
 
+    it('raise an error if trying to register a listener after a channel has been already created', () => {
+      createChannel();
+      let error;
+      try {
+        bridge.on('message1', () => {});
+      } catch (err) {
+        error = err;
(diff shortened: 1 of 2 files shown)


src/shared/test/bridge-test.js
```diff
@@ -168,14 +168,39 @@ describe('shared/bridge', () => {
 
   describe('#on', () => {
     it('adds a method to the method registry', () => {
-      createChannel();
       bridge.on('message1', sandbox.spy());
+      createChannel();
       assert.isFunction(bridge.channelListeners.message1);
     });
 
+    it('raise an error if trying to register a listener if a channel has been already created', () => {
```
robertknight, line 176:
```suggestion
    it('raise an error if trying to register a listener after a channel has been already created', () => {
```

robertknight (approved):
LGTM, assuming CI passes.

src/shared/test/bridge-test.js
esanzgar: 👍🏼

PR #3598, opened 2021-07-19
`RPC`-`Bridge` integration test

esanzgar:
We have agreed that `RPC` and `Bridge` classes would benefit from a more
realistic testing. These tests exercise `RPC` and `Bridge` code together
using `MessagePort`.

A second commit contains a small change in the current behaviour of `Bridge`:

Previously, executing the `call` method after the channels were
destroyed caused an asynchronous timeout and the promise to be
*resolved* with value of `[null, null...]`.

With this change, the `call` method returns immediately with an empty
array of results. This behaviour is similar to what happens trying to
use the `call` method before `createChannel`, like this:   https://github.com/hypothesis/client/blob/51f6d3ac02857e17c61665b08efa414a3f965b32/src/shared/test/integration/rpc-bridge-test.js#L158-L162

### DIFF
--- a/src/shared/test/bridge-test.js
+++ b/src/shared/test/bridge-test.js
@@ -99,6 +99,11 @@ describe('shared/bridge', () => {
   });
 
   describe('#call', () => {
+    it('returns an empty array calling a RPC method before a channel is created', async () => {
+      const results = await bridge.call('method1', 'params1');
+      assert.deepEqual(results, []);
+    });
+
     it('forwards the call to every created channel', () => {
       const channel = createChannel();
       channel.call.resetHistory();
--- a/src/shared/test/bridge-test.js
+++ b/src/shared/test/bridge-test.js
@@ -100,7 +100,7 @@ describe('shared/bridge', () => {
 
   describe('#call', () => {
     it('returns an empty array calling a RPC method before a channel is created', async () => {
-      const results = await bridge.call('method1', 'params1');
+      const results = await bridge.call('method1', 'param1');
       assert.deepEqual(results, []);
     });
 
--- a/src/shared/test/integration/rpc-bridge-test.js
+++ b/src/shared/test/integration/rpc-bridge-test.js
@@ -268,12 +268,10 @@ describe('rpc-bridge integration', () => {
       });
       clock.tick(1000);
 
-      // It timeouts using reciprocalBridge, too! This maybe a bit contour intuitive.
       reciprocalBridge.call('method1', 'params1', (err, results) => {
-        assert.deepEqual(results, [null]);
+        assert.deepEqual(results, []);
         assert.isNull(err);
         done();
       });
-      clock.tick(1000);
     }));
 });
--- a/src/shared/test/integration/rpc-bridge-test.js
+++ b/src/shared/test/integration/rpc-bridge-test.js
@@ -1,7 +1,6 @@
 import Bridge from '../../bridge';
 
 describe('RPC-Bridge integration', () => {
-  const sandbox = sinon.createSandbox();
   let clock;
   let port1;
   let port2;
@@ -26,45 +25,40 @@ describe('RPC-Bridge integration', () => {
 
   afterEach(() => {
     bridges.forEach(bridge => bridge.destroy());
-    sandbox.restore();
     clock.restore();
   });
 
-  context('establishing a connection', () => {
+  describe('establishing a connection', () => {
     it('should invoke Bridge `onConnect` callbacks after connecting', async () => {
       clock.restore();
       const bridge = createBridge();
       const reciprocalBridge = createBridge();
-
       let callbackCount = 0;
       const callback = () => {
         ++callbackCount;
       };
-
       bridge.onConnect(callback);
       bridge.onConnect(callback); // allows multiple callbacks to be registered
       reciprocalBridge.onConnect(callback);
 
       const channel = bridge.createChannel(port1);
-      reciprocalBridge.createChannel(port2);
-
+      const reciprocalChannel = reciprocalBridge.createChannel(port2);
       await waitForMessageDelivery();
+
       assert.equal(callbackCount, 3);
 
       // Additional calls to the RPC `connect` method are ignored
       channel.call('connect');
-      reciprocalBridge.call('connect');
-
+      reciprocalChannel.call('connect');
       await waitForMessageDelivery();
+
       assert.equal(callbackCount, 3);
     });
   });
 
-  context('sending and receiving RPC messages', () => {
-    let bridge;
-
-    beforeEach(() => {
-      bridge = createBridge();
+  describe('sending and receiving RPC messages', () => {
+    it('should invoke Bridge method handler on every channel when calling a RPC method', async () => {
+      const bridge = createBridge();
       const otherChannel = new MessageChannel();
       bridge.createChannel(port1);
       bridge.createChannel(otherChannel.port1);
@@ -74,23 +68,14 @@ describe('RPC-Bridge integration', () => {
       reciprocalBridge2.on('method1', (arg, cb) => cb(null, `${arg}bar`));
       reciprocalBridge1.createChannel(port2);
       reciprocalBridge2.createChannel(otherChannel.port2);
-    });
 
-    it('should invoke Bridge method handler on every channel when calling a RPC method (with Promise)', async () => {
-      const results = await bridge.call('method1', 'params1');
-      assert.deepEqual(results.sort(), ['params1foo', 'params1bar'].sort());
-    });
+      const results = await bridge.call('method1', 'param1');
 
-    it('should invoke Bridge method handler on every channel when calling a RPC method (with callback)', done => {
-      bridge.call('method1', 'params1', (err, results) => {
-        assert.deepEqual(results.sort(), ['params1foo', 'params1bar'].sort());
-        assert.isNull(err);
-        done();
-      });
+      assert.deepEqual(results.sort(), ['param1foo', 'param1bar'].sort());
     });
   });
 
-  context('errors and timeouts', () => {
+  describe('errors and timeouts', () => {
     it(`raises an error when the listener's callback fails`, async () => {
       const bridge = createBridge();
       const reciprocalBridge = createBridge();
@@ -101,10 +86,11 @@ describe('RPC-Bridge integration', () => {
 
       let error;
       try {
-        await reciprocalBridge.call('method1', 'params1');
+        await reciprocalBridge.call('method1', 'param1');
       } catch (err) {
         error = err;
       }
+
       assert.equal(error, errorMessage);
     });
 
@@ -115,11 +101,11 @@ describe('RPC-Bridge integration', () => {
       bridge.on('method1', (_arg, cb) => cb(errorMessage));
       bridge.createChannel(port1);
       const channel = reciprocalBridge.createChannel(port2);
-      sandbox.stub(channel, 'destroy').callThrough();
+      sinon.stub(channel, 'destroy').callThrough();
 
       let error;
       try {
-        await reciprocalBridge.call('method1', 'params1');
+        await reciprocalBridge.call('method1', 'param1');
       } catch (err) {
         error = err;
       }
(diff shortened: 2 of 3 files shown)


robertknight (commented):
Hi @esanzgar,

The main feedback I have here is that integration tests should try to avoid covering things that are effectively covered by unit tests and instead focus on the most important scenarios and failure modes that exercise the integration between the components. Since these integration tests are usually checking scenarios/activities rather than specific methods, the test descriptions will tend to reflect that. For example:

```js
it("should invoke Bridge `onConnect` callbacks after connecting") { ... }
it('should invoke Bridge method handlers when calling an RPC method', () => { ... });
it('should return the result of a successful call', () => { ... });
it('should return an error if an RPC call fails', () => { ... });
it('should close ports after destroying a Bridge', () => { ... });
```

It can be still be useful to group related tests together (eg. in "establishing a connection" or "calling RPC methods" blocks), and sometimes it will make sense to use the method/function that serves as the entry point to the scenario. If you're using a method, make sure it is clear what the class is though. In these tests all the `#methodName` describe blocks refer to methods of `Bridge`, but that may not be immediately obvious to a future reader who isn't as familiar with the code.

For this PR, I think what I would suggest to move forwards is to go through the tests, ask whether they are testing something that really involves integration between the components remove it if not. Then for the remaining tests revise the descriptions to either be more scenario focused or at least make sure it is clear which entry point (eg. `Bridge#call`) is being referred to.

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,277 @@
...
+
+  beforeEach(() => {
+    clock = sinon.useFakeTimers();
+    const channel = new MessageChannel();
+    port1 = channel.port1;
+    port2 = channel.port2;
+  });
+
+  afterEach(() => {
+    bridges.forEach(bridge => bridge.destroy);
```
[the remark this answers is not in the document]
esanzgar: Good catch! Thanks!

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,277 @@
...
+      reciprocalBridge.createChannel(port2);
+      channel.call('connect', () => {
+        assert.equal(callbackCount, 2);
+        done();
+      });
+    });
+  });
+
+  describe('#on', () => {
+    it('raise an error if trying to register a listener after a channel has been already created', () => {
```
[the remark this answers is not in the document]
esanzgar: I removed this test.

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,277 @@
...
+        assert.deepEqual(results.sort(), ['params1foo', 'params1bar'].sort());
+        assert.isNull(err);
+        done();
+      });
+    });
+
+    it('returns an empty array if call method before a channel is created', async () => {
+      const bridge = createBridge();
+      const results = await bridge.call('method1', 'params1');
+      assert.equal(results.length, 0);
```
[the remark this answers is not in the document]
esanzgar: Correct, I moved this test to the `Bridge` unit testing.

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,277 @@
...
+    port2 = channel.port2;
+  });
+
+  afterEach(() => {
+    bridges.forEach(bridge => bridge.destroy);
+    sandbox.restore();
+    clock.restore();
+  });
+
+  describe('#createChannel', () => {
```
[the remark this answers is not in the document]
esanzgar: I agree, I have changed the descriptions and reorganised the tests.

esanzgar:
@robertknight, based on your feedback, I have:

* changed the language of the test descriptions and focus on
  activities or common scenarios

* grouped related tests together based on the different activities

* moved one unit test to `bridge-test.js` and delete tests that are covered
  in unit tests.

You can see these changes in the last commit.

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# Date:      Tue Jul 27 18:37:41 2021 +0200
#
# On branch rpc-bridge-integration-test
# Changes to be committed:
#	modified:   src/shared/test/bridge-test.js
#	modified:   src/shared/test/integration/rpc-bridge-test.js
#

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,206 @@
...
+    port2 = channel.port2;
+  });
+
+  afterEach(() => {
+    bridges.forEach(bridge => bridge.destroy());
+    sandbox.restore();
+    clock.restore();
+  });
+
+  context('establishing a connection', () => {
```
robertknight, line 33: `describe` is more appropriate than `context` here. Functionally I believe they are exactly the same. However `context` is normally used to refer to a situation or circumstance (eg. "browser doesn't support feature X") whereas `describe` refers to _what_ is being tested, whether that is a specific class/function/method, as in most unit tests, or a process, as is the case here.

src/shared/test/bridge-test.js
```diff
@@ -99,6 +99,11 @@ describe('shared/bridge', () => {
   });
 
   describe('#call', () => {
+    it('returns an empty array calling a RPC method before a channel is created', async () => {
+      const results = await bridge.call('method1', 'params1');
```
robertknight, line 103:
```suggestion
      const results = await bridge.call('method1', 'param1');
```

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,206 @@
...
+      reciprocalBridge1.createChannel(port2);
+      reciprocalBridge2.createChannel(otherChannel.port2);
+    });
+
+    it('should invoke Bridge method handler on every channel when calling a RPC method (with Promise)', async () => {
+      const results = await bridge.call('method1', 'params1');
+      assert.deepEqual(results.sort(), ['params1foo', 'params1bar'].sort());
+    });
+
+    it('should invoke Bridge method handler on every channel when calling a RPC method (with callback)', done => {
```
robertknight, line 84: This is out of scope for this PR - but could we get rid of support for result callbacks in the `Bridge`? It looks to me like it duplicates the Promise return functionality and the Promise-based API is more convenient for consumers and easier to specify the type of.

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,206 @@
...
+      const bridge = createBridge();
+      const reciprocalBridge = createBridge();
+      const errorMessage = 'My error';
+      bridge.on('method1', (_arg, cb) => cb(errorMessage));
+      bridge.createChannel(port1);
+      reciprocalBridge.createChannel(port2);
+
+      let error;
+      try {
+        await reciprocalBridge.call('method1', 'params1');
```
robertknight, line 104:
```suggestion
        await reciprocalBridge.call('method1', 'param1');
```

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,206 @@
...
+      } catch (err) {
+        error = err;
+      }
+
+      assert.called(channel.destroy);
+      assert.equal(error, errorMessage);
+      assert.deepEqual(reciprocalBridge.links, []);
+    });
+
+    it('no longer send RPC messages to a Bridge channel that has received an error', async () => {
```
robertknight, line 132:
```suggestion
    it('no longer sends RPC messages to a Bridge channel that has received an error', async () => {
```

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,206 @@
...
+        assert.deepEqual(results, []);
+        assert.calledOnce(channel.call);
+      }
+    });
+
+    it('timeouts if the Bridge channel is not connected', done => {
+      const bridge = createBridge();
+      bridge.createChannel(port1);
+
+      bridge.call('method1', 'params1', (err, results) => {
```
robertknight, line 160: I suggest to use the Promise-based API for collecting results in all these tests. This will make them easier to update if we remove the callback-based API in future.

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,206 @@
...
+      const bridge = createBridge();
+      const reciprocalBridge = createBridge();
+      bridge.on('method1', (arg, cb) => cb(null, `${arg}foo`));
+      reciprocalBridge.on('method2', (arg, cb) => cb(null, `${arg}bar`));
+      bridge.createChannel(port1);
+      reciprocalBridge.createChannel(port2);
+
+      reciprocalBridge.destroy();
+
+      bridge.call('method2', 'params1', (err, results) => {
```
robertknight, line 192: As with the previous test, if we can use the Promise-based API here this will make it easier to remove the callback API in future.

robertknight (commented):
The general changes look good. The main feedback I had is to do with the APIs that `Bridge` provides for returning results. It provides both Promise-based and callback-based responses. I can't see a good reason to have both. In that case it makes sense to me to prefer the Promise-based API. If you're agreed, then I suggest using the Promise-based API for all tests except the one that is explicitly testing the callback API.

src/shared/test/integration/rpc-bridge-test.js
esanzgar: I am fine removing this test. It was just to demonstrated that the `error` is `null`, which is only possible to be tested in callback-based API (it is functionally the same as no error raised in the Promise-based).

esanzgar:
@robertknight , I changed the tests that used the callback-based API to test solely the `Promise`-based.

I also added a comment in the `Bridge.call` method to indicate the future intention to retire the callback-based API.

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,185 @@
+import Bridge from '../../bridge';
+
+describe('RPC-Bridge integration', () => {
+  const sandbox = sinon.createSandbox();
```
robertknight, line 4: I'm not sure if we actually need to use a custom sandbox here. This would normally only be needed if you wanted to temporarily patch properties of global objects (eg. a `window` method) for the duration of a test.

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,185 @@
...
+      bridge.createChannel(port1);
+      bridge.createChannel(otherChannel.port1);
+      const reciprocalBridge1 = createBridge();
+      const reciprocalBridge2 = createBridge();
+      reciprocalBridge1.on('method1', (arg, cb) => cb(null, `${arg}foo`));
+      reciprocalBridge2.on('method1', (arg, cb) => cb(null, `${arg}bar`));
+      reciprocalBridge1.createChannel(port2);
+      reciprocalBridge2.createChannel(otherChannel.port2);
+
+      const results = await bridge.call('method1', 'param1');
```
robertknight, line 76:
```suggestion
      const results = await bridge.call('method1', 'param1');
      
```

Having blank lines before and after the "action" phase of a test with many lines makes it easier to see which code is the setup, which code is the action being tested and which code is the assertions.

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,185 @@
...
+      bridge.createChannel(port1);
+      reciprocalBridge.createChannel(port2);
+
+      let error;
+      try {
+        await reciprocalBridge.call('method1', 'param1');
+      } catch (err) {
+        error = err;
+      }
+      assert.equal(error, errorMessage);
```
robertknight, line 96:
```suggestion

      assert.equal(error, errorMessage);
```

src/shared/test/integration/rpc-bridge-test.js
```diff
@@ -0,0 +1,185 @@
...
+
+      let error;
+      try {
+        await reciprocalBridge.call('method1', 'param1');
+      } catch (err) {
+        error = err;
+      }
+      assert.equal(error.message, errorMessage);
+
+      for (let i = 0; i < 5; ++i) {
```
robertknight, line 137: Having 5 iterations here seems a bit redundant. Surely one would be enough?

robertknight (approved):
LGTM.

I have some general thoughts that don't affect this PR:

I think we probably want to take a step back soon and look at the overall responsibilities of the `RPC` and `Bridge` classes and consider whether we really do need both of them. If we do, the distinction between them should be made clearer. Right now it's not easy to understand that looking at the JSDoc for the `Bridge` class.

Currently the differences that I can see are:

1. RPC supports a single channel whereas Bridge supports multiple channels and allows channels to be added and removed at any point
2. RPC requires all the method handlers to be registered up-front, whereas Bridge allows them to be added incrementally as long as they are all added before connections are created
3. RPC returns results via a callback, whereas Bridge wraps that to return a promise

(1) is an important distinction, whereas (2) and (3) may just be a result of history.

esanzgar:
> I think we probably want to take a step back soon and look at the overall responsibilities of the RPC and Bridge classes and consider whether we really do need both of them. If we do, the distinction between them should be made clearer. Right now it's not easy to understand that looking at the JSDoc for the Bridge class.
> [...]
> (1) is an important distinction, whereas (2) and (3) may just be a result of history.

I agree with you.

I have been thinking that if (1) is the main distinction, then maybe renaming the `Bridge` class to `Broadcaster` could reflect better the difference.

src/shared/test/integration/rpc-bridge-test.js
esanzgar: We don't use the sandbox at all 😄 .

Because this integration test derives from a unit test, I forgot to remove it.

src/shared/test/integration/rpc-bridge-test.js
esanzgar: I removed the loop.

PR #3600, opened 2021-07-22
Add manual test case for when Hypothesis is loaded into an iframe inside a shadow root

robertknight:
The first commit in this PR reorganizes the test cases on the home page of the dev server to split general HTML/PDF test documents from those which test for specific scenarios (non-standard configuration, documents with properties that may cause problems etc.):

Client dev server homepage reorganization

The second commit in this PR adds a manual test case to the dev server that exercises the problem preventing the client from working with [the new VitalSource book reader](https://github.com/hypothesis/client/issues/3590).

To test, restart the dev server and go to http://localhost:3000/document/host-in-shadow-root. Observe that the Annotate/Highlight toolbar appears when selecting text, but the sidebar does not.

### DIFF
--- a/dev-server/documents/html/host-in-shadow-root.mustache
+++ b/dev-server/documents/html/host-in-shadow-root.mustache
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>Guest frame in shadow root</title>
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
This is a nice way to approach validating this complex task. And I appreciate the organization of the documents in the dev server. Thanks!

PR #3616, opened 2021-07-26
Simplify construction of `AnnotationSync` service

esanzgar:
`CrossFrame` created an emitter and wrapped the `subscribe` and
`publish` methods into two functions that were passed as arguments to
the `AnnotationSync`'s constructor.

A simpler and more direct approach is to pass the event bus directly to
`AnnotationSync`.

### DIFF
--- a/src/annotator/annotation-sync.js
+++ b/src/annotator/annotation-sync.js
@@ -40,18 +40,29 @@ export default class AnnotationSync {
      */
     this.cache = {};
 
+    this.destroyed = false;
+
     // Relay events from the sidebar to the rest of the annotator.
     this.bridge.on('deleteAnnotation', (body, callback) => {
+      if (this.destroyed) {
+        callback(null);
+        return;
+      }
       const annotation = this._parse(body);
       delete this.cache[annotation.$tag];
       this._emitter.publish('annotationDeleted', annotation);
-      callback(null, this._format(annotation));
+      this._format(annotation);
+      callback(null);
     });
 
     this.bridge.on('loadAnnotations', (bodies, callback) => {
+      if (this.destroyed) {
+        callback(null);
+        return;
+      }
       const annotations = bodies.map(body => this._parse(body));
       this._emitter.publish('annotationsLoaded', annotations);
-      callback(null, annotations);
+      callback(null);
     });
 
     // Relay events from annotator to sidebar.
@@ -72,6 +83,10 @@ export default class AnnotationSync {
    * @param {AnnotationData[]} annotations
    */
   sync(annotations) {
+    if (this.destroyed) {
+      return;
+    }
+
     this.bridge.call(
       'sync',
       annotations.map(ann => this._format(ann))
@@ -127,6 +142,7 @@ export default class AnnotationSync {
   }
 
   destroy() {
+    this.destroyed = true;
     this._emitter.destroy();
   }
 }
--- a/src/annotator/annotation-sync.js
+++ b/src/annotator/annotation-sync.js
@@ -1,6 +1,8 @@
 /**
- * @typedef {import('../types/annotator').AnnotationData} AnnotationData
  * @typedef {import('../shared/bridge').default} Bridge
+ * @typedef {import('../types/annotator').AnnotationData} AnnotationData
+ * @typedef {import('../types/annotator').Destroyable} Destroyable
+ * @typedef {import('./util/emitter').EventBus} EventBus
  */
 
 /**
@@ -18,18 +20,16 @@
  *
  * It also listens for events from Guest when new annotations are created or
  * annotations successfully anchor and relays these to the sidebar app.
+ *
+ * @implements Destroyable
  */
 export default class AnnotationSync {
   /**
-   * @param {Bridge} bridge
-   * @param {Object} options
-   *   @param {(event: string, callback: (...args: any[]) => void) => void} options.on -
-   *     Function that registers a listener for an event from the rest of the
-   *     annotator
-   *   @param {(event: string, ...args: any[]) => void} options.emit -
-   *     Function that publishes an event to the rest of the annotator
+   * @param {EventBus} eventBus - Event bus for communicating with the annotator code (eg. the Guest)
+   * @param {Bridge} bridge - Channel for communicating with the sidebar
    */
-  constructor(bridge, options) {
+  constructor(eventBus, bridge) {
+    this._emitter = eventBus.createEmitter();
     this.bridge = bridge;
 
     /**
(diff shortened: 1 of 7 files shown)


src/annotator/annotation-sync.js
```diff
@@ -18,18 +20,16 @@
...
   /**
-   * @param {Bridge} bridge
-   * @param {Object} options
-   *   @param {(event: string, callback: (...args: any[]) => void) => void} options.on -
-   *     Function that registers a listener for an event from the rest of the
-   *     annotator
-   *   @param {(event: string, ...args: any[]) => void} options.emit -
-   *     Function that publishes an event to the rest of the annotator
+   * @param {EventBus} eventBus - enables intra-frame communication
+   * @param {Bridge} bridge - enables inter-frame communication
```
robertknight, line 29: Initially when I read these comments I misread the documentation for `eventBus` and `bridge` as being the same. I didn't notice the `intra-`/`inter-` distinction.

These comments describe what the `EventBus` and `Bridge` classes do generically, but not what they are for in the context of `AnnotationSync`. 

```
eventBus - Event bus for communicating with the annotator code (eg. the Guest)
bridge - Channel for communicating with the sidebar
```

src/annotator/annotation-sync.js
```diff
@@ -128,4 +125,8 @@ export default class AnnotationSync {
       msg: annotation,
     };
   }
+
+  destroy() {
+    this._emitter.destroy();
```
robertknight, line 130: Codecov is complaining about a missing test here. The test would be something like making sure that `AnnotationSync` does not respond to `beforeAnnotationCreated` events after `destroy` is called. Also I note that we're not disconnecting from the `Bridge` here. I guess that would be a pain to implement because the Bridge class doesn't support a way of disconnecting a specific set of callers. A workaround would be to have an internal flag in the `AnnotationSync` class that keeps track of whether `destroy` has been called.

src/annotator/test/annotation-sync-test.js
```diff
@@ -1,21 +1,19 @@
...
 
 import AnnotationSync from '../annotation-sync';
 
 describe('AnnotationSync', () => {
-  let createAnnotationSync;
+  let emitter;
   let fakeBridge;
-  let options;
+  let createAnnotationSync;
   let publish;
```
robertknight, line 9: Can you keep these variables sorted. Its easier to scan them and insert a new entry if they are.

src/annotator/test/annotation-sync-test.js
```diff
@@ -1,21 +1,19 @@
...
   const sandbox = sinon.createSandbox();
 
   beforeEach(() => {
-    const emitter = new EventEmitter();
-    const listeners = {};
+    const eventBus = new EventBus();
+    emitter = eventBus.createEmitter();
```
robertknight, line 14: This emitter is not currently cleaned up in `afterEach`.

src/annotator/test/annotation-sync-test.js
```diff
@@ -45,10 +37,10 @@ describe('AnnotationSync', () => {
 
   describe('#constructor', () => {
     context('when "deleteAnnotation" is published', () => {
```
robertknight, line 39: This is an issue with the existing code, but understanding these tests would be easier if the descriptions made it clear whether a particular event was an event coming from the sidebar or an event coming from the guest, like so:

```js
describe('handling events from sidebar', () => {
  describe('on "deleteAnnotation" event', ...)
});

describe('handling events from guest', () => {
  describe('on "beforeAnnotationCreated" event', ...);
});

The `describe('#constructor', ...)` block is not really useful. The constructor adds the listeners, but the tests are concerned with what happens when events are published rather than what directly happens when you call `new AnnotationSync`.

**Review by robertknight — CHANGES_REQUESTED** [review]
I think this change generally makes sense. I had some feedback on a few details and noticed a missing test for the added `destroy` method.

I also added some comments related to potentially removing the `CrossFrame` layer in future, but that's obviously outside the scope of this PR.

File: src/annotator/annotation-sync.js (hunk 2 above)
Comment (esanzgar) [reply src/annotator/annotation-sync.js:130]: Typically, the responsibility of destroying an object belongs to the class that created it. In this case, `CrossFrame` creates and destroys `Bridge`.

File: src/annotator/test/annotation-sync-test.js (hunk 3 above)
Comment (esanzgar) [reply src/annotator/test/annotation-sync-test.js:9]: I sorted in the order in which they are assigned in the `beforeEach`. I will sort them alphabetically.

File: src/annotator/test/annotation-sync-test.js (hunk 4 above)
Comment (esanzgar) [reply src/annotator/test/annotation-sync-test.js:14]: Good catch, thanks!

It is fixed.

File: src/annotator/annotation-sync.js (hunk 2 above)
Comment (esanzgar) [reply src/annotator/annotation-sync.js:145]: I have implemented the workaround you suggested in the last commit.

File: src/annotator/annotation-sync.js [hunk 6]
```diff
@@ -40,25 +40,32 @@ export default class AnnotationSync {
...
     this.cache = {};
 
-    this._on = options.on;
-    this._emit = options.emit;
+    this.destroyed = false;
 
     // Relay events from the sidebar to the rest of the annotator.
     this.bridge.on('deleteAnnotation', (body, callback) => {
+      if (this.destroyed) {
+        callback(new Error('AnnotationSync has been destroyed')); // this causes the bridge channel to be destroyed
```
robertknight, line 48: Destroying the bridge channel will affect other consumers of the `Bridge` instance, which is unexpected. I think it would be better to just return no result here. `FrameSyncService` in the sidebar doesn't actually use the result of this call, so perhaps we could just make this RPC handler return void in all cases.

src/annotator/test/annotation-sync-test.js
```diff
@@ -68,13 +61,14 @@ describe('AnnotationSync', () => {
         publish('deleteAnnotation', { msg: ann }, callback);
       });
 
-      it('deletes any existing annotation from its cache before calling emit', () => {
-        const ann = { id: 1, $tag: 'tag1' };
+      it('deletes any existing annotation from its cache before publish event to the annotator', done => {
```
robertknight, line 64:
```suggestion
      it('deletes any existing annotation from its cache before publishing event to the annotator', done => {
```

src/annotator/test/annotation-sync-test.js
```diff
@@ -169,4 +165,41 @@ describe('AnnotationSync', () => {
       assert.calledWith(fakeBridge.call, 'sync', [{ msg: ann, tag: ann.$tag }]);
     });
   });
+
+  describe('#destroy', () => {
+    it('respond with errors to events from the sidebar', () => {
```
robertknight, line 170: As noted above, I think causing the Bridge to be destroyed in this case is unexpected. Logically what we want to happen when `AnnotationSync` is destroyed is to disconnect the bridge. The simplest way to implement that currently is to make the event handlers do nothing.

robertknight (changes requested):
I think we want to avoid returning errors if `deleteAnnotations` and `loadAnnotations` are received after the `AnnotationSync` instance is destroyed, because this will cause other consumers of the Bridge instance to stop receiving messages. Logically what we want to do here is just disconnect from/ignore future messages.

src/annotator/annotation-sync.js
esanzgar: I have implemented your suggestion of returning no results in all cases.

robertknight (approved):
LGTM. This change is that it does couple `AnnotationSync` to the shared `EventBus`. In future we could potentially avoid that by making `AnnotationSync` its own event source. This would make sense if we decide that the `Guest` class should be the only consumer of these events.

I noted a test whose description needs to be updated due to the change in behavior of handling bridge calls after `AnnotationSync.destroy` is called.

src/annotator/test/annotation-sync-test.js
```diff
@@ -169,4 +165,38 @@ describe('AnnotationSync', () => {
...
```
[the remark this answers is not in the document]
esanzgar: Updated

PR #3617, opened 2021-07-26
Use ES2015 class methods

esanzgar:
The documentation for the `on` and `call` is referred to the `Bridge#call|on`.

Changed default export for named export.

### DIFF
--- a/src/annotator/cross-frame.js
+++ b/src/annotator/cross-frame.js
@@ -1,4 +1,4 @@
-import AnnotationSync from './annotation-sync';
+import { AnnotationSync } from './annotation-sync';
 import Bridge from '../shared/bridge';
 import * as frameUtil from './util/frame-util';
 import FrameObserver from './frame-observer';
--- a/src/annotator/cross-frame.js
+++ b/src/annotator/cross-frame.js
@@ -105,12 +105,9 @@ export class CrossFrame {
    * Subscribe to an event from the sidebar.
    *
    * @param {string} method
-   * @param {(...args: any[]) => void} listener -- Final argument is an optional
-   *   callback of the type: `(error: string|Error|null, ...result: any[]) => void`.
-   *   This callback must be invoked in order to respond (via `postMessage`)
-   *   to the other frame/s with a result or an error.
-   * @throws {Error} If trying to register a callback after a channel has already been created
-   * @throws {Error} If trying to register a callback with the same name multiple times
+   * @param {(...args: any[]) => void} listener
+   *
+   * @see {Bridge.on} for details.
    */
   on(method, listener) {
     this._bridge.on(method, listener);
@@ -119,12 +116,10 @@ export class CrossFrame {
   /**
    * Call an RPC method exposed by the sidebar to the annotator.
    *
-   * @param {string} method
-   * @param {any[]} args - Arguments to method. Final argument is an optional
-   *   callback with this type: `(error: string|Error|null, ...result: any[]) => void`.
-   *   This callback, if any, will be triggered once a response (via `postMessage`)
-   *   comes back from the other frame/s. If the first argument (error) is `null`
-   *   it means successful execution of the whole remote procedure call.
+   * @param {string} method - Name of remote method to call.
+   * @param {any[]} args
+   *
+   * @see {Bridge.call} for details.
    */
   call(method, ...args) {
     this._bridge.call(method, ...args);
--- a/src/annotator/cross-frame.js
+++ b/src/annotator/cross-frame.js
@@ -4,6 +4,7 @@ import * as frameUtil from './util/frame-util';
 import FrameObserver from './frame-observer';
 
 /**
+ * @typedef {import('../shared/frame-rpc').RPC} RPC
  * @typedef {import('../types/annotator').AnnotationData} AnnotationData
  * @typedef {import('../types/annotator').Destroyable} Destroyable
  * @typedef {import('./util/emitter').EventBus} EventBus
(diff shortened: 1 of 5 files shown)


src/annotator/cross-frame.js
```diff
@@ -31,10 +31,10 @@ export class CrossFrame {
    */
   constructor(element, eventBus, options) {
     const { config, server } = options;
-    const discovery = new Discovery(window, { server });
-    const bridge = new Bridge();
-    const annotationSync = new AnnotationSync(eventBus, bridge);
-    const frameObserver = new FrameObserver(element);
+    this.discovery = new Discovery(window, { server });
```
robertknight, line 34: Should these fields be private? If so they should use an underscore prefix per our conventions. When designing the API for a class I recommend to default to making fields private and reducing the API surface area unless you know of a good use case for exposing the field.

src/annotator/cross-frame.js
```diff
@@ -59,71 +59,88 @@ export class CrossFrame {
...
-     * is set up.
-     *
-     * @param {(...args: any[]) => void} callback
-     */
-    this.onConnect = callback => bridge.onConnect(callback);
+  /**
+   * Register a callback to be invoked once the connection to the sidebar
+   * is set up.
+   *
+   * @param {Parameters<Bridge['onConnect']>[0]} callback
```
robertknight, line 141: I preferred just duplicating the original type here (`(...args: any[]) => void`). This is easier for readers to see what the type is at a glance than mentally unpack the `Parameters<...>[0]` expression. In general I would tend to only use things like `Parameters<...>`/`ReturnType<...>` where the expression is dynamic or very complex.

robertknight (commented):
Changing this class to use ES methods is fine, but please take care to choose the privacy of fields appropriately and indicate private fields with an underscore prefix [1]. As a general design principle, I recommend to keep the API surface area of classes small by exposing only what is necessary.

[1] We might switch to using `#` in future as it is part of the language and might enable better tooling, but for now `_` is the convention.

src/annotator/cross-frame.js
esanzgar: Agree.

src/annotator/cross-frame.js
esanzgar: I prepended all internal fields with underscore on a follow up commit.

I suggest you merge first your pending PR https://github.com/hypothesis/client/pull/3611 and then I can resolve the conflicts.

robertknight (commented):
This branch will need to be rebased to account for the replacement of the `Discovery` class in master, but LGTM in principle.

robertknight:
It might make sense for the changes in this PR to come _before_  those in `decouple-annotation-sync-from-cross-frame`, since this PR is much more straightforward conceptually.

src/annotator/cross-frame.js
```diff
@@ -55,70 +56,87 @@ export class CrossFrame {
...
+  /**
+   * Subscribe to an event from the sidebar.
+   *
+   * @param {string} method
+   * @param {(...args: any[]) => void} listener -- Final argument is an optional
+   *   callback of the type: `(error: string|Error|null, ...result: any[]) => void`.
+   *   This callback must be invoked in order to respond (via `postMessage`)
+   *   to the other frame/s with a result or an error.
+   * @throws {Error} If trying to register a callback after a channel has already been created
+   * @throws {Error} If trying to register a callback with the same name multiple times
```
robertknight, line 113: Rather than repeating the documentation from the `Bridge#on` method, it would probably be safer to just write something like: 

```js
/**
 * Subscribe to an event from the sidebar.
 *
 * @see {Bridge.on} for details.
 */
```

This way the documentation won't get out of date if `Bridge.on` changes in future.

src/annotator/test/cross-frame-test.js
```diff
@@ -29,7 +29,7 @@ describe('CrossFrame', () => {
     proxyBridge = sinon.stub().returns(fakeBridge);
 
     $imports.$mock({
-      './annotation-sync': proxyAnnotationSync,
+      './annotation-sync': { AnnotationSync: proxyAnnotationSync },
```
robertknight, line 32: `proxyAnnotationSync` is a strange name in this context. The convention in our code for mocking a `FooBar` class is to call the instance mock `fakeFooBar` and the constructor/class mock `FakeFooBar`.

robertknight (changes requested):
This looks good except that I would suggest not to repeat all the `Bridge.on` documentation in `CrossFrame.on`, since it is likely to get out of date. Instead you can just use `@see {Bridge.on}` to point the reader at the relevant documentation.

src/annotator/test/cross-frame-test.js
esanzgar: Implemented in commit number 5

src/annotator/cross-frame.js
esanzgar: Implemented in commit number 4.

The only problem with this approach is that we lose the types for the parameters.

src/annotator/cross-frame.js
```diff
@@ -55,70 +56,76 @@ export class CrossFrame {
...
-     *
-     * @param {string} event
-     * @param {(...args: any[]) => void} callback
-     */
-    this.on = (event, callback) => bridge.on(event, callback);
+  /**
+   * Subscribe to an event from the sidebar.
+   *
+   * @see {Bridge.on} for details.
+   */
```
robertknight, line 108: Ah there was a misunderstanding here. I meant to suggest to leave the `@param {type} name` comments in place, but remove the explanatory comments.

src/annotator/cross-frame.js
```diff
@@ -55,70 +56,76 @@ export class CrossFrame {
...
-     *
-     * @param {string} method
-     * @param {any[]} args
-     */
-    this.call = (method, ...args) => bridge.call(method, ...args);
+  /**
+   * Call an RPC method exposed by the sidebar to the annotator.
+   *
+   * @see {Bridge.call} for details.
+   */
```
robertknight, line 117: See above comment about `Bridge.on`.

robertknight (approved):
Hi @esanzgar - There was a slight misunderstanding about removing the docs for `on` and `call`. I intended to keep the types but remove everything else. Otherwise LGTM.

src/annotator/cross-frame.js
esanzgar: I mended to include only the `@param` and the `@see` for more explanation

PR #3618, opened 2021-07-28
Remove `Discovery` class

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/3611**~~

After https://github.com/hypothesis/client/pull/3611 switches communication to use a new mechanism for establishing a connection between guest and sidebar frames, the previous "discovery" mechanism implemented by `src/shared/discovery.js` is no longer used. This PR this removes it.


esanzgar:
👍🏼 

What about keeping the class for a few days in case we need to revert back?

esanzgar (approved):

robertknight:
> What about keeping the class for a few days in case we need to revert back?

We could do. Although all we have to do in that case is just revert this PR. The main reason I can see to delay deleting code is if there is a risk that you make a bunch of other changes in the same area of the code which then has to be patched up if you revert an earlier change.

robertknight:
This is now rebased and ready to land. Per @esanzgar's previous comment, I'm going to let the changes from https://github.com/hypothesis/client/pull/3611 bake in production for a day or so before landing this.

robertknight:
All seems good in production, so I'm going to merge this.

PR #3620, opened 2021-07-30
Add tall iframe manual test that mimics VitalSource reader

robertknight:
Add a test case for tall iframes where the height of the frame is set to match the document height. This mimics how the new VitalSource Bookshelf reader presents book content. See https://github.com/hypothesis/client/issues/3590#issuecomment-886229080.


lyzadanger (approved):
This does indeed appear to generate a tall iFrame!

PR #3623, opened 2021-07-30
Replace query-string with URLSearchParams in APIService

robertknight:

It replaces the use of the query-string dependency with URLSearchParams in `sidebar/services/api.js`. There are no functional changes here. The new code is a little longer because we have to explicitly handle API parameter values which are arrays now. There is an existing test for this.


lyzadanger (approved):
This looks sensible. Seems like one to keep an eye on when/after it's deployed, though it seems due diligence has been done (better testing, e.g.)!

PR #3630, opened 2021-08-02
Refactor long promise chain to async/await in API client

robertknight:
This PR is some refactoring in preparation for improving the handling of a situation where API calls return 200 responses but the response body is empty, causing `Response.json()` to fail. I haven't confirmed with absolute certainty that this is the cause, but it looks like the most likely explanation for [some Sentry issues](https://sentry.io/organizations/hypothesis/issues/1274716662/events/72e201e05ffb4bc58d10d84dddc1047f/activity/?project=69811&query=is%3Aunresolved).

This PR refactors a lengthy promise chain in the function that executes API calls to use async/await, to make it easier to follow the control flow. There should be no functional changes.

Related issue: https://github.com/hypothesis/client/issues/3631

### DIFF
--- a/src/sidebar/services/api.js
+++ b/src/sidebar/services/api.js
@@ -103,88 +103,79 @@ function createAPICall(
   route,
   { getAccessToken, getClientId, onRequestStarted, onRequestFinished }
 ) {
-  return (params, data) => {
+  return async (params, data) => {
     onRequestStarted();
+    try {
+      const [linksMap, token] = await Promise.all([links, getAccessToken()]);
+      const descriptor = get(linksMap, route);
 
-    return Promise.all([links, getAccessToken()])
-      .then(([links, token]) => {
-        const descriptor = get(links, route);
-        const headers = {
-          'Content-Type': 'application/json',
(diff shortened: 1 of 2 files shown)


src/sidebar/services/api.js
```diff
@@ -99,75 +99,66 @@ function createAPICall(
...
+        // Fetch failures to be either handled in particular ways higher up
+        // or for them to be ignored in error reporting (see `sentry` config).
+        throw new Error(`Fetch operation failed for URL '${url}'`);
+      }
+
+      const status = response.status;
+
+      let responseBody;
+      if (status >= 200 && status !== 204 && status < 500) {
+        responseBody = await response.json();
```
robertknight, line 148: This is the `Response.json()` call which I believe is triggering errors due to the response having an empty body. Unfortunately the relevant Sentry reports are missing a useful stack trace.

esanzgar (approved):

src/sidebar/services/api.js
robertknight: Yes, that's exactly the plan.

PR #3644, opened 2021-08-04
Add missing types

esanzgar:
No functional change.

The added type has surfaced that `iframe.contentDocument` could be
`null`. This was no considered before. I didn't deviated from the
original behaviour and I casted the value. If in the future we found
issues we should see those in the Sentry reports.

### DIFF
--- a/src/annotator/util/frame-util.js
+++ b/src/annotator/util/frame-util.js
@@ -15,29 +15,45 @@ export function findFrames(container) {
   return Array.from(container.querySelectorAll('iframe[enable-annotation]'));
 }
 
-// Check if the iframe has already been injected
+/**
+ * Check if the iframe has already been injected
+ *
+ * @param {HTMLIFrameElement} iframe
+ */
 export function hasHypothesis(iframe) {
-  return '__hypothesis' in iframe.contentWindow;
+  const iframeWindow = /** @type {Window} */ (iframe.contentWindow);
+  return '__hypothesis' in iframeWindow;
 }
 
-// Inject embed.js into the iframe
-export function injectHypothesis(iframe, scriptUrl, config) {
+/**
+ * Inject the client's boot script into the iframe. The iframe must be from the
+ * same origin as the current window.
+ *
+ * @param {HTMLIFrameElement} iframe
+ * @param {string} scriptSrc
+ * @param {Record<string, any>} config
+ */
+export function injectHypothesis(iframe, scriptSrc, config) {
   const configElement = document.createElement('script');
   configElement.className = 'js-hypothesis-config';
   configElement.type = 'application/json';
   configElement.innerText = JSON.stringify(config);
 
-  const src = scriptUrl;
   const embedElement = document.createElement('script');
   embedElement.className = 'js-hypothesis-embed';
   embedElement.async = true;
-  embedElement.src = src;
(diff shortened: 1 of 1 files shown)


src/annotator/util/frame-util.js
```diff
@@ -15,29 +15,43 @@ export function findFrames(container) {
   return Array.from(container.querySelectorAll('iframe[enable-annotation]'));
 }
 
-// Check if the iframe has already been injected
+/**
+ * Check if the iframe has already been injected
+ *
+ * * @param {HTMLIFrameElement} iframe
```
robertknight, line 21:
```suggestion
 * @param {HTMLIFrameElement} iframe
```

src/annotator/util/frame-util.js
```diff
@@ -15,29 +15,43 @@ export function findFrames(container) {
...
+ */
 export function hasHypothesis(iframe) {
   return '__hypothesis' in iframe.contentWindow;
 }
 
-// Inject embed.js into the iframe
-export function injectHypothesis(iframe, scriptUrl, config) {
+/**
+ * Inject embed.js into the iframe
```
robertknight, line 28:
```suggestion
 * Inject the client's boot script into the iframe
```

It may not be obvious here what embed.js refers to, since h's `/embed.js` route has a rather generic name.

src/annotator/util/frame-util.js
```diff
@@ -15,29 +15,43 @@ export function findFrames(container) {
...
+ *
```
robertknight, line 29: I suggest adding an additional comment here that `iframe` must be same-origin with the current window. If we document this as a requirement of the function then the implementation could handle the `null` case for `contentDocument` by just throwing an error.

src/annotator/util/frame-util.js
```diff
@@ -46,18 +60,28 @@ export function isAccessible(iframe) {
   }
 }
 
+/**
+ * @param {HTMLIFrameElement} iframe
+ * @param {() => void} callback
+ */
 export function isDocumentReady(iframe, callback) {
```
robertknight, line 67: This is an existing issue, but it's not very clear from the function name and API what exactly `callback` does. A better API might be something like `waitForDocumentReady(iframe): Promise<void>`. Same with the function below.

robertknight (approved):
LGTM in the sense that this is better than before and documents some implicit assumptions about the same-originness of iframes.

The `isDocumentReady` and `isLoaded` functions have unintuitive names and APIs. The names suggest functions that return booleans or promises for booleans, when in reality they are functions that wait for the passed iframe to enter a certain state and then "return" (via a callback).

src/annotator/util/frame-util.js
esanzgar: I have added a comment.

I have done some testing to check if `iframe.contentDocument` can return `null` from same-origin iframes.  I checked in latest version of Firefox, Safari and Chrome, and `iframe.contentDocument` always returned a `Document` (even if it was an initial `about:blank`).

'injectHypothesis` effectively doesn't throw an error because it is used after `isAccessible` function , that checks the frame is from the same origin and hence accesible.

src/annotator/util/frame-util.js
esanzgar: I can changed the name in a follow up PR.

PR #3645, opened 2021-08-04
Extract functionality related to the Hypothesis injection

esanzgar:
Extracted from `CrossFrame` functionality that was related to the
injection of the Hypothesis client into its own class,
`HypothesisInjector`.

In addition, I simplified and modernised the `multi-frame-test.js`.

There are a couple of follow up questions:

- should the utility functions in `frameUtil` be consolidated in the
  `hypothesis-injector.js` file? Most of this functionality is only used
  in the `HypothesisInjector` or `FrameObserver` classes.

- should `HypothesisInjector` be combined with `FrameObserver`?


src/annotator/test/integration/multi-frame-test.js
```diff
@@ -24,43 +23,40 @@ describe('CrossFrame multi-frame scenario', () => {
...
   afterEach(() => {
     sandbox.restore();
-    crossFrame.destroy();
-    container.parentNode.removeChild(container);
-
-    $imports.$restore();
+    hypothesisInjectors.forEach(hypothesisInjector =>
+      hypothesisInjector.destroy()
+    );
+    container.remove();
```
esanzgar, line 37: I believe this should do the same as the previous line:
`container.parentNode.removeChild(container);`

src/annotator/frame-observer.js
```diff
@@ -21,6 +22,14 @@ export default class FrameObserver {
     );
   }
 
+  /**
+   * Registers two listeners: the first callback is fired when an Hypothesis frame
+   * is added, the second when an Hypothesis frame is removed.
+   * This method is expected to be called only once.
+   *
+   * @param {(frame: HTMLIFrameElement) => void } onFrameAddedCallback
```
robertknight, line 30:
```suggestion
   * @param {(frame: HTMLIFrameElement) => void} onFrameAddedCallback
```

With the improved JSDoc here the `Callback` suffix is now redundant IMO, so this could be simplified to `observe(onFrameAdded, onFrameRemoved)`.

src/annotator/hypothesis-injector.js
```diff
@@ -0,0 +1,78 @@
+import FrameObserver from './frame-observer';
+import * as frameUtil from './util/frame-util';
+
+/**
+ * @typedef {import('../shared/bridge').default} Bridge
+ * @typedef {import('../types/annotator').Destroyable} Destroyable
+ * */
```
robertknight, line 7:
```suggestion
 */
```

src/annotator/test/cross-frame-test.js
```diff
@@ -3,10 +3,12 @@ import { CrossFrame, $imports } from '../cross-frame';
 describe('CrossFrame', () => {
   let fakeAnnotationSync;
   let fakeBridge;
+  let fakeHypothesisInjector;
   let fakeEventBus;
 
   let proxyAnnotationSync;
   let proxyBridge;
+  let proxyHypothesisInjector;
```
robertknight, line 11: The `proxyHypothesisInjector` name is consistent with the above, but these variables should really be called `Fake<ClassName>` to match our convention for naming fakes for classes and class instances.

src/annotator/test/integration/multi-frame-test.js
robertknight: You're right. The use of the `element.parentNode.removeChild(element)` pattern in a few places is a legacy from when the test environment didn't support the `Element.remove` method.

robertknight (commented):
> should the utility functions in frameUtil be consolidated in the hypothesis-injector.js file? Most of this functionality is only used in the HypothesisInjector or FrameObserver classes.
>
> should HypothesisInjector be combined with FrameObserver?

Maybe. You're right that the "frame-util.js" and "FrameObserver" modules are not reused outside of `HypothesisInjector`, so there isn't a reuse case for separating the two. On this basis I think the decision comes down to whether it helps to separate these modules on the basis of:

- Dependencies / layering: The `HypothesisInjector` code has clear ties to the rest of the application whereas `FrameObserver` is an (almost) generic same-origin frame discovery utility, except for its references to the hard-coded `enable-annotation` attribute. The `frame-util` module on the other handle is muddled in this respect - some functions are kind of generic (the ones used by `FrameObserver`) and others are very context specific.
- Testability: Whether it helps testing different aspects of the code to have separate modules

My instinct from an initial read through is that the `frame-util` module is not really useful and it would make more sense to move the logic into the callers - either as helpers or just inlining the code into the callsite. There might still be a case for separation between `HypothesisInjector` and `FrameObserver` on the basis of dependencies, responsibilities and ease of testing.

src/annotator/test/cross-frame-test.js
esanzgar: Resolved in the last commit.

src/annotator/cross-frame.js
```diff
@@ -30,37 +29,13 @@ export class CrossFrame {
...
-      frameIdentifiers.delete(frame);
-    };
+    this._hypothesisInjector = new HypothesisInjector(
+      element,
+      this._bridge,
+      config
+    );
 
-    this._frameObserver.observe(injectIntoFrame, iframeUnloaded);
+    this._hypothesisInjector.init();
```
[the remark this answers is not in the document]
esanzgar: I have removed the `init` method.

esanzgar:
We have decided to move keep `HypothesisInjector` separated from `FrameObserver`. 

On a follow up PR, we will merge the functionality in `FrameUtil` either in `HypothesisInjector` or in `FrameObserver`.

src/annotator/test/integration/multi-frame-test.js
```diff
@@ -1,19 +1,18 @@
...
-import { CrossFrame, $imports } from '../../cross-frame';
+import { HypothesisInjector } from '../../hypothesis-injector';
 import { isLoaded } from '../../util/frame-util';
 
-describe('CrossFrame multi-frame scenario', () => {
-  let fakeAnnotationSync;
-  let fakeBridge;
-  let proxyAnnotationSync;
-  let proxyBridge;
+describe('HypothesisInjector-FrameObserver-frameUtil integration', () => {
```
robertknight, line 5: Naming integration tests after all the components involved is liable to get out of date after future refactoring. I would probably either use a description which refers to the overall process that is being tested (eg. "Anchoring integration test", "Threading integration test") or the main class that serves as the entry point for the functionality being tested (eg. "HypothesisInjector integration test")

src/annotator/test/integration/multi-frame-test.js
```diff
@@ -1,19 +1,18 @@
 import { DEBOUNCE_WAIT } from '../../frame-observer';
```
robertknight, line 1: This module could use a better name, eg. `integration/hypothesis-injector-test.js`.

robertknight (commented):
This generally looks good. I had some feedback on a few of the details.

src/annotator/frame-observer.js
```diff
@@ -37,22 +46,29 @@ export default class FrameObserver {
...
     if (FrameUtil.isAccessible(frame)) {
       FrameUtil.isDocumentReady(frame, () => {
-        frame.contentWindow.addEventListener('unload', () => {
+        const frameWindow = /** @type {Window} */ (frame.contentWindow);
+        frameWindow.addEventListener('unload', () => {
           this._removeFrame(frame);
         });
         this._handledFrames.push(frame);
-        this._onFrameAdded(frame);
+        this._onFrameAdded?.(frame);
```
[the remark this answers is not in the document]
esanzgar: `this._onFrameAdded` can't be `null`, I will cast and add a note. Thanks.

src/annotator/hypothesis-injector.js
```diff
@@ -0,0 +1,74 @@
...
+   * @param {HTMLIFrameElement} frame
+   */
+  _iframeUnloaded(frame) {
+    this._bridge.call('destroyFrame', this._frameIdentifiers.get(frame));
+    this._frameIdentifiers.delete(frame);
+  }
+
+  /**
+   * Remove the connection between the sidebar and annotator.
+   */
```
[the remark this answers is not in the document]
esanzgar: It is very true, however, I replaced by an alternative comment. If you wish I can deleted it, no problem.

src/annotator/test/integration/multi-frame-test.js
esanzgar: I agree, I replaced the name of the file.

src/annotator/frame-observer.js
```diff
@@ -37,22 +46,29 @@ export default class FrameObserver {
...
       // Could warn here that frame was not cross origin accessible
     }
   }
 
+  /**
+   * @param {HTMLIFrameElement} frame
+   */
   _removeFrame(frame) {
-    this._onFrameRemoved(frame);
+    this._onFrameRemoved?.(frame);
```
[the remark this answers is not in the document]
esanzgar: I have added a third commit that removes the current two step process to start the observation:

1. instantiate the `FrameObserver` constructor
2. `frameObserver.observe(callback1, callback2)`

Instead the callbacks are provided in the constructor and the observation starts immediately.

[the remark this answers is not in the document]
robertknight: This makes sense to me. I think the reason for the separate `observe` method was an attempt to follow the pattern of various browser `*Observer` APIs (MutationObserver etc.). If so it was an incorrect application of the pattern because in the browser APIs the callback is provided to the constructor and the `observe` method is used to add a target to observe.

src/annotator/frame-observer.js
```diff
@@ -10,23 +12,25 @@ let difference = (arrayA, arrayB) => {
 export const DEBOUNCE_WAIT = 40;
 
 export default class FrameObserver {
-  constructor(target) {
-    this._target = target;
+  /**
+   * @param {Element} element - element to observe
```
robertknight, line 16: This comment could be clearer about why this `element` is needed, ie. that it is the root of the DOM subtree to watch for the addition and removal of `<iframe>` elements.

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -24,43 +23,39 @@ describe('CrossFrame multi-frame scenario', () => {
...
 
   afterEach(() => {
     sandbox.restore();
-    crossFrame.destroy();
-    container.parentNode.removeChild(container);
-
-    $imports.$restore();
+    hypothesisInjectors.forEach(hypothesisInjector =>
+      hypothesisInjector.destroy()
+    );
```
robertknight, line 36: This is OK, although when the meaning of a variable in a small closure is very obvious (eg. in an array `map` or `forEach` callback) I tend to keep it short. eg. `hypothesisInjectors.forEach(injector => injector.destroy())`

src/annotator/hypothesis-injector.js
```diff
@@ -0,0 +1,73 @@
...
+    this._frameIdentifiers = new Map();
+    this._frameObserver = new FrameObserver(
+      element,
+      frame => this._injectIntoFrame(frame),
+      frame => this._iframeUnloaded(frame)
+    );
+  }
+
+  /**
+   * Disables the injection of the Hypothesis client into children iframe/s.
```
robertknight, line 36: This is now a useful comment 👍 

```suggestion
   * Disables the injection of the Hypothesis client into child iframes.
```

robertknight (approved):
This needs a rebase but otherwise LGTM.

PR #3646, opened 2021-08-05
Use new-style confirm prompt in all browsers

robertknight:
_Some recent discussion amongst the Twitterati reminded me that we never completed the work to transition `window.confirm` to a custom dialog. This PR finishes this off._
 
We previously implemented a custom confirm dialog for use in browsers
which disallow use of `window.confirm` in cross-origin iframes (Chrome 9x) but
kept `window.confirm` in other browsers while the design of the new
dialogs was polished.

The polishing has now been done, so this commit switches all confirm
prompts to use the new dialog. This makes the experience look better
and be consistent across browsers.

As a reminder, here is what the new dialogs look like:

Delete annotation prompt

Discard drafts prompt

### DIFF
--- a/src/shared/prompts.js
+++ b/src/shared/prompts.js
@@ -23,16 +23,6 @@ export async function confirm({
   message,
   confirmAction = 'Yes',
 }) {
-  const start = Date.now();
-
-  // Use the legacy `window.confirm` API where available until we've polished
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
Looks good locally!

PR #3647, opened 2021-08-05
Add information about loaded scripts to error reports

robertknight:
Add information to Sentry error reports about what JavaScript files were included on the page. The information will show up as the "loaded_scripts" key in the "Additional Data" section of Sentry reports.

I have a hypothesis that Sentry issues like [this one](https://sentry.io/organizations/hypothesis/issues/2528337318/events/e176f165836149b99d439fa71e69453c/?project=69811&query=is%3Aunresolved) might be caused by unwanted `<script>` tags injected by extension or the browser. We try to block most unexpected scripts using strict Content-Security-Policy settings, but extension and custom browser-injected scripts may be able to bypass this.

If this information turns out not to be useful I will just remove it.


esanzgar (approved):
👍🏼

PR #3648, opened 2021-08-05
Add code to debug ThreadList effect exception

robertknight:
Add code to debug an error [1] triggered during an effect in the
`ThreadList` component.

When the effect is run, an attempt to get a reference to a DOM node
rendered by the component using `getElementById` is failing. This could
be because:

 - The component has been unmounted (I didn't think effects could run
   after unmounting, but I'm not certain)
 - The component has been rendered but the DOM has not yet been attached
   to the document
 - The `visibleThreads` value captured by the effect does not match the
   most recently rendered output.

I couldn't reproduce the issue locally. The debugging code added here
should narrow down which of these is happening.

[1] https://sentry.io/organizations/hypothesis/issues/2554918407/ (I have seen some other reports as well with slightly different errors, depending on the browser. See https://sentry.io/organizations/hypothesis/issues/?project=69811&query=getComputedStyle&statsPeriod=14d)

### DIFF
--- a/src/sidebar/components/ThreadList.js
+++ b/src/sidebar/components/ThreadList.js
@@ -1,4 +1,10 @@
-import { useEffect, useLayoutEffect, useMemo, useState } from 'preact/hooks';
+import {
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'preact/hooks';
 import debounce from 'lodash.debounce';
 
 import { useStoreProxy } from '../store/use-store';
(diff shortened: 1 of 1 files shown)


esanzgar (approved):
I wonder if `useLayoutEffect` would be a better choice in this case, since it is triggered _after_ DOM mutations.

Alternatively, instead of searching for the element by `id`, maybe `visibleThreads` could be an array of `{component, ref}`, so there is not need for `document.getElementById`.

robertknight:
> I wonder if useLayoutEffect would be a better choice in this case, since it is triggered after DOM mutations.
>
> Alternatively, instead of searching for the element by id, maybe visibleThreads could be an array of {component, ref}, so there is not need for document.getElementById.

Both of those suggestions make sense to me. I'd still like to understand the problem with the current code better in case it teaches us something useful about when effects are or can be run.

PR #3649, opened 2021-08-05
Remove deprecated uses of `window.postMessage`

esanzgar:
After #3611 there is no need to support `window.postMessage` on `RPC` and `Bridge` classes.

### DIFF
--- a/src/shared/frame-rpc.js
+++ b/src/shared/frame-rpc.js
@@ -179,12 +179,12 @@ export class RPC {
 
         this._port.postMessage(message);
       };
-      this._methods[msg.method].call(this._methods, ...msg.arguments, callback);
+      this._methods[msg.method](...msg.arguments, callback);
     } else if ('response' in msg) {
       const cb = this._callbacks[msg.response];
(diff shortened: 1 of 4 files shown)


src/shared/frame-rpc.js
```diff
@@ -59,42 +59,15 @@ const PROTOCOL = 'frame-rpc';
  */
 export class RPC {
   /**
-   * Create an RPC client for sending RPC requests from `sourceFrame` to
-   * `destFrame`, and receiving RPC responses from `destFrame` to `sourceFrame`.
+   * Create an RPC client for sending and receiving RPC message using a
```
robertknight, line 62: The class documentation could use an update as well. eg. 

```
RPC provides remote procedure calls between frames.

It uses the Channel Messaging API [1] for inter-frame communication.

[1] https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API
```

robertknight (approved):
LGTM. Added a minor note about the RPC class documentation.

esanzgar:
I updated the class documentation.

PR #3650, opened 2021-08-05
Simplify code on `frame-rpc.js`

esanzgar:
The `methods` constructor 's argument could be an object that uses `this`.
However, we do not rely in this feature, so I am suggesting to simplify
the execution of the callbacks.


robertknight (approved):
LGTM. I think it is better not to have the capability to reference `this` within an RPC callback, as its not obvious what it refers to.

PR #3651, opened 2021-08-05
Register a `connect` listener at class instantiation

esanzgar:
Previously, if the user tried to register a listener called `connect`, it
could lead to problems and result in the channel not be connected.

With this change, we reserve the `connect` name for internal purposes,
and if a user tries to register the `connect` listener it would result
on an error been raised.

### DIFF
--- a/src/shared/bridge.js
+++ b/src/shared/bridge.js
@@ -16,6 +16,10 @@ export class Bridge {
     this.channelListeners = {};
     /** @type {Array<(channel: PortRPC) => void>} */
     this.onConnectListeners = [];
+
+    // `connect` is registered with a callback so it triggers a `postMessage`
+    // response from the reciprocal port.
+    this.on('connect', cb => cb());
   }
 
   /**
@@ -38,10 +42,7 @@ export class Bridge {
    * @return {PortRPC} - Channel for communicating with the reciprocal port.
    */
   createChannel(port) {
-    const listeners = { connect: cb => cb(), ...this.channelListeners };
-
-    // Set up a channel
-    const channel = new PortRPC(port, listeners);
+    const channel = new PortRPC(port, this.channelListeners);
 
     let connected = false;
     const ready = () => {


robertknight:
The change here seems reasonable, but looking at it made me question whether we need the whole `onConnect` mechanism any more.

`onConnect` is currently used for two purposes, both of which have better alternatives:

1. The guest in the host frame uses it to determine when the sidebar is ready and it publishes a `panelReady` message on the event bus in response, to which `Sidebar` responds by showing the UI. This is a rather roundabout way of doings things as the `Sidebar` class could just use its own `ready` promise.
2. The sidebar app uses this in `FrameSyncService` to add a new frame to the store and query its document information. However the `bridge.onConnect` call is just above the code that calls `bridge.createChannel`, which returns the same `RPC` instance that the `onConnect` handler receives.

A quick sketch of how `onConnect` could be eliminated by:

1. Passing the result of `bridge.createChannel` to `addFrame` in the sidebar
2. By eliminating the `panelReady` event in the guest and instead making the `Sidebar` update itself when its existing `ready` promise resolves

```diff
diff --git a/src/annotator/cross-frame.js b/src/annotator/cross-frame.js
index f1c6b8a31..4893f1000 100644
--- a/src/annotator/cross-frame.js
+++ b/src/annotator/cross-frame.js
@@ -112,13 +112,5 @@ export class CrossFrame {
      * @param {any[]} args
      */
     this.call = (method, ...args) => bridge.call(method, ...args);
-
-    /**
-     * Register a callback to be invoked once the connection to the sidebar
-     * is set up.
-     *
-     * @param {(...args: any[]) => void} callback
-     */
-    this.onConnect = callback => bridge.onConnect(callback);
   }
 }
diff --git a/src/annotator/guest.js b/src/annotator/guest.js
index 091b49f60..4e69ad038 100644
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -164,9 +164,10 @@ export default class Guest {
     // The "top" guest instance will have this as null since it's in a top frame not a sub frame
     this._frameIdentifier = config.subFrameIdentifier || null;
 
+    this.setVisibleHighlights(config.showHighlights === 'always');
+
     // Setup connection to sidebar.
     this.crossframe = new CrossFrame(this.element, eventBus, config);
-    this.crossframe.onConnect(() => this._setupInitialState(config));
     this._connectSidebarEvents();
 
     this._sideBySideActive = false;
@@ -271,11 +272,6 @@ export default class Guest {
     }
   }
 
-  _setupInitialState(config) {
-    this._emitter.publish('panelReady');
-    this.setVisibleHighlights(config.showHighlights === 'always');
-  }
-
   _connectAnnotationSync() {
     this._emitter.subscribe('annotationDeleted', annotation => {
       this.detach(annotation);
@@ -611,6 +607,10 @@ export default class Guest {
     this._emitter.publish('highlightsVisibleChanged', shouldShowHighlights);
   }
 
+  areHighlightsVisible() {
+    return this._visibleHighlights;
+  }
+
   /**
    * Return the scrollable element that contains the main document content.
    *
diff --git a/src/annotator/sidebar.js b/src/annotator/sidebar.js
index 49e6bd2fc..0dbf830c3 100644
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -112,13 +112,6 @@ export default class Sidebar {
 
     this._listeners = new ListenerCollection();
 
-    this._emitter.subscribe('panelReady', () => {
-      // Show the UI
-      if (this.iframeContainer) {
-        this.iframeContainer.style.display = '';
-      }
-    });
-
     this._emitter.subscribe('beforeAnnotationCreated', annotation => {
       // When a new non-highlight annotation is created, focus
       // the sidebar so that the text editor can be focused as
@@ -128,15 +121,6 @@ export default class Sidebar {
       }
     });
 
-    if (
-      config.openSidebar ||
-      config.annotations ||
-      config.query ||
-      config.group
-    ) {
-      this._emitter.subscribe('panelReady', () => this.open());
-    }
-
     // Set up the toolbar on the left edge of the sidebar.
     const toolbarContainer = document.createElement('div');
     this.toolbar = new ToolbarController(toolbarContainer, {
@@ -151,9 +135,11 @@ export default class Sidebar {
       this.toolbar.useMinimalControls = false;
     }
 
+    this.toolbar.highlightsVisible = guest.areHighlightsVisible();
     this._emitter.subscribe('highlightsVisibleChanged', visible => {
       this.toolbar.highlightsVisible = visible;
     });
+
     this._emitter.subscribe('hasSelectionChanged', hasSelection => {
       this.toolbar.newAnnotationType = hasSelection ? 'annotation' : 'note';
     });
@@ -210,6 +196,21 @@ export default class Sidebar {
         }
       });
     });
+
+    // Show the sidebar once it has loaded.
+    this.ready.then(() => {
+      if (this.iframeContainer) {
+        this.iframeContainer.style.display = '';
+      }
+      if (
+        config.openSidebar ||
+        config.annotations ||
+        config.query ||
+        config.group
+      ) {
+        this.open();
+      }
+    });
   }
 
   destroy() {
diff --git a/src/shared/bridge.js b/src/shared/bridge.js
index 70dfa9660..2417282c4 100644
--- a/src/shared/bridge.js
+++ b/src/shared/bridge.js
@@ -14,8 +14,6 @@ export default class Bridge {
     this.links = [];
     /** @type {Record<string, (...args: any[]) => void>} */
     this.channelListeners = {};
-    /** @type {Array<(channel: RPC) => void>} */
-    this.onConnectListeners = [];
   }
 
   /**
@@ -39,25 +37,8 @@ export default class Bridge {
    */
   createChannel(port) {
     const listeners = { connect: cb => cb(), ...this.channelListeners };
-
-    // Set up a channel
     const channel = new RPC(port, listeners);
-
-    let connected = false;
-    const ready = () => {
-      if (connected) {
-        return;
-      }
-      connected = true;
-      this.onConnectListeners.forEach(cb => cb(channel));
-    };
-
-    // Fire off a connection attempt
-    channel.call('connect', ready);
-
-    // Store the newly created channel in our collection
     this.links.push(channel);
-
     return channel;
   }
 
@@ -151,14 +132,4 @@ export default class Bridge {
     this.channelListeners[method] = listener;
     return this;
   }
-
-  /**
-   * Add a listener to be called upon a new connection.
-   *
-   * @param {(channel: RPC) => void} listener
-   */
-  onConnect(listener) {
-    this.onConnectListeners.push(listener);
-    return this;
-  }
 }
diff --git a/src/sidebar/services/frame-sync.js b/src/sidebar/services/frame-sync.js
index c16ff7762..7920b0ee8 100644
--- a/src/sidebar/services/frame-sync.js
+++ b/src/sidebar/services/frame-sync.js
@@ -1,5 +1,6 @@
 import debounce from 'lodash.debounce';
 
+import { RPC } from '../../shared/frame-rpc';
 import bridgeEvents from '../../shared/bridge-events';
 import { isReply, isPublic } from '../helpers/annotation-metadata';
 import { watch } from '../util/watch';
@@ -220,6 +221,8 @@ export class FrameSyncService {
      * Query the Hypothesis annotation client in a frame for the URL and metadata
      * of the document that is currently loaded and add the result to the set of
      * connected frames.
+     *
+     * @param {RPC} channel
      */
     const addFrame = channel => {
       channel.call('getDocumentInfo', (err, info) => {
@@ -236,8 +239,6 @@ export class FrameSyncService {
       });
     };
 
-    this._bridge.onConnect(addFrame);
-
     // Listen for messages from new guest frames that want to connect.
     //
     // The message will include a `MessagePort` to use for communication with
@@ -254,7 +255,8 @@ export class FrameSyncService {
         return;
       }
       const port = e.ports[0];
-      this._bridge.createChannel(port);
+      const channel = this._bridge.createChannel(port);
+      addFrame(channel);
     });
 
     // Notify host frame that it is ready for guests to connect to it.
```

esanzgar:
@robertknight, I understand your argument, however I would suggest we delay the removal of the `onConnect` method until the new inter-frame communication is setup.

I would suggest we accept this initial simplification first. Would you agree on that?

src/shared/bridge.js
```diff
@@ -16,6 +16,8 @@ export default class Bridge {
     this.channelListeners = {};
     /** @type {Array<(channel: RPC) => void>} */
     this.onConnectListeners = [];
+
+    this.on('connect', cb => cb());
```
robertknight, line 20: This could do with a comment here indicating why 'connect' is being registered with an apparently no-op listener.

robertknight (approved):
If you want to merge this as an incremental step that's OK with me, though I really think the whole `onConnect` mechanism is something we should be able to remove soon.

src/shared/bridge.js
esanzgar: I added a comment about the callback.

PR #3652, opened 2021-08-05
Replaced `container.parentNode.removeChild(...)`

esanzgar:
`container.parentNode.removeChild(...)` is an old pattern. The simpler `Element.remove` is currently supported by our
test environment.

### DIFF
--- a/src/annotator/test/guest-test.js
+++ b/src/annotator/test/guest-test.js
@@ -846,7 +846,7 @@ describe('Guest', () => {
     });
 
     afterEach(() => {
-      document.body.removeChild(el);
+      el.remove();
     });
 
     it("doesn't mark an annotation lacking targets as an orphan", () => {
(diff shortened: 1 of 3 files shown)


robertknight (approved):

robertknight:
A quick grep of the codebase turned up some other uses of `removeChild` that could be eliminated:

```
[I]  ~/h/client > ag removeChild src/
src/annotator/test/integration/multi-frame-test.js
49:    container.parentNode.removeChild(container);

src/annotator/test/range-util-test.js
38:    testNode.parentElement.removeChild(testNode);

src/annotator/test/integration/anchoring-test.js
75:    container.parentNode.removeChild(container);

src/annotator/test/guest-test.js
849:      document.body.removeChild(el);
```

esanzgar:
Thanks for pointing me to the other cases. I modified all, except  `multi-frame-test.js` for which there is a pending PR (that collaterally modifies this issue).

PR #3654, opened 2021-08-06
Change default export to named export for `Bridge`

esanzgar:
Following our conventions only Preact components should have default
exports.

### DIFF
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -108,7 +108,7 @@ import LaunchErrorPanel from './components/LaunchErrorPanel';
 import { ServiceContext } from './service-context';
 
 // Services.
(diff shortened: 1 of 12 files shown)


src/sidebar/index.js
```diff
@@ -108,7 +108,7 @@ import LaunchErrorPanel from './components/LaunchErrorPanel';
 import { ServiceContext } from './service-context';
 
 // Services.
-import bridgeService from '../shared/bridge';
+import { Bridge as BridgeService } from '../shared/bridge';
```
robertknight, line 111:
```suggestion
import { Bridge } from '../shared/bridge';
```

The other services have a `Service` suffix because the actual class does.

robertknight (approved):

PR #3663, opened 2021-08-09
Remove `waitForMessageDelivery` in `rpc-bridge-test.js`

esanzgar:
The test that used this utility function was failing randomly. It seems
that the `onConnect` callback is not always been triggered on the next
JS cycle.

### DIFF
--- a/src/shared/test/integration/rpc-bridge-test.js
+++ b/src/shared/test/integration/rpc-bridge-test.js
@@ -12,10 +12,6 @@ describe('PortRPC-Bridge integration', () => {
     return bridge;
   }
 
-  function waitForMessageDelivery() {
-    return new Promise(resolve => setTimeout(resolve, 0));
-  }
-
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


esanzgar:
An example or such failures:
https://jenkins.hypothes.is/job/client/job/dependabot%252Fnpm_and_yarn%252Fpuppeteer-10.2.0/1/console

```
Chrome Headless 86.0.4240.111 (Linux x86_64) PortRPC-Bridge integration establishing a connection should invoke Bridge `onConnect` callbacks after connecting FAILED
	AssertionError: expected 0 to equal 3
	    at Context.<anonymous> (/tmp/shared/test/integration/rpc-bridge-test.js:48:14 <- /tmp/6e96756abf2b883e442c6f3faacf46d5.browserify.js:241720:14)
```

robertknight (approved):
I haven't analyzed this carefully, but the revised approach looks safer to me.

PR #3665, opened 2021-08-09
Add debug code to confirm render / effect mismatch

robertknight:
It appears that an effect in `ThreadList` can be called (see [1]) when the
variables captured by the effect do not match the current rendered
output. Add some debug code to verify more directly that this is the
case.

[1] https://sentry.io/organizations/hypothesis/issues/2562006490/

### DIFF
--- a/src/sidebar/components/ThreadList.js
+++ b/src/sidebar/components/ThreadList.js
@@ -173,6 +173,10 @@ function ThreadList({ threads }) {
     };
   }, []);
 
+  // Debugging code related to https://sentry.io/organizations/hypothesis/issues/2554918407/
+  const lastVisibleThreads = useRef(visibleThreads);
+  lastVisibleThreads.current = visibleThreads;
+
   const rootRef = useRef(/** @type {HTMLDivElement|null} */ (null));
 
   // When the set of visible threads changes, recalculate the real rendered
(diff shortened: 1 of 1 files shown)


src/sidebar/components/ThreadList.js
```diff
@@ -204,8 +208,14 @@ function ThreadList({ threads }) {
             );
           }
 
-          // Check if effect was called with `visibleThreads` value which does
-          // not match most recently rendered `visibleThreads`.
```
robertknight, line 208: The old code _suggested_ that `visibleThreads` did not match the `visibleThreads` for the most recent render, but I wanted to be more sure of this.

esanzgar (approved):

robertknight:
To recap some discussion on Slack yesterday, I managed to reproduce the "Element not found" error sporadically on a PDF which had annotations which caused an error during rendering (see https://github.com/hypothesis/client/pull/3667).

I think what is happening here is that the earlier rendering error was causing Preact's internal state to become corrupted.  This resulted in a mismatch between the effect being run and the component output (ie. the effect callback closed over variables that were different from those used to generate the rendered output).

The error fixed in https://github.com/hypothesis/client/pull/3667 was easy to reproduce, but for some reason it didn't get reported to Slack. If it had been, we could have made a connection between the two errors based on the affected URLs.

Issue #3666, opened 2021-08-09
Invalid `uri` field in annotations causes rendering error in `AnnotationHeader`


robertknight:
**Steps to reproduce:**

Go to https://hypothes.is/a/YinbZO0iT1qV5a7fhGAlxw and observe that the client fails to load. In the console the following error is reported:

```
Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL
    at r (annotation-metadata.js:17)
    at i (annotation-metadata.js:74)
    at n.domainAndTitle (annotation-metadata.js:42)
    at v.constructor (AnnotationHeader.js:93)
    at v.D [as render] (preact.js:1)
```

For some reason this error is not appearing in Sentry.

The JSON data returned by h for this annotation is (API link: https://hypothes.is/api/annotations/YinbZO0iT1qV5a7fhGAlxw) has a `uri` field which contains `[object URL]`:

```json
{
    "consumer": "00000000-0000-0000-0000-000000000000",
    "id": "YinbZO0iT1qV5a7fhGAlxw",
    "created": "2015-03-12T15:20:25.080611+00:00",
    "updated": "2015-03-12T15:20:25.080626+00:00",
    "user": "acct:raz@hypothes.is",
    "uri": "[object URL]",
    "text": "This would be an amazing feature.",
    "tags": [],
    "group": "__world__",
    "permissions": {
        "read": [
            "group:__world__"
        ],
        "admin": [
            "acct:raz@hypothes.is"
        ],
        "update": [
            "acct:raz@hypothes.is"
        ],
        "delete": [
            "acct:raz@hypothes.is"
        ]
    },
    "target": [
        {
            "source": "[object URL]",
            "selector": [
                {
                    "type": "RangeSelector",
                    "endOffset": 124,
                    "startOffset": 0,
                    "endContainer": "/ul[3]/li[5]/p[1]/font[1]",
                    "startContainer": "/ul[3]/li[5]/p[1]/font[1]"
                },
                {
                    "end": 3446,
                    "type": "TextPositionSelector",
                    "start": 3324
                },
                {
                    "type": "TextQuoteSelector",
                    "exact": "Being able to follow a stream of their own, an author’s or a group’s annotations, independent of the websites they are on.",
                    "prefix": "ation list and the annotations.",
                    "suffix": "Being able to follow someone el"
                },
                {
                    "type": "FragmentSelector",
                    "value": ""
                }
            ]
        }
    ],
    "document": {
        "title": [
[...]

PR #3667, opened 2021-08-09
Handle annotation URI parsing errors

robertknight:
Since h has very liberal URI parsing compared to the browser's URL
constructor, as well as some completely invalid data on old annotations,
the client needs to handle errors parsing annotation `uri` values.

We may in future clean up the obviously-invalid URIs in h and introduce stricter validation, but that is a bigger project.

Surprisingly `annotation-metadata.js` is the only place I found in the
client that attempts to do this, everywhere else is just using the `uri`
field as a string.

Fixes https://github.com/hypothesis/client/issues/3666

### DIFF
--- a/src/sidebar/helpers/annotation-metadata.js
+++ b/src/sidebar/helpers/annotation-metadata.js
@@ -6,23 +6,32 @@
 /** @typedef {import('../../types/api').TextPositionSelector} TextPositionSelector */
 /** @typedef {import('../../types/api').TextQuoteSelector} TextQuoteSelector */
 
-/** Extract a URI, domain and title from the given domain model object.
+/**
+ * Extract document metadata from an annotation.
  *
  * @param {Annotation} annotation
- *
  */
 export function documentMetadata(annotation) {
   const uri = annotation.uri;
 
-  let domain = new URL(uri).hostname;
-  let title = domain;
-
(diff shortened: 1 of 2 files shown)


src/sidebar/helpers/annotation-metadata.js
```diff
@@ -6,25 +6,34 @@
 /** @typedef {import('../../types/api').TextPositionSelector} TextPositionSelector */
 /** @typedef {import('../../types/api').TextQuoteSelector} TextQuoteSelector */
 
-/** Extract a URI, domain and title from the given domain model object.
```
robertknight, line 9: The terminology "domain model object" is not something we use any more in the client. We just say "an annotation", "a group" etc.

src/sidebar/helpers/annotation-metadata.js
```diff
@@ -107,7 +117,7 @@ function titleTextFromAnnotation(annotation) {
 /**
  * Return `true` if the given annotation is a reply, `false` otherwise.
  *
- *  @param {Annotation} annotation - An annotation domain model object.
+ * @param {Annotation} annotation
```
robertknight, line 120: The comment here added no value.

esanzgar (approved):

PR #3668, opened 2021-08-10
Remove `isLoaded` from `frame-util.js`

esanzgar:
I believe it would be faster to rely on `DOMContentLoaded` event
(`isDocumentReady` function). This event triggers when the DOM is fully
loaded and parsed but without waiting for the stylesheets, images, and
subframes to finish loading [1].

[1] https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event

### DIFF
--- a/src/annotator/hypothesis-injector.js
+++ b/src/annotator/hypothesis-injector.js
@@ -50,7 +50,7 @@ export class HypothesisInjector {
       return;
     }
 
-    frameUtil.isLoaded(frame, () => {
+    frameUtil.isDocumentReady(frame, () => {
       // Generate a random string to use as a frame ID. The format is not important.
       const subFrameIdentifier = Math.random().toString().replace(/\D/g, '');
       this._frameIdentifiers.set(frame, subFrameIdentifier);
--- a/src/annotator/hypothesis-injector.js
+++ b/src/annotator/hypothesis-injector.js
@@ -28,8 +28,8 @@ export class HypothesisInjector {
     this._frameIdentifiers = new Map();
     this._frameObserver = new FrameObserver(
       element,
-      frame => this._injectIntoFrame(frame),
-      frame => this._iframeUnloaded(frame)
+      frame => this._addHypothesis(frame),
+      frame => this._removeHypothesis(frame)
     );
   }
 
@@ -49,7 +49,7 @@ export class HypothesisInjector {
    *
    * @param {HTMLIFrameElement} frame
    */
-  _injectIntoFrame(frame) {
+  _addHypothesis(frame) {
     if (frameUtil.hasHypothesis(frame)) {
       return;
     }
@@ -69,7 +69,7 @@ export class HypothesisInjector {
   /**
    * @param {HTMLIFrameElement} frame
    */
-  _iframeUnloaded(frame) {
+  _removeHypothesis(frame) {
     this._bridge.call('destroyFrame', this._frameIdentifiers.get(frame));
     this._frameIdentifiers.delete(frame);
   }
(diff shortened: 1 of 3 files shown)


src/annotator/hypothesis-injector.js
```diff
@@ -41,33 +41,32 @@ export class HypothesisInjector {
   }
 
   /**
-   * Inject Hypothesis client into a newly-discovered iframe.
+   * Inject Hypothesis client into a newly-discovered iframe. The DOM's iframe
+   * is guaranteed to be accessible and ready.
```
robertknight, line 45: The second sentence here implies that this is a guarantee made by _this method_. What you presumably mean however is that this is a precondition of this method (ie. that the iframe must be "accessible" and "ready" before this method is called. It would be helpful to clarify briefly what "accessible" and "ready" mean here).

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -65,7 +65,7 @@ describe('HypothesisInjector integration test', () => {
     createHypothesisInjector();
 
     const validFramePromise = new Promise(resolve => {
-      isLoaded(validFrame, () => {
+      isDocumentReady(validFrame, () => {
```
robertknight, line 68: Outside the scope of this PR, but these tests would have a more straightforward control flow if `isDocumentReady` returned a promise.

robertknight (changes requested):
Waiting for only `DOMContentLoaded` before injecting the client into iframes should be workable as long as the iframe contains static HTML and is not an SPA. This limitation was also true of waiting for the `load` event.

I think the documentation for `_addHypothesis` was confusing though as it implies that this method ensures the frame was ready before injecting the client, whereas it now _requires_ that the frame is already "ready".

src/annotator/hypothesis-injector.js
esanzgar: Yes, that's what I meant. `FrameObserver._addFrame` wrapper around `HypothesisInjector._addHypothesis` guarantees the precondition (DOM ready) is fulfilled.

src/annotator/test/integration/hypothesis-injector-test.js
esanzgar: I am currently working on this.

esanzgar:
I have changed the comment in the `_addHypothesis` method. Let me know what you think. Thanks.

src/annotator/test/integration/hypothesis-injector-test.js
esanzgar: See https://github.com/hypothesis/client/pull/3670

src/annotator/hypothesis-injector.js
```diff
@@ -43,31 +43,33 @@ export class HypothesisInjector {
   /**
    * Inject Hypothesis client into a newly-discovered iframe.
    *
+   * IMPORTANT: This method requires that the iframe is "accessible"
+   * (frame.contentDocument|contentWindow is reachable) and "ready" (DOM content
+   * is interactive) before the method is called.
```
robertknight, line 48: This is better - I think I would suggest to replace "reachable" by "not null", to be more specific and "is interactive" by "has been loaded". I think it is ambiguous whether the document is considered "interactive" at the point when `DOMContentLoaded` has fired but `load` has not (eg. on a website with many images, the page might not be useful from the user's point of view at this point, but Hypothesis can still function).

robertknight (approved):
LGTM with a couple of minor suggestions.

PR #3669, opened 2021-08-10
Remove ThreadList debugging code

robertknight:
I believe we know what was causing the issue in ThreadList where an
effect callback was sometimes called with closed-over variables which did not
match the most recent render. See [1] for details. The issue really
needs to be resolved in a more general way than trying to work around it
in individual components.

This commit removes code added for debugging this and also adds a check for a
legitimate way (ie. without anything going wrong inside Preact) that the problem
could occur, although it should never happen in the actual app as we
never render `ThreadList` in a container that is disconnected from the
document.

[1] https://github.com/hypothesis/client/pull/3665#issuecomment-895857072

### DIFF
--- a/src/sidebar/components/ThreadList.js
+++ b/src/sidebar/components/ThreadList.js
@@ -1,10 +1,4 @@
-import {
-  useEffect,
-  useLayoutEffect,
-  useMemo,
-  useRef,
-  useState,
-} from 'preact/hooks';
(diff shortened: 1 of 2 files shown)


esanzgar (approved):
I believe you need to remove the `useRef` import.

robertknight:
> I believe you need to remove the useRef import.

Oh, damn, yes. Will do.

Issue #530, opened 2017-09-01
Multiple issues caused by ad iframes on news website


robertknight:
A user reported an issue when using the client on https://www.theguardian.com/world/2016/nov/09/western-civilisation-appiah-reith-lecture where it would appear to "reload" repeatedly.

Original report: https://hypothesis.zendesk.com/agent/tickets/1407

Digging into the problem, it looks like this page has a large number (~10) of ad-generated iframes which are loaded dynamically while scrolling the page. This causes a number of issues with the client's support for iframes. Several of these are race conditions so they will not reliably reproduce all the time.

1. Sometimes the client can get injected into an iframe multiple times. I believe this is caused by `FrameObserver`'s `_addFrame` not immediately registering the iframe as "handled" if the frame has not yet finished loading. Therefore `_addFrame` can be called for the same frame multiple times, and thus the `onFrameAdded` callback can be triggered for the same frame multiple times.
2. The support for multiple iframes in the sidebar code is incomplete in some ways. See the explanation here: https://github.com/hypothesis/client/issues/527#issuecomment-326627225
3. The iframes into which we are injecting the client consist mostly of a single large image or another iframe, so there isn't really any point enabling annotation for them. We're just burning CPU cycles and using data. We have heuristics to filter the set of iframes into which the client is injected, but these are currently based purely on size. A large image/video ad is big enough to match the criteria.

Capturing the above as a checklist:

- [ ] Client injected into same iframe multiple times
- [ ] Non-deterministic anchoring messaging handling
- [x] Client injected into iframes unnecessarily

dwhly:
Also happening on this page https://www.wired.com/story/how-apple-finally-made-siri-sound-more-human?mbid=nl_090717_daily

Load in FF, activate bookmarklet, as ads cycle, they will cause annotation cards to load and unload over and over again.  cc @heatherstaines

robertknight:
Hypothesis v1.40 works around this issue for now by limiting annotation of iframes to those which are explicitly opted in by the site owner via an `enable-annotation` attribute on the `<iframe>` element.

The underlying issues mentioned above still exist in the codebase and I'll file separate issues for them.

robertknight:
Closing in favour of #539 and #540.

PR #3670, opened 2021-08-10
Move frame util functionality

esanzgar:
* Move `isAccessible`, `findFrames` and `isDocumentReady` functions to
  `frame-observer.js`, and `hasHypothesis` and `injectHypothesis` functions to
  `hypothesis-injector.js`.
* Rename and move `src/annotator/util/test/frame-util-test.js` to
`src/annotator/test/frame-observer-test.js` and add unit testing.
* Rename `isDocumentReady` to `onDocumentReady`, and convert it to a Promise-based function.


src/annotator/frame-observer.js
```diff
@@ -22,8 +15,8 @@ export default class FrameObserver {
     this._element = element;
     this._onFrameAdded = onFrameAdded;
     this._onFrameRemoved = onFrameRemoved;
-    /** @type {HTMLIFrameElement[]} */
-    this._handledFrames = [];
+    /** @type {Set<HTMLIFrameElement>} */
+    this._handledFrames = new Set();
```
robertknight, line 19: 👍

src/annotator/frame-observer.js
```diff
@@ -45,16 +38,11 @@ export default class FrameObserver {
...
-          this._removeFrame(frame);
-        });
-        this._handledFrames.push(frame);
-        this._onFrameAdded(frame);
-      });
+  async _addFrame(frame) {
+    if (isAccessible(frame)) {
+      await onDocumentReady(frame);
+      this._handledFrames.add(frame);
+      this._onFrameAdded(frame);
```
robertknight, line 45: Adding a `console.warn` in the else block would be useful, as long as this method is only being called for frames which have the `enable-annotation` attribute, so as not to generate a lot of console spam on ad-heavy web pages that contain many iframes.

src/annotator/frame-observer.js
```diff
@@ -45,16 +38,11 @@ export default class FrameObserver {
...
-        const frameWindow = /** @type {Window} */ (frame.contentWindow);
-        frameWindow.addEventListener('unload', () => {
-          this._removeFrame(frame);
-        });
-        this._handledFrames.push(frame);
-        this._onFrameAdded(frame);
-      });
+  async _addFrame(frame) {
+    if (isAccessible(frame)) {
+      await onDocumentReady(frame);
```
robertknight, line 43: Is there a timeout in place here?

src/annotator/frame-observer.js
```diff
@@ -65,22 +53,71 @@ export default class FrameObserver {
...
+ */
+export function onDocumentReady(iframe) {
+  return new Promise(resolve => {
+    const iframeDocument = /** @type {Document} */ (iframe.contentDocument);
+    if (iframeDocument.readyState === 'loading') {
+      iframeDocument.addEventListener('DOMContentLoaded', () => {
+        resolve();
+      });
+    } else {
+      resolve();
```
robertknight, line 103: I wondered whether calling `resolve` in all cases where `readyState` is not `loading` was appropriate. The only other states are `interactive` or `complete` though, so that's OK.

src/annotator/frame-observer.js
```diff
@@ -65,22 +53,71 @@ export default class FrameObserver {
...
+ *
+ * Eventually we may want annotation to be enabled by default for iframes that
+ * pass certain tests. However we need to resolve a number of issues before we
+ * can do that. See https://github.com/hypothesis/client/issues/530
+ *
+ * @param {Element} container
+ * @return {Set<HTMLIFrameElement>}
+ */
+export function findFrames(container) {
+  return new Set(container.querySelectorAll('iframe[enable-annotation]'));
```
robertknight, line 122: I wondered whether all the browsers we use support treating the `NodeList` returned by `querySelectorAll` as an iterable. It [looks like they do](https://developer.mozilla.org/en-US/docs/Web/API/NodeList#browser_compatibility).

src/annotator/hypothesis-injector.js
```diff
@@ -74,3 +73,37 @@ export class HypothesisInjector {
     this._frameIdentifiers.delete(frame);
   }
 }
+
+/**
+ * Check if the iframe has already been injected
```
robertknight, line 78:
```suggestion
 * Check if the Hypothesis client has already been injected into an iframe
```

src/annotator/test/frame-observer-test.js
```diff
@@ -0,0 +1,42 @@
+import { findFrames } from '../frame-observer';
```
robertknight, line 1: Hmm, we're missing unit tests for `FrameObserver`. Is that currently covered only by an integration test?

src/annotator/frame-observer.js
```diff
@@ -65,22 +53,71 @@ export default class FrameObserver {
...
+ * `enable-annotation` attribute.
+ *
+ * Eventually we may want annotation to be enabled by default for iframes that
+ * pass certain tests. However we need to resolve a number of issues before we
+ * can do that. See https://github.com/hypothesis/client/issues/530
+ *
+ * @param {Element} container
+ * @return {Set<HTMLIFrameElement>}
+ */
+export function findFrames(container) {
```
robertknight, line 121: `findFrames` is only used internally within `FrameObserver` and the implementation is pretty trivial. It is also inherently coupled to `FrameObserver` because the CSS selector here and the `MutationObserver` configuration need to be in-sync. I think it would make sense to inline the logic into `discoverFrames` and move the documentation here to the `FrameObserver` class.

robertknight (commented):
I had an initial look over this and left a few comments. The general direction looks good.

src/annotator/frame-observer.js
esanzgar: 👍🏼

src/annotator/test/frame-observer-test.js
esanzgar: Yes, only integration test.

src/annotator/frame-observer.js
esanzgar: 👍🏼

src/annotator/test/frame-observer-test.js
esanzgar: I added unit tests for `FrameObserver`.

src/annotator/frame-observer.js
esanzgar: I discover an 'uninitialized' `readyState` in Firefox and other subtleties. See https://github.com/hypothesis/client/pull/3686

esanzgar:
There are two lines that were not covered by the tests. I added `/* istanbul ignore next */` statements on https://github.com/hypothesis/client/pull/3686.

src/annotator/test/frame-observer-test.js
```diff
@@ -0,0 +1,138 @@
...
+
+    assert.calledOnce(onFrameRemoved);
+    assert.calledWith(onFrameRemoved, frame);
+  });
+
+  // This test doesn't work. Surprisingly, `isAccessible` returns `true` even
+  // thought the iframe is from a different domain. My suspicion is that the
+  // iframe is accessed at a time where the loading has not yet started and it
+  // is therefore accessible.
+  // it("doesn't triggers onFrameAdded when annotatable iframe is from a different domain", async () => {
```
robertknight, line 131: This seems like something worth digging into to understand what is going on, in case there are ramifications we ought to be aware of.

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -45,217 +42,162 @@ describe('HypothesisInjector integration test', () => {
...
-            'expected dynamically added frame to be modified'
-          );
-          resolve();
-        });
-      });
-    });
+    await waitForFrameObserver();
+    await onDocumentReady(frame);
+    assert.isTrue(
+      frame.contentDocument.body.hasChildNodes(),
```
robertknight, line 127: To check my understanding - this test is assuming that the frame initially had an empty `<body>` and that this will change after the client is embedded? This seems brittle to me. I think it would be better to test more directly for one or more of the elements that we expect the client to inject. I'd also recommend encapsulating that test in a helper so you can change it in one place easily.

src/annotator/frame-observer.js
```diff
@@ -1,16 +1,19 @@
...
+/**
+ * FrameObserver detects iframes added and deleted from the document.
+ *
+ * To enable annotation, an iframe must be opted-in by adding the
+ * `enable-annotation` attribute.
+ *
+ * Eventually we may want annotation to be enabled by default for iframes that
+ * pass certain tests. However we need to resolve a number of issues before we
+ * can do that. See https://github.com/hypothesis/client/issues/530
+ */
```
robertknight, line 16: Now would probably be a good time to update `FrameObserver` to not be a default export.

robertknight (changes requested):
This is looking good overall. I left feedback on various details. Regarding the commented-out test, I think it is worth digging into the issue to find out what is going on at least. What gets done about it will depend on what the issue turns out to be. Let me know if you want any help with that.

src/annotator/frame-observer.js
esanzgar: I decided to split the PR in two, it is in the following up https://github.com/hypothesis/client/pull/3686

src/annotator/frame-observer.js
esanzgar: I will add this in a follow up commit

src/annotator/hypothesis-injector.js
esanzgar: Apologies for that. I accepted the commit and GH automatically marked as resolved. However, I forgot to pull the commit to my local repo and it was lost.

I have fixed it.

src/annotator/frame-observer.js
```diff
@@ -1,16 +1,19 @@
...
-
+/**
+ * FrameObserver detects iframes added and deleted from the document.
+ *
+ * To enable annotation, an iframe must be opted-in by adding the
+ * `enable-annotation` attribute.
+ *
+ * Eventually we may want annotation to be enabled by default for iframes that
+ * pass certain tests. However we need to resolve a number of issues before we
+ * can do that. See https://github.com/hypothesis/client/issues/530
```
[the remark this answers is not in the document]
esanzgar: Do you want me to remove the comment?

src/annotator/frame-observer.js
```diff
@@ -64,23 +66,59 @@ export default class FrameObserver {
...
     this._onFrameRemoved(frame);
-
-    // Remove the frame from our list
-    this._handledFrames = this._handledFrames.filter(x => x !== frame);
   }
 
   _discoverFrames() {
-    let frames = FrameUtil.findFrames(this._element);
+    let frames = new Set(
+      /** @type {NodeListOf<HTMLIFrameElement> } */ (
```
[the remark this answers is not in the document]
esanzgar: Correct: `querySelectorAll` returns a `NodeListOf<Element>`

src/annotator/test/frame-observer-test.js
```diff
@@ -0,0 +1,138 @@
...
+
+    assert.notCalled(onFrameAdded);
+  });
+
+  it('removal of the annotatable iframe triggers onFrameRemoved', done => {
+    sinon.stub(frameObserver, '_removeFrame').callThrough();
+    const frame = createAnnotatableIFrame();
+
+    waitForFrameObserver()
+      .then(() => {
```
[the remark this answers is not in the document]
esanzgar: I understand that the test is a bit cumbersome.

I tried to simplified the test with async/await, but I couldn't. I need to test these steps:
- execution of `onFrameAdded`
- trigger the removal of the iframe
- execution of `onFrameRemoved`

However, I need to register a listener of iframe's `unload` _before_ the removal the iframe (step 2).

src/annotator/test/frame-observer-test.js
```diff
@@ -0,0 +1,138 @@
...
+
+  it("doesn't triggers onFrameAdded when non-annotatable iframes are added", async () => {
+    createAnnotatableIFrame('dummy-attribute');
+    await waitForFrameObserver();
+
+    assert.notCalled(onFrameAdded);
+  });
+
+  it('removal of the annotatable iframe triggers onFrameRemoved', done => {
+    sinon.stub(frameObserver, '_removeFrame').callThrough();
```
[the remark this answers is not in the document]
esanzgar: I am happy to remove this.

The reason why I added this is because there are two paths for executing the `_removeFrame` method. I wanted to check that only one path is taken and the method is only called once.

src/annotator/test/frame-observer-test.js
esanzgar: I did: https://github.com/hypothesis/client/pull/3686

src/annotator/frame-observer.js
esanzgar: Yes, I noticed that. I have done it in a follow up commit.

src/annotator/test/integration/hypothesis-injector-test.js
esanzgar: I did not change the test itself, but I will improve it.

src/annotator/frame-observer.js
```diff
@@ -64,23 +66,59 @@ export default class FrameObserver {
...
+    return !!iframe.contentDocument;
+  } catch (e) {
+    return false;
+  }
+}
+
+/**
+ * Resolves a Promise when the iframe's DOM is ready (loaded and parsed)
+ *
+ * @param {HTMLIFrameElement} iframe
```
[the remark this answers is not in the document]
esanzgar: In the follow up PR I added a `@trhow` statement: https://github.com/hypothesis/client/pull/3686

src/annotator/frame-observer.js
```diff
@@ -64,23 +66,59 @@ export default class FrameObserver {
...
+ * @return {Promise<void>}
+ */
+export function onDocumentReady(iframe) {
+  return new Promise(resolve => {
+    const iframeDocument = /** @type {Document} */ (iframe.contentDocument);
+    if (iframeDocument.readyState === 'loading') {
+      iframeDocument.addEventListener('DOMContentLoaded', () => {
+        resolve();
```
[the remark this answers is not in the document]
esanzgar: https://github.com/hypothesis/client/pull/3670#issuecomment-900067996

src/annotator/test/frame-observer-test.js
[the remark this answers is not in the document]
esanzgar: I have simplified the test in the last commit.

src/annotator/frame-observer.js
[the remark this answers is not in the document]
robertknight: Its probably worth keeping a record of why we don't automatically inject into iframes. Maybe change the comment to something like "We require the `enable-annotation` attribute to avoid the overhead of loading the client into frames which are not useful to annotate. See https://github.com/hypothesis/client/issues/530".

src/annotator/test/frame-observer-test.js
robertknight: Since you know what causes the issue and there is an open PR, can you remove the commented-out code from this PR.

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -9,13 +8,35 @@ describe('HypothesisInjector integration test', () => {
...
+  }
+
+  function getHypothesisScript(iframe) {
+    return iframe.contentDocument.querySelector(
+      'script[src="data:,Hypothesis"]'
+    );
+  }
+
+  function createHypothesisInjector() {
+    const hypothesisInjector = new HypothesisInjector(
```
robertknight, line 25:
```suggestion
    const injector = new HypothesisInjector(
```

Since this is just a local variable my preference would be to use a slightly more succinct name.

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -35,227 +56,158 @@ describe('HypothesisInjector integration test', () => {
...
-        isDocumentReady(frame, () => {
-          assert(
-            frame.contentDocument.body.hasChildNodes(),
-            'expected dynamically added frame to be modified'
-          );
-          resolve();
-        });
-      });
-    });
+    // Add a iframe to the DOM
```
robertknight, line 131:
```suggestion
    // Add an iframe to the DOM
```

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -35,227 +56,158 @@ describe('HypothesisInjector integration test', () => {
...
+
+    assert.isNotNull(
+      getHypothesisScript(iframe),
+      'expected dynamically added iframe to include the Hypothesis script'
+    );
   });
 
-  it('detects a frame dynamically added, removed, and added again', () => {
-    // Initialize with no initial frame
+  it('detects a iframe dynamically added, removed, and added again', async () => {
```
robertknight, line 184:
```suggestion
  it('detects an iframe dynamically added, removed, and added again', async () => {
```

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -35,227 +56,158 @@ describe('HypothesisInjector integration test', () => {
...
-                  'expected dynamically added frame to be modified'
-                );
-                resolve();
-              });
-            });
-          });
-        });
-      });
-    });
+    // Add a iframe to the DOM
```
robertknight, line 188:
```suggestion
    // Add an iframe to the DOM
```

robertknight (approved):
Sorry for the delay on this. I had a few minor comments but let's get this merged and the follow-up PR rebased.

Issue #3631, opened 2021-08-02
Improve handling of network errors when making API requests


robertknight:
Since the recent changes to inter-frame communication in the client the volume of noise in Sentry had now dropped to manageable levels, to the point that we could soon consider turning on internal notifications of new reports again.

Looking at the [current reports](https://sentry.io/organizations/hypothesis/issues/?project=69811&query=is%3Aunresolved) for the client project in Sentry, most of the remaining issues are transient network problems. These include:

- 5xx responses from some service in between the client and the backend
- Canceled network requests (I think that's what [this Sentry issue](https://sentry.io/organizations/hypothesis/issues/1205785027/?project=69811&query=is%3Aunresolved) is)
- 200 responses that are expected to have a JSON body but which actually have an empty or invalid body. See [this issue](https://sentry.io/organizations/hypothesis/issues/1274716662/?project=69811&query=is%3Aunresolved) and also [error suppressions in the Sentry config](https://github.com/hypothesis/client/blob/6642f90c4f598547253b021ca254703bb64754a6/src/sidebar/util/sentry.js#L64)

We should look into either improving how these are handled or suppressing the exception if there isn't anything useful we can do about them. The hypothesized issue with "successful" responses that have empty JSON bodies needs to be verified in different browsers to confirm that these are the cause of the problem.

PR #3671, opened 2021-08-10
Add `fetchJSON` utility to standardize network errors 

robertknight:
This PR adds a `fetchJSON` utility that wraps the browser's `fetch` API in order to standardize the handling of various kinds of request failure and produce errors which are more useful downstream. The issues which motivated this change are:

- The format of errors thrown when a `fetch` request fails varies depending on the error and browser. This makes it difficult to handle them consistently downstream (eg. filtering in Sentry)
- We don't currently handle failure to parse JSON responses in various places, but we see in Sentry that it does occasionally happen that a 200 response which should have a JSON body does not.
- There is some duplication of code in making requests that return JSON responses and processing the result

The `fetchJSON` utility throws a `FetchError` error in all the above cases which consumers can test for.

- The first commit adds the `fetchJSON` utility in src/sidebar/util/fetch.js.
- The subsequent commits update services and utilities in the client that call `fetch` to use `fetchJSON` instead.
- The last commit simplifies the Sentry error filtering configuration to only look for the standardized error responses.

I'm probably going to split this into some smaller chunks, but I've put it all in a single draft for initial discussion.

Fixes https://github.com/hypothesis/client/issues/2184
Part of https://github.com/hypothesis/client/issues/3631

### DIFF
--- a/src/sidebar/util/fetch.js
+++ b/src/sidebar/util/fetch.js
@@ -9,13 +9,15 @@
  */
 export class FetchError extends Error {
   /**
+   * @param {string} url - The URL that was requested. This may be different
+   *   than the final URL of the response if a redirect happened.
    * @param {Response|null} response - The response to the `fetch` request or
    *   `null` if the fetch failed
    * @param {string} [reason] - Additional details about the error. This might
    *   include context of the network request or a server-provided error in
    *   the response.
    */
-  constructor(response, reason = '') {
+  constructor(url, response, reason = '') {
     let message = 'Network request failed';
     if (response) {
       message += ` (${response.status})`;
@@ -25,6 +27,7 @@ export class FetchError extends Error {
     }
     super(message);
 
+    this.url = url;
     this.response = response;
     this.reason = reason;
   }
@@ -49,7 +52,7 @@ export async function fetchJSON(url, init) {
     // If the request fails for any reason, wrap the result in a `FetchError`.
     // Different browsers use different error messages for `fetch` failures, so
     // wrapping the error allows downstream clients to handle this uniformly.
-    throw new FetchError(null, err.message);
+    throw new FetchError(url, null, err.message);
   }
 
   if (response.status === 204 /* No Content */) {
@@ -62,14 +65,14 @@ export async function fetchJSON(url, init) {
   try {
     data = await response.json();
   } catch (err) {
-    throw new FetchError(response, 'Failed to parse response');
+    throw new FetchError(url, response, 'Failed to parse response');
   }
 
   // If the HTTP status indicates failure, attempt to extract a server-provided
   // reason from the response, assuming certain conventions for the formatting
   // of error responses.
   if (!response.ok) {
-    throw new FetchError(response, data?.reason);
+    throw new FetchError(url, response, data?.reason);
   }
 
   return data;
--- a/src/sidebar/util/fetch.js
+++ b/src/sidebar/util/fetch.js
@@ -1,0 +1,76 @@
+/**
+ * An error indicating a failed network request.
+ *
+ * Failures that this error can represent include:
+ *
+ *  - Failures to send an HTTP request
+ *  - Requests that returned non-2xx responses
+ *  - Failures to parse the response in the expected format (eg. JSON)
+ */
+export class FetchError extends Error {
+  /**
+   * @param {Response|null} response - The response to the `fetch` request or
+   *   `null` if the fetch failed
+   * @param {string} [reason] - Additional details about the error. This might
+   *   include context of the network request or a server-provided error in
+   *   the response.
+   */
+  constructor(response, reason = '') {
+    let message = 'Network request failed';
+    if (response) {
+      message += ` (${response.status})`;
+    }
+    if (reason) {
+      message += `: ${reason}`;
+    }
+    super(message);
+
+    this.response = response;
+    this.reason = reason;
+  }
+}
+
+/**
+ * Execute a network request and return the parsed JSON response.
+ *
+ * Throws a `FetchError` if making the request fails or the request returns
+ * a non-2xx response.
+ *
+ * Returns `null` if the request returns a 204 (No Content) response.
+ *
+ * @param {string} url
+ * @param {RequestInit} [init] - Parameters for `fetch` request
+ */
+export async function fetchJSON(url, init) {
+  let response;
+  try {
+    response = await fetch(url, init);
+  } catch (err) {
+    // If the request fails for any reason, wrap the result in a `FetchError`.
+    // Different browsers use different error messages for `fetch` failures, so
+    // wrapping the error allows downstream clients to handle this uniformly.
+    throw new FetchError(null, err.message);
+  }
+
+  if (response.status === 204 /* No Content */) {
+    return null;
+  }
+
+  // Attempt to parse a JSON response. This may fail even if the status code
+  // indicates success.
+  let data;
+  try {
+    data = await response.json();
+  } catch (err) {
+    throw new FetchError(response, 'Failed to parse response');
+  }
+
+  // If the HTTP status indicates failure, attempt to extract a server-provided
+  // reason from the response, assuming certain conventions for the formatting
+  // of error responses.
+  if (!response.ok) {
+    throw new FetchError(response, data?.reason);
+  }
+
+  return data;
+}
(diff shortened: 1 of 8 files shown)


src/sidebar/util/sentry.js
```diff
@@ -54,21 +54,10 @@ export function init(config) {
...
-
-      // Ignore network request failures due to empty JSON bodies.
-      // TODO - Consider removing these once `fetch` callers have been changed
-      // to handle `Response.json()` calls failing.
-      'JSON.parse: unexpected end of data', // Firefox
-      'Unexpected end of JSON input', // Opera Mobile
-
-      // Ignore network request cancellations
-      'AbortError: The operation was aborted.', // Firefox
+      'Network request failed', // Standard message prefix for `FetchError` errors
```
robertknight, line 60: Here we see the benefit of standardizing the format of network errors. We could also potentially do an `instanceof FetchError` check in this module and customize the reporting logic.

src/sidebar/util/fetch.js
```diff
@@ -0,0 +1,76 @@
...
+
+    this.response = response;
+    this.reason = reason;
+  }
+}
+
+/**
+ * Execute a network request and return the parsed JSON response.
+ *
+ * Throws a `FetchError` if making the request fails or the request returns
```
esanzgar, line 36:
```suggestion
 * @throws {FetchError} if...
 ```

File: src/sidebar/util/fetch.js [hunk 3]
```diff
@@ -0,0 +1,76 @@
...
+ * @param {RequestInit} [init] - Parameters for `fetch` request
+ */
+export async function fetchJSON(url, init) {
+  let response;
+  try {
+    response = await fetch(url, init);
+  } catch (err) {
+    // If the request fails for any reason, wrap the result in a `FetchError`.
+    // Different browsers use different error messages for `fetch` failures, so
+    // wrapping the error allows downstream clients to handle this uniformly.
```
esanzgar, line 51: 👍🏼

src/sidebar/util/fetch.js
```diff
@@ -0,0 +1,76 @@
...
+    throw new FetchError(null, err.message);
```
esanzgar, line 52: I think sometimes `message` is not defined, but it can't cause any problem given the logic of `FetchError` constructor.

src/sidebar/util/fetch.js
```diff
@@ -0,0 +1,76 @@
...
+export class FetchError extends Error {
+  /**
+   * @param {Response|null} response - The response to the `fetch` request or
+   *   `null` if the fetch failed
+   * @param {string} [reason] - Additional details about the error. This might
+   *   include context of the network request or a server-provided error in
+   *   the response.
+   */
+  constructor(response, reason = '') {
+    let message = 'Network request failed';
```
esanzgar, line 19: Capturing the `URL` would be essential.

src/sidebar/util/fetch.js
```diff
@@ -0,0 +1,76 @@
...
+    return null;
+  }
+
+  // Attempt to parse a JSON response. This may fail even if the status code
+  // indicates success.
+  let data;
+  try {
+    data = await response.json();
+  } catch (err) {
+    throw new FetchError(response, 'Failed to parse response');
```
esanzgar, line 65:
```suggestion
    throw new FetchError(response, error.message ?? 'Failed to parse response');
```
The above could help to identify what in the response was not parseable, for example:
'Unexpected token < in JSON at position 6'

src/sidebar/util/test/fetch-test.js
```diff
@@ -0,0 +1,83 @@
...
+
+      let err;
+      try {
+        await fetchJSON('https://example.com');
+      } catch (e) {
+        err = e;
+      }
+
+      assert.instanceOf(err, FetchError);
+      assert.include(err.message, 'Network request failed: Fetch failed');
```
esanzgar, line 40: It would be important to know the which `URL` failed.

src/sidebar/util/test/fetch-test.js
```diff
@@ -0,0 +1,83 @@
...
+      assert.instanceOf(err, FetchError);
+      assert.equal(
+        err.message,
+        'Network request failed (200): Failed to parse response'
+      );
+    });
+
+    it('throws a FetchError if the response has a non-2xx status code', async () => {
+      fakeResponse.status = 400;
+      fakeResponse.json.resolves({ reason: 'server error' });
```
esanzgar, line 66: 400 are client errors, 500 are server errors

esanzgar (commented):
I left a few comments.

src/sidebar/util/fetch.js
esanzgar: It is not only 'Network` errors here. The network can be fine and fail the URL parameters, parsing response, etc.

esanzgar: I guess one way to indirectly get the URL is by looking at the stacktrace.

src/sidebar/util/test/fetch-test.js
robertknight: This wasn't clear. "server error" was intended here as shorthand for "a server-generated error message in the response". I've changed the status code and reason message to a more realistic 404 example.

src/sidebar/util/test/fetch-test.js
robertknight: I have added the request URL to `FetchError`'s properties.

src/sidebar/util/fetch.js
robertknight: Honestly I don't think the specific error message will be all that useful. From what I've seen in Sentry the most common reason for `response.json()` to fail by far is that the body is missing/empty/unavailable to read. Even with an error like this, you really would need the full response body to make sense of it.

src/sidebar/util/fetch.js
robertknight: Adding the request URL to the error seems like a good idea, so I've done that.

src/sidebar/util/fetch.js
robertknight: `message` should always be defined if the exception is an `Error` instance, even if no message was passed to the constructor. In this context since `fetch` is a browser-native API any exceptions it throws should be an `Error` instance.

src/sidebar/util/fetch.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * An error indicating a failed network request.
+ *
+ * Failures that this error can represent include:
+ *
+ *  - Failures to send an HTTP request
+ *  - Requests that returned non-2xx responses
+ *  - Failures to parse the response in the expected format (eg. JSON)
+ */
+export class FetchError extends Error {
```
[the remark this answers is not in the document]
robertknight: I chose the `FetchError` name to match the `fetchJSON` operation. `FetchJSONError` seemed a bit unwieldy and not appropriate if we wanted to reuse this for requests that return non-JSON responses.  This error intentionally covers a variety of errors (networking - where there is no response, failure response, failure to parse response) so I thought it better to name the error after the operation rather than the kind of error.

src/sidebar/util/fetch.js
robertknight: I have revised the JSDoc comment here, mainly to better explain why `fetchJSON` is preferred to calling `fetch` directly. I did also move the exception details into an `@throws` comment, although I was disappointed to see that TypeScript doesn't parse these at all, even just to verify that the type is valid and make it linkable.

src/sidebar/util/fetch.js
esanzgar: Have you considered using the `cause` property like you did in `TokenError`?

`constructor(url, response, cause) {`

esanzgar (approved):
Thanks for adding the `url` field in the `FetchError` class. It is not in the reflected in the `message` but I believe it would be reachable in Sentry.

I have suggested to use a `cause` field in `FetchError` as you did in `TokenError`.

robertknight:
> I have suggested to use a cause field in FetchError as you did in TokenError.

We could add this, although in this context the amount of added information is going to be pretty limited. A typical network error in Chrome is just reported to JavaScript as a `TypeError` with message `Failed to fetch` (!).

I'll get the current version of this merged and then maybe add that later if it looks like it will be useful.

esanzgar:
> I'll get the current version of this merged and then maybe add that later if it looks like it will be useful.

👍🏼

PR #3675, opened 2021-08-11
Refactor token fetching and error handling in OAuthClient

robertknight:
Prepare for upcoming improvements to handling of network errors across
the client by refactoring some `fetch` requests and error handling in
OAuthClient.

 - Extract duplicated logic for fetching an access token into a
   `_getAccessToken` method.
 - Create a `TokenError` class to wrap all errors that may be thrown
   when fetching or revoking an access token. This makes it easier for
   downstream consumers to handle such errors in a uniform way.

   The `TokenError` error links to the network or other error that
   caused it via the `cause` property, following the convention
   currently being standardized in https://github.com/tc39/proposal-error-cause.
- Remove a `generateState` test seam in favor of using `$imports.$mock`. This addresses a code coverage gap.
   
See https://github.com/hypothesis/client/pull/3671 for more background on where this is going and why.

### DIFF
--- a/src/sidebar/util/oauth-client.js
+++ b/src/sidebar/util/oauth-client.js
@@ -1,33 +1,27 @@
-import * as random from './random';
+import { generateHexString } from './random';
 
 /**
  * An object holding the details of an access token from the tokenUrl endpoint.
+ *
  * @typedef {Object} TokenInfo
- * @property {string} accessToken  - The access token itself.
- * @property {number} expiresAt    - The date when the timestamp will expire.
- * @property {string} refreshToken - The refresh token that can be used to
- *                                   get a new access token.
+ * @prop {string} accessToken  - The access token itself.
+ * @prop {number} expiresAt    - The date when the timestamp will expire.
+ * @prop {string} refreshToken - The refresh token that can be used to
+ *                               get a new access token.
  */
 
 /**
- * Return a new TokenInfo object from the given tokenUrl endpoint response.
- * @param {Response} response - The HTTP response from a POST to the tokenUrl
- *                            endpoint.
- * @returns {Promise<TokenInfo>}
+ * Error thrown if fetching or revoking an access token fails.
  */
-function tokenInfoFrom(response) {
-  return response.json().then(data => {
-    return {
-      accessToken: data.access_token,
-
-      // Set the expiry date to some time slightly before that implied by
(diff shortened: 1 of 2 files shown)


src/sidebar/util/oauth-client.js
```diff
@@ -213,11 +184,44 @@ export default class OAuthClient {
...
     };
-    return fetch(url, {
+    const response = await fetch(url, {
       method: 'POST',
       headers,
       body: params.toString(),
     });
+
+    if (response.status !== 200) {
+      throw new Error('Request failed');
```
esanzgar, line 194: Maybe a bit more specific? (there are many other network errors, offline, etc.)

```suggestion
      throw new Error('Request failed with a non-200 status');
```

src/sidebar/util/oauth-client.js
```diff
@@ -213,11 +184,44 @@ export default class OAuthClient {
...
+    }
+
+    return response.json();
+  }
+
+  /**
+   * Fetch an OAuth access token.
+   *
+   * @param {Record<string, string>} data - Parameters for form POST request
```
esanzgar, line 203: Adding the return may not be that necessary. I don't know. 
```suggestion
   * @param {Record<string, string>} data - Parameters for form POST request
    * @returns {Promise<{accessToken: string, expiresAt: number, refreshToken: string}>}
```

esanzgar:
Would you need a custom `toString` method in the `TokenError` to expose the cause in Sentry?

src/sidebar/util/oauth-client.js
robertknight: I have added the status code to the error, although this is expected to be pretty short-lived code that is replaced in the next PR in the series.

src/sidebar/util/oauth-client.js
robertknight: I added an explicit `@return` here using the existing `TokenInfo` type.

robertknight:
> Would you need a custom toString method in the TokenError to expose the cause in Sentry?

Apparently Sentry already supports the use of a `cause` property on errors to chain them. It is implemented by the [LinkedErrors "integration"](https://docs.sentry.io/platforms/javascript/configuration/integrations/default/#linkederrors) which is part of the set of integrations/plugins enabled by default.

PR #3679, opened 2021-08-13
Allow for approximate equality in range-util tests

robertknight:
Several tests in range-util-test.js started failing locally after the
most recent Puppeteer upgrade with actual and expected values differing
by up to 1.0.

Change the tests to check for approximate equality, which is all that is
required for this use case.


lyzadanger (approved):
This fixes the glitch, thank you.

PR #3680, opened 2021-08-16
Replace `whitelistUrls` in Sentry config with `allowUrls`

robertknight:
Replace the deprecated `whitelistUrls` config option in Sentry with
`allowUrls`. `whitelistUrls` was deprecated in Sentry v5.18.0.

Also revise the comments to be more succinct and note what happens if an
exception does not have a useful stack trace. That behavior is not documented
but I can see it from [the code](https://github.com/getsentry/sentry-javascript/blob/74a08294122a0a900b9299919a3dd2d5a8a6e694/packages/core/src/integrations/inboundfilters.ts#L144) (note `!url` test at start of line, where `url` refers to the URL from the stack trace).

See https://docs.sentry.io/platforms/javascript/configuration/options/#allow-urls.

### DIFF
--- a/src/sidebar/util/sentry.js
+++ b/src/sidebar/util/sentry.js
@@ -36,21 +36,20 @@ function currentScriptOrigin() {
  * @param {SentryConfig} config
  */
 export function init(config) {
-  // Only send events for errors which can be attributed to our code. This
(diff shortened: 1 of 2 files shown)


esanzgar (approved):
It looks OK.

PR #3686, opened 2021-08-17
Improve the logic that checks for iframe's document ready

esanzgar:
In certain situations, the logic to check when the iframe's document is
ready didn't work as expected. The reason for that is that browsers
create an initial blank document before the actual document loads.
This blank document has a `readyState` of 'complete' in Safari and
Chrome and of 'uninitialized' in Firefox. This blank document can be
mistaken by the final document. I was able to reproduce a situation
where the Hypothesis client was injected into the blank document. Then,
the blank document was replaced by the actual document and the
Hypothesis client was not not longer embedded.

In addition, this commit:

* renames `frame` to `iframe`
* renames `_handledFrames` to _annotatableIFrames`
* removes `isAccessible` (no longer neede)
* enables a test that was commented previously

Here is a video that demonstrated this issue:

https://user-images.githubusercontent.com/8555781/129683454-953f744a-032e-499d-934d-446d6933d534.mov

The last iframe that was created didn't have the Hypothesis client embedded
because the reason explained above.


src/annotator/frame-observer.js
```diff
@@ -120,15 +120,11 @@ export function onDocumentReady(iframe) {
...
     /* istanbul ignore next */
     if (
-      // @ts-ignore
-      state === 'uninitialized' ||
-      (state === 'complete' &&
-        iframe.hasAttribute('src') &&
-        iframeDocument.head.hasChildNodes() === false &&
-        iframeDocument.body.hasChildNodes() === false)
+      iframeDocument.location.href === 'about:blank' &&
+      iframe.hasAttribute('src')
```
esanzgar, line 126: `src` could be `about:blank` in which this logic will fail. I don't see any logic for somebody using `src` in that way.

robertknight: > I don't see any logic for somebody using src in that way.

We could check for a `src` value that is different from `about:blank`. I wonder what happens if the `src` value is a `data:` or `blob:` URL that can be resolved synchronously?

src/annotator/frame-observer.js
```diff
@@ -83,47 +87,58 @@ export class FrameObserver {
...
-      resolve();
+      return;
     }
+
+    if (state === 'loading') {
+      iframeDocument.addEventListener('DOMContentLoaded', () => resolve());
+      return;
+    }
+
+    // state is 'interactive' or 'complete';
```
robertknight, line 141:
```suggestion
    // State is 'interactive' or 'complete';
```

src/annotator/test/frame-observer-test.js
```diff
@@ -4,140 +4,191 @@ import {
...
-  }
-
-  beforeEach(() => {
-    container = document.createElement('div');
-    document.body.appendChild(container);
-    onFrameAdded = sinon.stub();
-    onFrameRemoved = sinon.stub();
-    frameObserver = new FrameObserver(container, onFrameAdded, onFrameRemoved);
-  });
+describe('annotator/frame-observer.js', () => {
```
robertknight, line 7:
```suggestion
describe('annotator/frame-observer', () => {
```

src/annotator/test/frame-observer-test.js
```diff
@@ -4,140 +4,191 @@ import {
...
+      container.remove();
+      frameObserver.disconnect();
+      sandbox.restore();
+    });
 
-  it('triggers onFrameAdded when an annotatable iframe is added', async () => {
-    let frame = createAnnotatableIFrame();
-    await waitForFrameObserver();
+    it('triggers onFrameAdded when an annotatable iframe is added', async () => {
+      let iframe = createAnnotatableIFrame();
```
robertknight, line 55: I'm not sure about the `frame` => `iframe` change here. We tend to use the term `frame` in most places (eg. in the sidebar) following the example of `window.frames` and other browser APIs.

src/annotator/test/frame-observer-test.js
```diff
@@ -4,140 +4,191 @@ import {
...
-
-  //   assert.notCalled(onFrameAdded);
-  // });
+  describe('onDocumentReady', () => {
+    let fakeIFrame;
+    let fakeIFrameDocument;
+
+    beforeEach(() => {
+      fakeIFrameDocument = {
+        addEventListener: sinon.stub(),
```
robertknight, line 158: In recent versions of a Chrome a more convenient way to make a fake browser object that handles events is to create a new `EventTarget` instance or subclass `EventTarget`.

src/annotator/test/frame-observer-test.js
```diff
@@ -4,140 +4,191 @@ import {
...
+        readyState: 'loading',
+        location: {
+          href: 'about:blank',
+        },
+      };
+
+      fakeIFrame = {
```
robertknight, line 165: I'd probably suggest to create a fake generic DOM element here and add the `contentWindow` property. Then you're only having to take the `contentWindow` part.

robertknight (commented):
Generally looks good.

In most existing places in the code we tend to use the term `frame` rather than `iframe`. I think we should probably stick with that convention for consistency, even though we almost never deal with other kinds of frame (`<frameset>` is deprecated in browsers. Are there any other kinds of frame?)

src/annotator/frame-observer.js
```diff
@@ -83,47 +87,64 @@ export class FrameObserver {
...
+    if (
+      // @ts-ignore
+      state === 'uninitialized' ||
+      (state === 'complete' &&
+        iframe.hasAttribute('src') &&
+        iframeDocument.head.hasChildNodes() === false &&
+        iframeDocument.body.hasChildNodes() === false)
+    ) {
+      // Unfortunately, listening for 'DOMContentLoaded' on the iframeDocument
+      // doesn't work. Instead, we need to wait for a 'load' event to be triggered.
```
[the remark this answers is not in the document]
esanzgar: It makes sense, thanks.

src/annotator/test/frame-observer-test.js
esanzgar: I have taken a middle ground here: `frame-observer.js` uses `frame` for the code (not for the comments) but `frame-observer-test.js` uses `iframe`. That way I leave the code as generic as possible, but the test that is the application on a concrete scenario, more specific.

Are you happy with this? Otherwise, I am happy to revert all to `frame`. Please, let me know.

src/annotator/test/frame-observer-test.js
esanzgar: 👍🏼 . Incorporated this feedback in the last commit.

src/annotator/test/frame-observer-test.js
esanzgar: I tried it in this way, but it didn't work:

```
      fakeIFrame = document.createElement('iframe');
      fakeIFrame.contentDocument = fakeIFrameDocument;  // TypeError: Cannot set property contentDocument of #<HTMLIFrameElement> which has only a getter
      fakeIFrame.setAttribute('src', 'http://my.dummy');
```

How do you suggest to create this fake generic DOM element?

robertknight: > How do you suggest to create this fake generic DOM element?

I meant using `document.createElement('div')` or some other general-purpose non-iframe element which doesn't have its own `contentDocument` and `contentWindow` accessors, so you can define custom ones.

esanzgar: Yes, no problem, I can do that. Done.

esanzgar:
Robert, I changed  `iframe` back to `frame` in `frame-observer.js` but I left `iframe` in `frame-observer-test.js`. 

See: https://github.com/hypothesis/client/pull/3686/commits/5c6ff08771417d54ca7176d32de9a625b3fbfd44

I did that to keep the code base as general as possible. However, on the tests, I feel it is clearer and more concrete to refer to the name of the HTML element. It also makes `frame-observer-test.js` similar to `hypothesis-injector-test.js` (concrete reference to `iframe`) and `frame-observer.js` similar to `hypothesis-injector.js` (more general `frame`).

If you are not happy with that, let me know and I will change all the `iframe` words to `frame` in the tests too.

robertknight (approved):

Issue #3674, opened 2021-08-11
Mismatch between selection and highlighted text in PDF.js v2.11.106


robertknight:
When testing an upgrade to the latest version of PDF.js (v2.11.106) in the browser extension I encountered an issue where the highlights created for an annotation no longer match the originally selected text.

<img width="1022" alt="PDF js v2 11 text selection issue" src="https://user-images.githubusercontent.com/2458/129000040-bc7ed70c-6ba4-4077-b1b9-bfbd55133243.png">

In the above screenshot, the annotation quote matches the text I originally selected, but the highlight does not.

The PDF I'm using for testing here is https://cogcomp.seas.upenn.edu/papers/RRDA11.pdf.

I also observed an issue where there was quite noticeable latency between clicking the "Annotate" button for the first time in a session and the annotation being created. The latency seems to be reduced when creating subsequent annotations. In a long book this could be confusing for the user as they may think that the "Annotate" button did not work. Likely related to this issue, I notice that the bucket bar is very laggy in updating compared to the production client.

robertknight:
Aside from the issue shown above, there is also a more subtle problem with the way PDF anchoring currently falls back to quote-based anchoring in the event that the position selector no longer matches. See https://github.com/hypothesis/client/issues/3705.

robertknight:
Closing this as https://github.com/hypothesis/client/issues/3705 has now been resolved.

PR #3687, opened 2021-08-17
Fix anchoring issues in newest PDF.js releases

robertknight:
This PR fixes the issue described in https://github.com/hypothesis/client/issues/3674 which occurs in PDF.js releases after v2.9.359, which include the text rendering changes from https://github.com/mozilla/pdf.js/pull/13257. The changes in this PR are backwards compatible with earlier versions of PDF.js.

The problem is described in detail in the second commit in this PR, but briefly the issue is that there was a mismatch between the page text fetched using PDF.js's APIs and the contents of the transparent text layer element that PDF.js creates for rendered pages. The client relies on these being the same.

In addition to resolving the problem for the current version of PDF.js I also added a check that will make the issue easier to debug if this mismatch happens again in later versions.

Part of https://github.com/hypothesis/client/issues/3674 

- [x] Update tests to cover old and new versions of PDF.js, since we'll need to support both
- [x] File a separate PR to update the version of PDF.js used by the dev server
- [x] Create separate issue for handling of line breaks (https://github.com/hypothesis/client/issues/3693)

----

**Testing:**

The easiest way to test these changes locally is to create a temporary local branch that combines the changes from this branch and https://github.com/hypothesis/client/pull/3691. If you open up any of the test PDF assignments in the dev server, check that you can highlight some text and create an annotation, that the highlight is created in the correct place and that clicking on an annotation card scrolls to the correct location.

Alternatively you can build a browser extension that uses a build from this branch of the client combined with https://github.com/hypothesis/browser-extension/pull/631.

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -180,7 +180,7 @@ function getPageTextContent(pageIndex) {
     });
     let items = textContent.items;
 
-    // Older versions of PDF.js did not create elements in the text layer for
+    // Versions of PDF.js < v2.9.359 did not create elements in the text layer for
     // text items that contained all-whitespace strings. Newer versions (after
     // https://github.com/mozilla/pdf.js/pull/13257) do. The same commit also
     // introduced the `hasEOL` property to text items, so we use the absence
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -156,22 +156,30 @@ export async function documentHasText() {
 /**
  * Return the text of a given PDF page.
  *
+ * The page text returned by this function should match the `textContent` of the
+ * text layer element that PDF.js creates for rendered pages. This allows
+ * offsets computed in the text to be reused as offsets within the text layer
+ * element's content. This is important to create correct Ranges for anchored
+ * selectors.
+ *
  * @param {number} pageIndex
  * @return {Promise<string>}
  */
-async function getPageTextContent(pageIndex) {
+function getPageTextContent(pageIndex) {
+  // If we already have or are fetching the text for this page, return the
+  // existing result.
   const cachedText = pageTextCache[pageIndex];
   if (cachedText) {
     return cachedText;
   }
 
-  // Join together PDF.js `TextItem`s representing pieces of text in a PDF page.
-  const joinItems = items => {
-    let itemStrings = items.map(item => item.str);
+  const getPageText = async () => {
+    const pageView = await getPageView(pageIndex);
+    const textContent = await pageView.pdfPage.getTextContent({
+      normalizeWhitespace: true,
+    });
+    let items = textContent.items;
 
-    // We want the text returned by `getPageTextContent` to match the `textContent`
-    // of the transparent text layer, so that text offsets match up.
-    //
     // Older versions of PDF.js did not create elements in the text layer for
     // text items that contained all-whitespace strings. Newer versions (after
     // https://github.com/mozilla/pdf.js/pull/13257) do. The same commit also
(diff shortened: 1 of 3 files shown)


src/annotator/anchoring/pdf.js
```diff
@@ -436,21 +459,17 @@ export function anchor(root, selectors) {
...
       return findPage(position.start).then(({ index, offset, textContent }) => {
         const start = position.start - offset;
         const end = position.end - offset;
         const length = end - start;
 
-        checkQuote(textContent.substr(start, length));
+        const matchedText = textContent.substr(start, length);
+        if (quote && quote.exact !== matchedText) {
+          throw new Error('quote mismatch');
+        }
```
robertknight, line 472: There is no change to the logic here, I just inlined `checkQuote` to remove a misleading `range` parameter which is, in fact, a `string` rather than a `Range`.

src/annotator/anchoring/pdf.js
```diff
@@ -155,38 +156,45 @@ export async function documentHasText() {
...
+ * element's content. This is important to create correct Ranges for anchored
+ * selectors.
+ *
  * @param {number} pageIndex
  * @return {Promise<string>}
  */
-async function getPageTextContent(pageIndex) {
+function getPageTextContent(pageIndex) {
+  // If we already have or are fetching the text for this page, return the
+  // existing result.
```
robertknight, line 170: Unrelated to the fix for the anchoring issues, I added some comments about the way this function interacts with the page text cache.

robertknight:
This PR fixes the most glaring issue, but I realized today we have a more subtle problem that needs to be addressed. See https://github.com/hypothesis/client/issues/3705.

src/annotator/anchoring/test/fake-pdf-viewer-application.js
```diff
@@ -61,14 +81,25 @@ class FakePDFPageProxy {
       );
     }
 
+    const makeTextItem = str => {
```
esanzgar, line 84:
```suggestion
    /** @param {string} str */
    const makeTextItem = str => {
```

src/annotator/anchoring/pdf.js
```diff
@@ -155,38 +156,45 @@ export async function documentHasText() {
...
+  const getPageText = async () => {
     const pageView = await getPageView(pageIndex);
     const textContent = await pageView.pdfPage.getTextContent({
       normalizeWhitespace: true,
     });
-    return joinItems(textContent.items);
-  };
+    let items = textContent.items;
+
+    // Older versions of PDF.js did not create elements in the text layer for
```
esanzgar, line 182:
```suggestion
    // Versions of PDF.js < v2.9.359 did not create elements in the text layer for
```

src/annotator/anchoring/pdf.js
```diff
@@ -155,38 +156,45 @@ export async function documentHasText() {
...
+    // text items that contained all-whitespace strings. Newer versions (after
```
esanzgar, line 183:
```suggestion
    // text items that contained all-whitespace strings. Versions >= v2.9.359
```

esanzgar (approved):
All looks good to me.

src/annotator/anchoring/test/pdf-test.js
```diff
@@ -49,8 +53,9 @@ describe('annotator/anchoring/pdf', () => {
    *
    * @param {string[]} content -
    *   Array containing the text content of each page of the loaded PDF document
+   * @param {import('./fake-pdf-viewer-application').PDFJSConfig} [config]
```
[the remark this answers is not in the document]
robertknight: Yes. TypeScript doesn't currently check test files, but this is still useful for developers.

src/annotator/anchoring/pdf.js
```diff
@@ -155,38 +156,45 @@ export async function documentHasText() {
...
+ * selectors.
+ *
  * @param {number} pageIndex
  * @return {Promise<string>}
  */
-async function getPageTextContent(pageIndex) {
+function getPageTextContent(pageIndex) {
+  // If we already have or are fetching the text for this page, return the
+  // existing result.
   if (pageTextCache[pageIndex]) {
```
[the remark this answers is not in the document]
robertknight: > This condition will always return true since this 'Promise<string>' is always defined.ts(2801)
pdf.js(171, 7): Did you forget to use 'await'?

This warning occurs when using TypeScript 4.4+. When I created this branch we were still using 4.3. I fixed the issue yesterday when merging a change from Dependabot that upgrades master to 4.4.

src/annotator/anchoring/pdf.js
```diff
@@ -155,38 +156,45 @@ export async function documentHasText() {
...
-  // Fetch the text content for a given page as a string.
-  const getTextContent = async pageIndex => {
+  const getPageText = async () => {
     const pageView = await getPageView(pageIndex);
     const textContent = await pageView.pdfPage.getTextContent({
       normalizeWhitespace: true,
     });
-    return joinItems(textContent.items);
-  };
+    let items = textContent.items;
```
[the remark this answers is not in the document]
robertknight: The lines below need to access other properties on items in the array, which is why items is a `TextContentItem[]` rather than `string[]` here.

src/annotator/anchoring/pdf.js
robertknight: I've made the above change to specify which version counts as "older" and "newer", so I won't repeat it here.

src/annotator/anchoring/test/fake-pdf-viewer-application.js
```diff
@@ -220,15 +255,16 @@ export default class FakePDFViewerApplication {
    * @param {Options} options
    */
   constructor(options) {
```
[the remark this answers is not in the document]
robertknight: I left this as is for the time being. I don't have a super strong opinion here, but once you get into nested destructuring I'm not sure its necessarily more readable than not using destructuring at all.

PR #3691, opened 2021-08-18
Update PDF.js to the latest release (v2.11.213)

robertknight:
Update PDF.js in the dev server to the newest release. This is the result of running the `scripts/update-pdfjs` script. I didn't change anything else.

You can see the version number of PDF.js in use by looking in the browser console.

There is an issue with anchoring and creating annotations in this release which is fixed by https://github.com/hypothesis/client/pull/3687.


robertknight:
https://github.com/hypothesis/client/pull/3743 resolves an issue with anchoring some old annotations after we switch our various viewers to the newest PDF.js release, so I think this is unblocked now.

robertknight:
I bumped the PDF.js version again so that the client's dev server matches the browser extension and Via.

esanzgar (approved):

PR #3707, opened 2021-08-24
Require quote selectors when anchoring in PDFs

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/3687**~~

Changes in text rendering across PDF.js versions can render position
selectors invalid. Therefore any anchoring done with position selectors must
be checked against the quote, as we do with HTML annotations.

This commit disallows anchoring using only position selectors in PDFs
and restructures `anchor` control flow using async/await to make it
easier to follow.

We have been capturing quote selectors with PDF annotations forever, so
there should be no impact on old annotations.

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -446,57 +446,71 @@ function prioritizePages(position) {
 /**
  * Anchor a set of selectors to a DOM Range.
  *
+ * `selectors` must include a `TextQuoteSelector` and may include other selector
+ * types.
+ *
  * @param {HTMLElement} root
- * @param {Selector[]} selectors - Selector objects to anchor
+ * @param {Selector[]} selectors
  * @return {Promise<Range>}
  */
-export function anchor(root, selectors) {
-  const position = /** @type {TextPositionSelector|undefined} */ (
-    selectors.find(s => s.type === 'TextPositionSelector')
-  );
+export async function anchor(root, selectors) {
   const quote = /** @type {TextQuoteSelector|undefined} */ (
     selectors.find(s => s.type === 'TextQuoteSelector')
   );
 
-  /** @type {Promise<Range>} */
-  let result = Promise.reject('unable to anchor');
-
-  if (position) {
-    result = result.catch(() => {
-      return findPage(position.start).then(({ index, offset, textContent }) => {
-        const start = position.start - offset;
-        const end = position.end - offset;
-        const length = end - start;
-
-        const matchedText = textContent.substr(start, length);
-        if (quote && quote.exact !== matchedText) {
-          throw new Error('quote mismatch');
-        }
-
-        return anchorByPosition(index, start, end);
-      });
-    });
+  // The quote selector is required in order to check that text position
+  // selector results are still valid.
+  if (!quote) {
+    throw new Error('No quote selector found');
   }
 
-  if (quote) {
-    result = result.catch(() => {
+  const position = /** @type {TextPositionSelector|undefined} */ (
+    selectors.find(s => s.type === 'TextPositionSelector')
+  );
+
+  if (position) {
+    // If we have a position selector, try using that first as it is the fastest
+    // anchoring method.
+    try {
+      const { index, offset, textContent } = await findPage(position.start);
+      const start = position.start - offset;
+      const end = position.end - offset;
+      const length = end - start;
+
+      const matchedText = textContent.substr(start, length);
+      if (quote.exact !== matchedText) {
+        throw new Error('quote mismatch');
+      }
+
+      const range = await anchorByPosition(index, start, end);
+      return range;
+    } catch {
+      // Fall back to quote selector
+    }
+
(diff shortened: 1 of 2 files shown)


src/annotator/anchoring/pdf.js
```diff
@@ -446,57 +446,71 @@ function prioritizePages(position) {
...
+    try {
       if (
-        position &&
         quotePositionCache[quote.exact] &&
         quotePositionCache[quote.exact][position.start]
       ) {
         const { pageIndex, anchor } =
           quotePositionCache[quote.exact][position.start];
-        return anchorByPosition(pageIndex, anchor.start, anchor.end);
+        const range = await anchorByPosition(
```
esanzgar, line 500:
```suggestion
        return anchorByPosition(
```

This matches the style of the last return statement in this function.

src/annotator/anchoring/pdf.js
```diff
@@ -446,57 +446,71 @@ function prioritizePages(position) {
...
+      const matchedText = textContent.substr(start, length);
+      if (quote.exact !== matchedText) {
+        throw new Error('quote mismatch');
+      }
 
-        return anchorByPosition(index, start, end);
-      });
-    });
-  }
+      const range = await anchorByPosition(index, start, end);
```
esanzgar, line 485:
```suggestion
      return anchorByPosition(index, start, end);
```

This matches the last return statement in this function.

esanzgar: > This commit disallows anchoring using only position selectors in PDFs.

It seems that if `position` is available it returns the range by `anchorByPosition`. It doesn't seem to be any other condition check.

esanzgar (commented):
Robert, I wonder if the flow could be simplified in something like this:

```
  if (position) {
    try {
      // If we have a position selector, try using that first as it is the fastest
      // anchoring method.
      return searchByPosition(position, quote);
    } catch {
      // If anchoring with the position failed, check for a cached quote-based
      // match using the quote + position as a cache key.
      return searchInCache(position, quote);
    }
  }
```

I don't fully understand how this changes 
> disallows anchoring using only position selectors in PDFs.

esanzgar (approved):
Now, I understand that the changes: before if a `position` was  specified and the anchoring failed the promise was rejected, while now if the anchoring failed it will _also_ try to find the `quote` (as it is also done in the case of HTML anchoring).

src/annotator/anchoring/pdf.js
```diff
@@ -446,57 +446,71 @@ function prioritizePages(position) {
...
-      });
-    });
+    } catch {
+      // Fall back to uncached quote selector match
+    }
   }
 
-  return result;
+  const pageIndices = await prioritizePages(position?.start ?? 0);
+  return findInPages(pageIndices, quote, position);
```
esanzgar, line 513: Correct me if I am wrong, Robert: if the anchoring by position fails and starts searching the quote using `findInPages`, it has the potential to slow down the application? Would `findInPages` will fetch text in the neighbouring pages or in the whole document to find a match? (Sorry, I haven't look at the code).

src/annotator/anchoring/pdf.js
```diff
@@ -446,57 +446,71 @@ function prioritizePages(position) {
...
+    } catch {
+      // Fall back to quote selector
+    }
 
-  if (quote) {
-    result = result.catch(() => {
+    // If anchoring with the position failed, check for a cached quote-based
+    // match using the quote + position as a cache key.
+    try {
       if (
```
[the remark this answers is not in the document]
robertknight: This is true (`position.start` would throw if it were undefined), but I think it makes it easier to follow the logic of the function by having an explicit check, as it makes it clear that the whole block is not applicable if there is no `position`. Also it allows TypeScript to see that `position` is defined inside the block.

src/annotator/anchoring/pdf.js
robertknight: There's a subtle problem with that! Dropping the `await` means that any errors thrown by `anchorByPosition` will no longer cause control to move to the `catch` block below. Changing the code to `return await anchorByPosition(...)` would work, but I thought naming the result was helpful to the reader.

src/annotator/anchoring/pdf.js
robertknight: > Correct me if I am wrong, Robert: if the anchoring by position fails and starts searching the quote using findInPages, it has the potential to slow down the application?

You're right. Searching by quote is slower than searching by position. The most significant overheads are:

1. Getting PDF.js to extract text from additional pages in the document. We mitigate this by caching the text for each page for future calls to `anchor`.
2. Doing an approximate string search for the quote. We mitigate this by first trying to find an exact match, using `string.indexOf(...)`, which is very fast. If we do need to fall back to a fuzzy/approximate search, we rely on the approximate searching implementation being highly optimized.

Of these, (1) is the most slowest part. Fortunately PDF.js does this work in a web worker, so the browser is not unresponsive while it is happening. We may want to show some kind of activity/progress while it is happening in very long documents however.

PR #3708, opened 2021-08-25
Refactor helpers in anchoring/pdf.js to async/await

robertknight:
**Depends on https://github.com/hypothesis/client/pull/3706**

This is some follow-up refactoring to make some of the PDF anchoring code easier to follow. See individual commits for details, but in summary:

- Convert some functions that were conceptually async iteration over all pages in a document to use async/await instead of Promise/chains + recursion
- Remove an unused typedef
- Convert the remaining non-Map cache to use a Map as well

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -16,14 +16,6 @@ import { TextQuoteAnchor } from './types';
  */
 
 /**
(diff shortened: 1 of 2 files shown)


robertknight:
This PR is superceded by https://github.com/hypothesis/client/pull/3736.

PR #3732, opened 2021-09-06
Add cross-origin annotatable iframe dev server scenario

### LINKED ISSUE #3629: Create a different origin guest dev scenario
For the new #3533, it would be important to add a `sidebar` <-> `guest` communication scenario in the dev server where the `guest` frame is from another origin.

esanzgar:
The current inter-frame communication doesn't work if an annotatable
(guest) iframe is from a different origin than the host frame (see
https://github.com/hypothesis/client/pull/3611#issue-696947005). This
will be fixed in a more comprehensive overhaul of the inter-face
communication (see https://github.com/hypothesis/client/issues/3533).

Meanwhile, I add a scenario into the local dev server where the
annotatable iframe is from a different origin than the host frame. For
this, I needed to spawn an additional dev server at port 3002):

```
[11:32:50] Dev web server started at http://localhost:3000/
[11:32:50] Dev web server started at http://localhost:3002/
```

Close #3629

### DIFF
--- a/dev-server/documents/html/cross-origin-iframe.mustache
+++ b/dev-server/documents/html/cross-origin-iframe.mustache
@@ -26,9 +26,9 @@
     <iframe></iframe>
     <script>
       // Set a different port for the iframe `src` property to make it cross-origin.
-      const url = new URL(document.location);
-      const port = url.port === '3000' ? '3002' : '3000';
-      const src = `//${url.hostname}:${port}/document/doyle-embedded`;
+      const { port, hostname } = document.location;
+      const newPort = port === '3000' ? '3002' : '3000';
+      const src = `//${hostname}:${newPort}/document/doyle-embedded`;
       document.querySelector('iframe').setAttribute('src', src);
     </script>
     {{{hypothesisScript}}}
--- a/dev-server/documents/html/cross-origin-iframe.mustache
+++ b/dev-server/documents/html/cross-origin-iframe.mustache
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Cross-origin guest iframe</title>
+    <style>
+      body {
+        font-family: sans-serif;
+      }
+      iframe {
+        width: 75%;
+        height: 300px;
+        resize: both;
+        overflow: auto;
+      }
+    </style>
+
+  </head>
+  <body>
+    <h1>Cross-origin guest iframe</h1>
+    <p>
+      The cross-origin iframe below should behave like a normal guest iframe. It should
+      be possible to select text and add annotations.
+    </p>
+    <iframe></iframe>
+    <script>
+      // Set a different port for the iframe `src` property to make it cross-origin.
+      const url = new URL(document.location);
+      const port = url.port === '3000' ? '3002' : '3000';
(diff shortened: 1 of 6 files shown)


gulpfile.js
```diff
@@ -316,6 +316,8 @@ gulp.task('serve-package', () => {
 
 gulp.task('serve-test-pages', () => {
   serveDev(3000, { clientUrl: `//{current_host}:3001/hypothesis` });
+  // Starts an additional dev web server to test cross-origin functionality
+  serveDev(3002, { clientUrl: `//{current_host}:3001/hypothesis` });
```
esanzgar, line 320: I think this is useful in other context, like for example serving PDFs files from another URL and checking that the fingerprint association allows still to find the annotations.

dev-server/documents/html/cross-origin-iframe.mustache
```diff
@@ -0,0 +1,35 @@
...
+    <p>
+      The cross-origin iframe below should behave like a normal guest iframe. It should
+      be possible to select text and add annotations.
+    </p>
+    <iframe></iframe>
+    <script>
+      // Set a different port for the iframe `src` property to make it cross-origin.
+      let port = new URL(document.location).port === '3000' ? '3002' : '3000';
+      let url = '//localhost:' + port + '/document/burns-embedded';
+      document.querySelector('iframe').setAttribute('src', url);
```
esanzgar, line 31: It is not the most terse JS because JS is not transpiled and I wanted to make it compatible with all the browsers we support. 
Related to #2900

robertknight:
Starting an additional local server on a different port seems reasonable. Could also be useful for testing anchoring based on non-URL identifiers (eg. DOIs).

I suggest using a smaller document for this test unless you really need a long document. Very long files run into issues with things like GitHub's diff viewer not displaying the diff by default.

gulpfile.js
robertknight: Yes, this seems useful.

esanzgar:
I changed the document for `The Disappearance of Lady Carfax, by Arthur Conan Doyle` which is smaller.

robertknight (approved):
LGTM. I think it would be useful to adjust the messages that get logged when the dev server starts up, to make it more obvious what the difference is between the different servers. Currently the messages read:

```
[10:08:01] Package served at http://localhost:3001/hypothesis
[10:08:01] Dev web server started at http://localhost:3000/
[10:08:01] Dev web server started at http://localhost:3002/
```

I think it would be better if these messages read something like:

```
[10:08:01] Client boot script served at http://localhost:3001/hypothesis
[10:08:01] Primary dev server started at http://localhost:3000/
[10:08:01] Alternate dev server started at http://localhost:3002/
```

dev-server/documents/html/cross-origin-iframe.mustache
```diff
@@ -0,0 +1,35 @@
...
+  <body>
+    <h1>Cross-origin guest iframe</h1>
+    <p>
+      The cross-origin iframe below should behave like a normal guest iframe. It should
+      be possible to select text and add annotations.
+    </p>
+    <iframe></iframe>
+    <script>
+      // Set a different port for the iframe `src` property to make it cross-origin.
+      let port = new URL(document.location).port === '3000' ? '3002' : '3000';
```
robertknight, line 29: You can use `document.location.port` here, since there is a `Location.port` property.

PR #3738, opened 2021-09-06
Fix type issue with the latest frontend-shared version

esanzgar:
In the latest version of the frontent-shared package, `extraRoute` has a
mandatory property `group` that must be either
`"home" | "foundations" | "patterns" | "components"`;

There are three ways to fix this:

* make the `group` property optional in the frontend-shared package (I
  don't know if that's a good idea).

* cast the `string` type to one of the allowed types: `    group:
  /** @type {'components'} */ ('components'),`

* convert the `string` type to const without casting (this PR).

### DIFF
--- a/dev-server/ui-playground/index.js
+++ b/dev-server/ui-playground/index.js
@@ -3,11 +3,13 @@ import ButtonPatterns from './components/ButtonPatterns';
 
 import sidebarIcons from '../../src/sidebar/icons';
 
+/** @type {import('@hypothesis/frontend-shared/lib/pattern-library').PlaygroundRoute[]} */
 const extraRoutes = [
   {
     route: '/buttons',
     title: 'Buttons',
     component: ButtonPatterns,
+    group: 'components',
   },
 ];
 


dev-server/ui-playground/index.js
```diff
@@ -3,11 +3,15 @@ import ButtonPatterns from './components/ButtonPatterns';
 
 import sidebarIcons from '../../src/sidebar/icons';
 
+/** @type {import('@hypothesis/frontend-shared/lib/pattern-library').PlaygroundRoute['group']} */
```
robertknight, line 6: It'll be easier to add the cast to the `extraRoute` definition:

```js
/** @type {import('@hypothesis/frontend-shared/lib/pattern-library').PlaygroundRoute[]} */
const extraRoutes = [
  {
    route: '/buttons',
    title: 'Buttons',
    component: ButtonPatterns,
    group: 'components',
  },
];
 ```

This way you only have to specify a type once if there are multiple fields like `group`. You'll also get an earlier error if the `extraRoutes` object has the wrong structure.

robertknight (commented):
I suggest moving the cast to a different place, both to reduce the amount of annotations needed in future and to get more helpful error messages.

dev-server/ui-playground/index.js
esanzgar: Much better, thanks.

An alternative approach that will be available with typescript 4.5 is:

` group: /** @type {const} */ ('components')`

https://github.com/microsoft/TypeScript/issues/30445

robertknight (approved):

dev-server/ui-playground/index.js
robertknight: Ah, that's useful to know. I think I'd still prefer to type `extraRoutes` in this particular case in order to communicate intent to readers.

lyzadanger (approved):
I think this change is fine and innocuous for the moment.

I think it would likely be more ergonomic for authors in the future if the package weren't so persnickity about groups and their type. Perhaps a sensible default fallback for `extraRoutes` that lack a `group`? I'll track this (in `frontend-shared`).

This linting grumpiness is just popping up now in the newest version of the package, which suggests that this is a side effect of the updated type-declaration generation? That is, typing for this hasn't been touched directly in some time. Any insight you might have on that would be helpful!

esanzgar:
@lyzadanger, you are right this is caused by changes in `frontend-shared@3.8.1`.

Previously, only files that were 'connected' to `src/index.js` because of the rule to create types: `tsc --allowJs --declaration --emitDeclarationOnly --outDir lib src/index.js`. That caused the `pattern-library` to not have any typings.

In version 3.8.1, type definitions are generated for all JS files: `"include": ["**/*.js"],` based on the `tsconfig.json`

PR #3739, opened 2021-09-07
Support new PDF.js fingerprint API

### LINKED ISSUE #3673: PDF.js warning about deprecated `PDFDocumentProxy.fingerprint` usage
After upgrading the version of PDF.js in the browser extension to v2.11.106, I am seeing this warning in the console when the PDF viewer loads:

```
pdf.js:479 Deprecated API usage: `PDFDocumentProxy.fingerprint`, please use `PDFDocumentProxy.fingerprints` instead.
```

Looking at [the code](https://github.com/mozilla/pdf.js/blob/43060d39adda43755f257982b1e9c4a986743090/src/display/api.js#L721) that emits the warning, the `fingerprints` property returns an array and the first entry is equivalent to the deprecated `fingerprint` property.

See https://github.com/mozilla/pdf.js/pull/13661 for more information on what the contents of the new `fingerprints` property are. Based on a comment there they don't expect to remove the deprecated property soon:

> Please note: While I'm not sure if we'll ever be able to remove the old PDFDocumentProxy.fingerprint getter, given that it's existed since "forever", that probably isn't a big deal given that it's now limited to only GENERIC-builds.

robertknight:
In https://github.com/mozilla/pdf.js/pull/13661 the API for retrieving
PDF fingerprints (aka. File Identifiers) changed. In generic builds of
PDF.js the old API remains available, but not in the non-generic one
that Firefox's built-in viewer uses.

This commit makes Hypothesis use the new API if available or the old API
otherwise. The fingerprint value should be the same in both cases.

Fixes #3673

----

**Testing:**

Example PDF for testing: https://arxiv.org/ftp/arxiv/papers/2106/2106.04651.pdf, which has fingerprint 835121d1b268bd9c323505dd56851cc0.

1. Install the latest version of Firefox and open a PDF in the native viewer. Use the bookmarklet to load the production Hypothesis client. Click on the "?" button => "About this version" in the Hypothesis sidebar and observe that the fingerprint is displayed as N/A, whereas in Via or the Chrome extension a valid fingerprint should be shown.
2. Switch to this branch and use the bookmarklet to load the development client into Firefox's copy of PDF.js. The fingerprint that the client shows should now match the one that Via displays.

### DIFF
--- a/src/annotator/integrations/pdf-metadata.js
+++ b/src/annotator/integrations/pdf-metadata.js
@@ -16,6 +16,11 @@ import { normalizeURI } from '../util/url';
  * @prop {string} documentFingerprint - The fingerprint of this PDF. This is
  *   referred to as the "File Identifier" in the PDF spec. It may be a hash of
  *   part of the content if the PDF file does not have a File Identifier.
+ *
+ *   PDFs may have two file identifiers. The first is the "original" identifier
+ *   which is not supposed to change if the file is updated and the second
+ *   one is the "last modified" identifier. This property is the original
+ *   identifier.
  */
 
 /**
@@ -117,7 +122,7 @@ export class PDFMetadata {
     return this._loaded.then(app => {
       let uri = getPDFURL(app);
       if (!uri) {
-        uri = fingerprintToURN(app.pdfDocument.fingerprint);
+        uri = fingerprintToURN(getFingerprint(app));
       }
       return uri;
     });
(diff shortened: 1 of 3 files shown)


esanzgar (approved):
👍🏼 

In the future it could be advantageous to store both fingerprint identifiers. I could imaging a feature that allows to show only the annotations  that apply  to a the specific version of the PDF. Something to think about.

robertknight:
> In the future it could be advantageous to store both fingerprint identifiers.

Yes. The second fingerprint is intended to be a version identifier of sorts. We don't currently have a concept of a "document version" anywhere in h, and a standardized method of storing that version as part of the annotation's metadata, so we'd need to think about that at the same time.

PR #3741, opened 2021-09-07
Provide `externalContainerSelector` config in a different way

esanzgar:
Provide `externalContainerSelector` config in a different way

Instead of override the window.hypothesisConfig, I provide the
`externalContainerSelector` through a JSON file. This allows a more
incremental way to add configuration options.


robertknight:
>  I provide the externalContainerSelector through a JSON file

You mean JSON _script tag_, right?

> This allows a more incremental way to add configuration options.

The change seems fine. Do you have planned changes that will rely on this?

robertknight:
> No, immediately, However, I thought this changes allows to more easily and intuitively test additional options. For example, in addition to testing the externalContainerSelector, one could temporarily modify client-config.mustache (which is the most common place to modify client options) to test combination of options: externalContainerSelector and openSidebar, or externalContainerSelector and theme, for example.

Were you thinking that the settings in `window.hypothesisConfig` are exclusive and override the settings from JSON tags? The settings from `window.hypothesisConfig` are not exclusive unless the `ignoreOtherConfiguration` option is specified. Rather they are layered on top of settings from JSON script tags.

For the current settings that need to be configured in `sidebar-external-container.mustache` either method of specifying the settings works equally well. As I see it the main advantage of `window.hypothesisConfig` is that it provides more flexibility as you can generate the config from runtime information. The main advantage of JSON script tags by contrast is security if you are generating the config and then injecting it into a template, since there is no chance of it being executed as JS code.

esanzgar:
If there are two `window.hypothesisConfig`, like in this case, the last one wins.

```
    {{{hypothesisScript}}}   <- defines `window.hypothesisConfig`
    <script>
      window.hypothesisConfig = function () {    <- it overrides the previous one
        return {
          externalContainerSelector: '#my-sidebar',
        };
      };
    </script>
```

I prefer having the `externalContainerSelector` provided in a JSON script tag and _supplemented_ by the `window.hypothesisConfig` that its is defined with `{{{hypothesisScript}}}`.

PR #3742, opened 2021-09-07
Fix security error mixing cross-origin frames and promises in old browsers

robertknight:
Fix an issue where the sidebar failed to appear in Safari <= 11 and Chrome
63 when returning a cross-origin Window from a Promise `then` callback.
In these browsers an exception is triggered when the browser tries to
test if the return value is a Promise that can be unwrapped.

Based on bug reports in external trackers, it looks like this affects Firefox <= 59 as well.
Edge 17 and 18 also seem to be affected.

The issue was resolved in more recent browsers by adding an undefined
`then` property to cross origin objects. [1]

[1] https://github.com/whatwg/dom/issues/536

### DIFF
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -68,9 +68,9 @@ function init() {
     sidebar = new Sidebar(document.body, eventBus, guest, getConfig('sidebar'));
 
     // Expose sidebar window reference for use by same-origin guest frames.
-    window_.__hypothesis.sidebarWindow = sidebar.ready.then(
-      () => sidebar.iframe.contentWindow
-    );
+    window_.__hypothesis.sidebarWindow = sidebar.ready.then(() => [
+      sidebar.iframe.contentWindow,
+    ]);
   }
 
   // Clear `annotations` value from the notebook's config to prevent direct-linked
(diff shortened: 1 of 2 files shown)


robertknight:
The code that triggers this issue has been live for over a month (it was committed in 036fb8c6bd on 2021-07-26 and shipped 3 days later) and it was only noticed by @esanzgar doing some internal testing.

This makes me think that we can probably update our supported browser baselines pretty soon to versions that are at least newer than all the affected versions here (ie. require Chrome >= 64, Firefox >= 60, Edge >= 79, Safari >= 12).

esanzgar (approved):
I confirm this PR fixes the issue in the following web browsers:
* Firefox 53
* Chrome 57
* Safari 10

It is not however working for Edge 17 and 18. There is a `SyntaxError` throw but it maybe unrelated.

robertknight:
> It is not however working for Edge 17 and 18. There is a SyntaxError throw but it maybe unrelated.

Yes. I confirm there appears to be an additional issue that affects these browsers. I confirmed that they would also be affected by this issue though. Given [low usage](https://github.com/hypothesis/client/issues/3744), I'm going to suggest that we drop support for legacy Edge entirely.

PR #3748, opened 2021-09-13
Improve speed of the Hypothesis client injection

esanzgar:
For big documents the `window.onload` event can take a while to trigger.
The Hypothesis client can be injected as soon as the document is in
`readyState === 'interactive'`, or when DOMContentLoaded` is fired,
which is the same.

In this PR instead of waiting for `window.onload`, we use a polling
logic to check the `readyState` or to default to the `DOMContentLoaded`
event.

For a medium document, on Chrome, I have noticed the logic improves the
speed by 200ms (from 791ms to 564ms). I didn't observe improvements on
Firefox and Safari (both are substantially faster than Chrome). On large
documents, this new logic should result in speed improvement, on all web
browsers.

### DIFF
--- a/src/annotator/frame-observer.js
+++ b/src/annotator/frame-observer.js
@@ -108,7 +108,7 @@ export class FrameObserver {
  * @throws {Error} if trying to access a document from a cross-origin iframe
  */
 export function onDocumentReady(frame) {
-  return new Promise(resolve => {
+  return new Promise((resolve, reject) => {
     // @ts-expect-error
     const frameDocument = frame.contentWindow.document;
     const { readyState, location } = frameDocument;
@@ -124,11 +124,7 @@ export function onDocumentReady(frame) {
       frame.hasAttribute('src') &&
       frame.src !== 'about:blank'
     ) {
-      // Unfortunately, listening for 'DOMContentLoaded' on the iframeDocument
-      // doesn't work. Instead, we need to wait for a 'load' event to be triggered.
-      frame.addEventListener('load', () => {
-        resolve();
-      });
+      setTimeout(() => onDocumentReady(frame).then(resolve).catch(reject), 10);
       return;
     }
 
--- a/src/annotator/frame-observer.js
+++ b/src/annotator/frame-observer.js
@@ -107,7 +107,7 @@ export class FrameObserver {
  * @return {Promise<void>}
  * @throws {Error} if trying to access a document from a cross-origin iframe
  */
-export function onDocumentReady(frame) {
+export function onDocumentReady(frame, counter = 0) {
   return new Promise((resolve, reject) => {
     // @ts-expect-error
     const frameDocument = frame.contentWindow.document;
@@ -124,7 +124,19 @@ export function onDocumentReady(frame) {
       frame.hasAttribute('src') &&
       frame.src !== 'about:blank'
     ) {
-      setTimeout(() => onDocumentReady(frame).then(resolve).catch(reject), 10);
+      setTimeout(() => {
+        // If `onDocumentReady` is recursively called more than 100 times
+        // (10ms * 100 = 1s) exit.
+        if (counter >= 99) {
+          reject(new Error('Annotatable iframe took too long to load'));
+          return;
+        }
+
+        onDocumentReady(frame, counter + 1)
+          .then(resolve)
+          .catch(reject);
+      }, 10);
+
       return;
     }
 
--- a/src/annotator/test/frame-observer-test.js
+++ b/src/annotator/test/frame-observer-test.js
@@ -12,6 +12,7 @@ describe('annotator/frame-observer', () => {
     let onFrameRemoved;
     const sandbox = sinon.createSandbox();
 
+    // FrameObserver waits a few milliseconds (DEBOUNCE_WAIT) to detect changes in the DOM
     function waitForFrameObserver() {
       return new Promise(resolve => setTimeout(resolve, DEBOUNCE_WAIT + 10));
     }
@@ -142,6 +143,10 @@ describe('annotator/frame-observer', () => {
       iframe.setAttribute('src', 'http://cross-origin.dummy');
       let expectedError;
 
+      // In this particular case waiting for the FrameObserver to detect the new
+      // iframe may not be enough. Because the browser fetches the URL in `src`
+      // (it is not reachable) it could take longer, that's why we, in addition,
+      // wait for the iframe's document to completely load.
       try {
         await onDocumentReady(iframe);
       } catch (error) {
@@ -149,13 +154,14 @@ describe('annotator/frame-observer', () => {
       }
       await waitForFrameObserver();
 
-      assert.isDefined(expectedError);
       assert.notCalled(onFrameAdded);
+      assert.isDefined(expectedError); // Cross-origin error
       assert.calledOnce(console.warn);
     });
   });
 
   describe('onDocumentReady', () => {
+    let clock;
     let fakeIFrame;
     let fakeIFrameDocument;
 
@@ -170,20 +176,40 @@ describe('annotator/frame-observer', () => {
     }
 
     beforeEach(() => {
+      clock = sinon.useFakeTimers();
       fakeIFrameDocument = new FakeIFrameDocument();
       fakeIFrame = document.createElement('div');
       fakeIFrame.contentWindow = { document: fakeIFrameDocument };
       fakeIFrame.setAttribute('src', 'http://my.dummy');
     });
 
+    afterEach(() => {
+      clock.restore();
+    });
+
     it('waits for the iframe to change the location and readyState if the initial document is blank', () => {
       fakeIFrameDocument.location.href = 'about:blank';
       const onReady = onDocumentReady(fakeIFrame);
 
-      fakeIFrameDocument.location.href = 'http://my.dummy';
-      fakeIFrameDocument.readyState = 'complete';
+      // After the initial 'about:blank' document, a new document is loaded.
+      const newDocument = new FakeIFrameDocument();
+      newDocument.location.href = 'http://my.dummy';
+      newDocument.readyState = 'complete';
+      fakeIFrame.contentWindow = { document: newDocument };
+      clock.tick(10);
 
       return onReady;
+    });
+
+    it('rejects promise if the document stays in the initial blank document too long', () => {
+      fakeIFrameDocument.location.href = 'about:blank';
+      const promise = onDocumentReady(fakeIFrame).catch(error =>
+        assert.equal(error.message, 'Annotatable iframe took too long to load')
+      );
+
+      clock.tick(1000);
+
+      return promise;
     });
 
     it('waits for the iframe DOMContentLoaded event to be triggered if the document is loading', () => {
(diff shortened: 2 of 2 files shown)


src/annotator/frame-observer.js
```diff
@@ -126,11 +126,7 @@ export function onDocumentReady(frame) {
       frame.hasAttribute('src') &&
       frame.src !== 'about:blank'
     ) {
-      // Unfortunately, listening for 'DOMContentLoaded' on the iframeDocument
-      // doesn't work. Instead, we need to wait for a 'load' event to be triggered.
-      frame.addEventListener('load', () => {
-        resolve();
-      });
+      setTimeout(() => onDocumentReady(frame).then(resolve).catch(reject), 10);
```
robertknight, line 129: Are there situations where this could loop indefinitely? If so it might be wise to add a timeout to the whole process. Also, how long does the initial navigation typically take? I'm wondering how many recursive calls to `onDocumentReady` will happen in the typical case.

src/annotator/test/frame-observer-test.js
```diff
@@ -140,10 +140,16 @@ describe('annotator/frame-observer', () => {
...
 
-      await onDocumentReady(iframe);
+      try {
+        await onDocumentReady(iframe);
+      } catch (error) {
+        expectedError = error;
+      }
       await waitForFrameObserver();
 
+      assert.isDefined(expectedError);
```
robertknight, line 152: What is the actual error that is thrown here?

src/annotator/test/frame-observer-test.js
```diff
@@ -140,10 +140,16 @@ describe('annotator/frame-observer', () => {
...
       iframe.setAttribute('src', 'http://cross-origin.dummy');
+      let expectedError;
 
-      await onDocumentReady(iframe);
+      try {
+        await onDocumentReady(iframe);
+      } catch (error) {
+        expectedError = error;
+      }
       await waitForFrameObserver();
```
robertknight, line 150: Unrelated to the changes in this PR, the `waitForFrameObserver` helper would benefit from a comment to clarify what exactly it is waiting for.

src/annotator/test/frame-observer-test.js
```diff
@@ -140,10 +140,16 @@ describe('annotator/frame-observer', () => {
     it("doesn't trigger onFrameAdded when annotatable iframe is from a different domain", async () => {
       const iframe = createAnnotatableIFrame();
       iframe.setAttribute('src', 'http://cross-origin.dummy');
+      let expectedError;
 
-      await onDocumentReady(iframe);
+      try {
+        await onDocumentReady(iframe);
+      } catch (error) {
+        expectedError = error;
```
robertknight, line 148: So the fact that `onDocumentReady` resolved previously with a cross-origin iframe was a bug?

Also, can you clarify why the `onDocumentReady` call is needed here? The other tests just use the `await waitForFrameObserver()` step before checking whether the frame was injected or not.

src/annotator/test/frame-observer-test.js
```diff
@@ -170,13 +176,14 @@ describe('annotator/frame-observer', () => {
       fakeIFrame.setAttribute('src', 'http://my.dummy');
     });
 
-    it('waits for the iframe load event to be triggered if the document is blank', () => {
+    it('waits for the iframe to change the location and readyState if the initial document is blank', () => {
```
robertknight, line 179: When the navigation from `about:blank` to the URL specified in the `src` attribute happens, _the initial document is replaced_. In other words, `frame.contentDocument` (and `fakeIFrameDocument` here) changes to a different object. I think that needs to either be simulated correctly, or the test should be reworked to avoid relying on a fake.

robertknight (commented):
I had an initial look over this and had a few comments and queries.

src/annotator/test/frame-observer-test.js
esanzgar: Would this be OK?
`// FrameObserver waits a few milliseconds (DEBOUNCE_WAIT) to detect changes in the DOM`

src/annotator/test/frame-observer-test.js
esanzgar: > So the fact that onDocumentReady resolved previously with a cross-origin iframe was a bug?

No, it wasn't. The cross-origin iframe fired the `load` event normally. The expected error would raised later, when trying to execute the `_onFrameAdded` function to inject the client. However, that line is wrapped in a `try/catch` and we tested the `console.warn` execution. With the changes in this PR, however, the error is thrown in the `onDocumentReady`.

> Also, can you clarify why the onDocumentReady call is needed here? 

It is not strictly necessary. I added `onDocumentReady` because the iframe may not load  in 50s (`waitForFrameObserver`). It may take a few more milliseconds more to do to try to fetch a non-existent URL.

I have remove it anyway because it doesn't seem necessary.

src/annotator/test/frame-observer-test.js
esanzgar: I correct it.

src/annotator/frame-observer.js
esanzgar: Actually, I was able to create a situation where the loop runs indefinitely. In Safari (and sometimes in Chrome), if the iframe has an unreachable site in the `src` (like 'http://unreachable.gov'), the iframe stays in the initial `about:blank` document, causing an infinite loop. 

Here is a video:

https://user-images.githubusercontent.com/8555781/133477444-daf92eff-eba7-4d49-9bd9-085cf5bcc99a.mov

I include an additional commit that includes a fix for that.

src/annotator/test/frame-observer-test.js
esanzgar: Actually, I will put it back because sometimes it does take a bit longer than 50s:

```
Chrome Headless 93.0.4577.0 (Linux x86_64) annotator/frame-observer FrameObserver doesn't trigger onFrameAdded when annotatable iframe is from a different domain FAILED
	AssertError: expected warn to be called once but was called 0 times
	    at Function.fail (/home/runner/work/client/client/node_modules/sinon/pkg/sinon.js:174:25)
	    at failAssertion (/home/runner/work/client/client/node_modules/sinon/pkg/sinon.js:120:20)
	    at Function.assert.<computed> [as calledOnce] (/home/runner/work/client/client/node_modules/sinon/pkg/sinon.js:149:17)
	    at Context.<anonymous> (/tmp/annotator/test/frame-observer-test.js:147:14 <- /tmp/31ce315758c0f34b60d4b4aa2b02be64.browserify.js:222901:14)
Chrome Headless 93.0.4577.0 (Linux x86_64): Executed 456 of 2894 (1 FAILED) (0 secs / 1.279 secs)
WARN: 'Unable to inject the Hypothesis client (from 'http://localhost:9876/context.html' into a cross-origin frame 'http://cross-origin.dummy/')'
```

You can see above that the `console.warn` function has been executed (from the catch block) but the after the test has finished.

In this example the CI in Jenkins finished OK, while in GH didn't (with the above error):
Screenshot 2021-09-16 at 10 45 49

src/annotator/frame-observer.js
```diff
@@ -97,33 +101,51 @@ export class FrameObserver {
...
+    // replaced by the final document.
+    if (
+      location.href === 'about:blank' &&
+      frame.hasAttribute('src') &&
+      frame.src !== 'about:blank'
+    ) {
+      setTimeout(() => {
+        // If `onDocumentReady` is recursively called more than 100 times
+        // (10ms * 100 = 1s) exit.
+        if (counter >= 99) {
```
robertknight, line 130: This might end up being a little short if the system is slow and under heavy load. We can easily adjust it though.

src/annotator/test/frame-observer-test.js
robertknight: > It may take a few more milliseconds more to do to try to fetch a non-existent URL.

How about using an invalid local URL instead (eg. `http://localhost:$PORT` where $PORT is some value we can be decently sure is not likely to be in use)?

src/annotator/test/frame-observer-test.js
```diff
@@ -4,128 +4,230 @@ import {
...
+      const onReady = onDocumentReady(fakeIFrame);
+
+      // After the initial 'about:blank' document, a new document is loaded.
+      const newDocument = new FakeIFrameDocument();
+      newDocument.location.href = 'http://my.dummy';
+      newDocument.readyState = 'complete';
+      fakeIFrame.contentWindow = { document: newDocument };
+      clock.tick(10);
+
+      return onReady;
```
robertknight, line 201: This check verifies that `onDocumentReady`'s result does resolve, but you could still have a bug where it resolves too soon (before the document has changed). In other words, this doesn't check the "waits for" part.

src/annotator/test/frame-observer-test.js
```diff
@@ -4,128 +4,230 @@ import {
...
+
+    it('rejects promise if the document stays in the initial blank document too long', () => {
+      fakeIFrameDocument.location.href = 'about:blank';
+      const promise = onDocumentReady(fakeIFrame).catch(error =>
+        assert.equal(error.message, 'Annotatable iframe took too long to load')
+      );
+
+      clock.tick(1000);
+
+      return promise;
```
robertknight, line 212: If `onDocumentReady` incorrectly resolves rather than rejects, this test will still pass. This is why we commonly use the `let error; try { await operation(); } catch (e) { error = e }; /* check error */` pattern for checking for promise rejections.

src/annotator/test/frame-observer-test.js
```diff
@@ -4,128 +4,230 @@ import {
...
+    });
+
+    it('waits for the iframe DOMContentLoaded event to be triggered if the document is loading', () => {
+      fakeIFrameDocument.location.href = 'about:srcdoc';
+      fakeIFrameDocument.readyState = 'loading';
+      const onDOMContentLoaded = onDocumentReady(fakeIFrame);
+
+      fakeIFrameDocument.dispatchEvent(new Event('DOMContentLoaded'));
+
+      return onDOMContentLoaded;
```
robertknight, line 222: This checks that the promise resolves, but not which order it resolves in.

src/annotator/test/frame-observer-test.js
```diff
@@ -4,128 +4,230 @@ import {
...
+    });
+
+    it("resolves immediately if document is 'complete' or 'interactive'", () => {
+      fakeIFrameDocument.location.href = 'about:srcdoc';
+      fakeIFrameDocument.readyState = 'complete';
+      const onReady = onDocumentReady(fakeIFrame);
```
robertknight, line 228:
```suggestion
      return onDocumentReady(fakeIFrame);
```

`onReady` variable seems superfluous here.

robertknight (commented):
The timeout logic looks reasonable. I found some potential issues in tests that could pass when they shouldn't and also had a suggestion about the potentially variable delay in loading an invalid URL (presumably due to DNS lookups).

src/annotator/frame-observer.js
esanzgar: For a Chrome and Firefox any document (of any size) takes 1 or 2 calls.

For Safari, it seems that it is related to the size of the document. For medium documents takes 1 or 2 recursive calls. For large documents (https://html.spec.whatwg.org/), it can take up to 40 calls with my internet speed. What would happened when offline?

esanzgar:
I have encountered cases where the mechanism proposed in this PR to speed the Hypothesis client injection resulted in failures to inject the client. I would rather be a few milliseconds slower and 100% robust than marginally faster and flaky.

PR #3757, opened 2021-09-13
Use maps for PDF text and anchoring caches

robertknight:
Using maps is a little faster and avoids the need for `hasOwnProperty` checks (which we weren't doing, but probably should have). The quote position cache has been converted to a single level map as this is more ergonomic to work with and the second level maps almost always had exactly one entry in them.

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -42,21 +42,31 @@ export const RenderingStates = {
 /**
  * Map of page index to page text content.
  *
- * @type {Record<number,Promise<string> | undefined>}
+ * @type {Map<number,Promise<string> | undefined>}
  */
-let pageTextCache = {};
+const pageTextCache = new Map();
 
 /**
- * A cache that maps a `(quote, text offset in document)` key to a specific
+ * A cache that maps a `{quote}:{offset}` key to a specific
  * location in the document.
  *
  * The components of the key come from an annotation's selectors. This is used
  * to speed up re-anchoring an annotation that was previously anchored in the
  * current session.
  *
- * @type {Object<string, Object<number, PdfTextRange>>}
+ * @type {Map<string, PdfTextRange>}
  */
-let quotePositionCache = {};
+const quotePositionCache = new Map();
+
+/**
+ * Return a cache key for lookups in `quotePositionCache`.
+ *
+ * @param {string} quote
+ * @param {number} [pos] - Offset in document text
+ */
+function quotePositionCacheKey(quote, pos) {
+  return `${quote}:${pos}`;
+}
 
 /**
  * Return offset of `node` among its siblings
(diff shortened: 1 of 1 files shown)


src/annotator/anchoring/pdf.js
```diff
@@ -42,21 +42,31 @@ export const RenderingStates = {
 /**
  * Map of page index to page text content.
  *
- * @type {Record<number,Promise<string> | undefined>}
+ * @type {Map<number,Promise<string> | undefined>}
```
esanzgar, line 45:
```suggestion
 * @type {Map<number,Promise<string>}
```

src/annotator/anchoring/pdf.js
```diff
@@ -42,21 +42,31 @@ export const RenderingStates = {
...
  */
-let pageTextCache = {};
+const pageTextCache = new Map();
 
 /**
- * A cache that maps a `(quote, text offset in document)` key to a specific
+ * A cache that maps a `{quote}:{offset}` key to a specific
```
esanzgar, line 50: This is a good idea to reduce the number of _collisions_. Several annotation could have the same quote but could be anchored in different places.

However, I feel we should reduce further the possibility of hitting the wrong cache by adding the start and end offset of the anchored text: 

```
function quotePositionCacheKey(quote, start, end) {
  return `${quote}:${start}:${end}`;
```

Imaging two annotations that start in the same position, but finish in different one, with the proposed change, each will have their own cache.

Furthermore, we have different offsets systems depending of the PDF.js version. Adding as much information as possible in the cache key, could further reduce possible _collisions_.

src/annotator/anchoring/pdf.js
```diff
@@ -541,12 +549,10 @@ export async function anchor(root, selectors) {
     // If anchoring with the position failed, check for a cached quote-based
     // match using the quote + position as a cache key.
     try {
-      if (
-        quotePositionCache[quote.exact] &&
-        quotePositionCache[quote.exact][position.start]
-      ) {
-        const { pageIndex, anchor } =
-          quotePositionCache[quote.exact][position.start];
+      const cacheKey = quotePositionCacheKey(quote.exact, position.start);
```
esanzgar, line 552: On a future PR, I am a proponent of checking the cache first, by moving this try/catch above the previous one.

esanzgar (changes requested):
In line 182, your are accessing a field of the `pageTextCache` instead of the cache. It should be:
```
  pageTextCache.get(pageIndex);
```

I added a few other suggestions.

src/annotator/anchoring/pdf.js
robertknight: We discussed on Slack, but to summarize: The end offset is determined by the start offset and quote length, so in this case it would always be the same for given `quote` and `start` arguments.

src/annotator/anchoring/pdf.js
robertknight: Yes, this would make sense. It would be faster as it would avoid work to recompute the page index. Currently only quote-based anchoring populates this cache, but it would make sense to populate it for position-based anchoring too.

esanzgar (approved):
👍🏼

PR #3758, opened 2021-09-14
Improve visibility of focused PDF highlights

robertknight:
When the same text is highlighted multiple times in a PDF, it can be hard to
distinguish focused highlights as the `<rect>`s for focused highlights
could be obscured by `<rect>`s for overlapping unfocused highlights.

SVG elements are painted in the order they appear in the document and don't
yet support a z-index property [1], so this commit fixes the issue by moving
`<rect>`s to be last amongst siblings when focused.

[1] The SVG 2 spec does support a z-index attribute, but none of the major
    browsers support it yet. See https://bugs.webkit.org/show_bug.cgi?id=90738,
    https://bugs.chromium.org/p/chromium/issues/detail?id=670177 and
    https://bugzilla.mozilla.org/show_bug.cgi?id=360148.

---

These screenshots show the effect on text that has several (3+) overlapping highlights. The annotation card on the right corresponds to the first highlight from this set which is drawn first.

**All highlights unfocused:**

PDF highlight unfocused

**Bottom highlight focused (before):**

PDF highlight focused - original

The blue color for the focused highlight is obscured by the yellow unfocused highlights drawn above it.

**Bottom highlight focused (after):**

PDF highlight focused - improved

The focused highlight's SVG `<rect>` has been moved to the top of the z-index so it is now easier to see.

### DIFF
--- a/src/annotator/highlighter.js
+++ b/src/annotator/highlighter.js
@@ -331,6 +331,16 @@ export function setHighlightsFocused(highlights, focused) {
     // `<hypothesis-highlight>` element.
     if (h.svgHighlight) {
       h.svgHighlight.classList.toggle('is-focused', focused);
+
+      // Ensure that focused highlights are drawn above un-focused highlights
+      // on the same page.
+      //
+      // SVG elements are rendered in document order so to achieve this we need
+      // to move the element to be the last child of its parent.
+      if (focused) {
+        const parent = /** @type {SVGElement} */ (h.svgHighlight.parentNode);
+        parent.append(h.svgHighlight);
+      }
     } else {
       h.classList.toggle('hypothesis-highlight-focused', focused);
     }
--- a/src/annotator/test/highlighter-test.js
+++ b/src/annotator/test/highlighter-test.js
@@ -485,6 +485,23 @@ describe('annotator/highlighter', () => {
       );
     });
 
+    it('raises focused highlights in PDFs', () => {
+      const root = document.createElement('div');
+      render(<PdfPage />, root);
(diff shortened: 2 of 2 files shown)


src/annotator/highlighter.js
```diff
@@ -331,6 +331,16 @@ export function setHighlightsFocused(highlights, focused) {
...
       h.svgHighlight.classList.toggle('is-focused', focused);
+
+      // Ensure that focused highlights are drawn above un-focused highlights
+      // on the same page.
+      //
+      // SVG elements are rendered in document order so to achieve this we need
+      // to move the element to be the last child of its parent.
+      if (focused) {
+        const parent = /** @type {SVGElement} */ (h.svgHighlight.parentNode);
+        parent.append(h.svgHighlight);
```
esanzgar, line 342: There is not much difference here (https://developer.mozilla.org/en-US/docs/Web/API/Element/append), but I tend to use more `appendChild` because it feels safer.

esanzgar (approved):
Very nice!

esanzgar:
I didn't observed what Robert described. However, I saw several blue shades when focusing on the different annotations:

https://user-images.githubusercontent.com/8555781/133293812-3ac6011e-d953-4eac-b9e4-e3dcbf1d99a3.mov


This PR produce a more consistent blue color when focusing on annotations at the cost of a slight flickering because of the reordering of elements:

https://user-images.githubusercontent.com/8555781/133294002-e341a32e-0105-4b56-810b-10ad8ce97fd7.mov

PR #3762, opened 2021-09-17
Add documentation to test cases

esanzgar:
* add information about why to use `window.load` instead of
  `document.DOMContentLoaded` to check when a document is ready

* use an invalid local URL (http://localhost:1) to speed DNS lookup in
  the test

* include some comments about the use of `waitForFrameObserver` in the
  tests


src/annotator/frame-observer.js
```diff
@@ -124,8 +124,9 @@ export function onDocumentReady(frame) {
       frame.hasAttribute('src') &&
       frame.src !== 'about:blank'
     ) {
-      // Unfortunately, listening for 'DOMContentLoaded' on the iframeDocument
-      // doesn't work. Instead, we need to wait for a 'load' event to be triggered.
+      // Listening to `document.DOMContentLoaded` doesn't work because the
+      // document is replaced. On the other hand, listening to `window.load`
+      // works because window doesn't change.
```
esanzgar, line 129: I could add a reference to https://github.com/hypothesis/client/pull/3748 about the failed efforts to avoid `window.load`

src/annotator/frame-observer.js
```diff
@@ -124,8 +124,9 @@ export function onDocumentReady(frame) {
       frame.hasAttribute('src') &&
       frame.src !== 'about:blank'
     ) {
-      // Unfortunately, listening for 'DOMContentLoaded' on the iframeDocument
-      // doesn't work. Instead, we need to wait for a 'load' event to be triggered.
+      // Listening to `document.DOMContentLoaded` doesn't work because the
```
robertknight, line 127:
```suggestion
      // Listening to `DOMContentLoaded` on the frame's document doesn't work because the
```

I guess you were trying to use "document.DOMContentLoaded" as a shorthand, but it is ambiguous as to what "document" refers to (eg. it _doesn't_ mean the `document` global).

src/annotator/frame-observer.js
```diff
@@ -124,8 +124,9 @@ export function onDocumentReady(frame) {
...
+      // document is replaced. On the other hand, listening to `window.load`
```
robertknight, line 128:
```suggestion
      // document is replaced. On the other hand, listening the frame's `load`
```

It is important to be clear that this refer's to the _frame's_ `load` event, not the load event on a Window (either the `window` global or `frame.contentWindow`). The frame's `contentWindow` changes when the document does. It is only the frame itself that does not change.

robertknight (commented):
I suggested some small revisions to the comment about listening for `DOMContentLoaded` vs `load`. The important concept to understand is the difference in lifetime of the frame's `contentDocument` and `contentWindow` (these change after a navigation) compared to the frame itself (does not change).

src/annotator/frame-observer.js
esanzgar: Thank you for the clarification.

So if I understand this correctly this is the hierarchy:

```
global Window (with load event)
|
|-> document (with a load event)
        | 
        |-> frame (with a load event) ** this is the one that  we need **
                 |
                 |-> window (with a load event)
                        |
                        | - > document (with a load event)
```

PR #3764, opened 2021-09-17
Simplify PDF viewer initialization in the dev server

robertknight:
Remove logic that is no longer required with the current Hypothesis
client and PDF.js version.

See https://github.com/hypothesis/browser-extension/pull/658.

### DIFF
--- a/dev-server/static/scripts/pdfjs-init.js
+++ b/dev-server/static/scripts/pdfjs-init.js
@@ -1,62 +1,21 @@
 'use strict';
 
-// Note: This file is not transpiled.
+/* global PDFViewerApplication, PDFViewerApplicationOptions */
+
 // Listen for `webviewerloaded` event to configure the viewer after its files
 // have been loaded but before it is initialized.
 document.addEventListener('webviewerloaded', () => {
-  const appOptions = window.PDFViewerApplicationOptions;
   const app = window.PDFViewerApplication;
 
-  // Ensure that PDF.js viewer events such as "documentloaded" are dispatched
-  // to the DOM. The client relies on this.
-  appOptions.set('eventBusDispatchToDOM', true);
-
-  // Disable preferences support, as otherwise this will result in `eventBusDispatchToDOM`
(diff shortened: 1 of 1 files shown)


dev-server/static/scripts/pdfjs-init.js
```diff
@@ -1,62 +1,21 @@
 'use strict';
 
-// Note: This file is not transpiled.
+/* global PDFViewerApplication, PDFViewerApplicationOptions */
```
esanzgar, line 3: It is not clear why you use `/* */` format for this comment.
```suggestion
// PDFViewerApplication and PDFViewerApplicationOptions are defined globally
```

dev-server/static/scripts/pdfjs-init.js
```diff
@@ -1,62 +1,21 @@
...
-      typeof Promise === 'function' &&
-      typeof PromiseRejectionEvent === 'undefined'
-    ) {
-      window.PromiseRejectionEvent = function FakePromiseRejectionEvent() {
-        // core-js doesn't actually use this, it just tests for `typeof PromiseRejectionEvent`
-        console.warn('Tried to construct fake `PromiseRejectionEvent`');
-      };
-    }
-
+  PDFViewerApplication.initializedPromise.then(() => {
```
esanzgar, line 18: Since you are using  the global `PDFViewerApplication` here, I would remove the use of the `app` variable (lines 8 and 32).

esanzgar (approved):
I just leave a comment about removing the `app` variable.

dev-server/static/scripts/pdfjs-init.js
robertknight: This comment is for ESLint. See https://eslint.org/docs/rules/no-undef. It requires the use of `/* ... */`.

esanzgar: I asked because for some reason `eslint` doesn't complain if `/* global .... */` is removed.

PR #3765, opened 2021-09-20
Replace Object for object

esanzgar:
For the past few months, we have started to use the `object` type instead of `Object`. Both are almost interchangeable, but we prefer `object` for simplicity.

This PR substitutes all `Object` to `object` and `Object.<...>` to `Record<...>` and makes the code more consistent.

* First commit is an automatic transformation
* Second commit is a manual fix of a few types


src/sidebar/helpers/theme.js
```diff
@@ -62,7 +62,7 @@ export function applyTheme(themeProperties, settings) {
 
   themeProperties.forEach(themeProp => {
     const propertyName = supportedThemeProperties[themeProp];
-    const propertyValue = settings.branding[themeProp];
+    const propertyValue = settings.branding?.[themeProp];
```
esanzgar, line 65: The `if` statement above checks that is `setting.branding` is defined. The compiler unfortunately still complains. I decided to solved in this way (I could have ignore the error, or cast)

src/sidebar/store/modules/filters.js
```diff
@@ -46,7 +46,7 @@ import { createStoreModule } from '../create-store';
  */
 
 /**
- * @typedef {Record<FilterKey, FilterOption>} Filters
```
esanzgar: Maybe better:

```suggestion
 * @typedef {Record<FilterKey, FilterOption>|{}} Filters
 ```

File: src/sidebar/helpers/theme.js (hunk 1 above)
Comment (robertknight) [reply src/sidebar/helpers/theme.js:65]: The way I usually resolve this is by assigning to the property in the parent scope (eg. `const branding = settings.branding`) and then referencing that in the callback. This makes it easier for TS to see that the value can't change. In this context you could also just change the `forEach` to a `for ... of` loop.

File: src/sidebar/store/modules/filters.js [hunk 3]
```diff
@@ -46,7 +46,7 @@ import { createStoreModule } from '../create-store';
  */
 
 /**
- * @typedef {Object.<FilterKey, FilterOption>} Filters
+ * @typedef {{user?: FilterOption}} Filters
```
robertknight, line 49: Should this not be `Record<FilterKey, FilterOption>`?

robertknight (approved):
LGTM. I had a suggestion about the case where a property is checked for null-ness and then referenced from within a closure.

src/sidebar/store/modules/filters.js
esanzgar: Because we assign empty objects to filter there are two solutions:
* make the key optionals
* or define filter as `Record<FilterKey, FilterOption>|{}`

I went for the last option.

PR #3787, opened 2021-09-27
Replace deprecated method `String.prototype.substr`

esanzgar:
`String.protoype.substr` is a deprecated method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr). It is recommended to be replaced by `String.prototype.substring` method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring).

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -533,9 +533,8 @@ export async function anchor(root, selectors) {
       const { index, offset, text } = await findPageByOffset(position.start);
       const start = position.start - offset;
       const end = position.end - offset;
-      const length = end - start;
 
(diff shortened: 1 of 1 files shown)


robertknight (approved):
I wasn't sure why `substr` was deprecated, although `substring(...)` is more consistent with how eg. `slice` works. https://stackoverflow.com/a/52640366/434243 has some info. In this context `slice(...)` would have worked fine as well.

esanzgar:
I believe `subtr` will never disappear in practice, but as you said it is better to use methods that are in the ECMA standard and that uses the second argument consistently.

PR #3791, opened 2021-09-29
Allow integrations to indicate whether text can be annotated

### LINKED ISSUE #3790: Adder toolbar is shown when text is selected in PDF viewer UI
In PDFs only the document content should be annotatable. However the Annotate / Highlight toolbar is shown when selecting text in parts of the PDF.js UI such as the document metadata dialog, accessible from the chevron menu in the top-right corner of the toolbar. Clicking the Annotate button in this case results in a console error (`Unhandled Promise Rejection: Error: Selection is outside page text`) and no annotation appearing in the sidebar.

PDF viewer UI

In ebook readers we have a related use case where the reader consists of a container frame that handles navigating between chapters, and a child content frame that actually contains the book content. The annotation toolbar should never be shown in the content frame.

I think it would be useful if integrations could indicate whether specific content can be annotated or not.

robertknight:
Add a `canAnnotate` method to the `Integration` interface to indicate
whether a range is part of the annotatable content of the document. This
allows integrations to signal whether the Annotate/Highlight toolbar
should be shown for a given selection.

Implement it for PDFs by returning `true` if the range is part of the
text layer of a single page. Implement it for HTML documents by always
returning true.

This fixes an issue where the user would see the Annotate/Highlight toolbar if they selected text in the PDF.js UI such as in the document metadata dialog or table of contents.

In ebooks (eg. VitalSource, but also Readium and Epub.js) which consist of a "container" frame where the sidebar is loaded and a child "content" frame where the current book content is displayed, this mechanism could be used as a simple way to disable annotation within the container frame, by making `canAnnotate` always return `false`. There are other approaches we could take in this context (such as adding support for "host-only" frames that do not have a `Guest` instance). I'm still evaluating the options, but since this one is useful for single-frame apps (eg. PDF.js) I've implemented this one already.

Fixes #3790

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -570,10 +570,9 @@ export async function anchor(root, selectors) {
 /**
  * Prepare a DOM range for generating selectors and find the containing text layer.
  *
- * Throws if the range cannot be annotated.
- *
  * @param {Range} range
  * @return {[Range, Element]}
+ * @throws If the range cannot be annotated
  */
 function getTextLayerForRange(range) {
   // "Shrink" the range so that the start and endpoints are at offsets within


src/annotator/anchoring/pdf.js
```diff
@@ -568,17 +568,14 @@ export async function anchor(root, selectors) {
 }
 
 /**
- * Convert a DOM Range object into a set of selectors.
+ * Prepare a DOM range for generating selectors and find the containing text layer.
```
robertknight, line 571: I haven't written too much in the description about the return value, because this is a private helper in this module and if I did I'd largely be restating comments from the implementation.

src/annotator/anchoring/pdf.js
```diff
@@ -568,17 +568,14 @@ export async function anchor(root, selectors) {
...
  *
- * Converts a DOM `Range` object into a `[position, quote]` tuple of selectors
- * which can be saved with an annotation and later passed to `anchor` to
- * convert the selectors back to a `Range`.
+ * Throws if the range cannot be annotated.
```
esanzgar, line 573: I personally feel it is more obvious to annotate the function than to parse the comments.

```suggestion
 * @throws {Error} if the range cannot be annotated.
```

esanzgar (approved):
It works well, thanks!

src/annotator/anchoring/pdf.js
robertknight: Done. I think we can probably omit the `Error` type and just assume it if there is no custom error. One caveat also with types is that TypeScript doesn't do anything with them at present. It cannot for example verify that a function really throws what it claims to.

PR #3792, opened 2021-09-29
Remove "tall iframe" example from dev server

robertknight:
This was created to simulate a scenario in VitalSource's book reader,
but they changed their reader so that it no longer uses a tall frame.


lyzadanger (approved):
👍🏻 Thanks for cleaning up

PR #3796, opened 2021-09-30
Add mock VitalSource book reader to dev server

robertknight:
This adds a mock version of the VitalSource Bookshelf reader to the dev
server. It includes only the essential aspects of the viewer's structure that
will trigger and exercise VitalSource-specific functionality:

- A container frame. Annotation will be disabled in this frame once the integration is complete.
- A `<mosaic-book>` element with a shadow root
- The chapter content iframe, inside the shadow root

The structure is broadly similar to other ebook readers (eg. Readium, Epub.js), so we might be able to treat it as a proxy for those as well.

The book chapters used in the sample were extracted from a [project Gutenberg EPUB](https://www.gutenberg.org/ebooks/514).

----

**Testing:**

Start dev server and follow the new link in the homepage or go to http://localhost:3000/document/vitalsource.


dev-server/documents/html/vitalsource.mustache
```diff
@@ -0,0 +1,98 @@
...
+    <title>VitalSource book reader</title>
+    <style>
+      body {
+        font-family: sans-serif;
+      }
+    </style>
+    <script type="module">
+      /**
+       * Custom element that holds the book content frame. The class name is not
+       * important but the `<mosaic-book>` tag name triggers the VitalSource integration.
```
robertknight, line 15: Note there currently is no VitalSource integration, but it is coming soon.

dev-server/documents/html/vitalsource.mustache
```diff
@@ -0,0 +1,98 @@
...
+    <h1>Mock VitalSource Bookshelf reader</h1>
+    <p>
+      This page simulates the aspects of the <a href="https://bookshelf.vitalsource.com">VitalSource
+      Bookshelf</a> ebook reader which are relevant to Hypothesis. Other ebook readers
+      (eg. Readium and Epub.js) have a similar frame structure.
+    </p>
+
+    <p>
+      There is a container frame where the Hypothesis sidebar should be loaded and a child
+      content frame containing the current chapter content. The container frame should not be annotatable.
```
robertknight, line 85: The container frame _is_ currenly annotatable. This will change once the VS integration is implemented.

lyzadanger (approved):
Let's get this thing going...

dev-server/documents/html/vitalsource.mustache
robertknight: This implementation mimicks the essential elements of the VS UI that affect how the Hypothesis integration works. This is all written from scratch, nothing is taken from elsewhere except the name of the custom element.

PR #3797, opened 2021-10-01
Unify all `delay` functions into a common test utility

esanzgar:
Although the `delay` utility function is a small function that can
almost be in-lined, I decided to create a common `delay` function in
`wait.js` and replace all usage in tests by the new function.


robertknight (approved):
👍

PR #3799, opened 2021-10-04
Convert CommonJS requires in tests to ES imports

robertknight:
In preparation for modernizing the client's module bundling, convert remaining CommonJS `require` statements in tests to ES imports.

I also replaced a reference to the Node-specific `global` variable with ES-standard `globalThis` as a way to access the global object.

### DIFF
--- a/src/annotator/anchoring/test/html-baselines/index.js
+++ b/src/annotator/anchoring/test/html-baselines/index.js
@@ -21,15 +21,21 @@
 //     them as `<fixture name>.json` in this directory
 //  4. Add an entry to the fixture list below.
 
+import minimalDoc from './minimal.html';
+import minimalJSON from './minimal.json';
+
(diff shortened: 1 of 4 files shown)


esanzgar (approved):
I didn't know about `globalThis`. Thanks for that.

> The globalThis property provides a standard way of accessing the global this value (and hence the global object itself) across environments. Unlike similar properties such as window and self, it's guaranteed to work in window and non-window contexts.

When do you think we should use this? For example, `console.warn(...)` ,  or `globalThis.console.warn(...)`.

robertknight:
> When do you think we should use this? For example, console.warn(...) , or globalThis.console.warn(...).

The only real critical case I can think of is where we have some code that would normally access `window` but needs to work in both Window and non-Window contexts. Looking at this again I guess I could have changed the code to reference `window` instead of `global` and it would have solved the problem just as well.

PR #3806, opened 2021-10-05
Update supported browsers list

esanzgar:
All frontend Hypothesis app are going to have the same base line as
this:

https://github.com/hypothesis/bouncer/pull/469/commits/6a01199bfc196a8932be271f61c07029dbf886c5

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "watchify": "^4.0.0",
     "wrap-text": "^1.0.7"
   },
-  "browserslist": "chrome 57, edge 17, firefox 53, safari 10.1",
+  "browserslist": "chrome 70, firefox 70, safari 11.1",
   "browserify": {
     "transform": [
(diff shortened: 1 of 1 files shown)


esanzgar:
@robertknight, the `boot.js` bundle is transpiled using other browser list, is that still required/desired?

robertknight (approved):

robertknight:
> @robertknight, the boot.js bundle is transpiled using an alternative browser list, is that still required/desired?

The `boot.js` script is designed to run in older browsers so that it can notify the user that the current browser is unsupported. Currently it does this quietly via a console message. In the lms app we have something similar that uses a more visible message.

We might want to adjust the boot script's browser baseline at some point. This would enable us to drop the `babel-plugin-transform-async-to-promises` dependency for example. We should do that separately from this PR though.

PR #3807, opened 2021-10-06
Use separate channels for guest / host messages from sidebar

robertknight:
Use separate channels for sending messages to guests vs the host in the
sidebar, even for the common case when there is only one guest and it is
the same frame as the host.

This change makes it clear for readers which part of the annotator code
is intended to handle a particular message from the sidebar. It is also
a step towards supporting host frames that are not guests. This will be
needed in ebook readers where the host frame provides the navigation UI
and contains the frame displaying the book content, but should not be
annotatable itself.

 - Remove the `bridge` service in the sidebar. The `frameSync` service
   now provides the entry point for other services/components to make RPC calls
   to the host or guest frames. Currently the only use case is
   sending notifications to the host via `FrameSyncService.notifyHost`.

 - Create separate `Bridge` instances in `FrameSyncService` for sidebar
   <-> guest and sidebar <-> host communication. The sidebar <-> guest
   bridge works the same as before. The sidebar <-> host bridge
   is established by having `FrameSyncService` create a MessageChannel
   when sending the `hypothesisSidebarReady` notification to the host.

   The sidebar / host frames then add respective ports of this channel
   to a Bridge instance.

 - Change the various existing RPC calls between frames to use either
   the guest <-> sidebar or host <-> sidebar communication channels as
   appropriate


esanzgar:
> I thought we had already settled on "guest frame" or "guest" for short, mainly because this means it contains an instance of the Guest class and it contrasts with "host frame" or "host" for short.

I have no problem referring to annotatable frames as `guest` frames. The `Guest` class is instantiated in both the `host` and annotatable frames, that's probably why I was thinking on `guest` frames as a superset that refers to both the `host` frame or `annotatable` frame/s. That was my interpretation of this sentence from the PR description: 

> a step towards supporting host frames that are not guests

PR #3810, opened 2021-10-06
Replace Browserify with Rollup

robertknight:
Replace Browserify with Rollup as a module bundler. This brings several
advantages:

- It is a more modern bundler that has better support in the ecosystem,
  more active maintenence, fewer dependencies and is easier to write plugins for.
  The core team also maintain most of the plugins that we need.

- It has native support for modern platform features like ES modules,
  including dynamic import for lazy-loading of functionality.

- It generates smaller bundles due to its ability to do tree shaking and scope hoisting.

The new bundles are generated as ES modules rather than UMD / IIFE bundles.
Our target browsers now all support ES modules and this will enable us to use
native static/dynamic imports for code splitting or lazy-loading of
functionality in future.

The initial Rollup build generates one bundle per application (boot script,
annotator, sidebar). This simplifies the build process and minimizes production
bundle size, at the cost of making incremental updates during development
slightly slower.

Build performance is currently similar to or a little slower than Browserify.
We can optimize this by adding caching of transforms (mainly Babel) and
pre-building of a vendor bundle containing large dependencies.

As part of changing the bundler this also changes how the code is packaged
for tests.  We now do the bundling ourselves rather than using a Karma
integration. This makes it much easier to inspect the processed code and fix
problems. It also makes it easier to optimize the bundle building in future.


robertknight:
Prod bundle size before:

```
[I]  ~/h/client > exa --long --no-permissions --no-user --no-time build/scripts/ | grep '\.js$'
153k annotator.bundle.js
5.8k boot.bundle.js
260k katex.bundle.js
113k sentry.bundle.js
 76k showdown.bundle.js
340k sidebar.bundle.js
225k ui-playground.bundle.js
```

Where the sidebar app consists of the sidebar, sentry, katex and showdown bundles (340 + 113 + 260 + 76 = 789k)

Prod bundle size after:

```
[I]  ~/h/client > exa --long --no-permissions --no-user --no-time build/scripts/ | grep '\.js$'
114k annotator.bundle.js (-25%)
4.3k boot.bundle.js
651k sidebar.bundle.js (-17% of total sidebar app size)
221k ui-playground.bundle.js
```

Where the sidebar app is contained in only one bundle.

robertknight:
I did some experiments with adding caching at various points in the build process. What I learned is that there are worthwhile wins to be had from minimizing or caching:

1. Babel transformation by @rollup/plugin-babel
2. The JS parsing that Rollup does using [Acorn](https://github.com/acornjs/acorn)
3. The work that @rollup/plugin-commonjs does to support non-ES module dependencies

Of the above, the Babel transform looks like it will be simplest to do. Not sure about the others yet.

Also we could look at what we did with Browserify where large vendor dependencies were bundled separately in development. In production building vendor dependenceis + app code together maximizes the effectiveness of tree-shaking.

I will look at these after the initial conversion lands.

robertknight:
> What's the failure mode going forward for unsupported browsers?

That's the same as before. The only code that runs in unsupported browsers is the boot script, and that's compiled to a classic (non ES-module) script.

robertknight:
> One thing that would be helpful is if you could summarize all the need-to-know stuff for devs (e.g. the slow test, the remaining warning) so we have a quick reference of what to expect.

Sure. The slow test was fixed, so I think the most relevant remaining things are:

1. You will see a warning (`[DEP0148] DeprecationWarning: ...`) when running `make dev` in the most recent version of Node. I'm trying to talk with upstream to get this resolved.
2. When running `make dev` you will see `JS build starting...` and `JS build completed` messages in the log output. The "JS build completed" message is the signal that the client is built and ready to use.
3. You'll see the same messages as (2) whenever you make changes.
4. Running tests works the same as before, but the sequence of messages is a bit different. You'll see `Building test bundle... ($N files)` at the start, where `$N` is either the total number of files or the number matching the `--grep` pattern, if specified.

PR #3817, opened 2021-10-08
Change default export to name export

esanzgar:
As per team conventions, we only use default exports for Preact
components.


esanzgar:
I honestly don't see a value on this now that we have types.

esanzgar:
Let's try to replace this with types.

PR #3829, opened 2021-10-11
Types for bridge events

esanzgar:
I have created type definitions for all the event names that are
sent across the different frames using various `Bridge`s. It is based on
the previous `bridge-events.js`. I broke down the events in four
sections based on the direction of the messages:

* host -> sidebar events
* sidebar -> host events
* sidebar -> guest/s events
* guest -> sidebar events

For those events that didn't have a description I added one.

This is more stringent and less verbose than the previous lookup system.

### DIFF
--- a/src/types/bridge-events.d.ts
+++ b/src/types/bridge-events.d.ts
@@ -6,145 +6,141 @@
 /**
  * Events that the sidebar sends to the host
  */
-type SidebarToHostEvent = {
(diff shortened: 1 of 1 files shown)


esanzgar:
We have decided that event thought JavaScript enum are a little bit more verbose, they provide a level of indirection that is useful for documentation.

I will close this and replace it for another PR.

PR #3833, opened 2021-10-14
Re-enforce the use of enums for `Bridge` events

esanzgar:
In contrast to https://github.com/hypothesis/client/pull/3829, we have
decided to re-enforce the use of the lookup enums for the event names
used by `Bridge`.

I have added definitions for all the events that are sent across the
different frames using various `Bridge`s. I broke down the events into
four sections based on the direction of the messages:

* guest -> sidebar
* host -> sidebar
* sidebar -> guest/s
* sidebar -> host


esanzgar:
We close in favour of https://github.com/hypothesis/client/pull/3838

PR #3837, opened 2021-10-15
Re-fetch access token if needed when reconnecting WebSocket

robertknight:
The `Socket` WebSocket wrapper that handles automatically reconnecting after
an unexpected disconnection was configured with a static URL. Since the
access token for WS connections is part of the URL (in the
`access_token` query param) this meant that the access token would not
be re-fetched if it had expired. As a result, if the original token had
expired and then a reconnection happened, the user would be connected to
the WebSocket as an anonymous user, not the logged-in identity shown in
the UI. I noticed the issue due to client warnings about a mismatch between the WebSocket's user ID and the identity shown in the UI [^1].

This commit modifies the WebSocket wrapper so that it can fetch a new
URL on each connection attempt, and then modifies the client's real-time
service to use this capability so that each connection attempt always
includes a current access token, assuming the client is able to fetch
one.

 - Change `Socket` to accept a callback that returns the URL
   to connect to, rather than a fixed URL. This callback may be sync or
   async and is invoked before each connection attempt.

   This change resulted in significant changes to the tests because they
   previously relied on the initial connection occuring synchronously
   in the tests. That's no longer possible as the URL callback is async.

   Additionally it resulted in more straightforward code in `Socket` to
   not rely on the `retry` library for reconnection but just manually
   track the number of connection attempts and calculate a backoff based
   on that before reconnecting.

 - Move the current logic in `StreamerService` that generates a
   WebSocket URL, with authentication, into a callback that is passed to
   the `Socket` constructor
   
This issue fixed here may have been one cause of the problems described in https://github.com/hypothesis/private-issues/issues/79.

[^1]:  I noticed this happen after my laptop had been put to sleep and then woken up many hours later. Based on h's log output I think this event triggered a timeout that caused one of the WebSocket workers to shut down, disconnecting the client. Following this I observed that the client reconnected to the WebSocket without refetching an access token, although I would have expected any access token to have expired by then (since access tokens are only valid for one hour). Indeed a client action which triggered an API call did lead to a token re-fetch.

### DIFF
--- a/src/sidebar/test/websocket-test.js
+++ b/src/sidebar/test/websocket-test.js
@@ -252,4 +252,15 @@ describe('Socket (WebSocket wrapper)', () => {
       assert.equal(fakeSocket, initialSocket);
     });
   });
+
+  describe('#isConnected', () => {
+    it('returns true if WebSocket is connected', async () => {
+      const socket = new Socket(() => 'ws://test:1234');
+      assert.isFalse(socket.isConnected());
+
+      await awaitEvent(socket, 'open');
+
+      assert.isTrue(socket.isConnected());
+    });
+  });
 });


robertknight:
Something that is currently unresolved: Should the client attempt to automatically reconnect if the token fetch fails, eg due to a network error when hitting the `POST /api/token` endpoint? The current behavior on master and in this PR is that it doesn't. This means this kind of transient error is treated differently from transient errors which cause WebSocket disconnection.

PR #3839, opened 2021-10-15
Use `notifyHost` instead of `_hostRPC.call`

esanzgar:
It makes sense to reuse `notifyHost` even internally, for consistency.


esanzgar:
I agree there is not a strong reason, except to give a bit more consistency. I am ambivalent too.

PR #3858, opened 2021-10-20
PoC - Replace the central event bus in the annotator

robertknight:
**nb. This branch is not intended for merging.** I've opened this to discuss some changes I'm thinking of in the annotator to pave the way for supporting multiple guests (and in some temporary situations, no guests) in the sidebar controls and bucket bar. 

---

Change the way that major entities in the annotator notify other entities when
events happen. Instead of having a single shared event bus, make the individual
entities (`Guest`, `AnnotationSync`, `Notebook`) their own event sources.

This eliminates some indirection and makes the flow of events easier to follow.
For example `Guest` creates the `AnnotationSync` instance and can subscribe directly
to it rather than needing to communicate via a shared event bus.

The current motivation though is to allow the sidebar and bucket bar to know which
Guest events such as `anchorsChanged` and `hasSelectionChanged` are coming
from and also subscribe and unsubscribe from these events from specific guests.

The bucket bar and sidebar can subscribe to events from same-origin guests
directly. If we add support for cross-origin guests in future, we could
potentially handle this by creating proxies in the host frame.

Sub-tasks:

 - Change `AnnotationSync` to become an event source that emits
   `annotationDeleted` and `annotationsLoaded` events which `Guest` then
   listens for.

   To send new annotations from the `Guest` to the sidebar, instead of having
   `Guest` publish a `beforeCreateAnnotation` event on the event bus which
   `AnnotationSync` responds to, add a direct `createAnnotation` method to
   `AnnotationSync` which `Guest` then calls.

 - Change `Guest` to become an event source that emits `hasSelectionChanged` and
   `anchorsChanged` events which the `Sidebar` and `BucketBar` listen to.

   As part of this, move the `anchorsChanged` event handler from `Sidebar` to
   `BucketBar` since the `BucketBar` already has a reference to the `Guest`

 - Change the way that `Sidebar` opens the notebook and responds to it being
   closed.

   Instead of publishing an `openNotebook` event on the event bus to
   open the notebook, pass a reference to the Notebook directly to the
   Sidebar, which then calls it's `open` method when an `openNotebook` RPC call
   is received from the sidebar app.

   To handle closing of the notebook, instead of emitting a `closeNotebook` event
   on the event bus, make the `Notebook` an event source which emits a `closed`
   event, which is then handled by `Sidebar`.

 - Remove the `EventBus` class in `src/annotator/util/emitter.js` as it is now unused


robertknight:
The changes in this PR have been made obsolete by ongoing work related to inter-frame communication in other PRs.

Issue #3695, opened 2021-08-19
Enable and deploy the new inter-frame communication system


esanzgar:
After the implementation of core functionality for the inter-frame communication (#3627), enable and deploy it.

PR #3881, opened 2021-10-29
Implement `PortProvider` and `PortFinder`

esanzgar:
The combination of these two classes enable the discovery and
establishment of `MessageChannel` communication between pairs of frames.

It works with cross-origin frames.

It is the last step of this epic: https://github.com/hypothesis/client/issues/3533

Closes: https://github.com/hypothesis/client/issues/3695


robertknight:
> It is the last step of this epic: #3533
> Closes: #3695

This PR implements parts of a new inter-frame communication system but it doesn't make setup of the existing channels (guest <-> sidebar or host <-> sidebar) use them. While this does check off a box in https://github.com/hypothesis/client/issues/3533, I don't think we can reasonably close the logical "implement new inter-frame communication system" epic until that is done.

src/shared/port-provider.js
```diff
@@ -0,0 +1,269 @@
...
+ * 5. send reciprocal port to the `sidebar` frame using the
+ *    `host-sidebar(MessageChannel).sidebar(MessagePort)`
+ *
+ *
+ * In some situations, because `guest` iframe/s load in parallel to the `host`
+ * frame, we can not assume that the code in the `host` frame is executed before
+ * the code in a `guest` frame. Hence, we can't assume that `PortProvider` (in
+ * the `host` frame) is initialized before `PortFinder` (in the `guest` frame).
+ * Therefore, for the `PortFinder`, we implement a polling strategy (sending a
+ * message every N milliseconds) until a response is received.
```
robertknight, line 64: The fact that polling is used could be an implementation detail of `PortFinder`, except for the fact that `PortProvider` needs to allow for potentially receiving duplicate requests for a guest port.

src/shared/port-provider.js
```diff
@@ -0,0 +1,269 @@
...
+  /**
+   * Returns a port from a channel. Currently, only returns the `host` port from
+   * the `host-Sidebar` channel. Otherwise, it returns `null`.
+   *
+   * @param {object} options
+   *   @param {'host-sidebar'} options.channel
+   *   @param {'host'} options.port
+   */
+  getPort({ channel, port }) {
+    if (channel === 'host-sidebar' && port === 'host') {
```
robertknight, line 116: Presumably a hypothetical `host-notebook` channel is another possible future use case for this method.

robertknight (commented):
I've had an initial look through this and left some comments. So far I've focused only on the implementation and not the tests. The main high level comment I have is that I'd like to see a draft of the follow-up PR that actually replaces the existing channel setup mechanisms (eg. for guest <-> sidebar and host <-> sidebar) with the ones added here.

src/shared/port-finder.js
```diff
@@ -0,0 +1,145 @@
...
+   *   @param {Message} options.message
+   *   @param {(port: MessagePort) => void} options.resolve
+   *   @param {number} options.timeoutId
+   *   @param {number} [options.intervalId]
+   */
+  _handlePortOffer(
+    { data, ports },
+    { message, resolve, timeoutId, intervalId }
+  ) {
+    if (isMessageEqual(data, message)) {
```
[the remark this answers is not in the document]
esanzgar: I am happy to remove the `_handlePortOffer method`.

src/shared/port-finder.js
```diff
@@ -0,0 +1,145 @@
...
+      clearInterval(intervalId);
+      clearTimeout(timeoutId);
+      resolve(ports[0]);
+    }
+  }
+
+  destroy() {
+    this._listeners.removeAll();
```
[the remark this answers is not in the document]
esanzgar: Correct. One of the tests exemplify the timeout that is caused when `PorFinder#destroy()` is invoked.

src/shared/port-provider.js
```diff
@@ -0,0 +1,269 @@
...
+ * -------------------------------------------------------------------|-------------------------------------------------------------
+ *
+ * 2. listens to requests of `MessagePort` <---------------------------- 1. request a `MessagePort` using `window.parent.postMessage`
+ *                 |
+ *                 V
+ * 3. sends offers of `MessagePort` using `event.source.postMessage` ---> 4. listens to offers of `MessagePort`
+ *                 |
+ *                 V
+ * 5. send reciprocal port to the `sidebar` frame using the
+ *    `host-sidebar(MessageChannel).sidebar(MessagePort)`
```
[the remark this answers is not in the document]
esanzgar: I can add that comment.

src/shared/port-provider.js
```diff
@@ -0,0 +1,269 @@
...
+ *
+ * Channel nomenclature is `[frame1]-[frame2]` so that:
+ *   - `port1` should be owned by/transferred to `frame1`, and
+ *   - `port2` should be owned by/transferred to `frame2`
+ *
+ * @implements Destroyable
+ */
+export class PortProvider {
+  /**
+   * @param {string} hypothesisAppsURL
```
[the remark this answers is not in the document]
esanzgar: You are right, it needs a little bit more explanation.

src/shared/port-provider.js
```diff
@@ -0,0 +1,269 @@
...
+ * Therefore, for the `PortFinder`, we implement a polling strategy (sending a
+ * message every N milliseconds) until a response is received.
+ *
+ * Channel nomenclature is `[frame1]-[frame2]` so that:
+ *   - `port1` should be owned by/transferred to `frame1`, and
+ *   - `port2` should be owned by/transferred to `frame2`
+ *
+ * @implements Destroyable
+ */
+export class PortProvider {
```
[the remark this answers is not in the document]
esanzgar: Yes, you understood the workflow correctly. It is exemplified in: https://github.com/hypothesis/client/pull/3903

src/shared/port-provider.js
```diff
@@ -0,0 +1,269 @@
...
+   */
+  _sendPort(event, { allowedMessage, port, reciprocalPort }) {
+    const message = { ...allowedMessage, type: 'offer' };
+
+    const source = /** @type {Window} */ (event.source);
+
+    source.postMessage(message, event.origin, [port]);
+
+    if (reciprocalPort) {
+      if (['notebook-sidebar', 'guest-sidebar'].includes(message.channel)) {
```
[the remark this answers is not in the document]
esanzgar: Currently, to open the notebook, the sidebar sends a RPC message to the host. The host uses the emitter to communicate to the `Notebook` class, which causes the iframe containing the notebook app to display.

The notebook is currently opened via the sidebar (could change in the the future), so it makes sense to me that the notebook initiates the request. 

Adding a direct notebook-sidebar channel will enable to change groups (in either apps) without reloading their iframes, which is a slow process.

src/shared/port-finder.js
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * @typedef RequestPortOptions
+   * @prop {Channel} channel - requested channel
+   * @prop {Window} hostFrame - the frame where the hypothesis client is loaded.
+   *   It is used to send a `window.postMessage`.
+   * @prop {Port} port - requested port
+   * @prop {(reason: Error) => void} reject - execute the `Promise.reject` in case
+   *   the `host` frame takes too long to answer the request.
+   * @prop {(port: MessagePort) => void} resolve - execute the `Promise.resolve`
+   *   when `host` frame successfully answers the request.
```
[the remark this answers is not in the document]
esanzgar: I think we can eliminate the `_requestPort` method, similarly to what you suggested for PortProvider.

src/shared/port-provider.js
esanzgar: Are you suggesting to remove this comment from the documentation?

src/shared/port-provider.js
esanzgar: Yes, you are correct.

robertknight:
I see a typecheck error in CI related to `setTimeout`/`setInterval`. There is a proposed fix for that in https://github.com/hypothesis/client/pull/3913.

src/shared/port-finder.js
```diff
@@ -0,0 +1,145 @@
...
+
+    // The `host` frame maybe busy, that's why we should wait.
+    const timeoutId = window.setTimeout(() => {
+      clearInterval(intervalId);
+      reject(
+        new Error(`Unable to find '${port}' port on '${channel}' channel`)
+      );
+    }, MAX_WAIT_FOR_PORT);
+
+    // TODO: It would be nice to remove the listener after receiving the port.
```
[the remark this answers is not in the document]
esanzgar: Sometimes, I wished `ListenerCollection` would have a method to remove individual events, so I decided to solved in this way: https://github.com/hypothesis/client/pull/3914

esanzgar (commented):
I have acted on your input. I have streamlined both `PortProvider` and `PortFinder`. Please, let me know what you think.

src/shared/port-util.js
```diff
@@ -0,0 +1,65 @@
...
+  if (!isMessageValid(data)) {
+    return false;
+  }
+
+  try {
+    return (
+      JSON.stringify(data, Object.keys(data).sort()) ===
+      JSON.stringify(message, Object.keys(message).sort())
+    );
+  } catch (error) {
```
robertknight, line 62: If the error is unused it can be elided in modern JS (ie. `try { ... } catch { ... }`).

src/shared/port-provider.js
```diff
@@ -0,0 +1,269 @@
...
+    return true;
+  }
+
+  /**
+   * Send a message and a port to the corresponding destinations.
+   *
+   * @param {MessageEvent} event
+   * @param {Message} message - the message to be sent.
+   * @param {MessagePort} port - the port to be sent via `window#postMessage`
+   *   (the origin is set to match the MessageEvent's origin)frame that sends the initial request th
```
robertknight, line 145: This line is an implementation detail, I'd suggest to remove it.

src/shared/port-provider.js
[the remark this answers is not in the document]
robertknight: When the notebook is opened, my understanding is that two things would happen:

1. Sidebar sends request to host to open notebook
2. Sidebar sends request to notebook to sync groups or other view state

Step 1 will use the sidebar <-> host channel, but step 2 needs the sidebar <-> notebook channel. My thinking when I read this initially was that because the sidebar initiates step 2, it would be the one requesting creation of the sidebar <-> notebook channel. When the notebook opens, it would then request its end of the channel and process any messages that happen to already have been sent on the channel.

Does this match your thinking or were you thinking of a different flow?

robertknight (changes requested):
I've had a look through all of this now. I added lots of small comments, mostly about documentation, which should be easy to address, as well as a few questions. I did have some feedback about the `discover` API but I think we probably want to get this version landed first.

src/shared/test/port-provider-test.js
```diff
@@ -0,0 +1,260 @@
...
+
+      assert.calledOnceWithExactly(
+        window.postMessage,
+        { ...data, type: 'offer' },
+        window.location.origin,
+        [sinon.match.instanceOf(MessagePort)]
+      );
+    });
+
+    it('sends the reciprocal port of the `guest-sidebar` channel (via the sidebar port)', async () => {
```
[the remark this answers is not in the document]
esanzgar: I have changed the language so it doesn't refers to any specific channel, but to the to general cases you described. Let me know if that's enough.

esanzgar (commented):
Robert, I have incorporated your suggestions in the last two commits. Could you take a look at it? Please, let me know if anything needs to be corrected.

src/shared/test/port-provider-test.js
```diff
@@ -0,0 +1,285 @@
...
+      const hostPort = portProvider.getPort({
+        channel: 'host-sidebar',
+        port: 'host',
+      });
+      assert.exists(hostPort);
+    });
+  });
+
+  describe('#listen', () => {
+    it('ignores all port requests before `listen` is called', async () => {
```
robertknight, line 82: I was looking for a `listen` call in this test, but then realized it was in the setup code. I think it might make sense to call `listen` explicitly in each test, even though it will lead to more lines of code in total.

src/shared/test/port-provider-test.js
```diff
@@ -0,0 +1,285 @@
...
+      });
+      await delay(0);
+
+      assert.notCalled(window.postMessage);
+    });
+
+    [
+      // Disabled this check because it make axes-core to crash
+      // Reported: https://github.com/dequelabs/axe-core/pull/3249
+      //{ data: null, reason: 'if message is null' },
```
robertknight, line 134: I see that your fix was approved upstream :)

src/shared/test/port-provider-test.js
```diff
@@ -0,0 +1,285 @@
...
+      const data = {
+        channel: 'host-sidebar',
+        port: 'sidebar',
+        source,
+        type: 'request',
+      };
+      portFinderRequest({
+        data,
+      });
+      await delay(0);
```
robertknight, line 94: Since `portFinderRequest` is followed by an `await delay(0)` in almost all cases, it might make sense to make the delay part of that function.

src/shared/test/port-provider-test.js
```diff
@@ -0,0 +1,285 @@
...
+
+      assert.calledWith(
+        window.postMessage,
+        { ...data, type: 'offer' },
+        window.location.origin,
+        [sinon.match.instanceOf(MessagePort)]
+      );
+    });
+
+    it('responds to the first valid port request, ignore additional requests', async () => {
```
robertknight, line 211:
```suggestion
    it('responds to the first valid port request but ignores additional requests', async () => {
```

src/shared/test/port-util-test.js
```diff
@@ -0,0 +1,140 @@
...
+          extra: 'dummy', // additional
+        },
+        message: {
+          channel: 'host-sidebar',
+          port: 'guest',
+          source,
+          type: 'offer',
+        },
+        expectedResult: false,
+        reason: 'data has one additional property',
```
robertknight, line 98: I wonder whether we might ever see a situation where the browser or third-party code modifies messages and adds extra properties. I guess we'll find out when we deploy it.

robertknight (approved):
Good work 👍 - thanks for your patience with the review.

PR #3895, opened 2021-11-01
Fix KaTeX import

robertknight:
Per the release notes in https://github.com/hypothesis/client/pull/3887 (see notes for v0.14.0), KaTeX should be imported using a default import rather than namespace import.

This fixes an error I observed during the build, along with a failure to render math equations in annotations:

```
[12:19:55] Rollup warning: src/sidebar/render-markdown.js (121:29) 'renderToString' is not exported by 'node_modules/katex/dist/katex.mjs' (https://rollupjs.org/guide/en/#error-name-is-not-
exported-by-module)
```


lyzadanger (approved):
👍🏻

PR #3899, opened 2021-11-03
Remove elements created during test

esanzgar:
After each test is run DOM elements should be removed.

### DIFF
--- a/src/annotator/integrations/test/pdf-test.js
+++ b/src/annotator/integrations/test/pdf-test.js
@@ -104,6 +104,7 @@ describe('annotator/integrations/pdf', () => {
       pdfIntegration?.destroy();
       delete window.PDFViewerApplication;
       outerContainer.remove();
+      document.querySelector('hypothesis-banner')?.remove(); // <hypothesis-banner> elements are created outside the outerContainer
       $imports.$restore();
     });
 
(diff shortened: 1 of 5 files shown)


src/annotator/integrations/test/pdf-test.js
```diff
@@ -104,6 +104,7 @@ describe('annotator/integrations/pdf', () => {
       pdfIntegration?.destroy();
       delete window.PDFViewerApplication;
       outerContainer.remove();
+      document.querySelector('hypothesis-banner')?.remove(); // <hypothesis-banner> elements are created outside the outerContainer
```
robertknight, line 107: Since the `PDFIntegration` instance creates the banner, it should probably destroy it too in the `destroy` method.

src/annotator/integrations/test/pdf-test.js
```diff
@@ -288,11 +289,16 @@ describe('annotator/integrations/pdf', () => {
...
-        assert.ok(anchor.range);
-        assert.notCalled(fakeAnnotator.anchor);
+        try {
+          await triggerUpdate();
+
+          assert.equal(anchor.highlights.length, 1);
+          assert.ok(anchor.range);
+          assert.notCalled(fakeAnnotator.anchor);
+        } finally {
+          anchor.highlights[0].remove();
```
robertknight, line 300: I think this is OK for the moment, although we might want to do this as part of `PDFIntegration.destroy` as well in future.

robertknight (changes requested):
I suggest we might want to remove the `<hypothesis-banner>` element in `PDFIntegration.destroy`, which the tests already call. We _might_ want to do the same for highlights in future, but this requires more thought so I'd keep the change to remove highlights manually as you are doing here for the moment.

src/annotator/integrations/test/pdf-test.js
esanzgar: That's a good suggestion, I have added a second commit for that.

robertknight (approved):

PR #3900, opened 2021-11-03
Restore `sinon.useFakeTimers()`

esanzgar:
`sinon.useFakeTimers` should be restored, otherwise it has the potential to affect other tests.

### DIFF
--- a/src/sidebar/components/test/Menu-test.js
+++ b/src/sidebar/components/test/Menu-test.js
@@ -183,15 +183,19 @@ describe('Menu', () => {
       shouldClose ? 'closes' : "doesn't close"
     } when user performs a "${eventType}" (key: "${key}") on menu content`, () => {
       clock = sinon.useFakeTimers();
-      const wrapper = createMenu({ defaultOpen: true });
-      wrapper.find('.Menu__content').simulate(eventType, { key });
-      // The close event is delayed by a minimal amount of time in
-      // order to allow links to say in the DOM long enough to be
-      // followed on a click. Therefore, this test must simulate
-      // time passing in order for the menu to close.
-      clock.tick(1);
-      wrapper.update();
-      assert.equal(isOpen(wrapper), !shouldClose);
+      try {
+        const wrapper = createMenu({ defaultOpen: true });
+        wrapper.find('.Menu__content').simulate(eventType, { key });
+        // The close event is delayed by a minimal amount of time in
+        // order to allow links to say in the DOM long enough to be
+        // followed on a click. Therefore, this test must simulate
+        // time passing in order for the menu to close.
+        clock.tick(1);
+        wrapper.update();
+        assert.equal(isOpen(wrapper), !shouldClose);
+      } finally {
+        clock.restore();
+      }
     });
   });
 
--- a/src/sidebar/components/test/Menu-test.js
+++ b/src/sidebar/components/test/Menu-test.js
@@ -9,7 +9,6 @@ import { checkAccessibility } from '../../../test-util/accessibility';
 
 describe('Menu', () => {
   let container;
-  let clock;
 
   const TestLabel = () => 'Test label';
   const TestMenuItem = () => 'Test item';
@@ -39,10 +38,6 @@ describe('Menu', () => {
   afterEach(() => {
     $imports.$restore();
     container.remove();
-    if (clock) {
-      clock.restore();
-      clock = null;
-    }
   });
 
   it('opens and closes when the toggle button is clicked', () => {
@@ -182,7 +177,7 @@ describe('Menu', () => {
     it(`${
       shouldClose ? 'closes' : "doesn't close"
     } when user performs a "${eventType}" (key: "${key}") on menu content`, () => {
-      clock = sinon.useFakeTimers();
+      const clock = sinon.useFakeTimers();
       try {
         const wrapper = createMenu({ defaultOpen: true });
         wrapper.find('.Menu__content').simulate(eventType, { key });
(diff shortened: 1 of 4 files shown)


src/sidebar/services/test/session-test.js
```diff
@@ -172,30 +172,26 @@ describe('SessionService', () => {
...
         const session = createService();
-        return session
-          .load()
-          .then(() => {
-            return session.load();
-          })
-          .then(() => {
-            assert.calledOnce(fakeApi.profile.read);
-          });
+        session.load();
```
esanzgar, line 175: This change is unrelated, but I think it is simpler. Two consecutive calls to `session.load` doesn't invoke the reading of the profile because it is cached for X amount of time.

src/sidebar/components/test/Menu-test.js
```diff
@@ -183,15 +183,20 @@ describe('Menu', () => {
...
+        // The close event is delayed by a minimal amount of time in
+        // order to allow links to say in the DOM long enough to be
+        // followed on a click. Therefore, this test must simulate
+        // time passing in order for the menu to close.
+        clock.tick(1);
+        wrapper.update();
+        assert.equal(isOpen(wrapper), !shouldClose);
+      }
+      finally {
+        clock.restore();
```
robertknight, line 197: There is a module-level `afterEach` which has a `clock.restore()` call, so this should't be needed.

src/sidebar/services/test/session-test.js
```diff
@@ -172,30 +172,26 @@ describe('SessionService', () => {
...
+        session.load();
```
robertknight, line 176: `session.load()` is an async method so the test needs to await the result. Otherwise the ongoing async activity might interrupt some other test.

src/sidebar/services/test/session-test.js
```diff
@@ -172,30 +172,26 @@ describe('SessionService', () => {
...
-          .then(() => {
-            assert.calledTwice(fakeApi.profile.read);
-          });
+        try {
+          const session = createService();
+          session.load();
+          clock.tick(CACHE_TTL * 2);
+          session.load();
+        } finally {
+          clock.restore();
```
robertknight, line 191: There is an `afterEach` block which restores the clock after each test, so this is not needed.

src/sidebar/util/test/postmessage-json-rpc-test.js
```diff
@@ -148,18 +148,21 @@ describe('sidebar/util/postmessage-json-rpc', () => {
...
+          origin,
+          data: {
+            jsonrpc: '2.0',
+            id: messageId,
+            result: {},
+          },
+        });
+        await result;
+      } finally {
+        clock.restore();
```
robertknight, line 164: This `clock.restore()` however is needed.

robertknight (changes requested):
Changes in `MenuKeyboardNavigation` and `postmessage-json-rpc-test.js` look correct. The others are not needed however for reasons I explained in comments. Is there a way to detect whether fake clocks are currently active? If so it might be a good idea to put a check in the global `afterEach` callback in `src/sidebar/test/bootstrap.js`.

src/sidebar/services/test/session-test.js
esanzgar: Indeed.

src/sidebar/components/test/Menu-test.js
esanzgar: When a fake clocks is used on a single test, I think it is better to declare and restore the fake clock within the function. I have removed the block-level variable. I think this is beneficial from a readability perspective.

src/sidebar/services/test/session-test.js
```diff
@@ -170,32 +162,28 @@ describe('SessionService', () => {
...
-        return session
-          .load()
-          .then(() => {
-            return session.load();
-          })
-          .then(() => {
-            assert.calledOnce(fakeApi.profile.read);
-          });
+        await session.load();
+        await session.load();
```
esanzgar, line 168: This is unrelated, but I think it is clearer: two consecutive `load` operations triggers the reading of the profile only once.

src/sidebar/services/test/session-test.js
esanzgar: When a fake clocks is used on a single test, I think it is better to declare and restore the fake clock within the function. I have removed the block-level variable. I think this is beneficial from a readability perspective.

esanzgar (commented):
> Thinking about a general way to address this problem: Is there an API to test if fake clocks are currently active? If so it might be a good idea to put a check in the global afterEach callback in src/sidebar/test/bootstrap.js.

I see your reasoning here. However, given the nested structure of the tests I think it could lead to problems:

```
describe...
  let clock;

  beforeEach(() => {
   clock = sinon.useFakeTimers()
  });
 
 afterEach(() => {
   we would like to restore the clock if it has been assigned in this block.
 });

  describe... (nested)
  
  afterEach(()=>{
   restore automatically the clock set in the parent block?
  })
```
 
On the second commit I decided that when a fake clocks is used only on a single test, I think it is better to declare and restore the fake clock within the function. I have removed the block-level variable in two files. I think this is beneficial from a readability perspective.

src/sidebar/components/test/Menu-test.js
robertknight: OK, that sounds sensible to me.

robertknight (approved):
The changes look good to me. I agree that if fake clocks are only used in one or a minority of tests it seems nicer to localize the variable and cleanup.

PR #3902, opened 2021-11-03
Update two test dependencies to ESM versions

robertknight:
Update babel-plugin-mockable-imports and enzyme-adapter-preact-pure to
the newest versions which have ES module builds.

This is primarily to verify that these releases, which I maintain, work
in a real application.


robertknight:
Obama meme

PR #3903, opened 2021-11-03
Adoption of `PortProvider` and `PortFinder`

### LINKED ISSUE #3533: Epic: overhaul the inter-frame communication system

esanzgar:
This adoption simplifies the discovery of frames.

I did a little bit of clean-up in the tests: rearranged a few variables
alphabetically, and delete unneeded lines and variables.

Closes #3533

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -102,7 +102,8 @@ function resolveAnchor(anchor) {
  * loads Hypothesis (not all frames will be annotation-enabled). In one frame,
  * usually the top-level one, there will also be an instance of the `Sidebar`
  * class that shows the sidebar app and surrounding UI. The `Guest` instance in
- * each frame connects to the sidebar when {@link connectToSidebar} is called.
+ * each frame connects to the sidebar as part of its initialization, when
+ * {@link _connectSidebarEvents} is called.
  *
  * The anchoring implementation defaults to a generic one for HTML documents and
  * can be overridden to handle different document types.
@@ -118,10 +119,14 @@ export default class Guest {
    * @param {EventBus} eventBus -
    *   Enables communication between components sharing the same eventBus
    * @param {Record<string, any>} [config]
+   * @param {Window} [hostFrame] -
+   *   Host frame which this guest is associated with. This is expected to be
+   *   an ancestor of the guest frame. It may be same or cross origin.
    */
-  constructor(element, eventBus, config = {}) {
+  constructor(element, eventBus, config = {}, hostFrame = window) {
     this.element = element;
     this._emitter = eventBus.createEmitter();
+    this._hostFrame = hostFrame;
     this._highlightsVisible = false;
     this._isAdderVisible = false;
 
@@ -161,9 +166,7 @@ export default class Guest {
     // Set the frame identifier if it's available.
     // The "top" guest instance will have this as null since it's in a top frame not a sub frame
     /** @type {string|null} */
-    this._frameIdentifier = config.subFrameIdentifier;
-    this._hostFrame =
-      config.subFrameIdentifier === null ? window : window.parent;
+    this._frameIdentifier = config.subFrameIdentifier || null;
     this._portFinder = new PortFinder({
       hostFrame: this._hostFrame,
       source: 'guest',
@@ -181,10 +184,6 @@ export default class Guest {
     // in this frame.
     this._annotationSync = new AnnotationSync(eventBus, this._bridge);
     this._connectAnnotationSync();
-
-    // Discover and connect to the sidebar frame. All RPC events must be
-    // registered before creating the channel.
-    this._connectToSidebar();
 
     // Set up automatic and integration-triggered injection of client into
     // iframes in this frame.
@@ -316,7 +315,7 @@ export default class Guest {
     });
   }
 
-  _connectSidebarEvents() {
+  async _connectSidebarEvents() {
     // Handlers for events sent when user hovers or clicks on an annotation card
     // in the sidebar.
     this._bridge.on('focusAnnotations', (tags = []) => {
@@ -367,12 +366,9 @@ export default class Guest {
     this._bridge.on('setHighlightsVisible', showHighlights => {
       this.setHighlightsVisible(showHighlights);
     });
-  }
 
-  /**
-   * Attempt to connect to the sidebar frame.
-   */
-  async _connectToSidebar() {
+    // Discover and connect to the sidebar frame. All RPC events must be
+    // registered before creating the channel.
     const sidebarPort = await this._portFinder.discover('sidebar');
     this._bridge.createChannel(sidebarPort);
   }
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -177,12 +177,12 @@ export default class Guest {
      *
      * @type {Bridge<GuestToSidebarEvent,SidebarToGuestEvent>}
      */
-    this._bridge = new Bridge();
+    this._sidebarRPC = new Bridge();
     this._connectSidebarEvents();
 
     // Set up listeners for when the sidebar asks us to add or remove annotations
     // in this frame.
-    this._annotationSync = new AnnotationSync(eventBus, this._bridge);
+    this._annotationSync = new AnnotationSync(eventBus, this._sidebarRPC);
     this._connectAnnotationSync();
 
     // Set up automatic and integration-triggered injection of client into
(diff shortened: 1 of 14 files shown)


src/annotator/guest.js
```diff
@@ -361,7 +369,16 @@ export default class Guest {
     });
   }
 
+  /**
+   * Attempt to connect to the sidebar frame.
+   */
+  async _connectToSidebar() {
```
robertknight, line 375: There is a reference to the public `connectToSidebar` method in the description of the `Guest` class that needs to be removed. You could add a note in the docs for the constructor that the Guest will connect to sidebar as part of its initialization.

src/annotator/index.js
```diff
@@ -42,63 +40,26 @@ const sidebarLinkElement = /** @type {HTMLLinkElement} */ (
...
-      guest.connectToSidebar(frame, sidebarOrigin)
-    );
-  } else {
-    // eslint-disable-next-line no-console
-    console.warn(
-      `Hypothesis guest frame in ${location.origin} could not find a sidebar to connect to.
-Guest frames can only connect to sidebars in their same-origin parent frame.`
-    );
+    // Clear `annotations` value from the notebook's config to prevent direct-linked
+    // annotations from filtering the threads.
```
robertknight, line 56: This was a pre-existing issue, but this comment doesn't make sense any more. It talks about doing something with an "`annotations` value" but there is no code nearby that does this. I suggest to just remove it.

src/annotator/index.js
```diff
@@ -42,63 +40,26 @@ const sidebarLinkElement = /** @type {HTMLLinkElement} */ (
  * client is initially loaded, is also the only guest frame.
  */
 function init() {
-  // Create an internal global used to share data between same-origin guest and
-  // host frames.
-  window_.__hypothesis = {};
-
   const annotatorConfig = getConfig('annotator');
-  const hostFrame = annotatorConfig.subFrameIdentifier ? window.parent : window;
+  const isHostFrame = annotatorConfig.subFrameIdentifier === null;
```
robertknight, line 44: As noted in comments in the `Guest` class, I'm not convinced about pushing the knowledge about how to find the host frame in relation to the current frame down into the `Guest` class.

Previously the "is this the host frame?" and "where is the host frame?" logic was located in the same place, in the `init` function. Now it is split across this function and the `Guest` constructor.

src/annotator/test/integration/anchoring-test.js
```diff
@@ -1,7 +1,7 @@
 // Tests that the expected parts of the page are highlighted when annotations
 // with various combinations of selector are anchored.
 
-import Guest from '../../guest';
+import Guest, { $imports } from '../../guest';
```
robertknight, line 4: Since this is not a unit test which corresponds to a single module, where the meaning of `$imports` would be obvious, I think it would be a good idea to give this a more specific name, eg. `guestImports`. Also if you ever need to mock imports from additional modules in this test, you'll have a pattern to follow.

robertknight (changes requested):
I have had an initial look over this. Functionally everything seems to work as desired. Aside from the minor comments, a few high level issues I raised:

1. Where we put the logic that determines where the host frame is, in relation to the current frame. Related to this, the behavior of the `Guest` constructor changes significantly depending on the `config.subFrameIdentifier` property, but that's not captured at all in docs or argument types.
2. Where the `PortProvider` is constructed. This currently happens in the `Sidebar` class, but perhaps it should live at a higher level (ie. the entry point) since it is so central to how the application works and should be easy to find.
3. I raised a hazard regarding stubbing of `window.parent.postMessage` in guests. That doesn't break anything in the test runner that I can see at present, but I can see a risk of that happening in future.

esanzgar (commented):
@robertknight, I have added all your suggestions on a second commit.

src/annotator/index.js
```diff
@@ -13,14 +13,13 @@ import { registerIcons } from '@hypothesis/frontend-shared';
...
 registerIcons(iconSet);
 
+import { PortProvider } from '../shared/port-provider';
 import { getConfig } from './config/index';
 import Guest from './guest';
 import Notebook from './notebook';
 import Sidebar from './sidebar';
 import { EventBus } from './util/emitter';
 
-const window_ = /** @type {HypothesisWindow} */ (window);
```
robertknight, line 22: The `HypothesisWindow` typedef in this file is currently unused.

src/annotator/sidebar.js
```diff
@@ -56,17 +56,20 @@ export default class Sidebar {
    * @param {HTMLElement} element
    * @param {import('./util/emitter').EventBus} eventBus -
    *   Enables communication between components sharing the same eventBus
+   * @param {MessagePort} hostPort -
+   *   Enables communication with the sidebar frame
```
robertknight, line 60: It might be useful to add some additional context here. eg. "Host port for the host <-> sidebar communication channel. The sidebar app will request its side of the channel when it starts."

src/shared/port-provider.js
```diff
@@ -178,7 +178,7 @@ export class PortProvider {
   /**
    * Returns the `host` port from the `sidebar-host` channel.
    */
-  get sidebarPort() {
```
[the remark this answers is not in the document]
robertknight: You'll need to change this if other other host-side ports can be retrieved in future, such as a host <-> notebook port. I think something like `hostPortFor('sidebar')` might be a good API even if the 'sidebar' argument is currently redundant, as it makes it clearer to a reader what is at the other end of the port.

src/shared/port-provider.js
```diff
@@ -162,12 +162,12 @@ export class PortProvider {
...
 
     if (channel === 'guest-host' && message.frame1 === 'guest') {
-      this._emitter.emit('hostPortRequest', message.frame1, port2);
+      this._emitter.emit('hostFrameRequest', message.frame1, port2);
     }
   }
 
   /**
-   * @param {'hostPortRequest'} eventName
+   * @param {'hostFrameRequest'} eventName
```
robertknight, line 170: The important point for readers to know is that this event signals that a frame (currently the sidebar, but maybe other frames in future) has connected and the host may want to interact with it in some way. How about `frameConnected`?

src/sidebar/services/frame-sync.js
```diff
@@ -226,7 +234,7 @@ export class FrameSyncService {
   /**
    * Connect to the host frame and guest frame(s) in the current browser tab.
    */
-  connect() {
+  async connect() {
```
robertknight, line 237: Because `connect` is now async, any failures will no longer trigger the `catch` block at the bottom of `src/sidebar/index.js` and thus the exception won't be reported via `reportLaunchError`. I think fixing that would be too big of a change for this PR, but we should address it in future.

robertknight (approved):
I had some suggestions regarding `PortProvider` APIs and a few JSDoc comments. Otherwise the changes look good to me.

src/shared/port-provider.js
esanzgar: 👍🏼

PR #3904, opened 2021-11-05
Support annotation/note button toggle on multi-frame situations

esanzgar:


src/annotator/guest.js
```diff
@@ -315,6 +327,37 @@ export default class Guest {
     });
   }
 
+  async _connectHostEvents() {
+    this._hostRPC.on(
+      'deselectTextExcept',
+      /** @type {string} */ frameIdentifier => {
```
robertknight, line 333: `@type` doesn't seem to work in this context. This works though:

```
      'deselectTextExcept',
      /** @param {string} frameIdentifier */
      frameIdentifier => {
```

src/annotator/guest.js
```diff
@@ -315,6 +327,37 @@ export default class Guest {
...
+        // It would be nice to clear the shadow of the selection calling
+        // `document.getSelection()?.removeAllRanges();` however, because of the
+        // selection observer, that will trigger a call to `_onClearSelection({informHostFrame: true})`
+        // which in turn will remove the selection from the `Guest` that has the selection.
+      }
+    );
+
+    this._hostRPC.on(
+      'createAnnotationAt',
+      /** @type {string} */ frameIdentifier => {
```
robertknight, line 348: See note about `@type` above.

src/annotator/sidebar.js
```diff
@@ -65,6 +67,22 @@ export default class Sidebar {
   constructor(element, eventBus, guest, config = {}) {
     this._emitter = eventBus.createEmitter();
 
+    /**
+     * Tracks which `Guest` has a text selection. It uses the frame i«dentifier
```
robertknight, line 71:
```suggestion
     * Tracks which `Guest` has a text selection. It uses the frame identifier
```

src/annotator/sidebar.js
```diff
@@ -225,11 +244,36 @@ export default class Sidebar {
    * @param {MessagePort} port
    */
   onFrameConnected(source, port) {
-    if (source === 'sidebar') {
-      this._sidebarRPC.createChannel(port);
+    switch (source) {
```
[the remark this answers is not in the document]
robertknight: I'm fine with switches.

src/annotator/sidebar.js
```diff
@@ -225,11 +244,36 @@ export default class Sidebar {
...
+      case 'sidebar':
+        this._sidebarRPC.createChannel(port);
+        break;
     }
   }
 
+  _setupGuestEvents() {
+    this._guestRPC.on(
+      'textSelectedAt',
+      /** @type {string} */ frameIdentifier => {
```
robertknight, line 260: See comments about `@type` in guest.js. You'll need to use `@param` here I think.

src/annotator/test/guest-test.js
```diff
@@ -624,24 +624,70 @@ describe('Guest', () => {
...
+      fakeBridge.call.resetHistory();
+      handler('dummy');
 
-      assert.calledWith(callback, true);
+      assert.equal(guest.selectedRanges.length, 0);
+      assert.notCalled(fakeBridge.call);
     });
 
-    it('emits `hasSelectionChanged` event with argument `false` if selection is empty', () => {
+    it("doesn't unselects text if frame identifiers matches", () => {
```
robertknight, line 680:
```suggestion
    it("doesn't deselect text if frame identifiers matches", () => {
```

src/annotator/test/sidebar-test.js
```diff
@@ -217,30 +218,47 @@ describe('Sidebar', () => {
...
+
+      FakeToolbarController.args[0][1].createAnnotation();
+
+      assert.calledWith(fakeBridge.call, 'createAnnotationAt', 'main');
+    });
+
+    it('creates an annotation on another frame toolbar button is clicked', () => {
+      createSidebar({});
+
+      // make a text selection on another frame
```
robertknight, line 232:
```suggestion
      // Make a text selection in another frame
```

src/annotator/test/sidebar-test.js
```diff
@@ -588,6 +606,16 @@ describe('Sidebar', () => {
     });
   });
 
+  describe('#onFrameConnected', () => {
+    it('creates a channel to communicate to the guests', () => {
```
robertknight, line 610:
```suggestion
    it('creates a channel to communicate with the guests', () => {
```

src/annotator/guest.js
```diff
@@ -315,6 +327,37 @@ export default class Guest {
     });
   }
 
+  async _connectHostEvents() {
+    this._hostRPC.on(
```
robertknight, line 331: If the messages from the host could target a specific guest, that would remove the need for individual messages to filter based on frame identifier.

src/annotator/guest.js
```diff
@@ -315,6 +327,37 @@ export default class Guest {
...
+        this._onClearSelection({ informHostFrame: false });
+        // It would be nice to clear the shadow of the selection calling
+        // `document.getSelection()?.removeAllRanges();` however, because of the
+        // selection observer, that will trigger a call to `_onClearSelection({informHostFrame: true})`
+        // which in turn will remove the selection from the `Guest` that has the selection.
+      }
+    );
+
+    this._hostRPC.on(
+      'createAnnotationAt',
```
robertknight, line 347: As long as these messages include a frame identifier, `In` seems a better suffix (as in "create an annotation _in_ the frame" or "deselected text _in_ the frame").

robertknight (commented):
Functionally this works well. The main query I had in looking at this was whether it would make sense to rework it to avoid each individual message needing to identify the guest that the message came from or is going to. Currently each of the new messages to guests needs to include a frame identifier which the guest then has to filter on. If the host frame could send a message to specific guests that would avoid the need for this. Likewise if the host frame knew which guest a particular message was coming from, the message wouldn't need to include the frame identifier.

esanzgar:
@robertknight, your suggestion sounds very sensible, however I suggest we implement it on a follow up PR. Creating a version of `Bridge` that is able to send messages to specific frame, or all frames except one will require a few changes:

```
guestRPC.call('unselectText', {except: 'frameIdentifier'})
guestRPC.call('createAnnotation, annotation, {to: 'frameIdentifier'})
```

The `Bridge#call` signature will have to accommodate for (1) a variable lenghth array of arguments, (2) a callback (except if we convert it to Promise-based) and (2) a future option to target specific frames.

src/annotator/guest.js
esanzgar: Totally, agree.

esanzgar (commented):
I have incorporated your preliminary comments on a second commit.

src/annotator/guest.js
```diff
@@ -102,8 +104,9 @@ function resolveAnchor(anchor) {
  * loads Hypothesis (not all frames will be annotation-enabled). In one frame,
  * usually the top-level one, there will also be an instance of the `Sidebar`
  * class that shows the sidebar app and surrounding UI. The `Guest` instance in
- * each frame connects to the sidebar as part of its initialization, when
- * {@link _connectSidebarEvents} is called.
+ * each frame connects to the sidebar frame as part of its initialization, when
+ * {@link _connectSidebarEvents} is called. The `Guest` is also connected to the
+ * host frame using the {@link _connectHostEvents} method.
```
robertknight, line 109: I suggest removing references to internal methods, as these are liable to get out of date in response to future refactoring.

src/annotator/guest.js
```diff
@@ -315,6 +327,39 @@ export default class Guest {
...
+    this._hostRPC.on(
+      'unselectTextExceptIn',
+      /** @param {string} frameIdentifier */
+      frameIdentifier => {
+        if (this._frameIdentifier === frameIdentifier) {
+          return;
+        }
+
+        this._onClearSelection({ informHostFrame: false });
+        // It would be nice to clear the shadow of the selection calling
```
robertknight, line 340: I'm not sure what is meant by "shadow of the selection"? Do you mean the way some browsers render the selection in un-focused iframes in a grey color? That's not a shadow of the selection, that _is_ the selection.

A way to avoid this could be to set a flag to ignore the next selection update, which is then reset when the update is received. This would avoid the need to call `_onClearSelection` here explicitly.

src/annotator/sidebar.js
```diff
@@ -67,6 +69,22 @@ export default class Sidebar {
...
     this._emitter = eventBus.createEmitter();
 
+    /**
+     * Tracks which `Guest` has a text selection. It uses the frame identifier
+     * which uniquely identifies each `Guest`. If there is no text selected, the
+     * value is set to `null`.
+     *
+     * @type {null|string}
+     */
+    this._selectionAt = null;
```
robertknight, line 79:
```suggestion
    this._guestWithSelection = null;
```

This would make it more immediately obvious what this field is.

src/annotator/sidebar.js
```diff
@@ -67,6 +69,22 @@ export default class Sidebar {
   constructor(element, eventBus, guest, config = {}) {
     this._emitter = eventBus.createEmitter();
 
+    /**
+     * Tracks which `Guest` has a text selection. It uses the frame identifier
```
robertknight, line 73: To check my understanding - we're assuming that only one guest can have a selection at a time because the sidebar will unselect text in other guests?

src/annotator/sidebar.js
```diff
@@ -67,6 +69,22 @@ export default class Sidebar {
...
+     * which uniquely identifies each `Guest`. If there is no text selected, the
+     * value is set to `null`.
+     *
+     * @type {null|string}
```
robertknight, line 77: I tend to put the null/undefined type last when writing out types as it is the least-distinctive part of the type name.

src/annotator/guest.js
robertknight: A sketch of how this might work:

```diff
diff --git a/src/annotator/guest.js b/src/annotator/guest.js
index c4eb66b0f..ea0905894 100644
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -90,6 +90,10 @@ function resolveAnchor(anchor) {
   }
 }
 
+function clearSelection() {
+  /** @type {Selection} */ (document.getSelection()).removeAllRanges();
+}
+
 /**
  * `Guest` is the central class of the annotator that handles anchoring (locating)
  * annotations in the document when they are fetched by the sidebar, rendering
@@ -136,11 +140,11 @@ export default class Guest {
     this._adder = new Adder(this.element, {
       onAnnotate: async () => {
         await this.createAnnotation();
-        /** @type {Selection} */ (document.getSelection()).removeAllRanges();
+        clearSelection();
       },
       onHighlight: async () => {
         await this.createAnnotation({ highlight: true });
-        /** @type {Selection} */ (document.getSelection()).removeAllRanges();
+        clearSelection();
       },
       onShowAnnotations: anns => {
         this.selectAnnotations(anns);
@@ -154,6 +158,7 @@ export default class Guest {
         this._onClearSelection();
       }
     });
+    this._ignoreNextSelection = false;
 
     /**
      * The anchors generated by resolving annotation selectors to locations in the
@@ -335,12 +340,8 @@ export default class Guest {
         if (this._frameIdentifier === frameIdentifier) {
           return;
         }
-
-        this._onClearSelection({ informHostFrame: false });
-        // It would be nice to clear the shadow of the selection calling
-        // `document.getSelection()?.removeAllRanges();` however, because of the
-        // selection observer, that will trigger a call to `_onClearSelection({informHostFrame: true})`
-        // which in turn will remove the selection from the `Guest` that has the selection.
+        this._ignoreNextSelection = true;
+        clearSelection();
       }
     );
 
@@ -661,12 +662,15 @@ export default class Guest {
     this._adder.show(focusRect, isBackwards);
   }
 
-  _onClearSelection({ informHostFrame = true } = {}) {
+  _onClearSelection() {
     this._isAdderVisible = false;
     this._adder.hide();
     this.selectedRanges = [];
-    if (informHostFrame) {
+
+    if (!this._ignoreNextSelection) {
       this._hostRPC.call('textUnselectedIn', this._frameIdentifier);
+    } else {
+      this._ignoreNextSelection = false;
     }
   }
 ```

robertknight (commented):
I had a suggestion regarding clearing the selection in one guest when text is selected in another guest, avoiding a disconnect between the state of the selection in the guest (present) and the state of the adder (hidden).

src/annotator/guest.js
esanzgar: Should I remove the one that was already in the code?

src/annotator/sidebar.js
esanzgar: 👍🏼 (I have changed this type in the followup PR).

src/annotator/sidebar.js
esanzgar: Yes.

src/annotator/guest.js
esanzgar: I believe this proposal won't resolve the problem.

I propose we leave this as a follow up.

src/annotator/guest.js
robertknight: Yes please.

src/annotator/guest.js
robertknight: > I believe this proposal won't resolve the problem.

Why not? I did test this code locally.

robertknight:
@esanzgar - Can you rebase this on master. That should fix the Jenkins CI failure. Also I had a response re. clearing the selection above. The alternative proposal in https://github.com/hypothesis/client/pull/4014 could work, but I want to understand why you think the suggestion I posted in this comment wouldn't.

esanzgar:
> The alternative proposal in #4014 could work, but I want to understand why you think the suggestion I posted in this comment wouldn't.

I didn't fully understand your sketched solution. If you removed the selection (`/** @type {Selection} */ (document.getSelection()).removeAllRanges()`) from `onAnnotate` and `onHighlight` callbacks then it would need to be added on `_onClearSelection` method. That could result in an infinite loop, if not base case is introduced). Because of `_onClearSelection` is called from different parts of the code, it is also kind of difficult to know what was the value of `this._ignoreNextSelection` previous to the call.

I am happy to look at an alternative proposal to #4014.

robertknight:
> I didn't fully understand your sketched solution. If you removed the selection (/** @type {Selection} */ (document.getSelection()).removeAllRanges()) from onAnnotate and onHighlight callbacks then it would need to be added on _onClearSelection method. 

The changes to `onAnnotate` and `onHighlight` are just extracting duplicated code into a helper function. There is no functional to that part. It isn't necessary to call `_onClearSelection` directly in these handlers because the selection change is handled by SelectionObserver and reported to the guest. The way selection clearing in `onAnnotate` and `onHighlight` works is the same as if the user, or another piece of code on the page, cleared the selection:

1. User clears selection, or `onAnnotate` / `onHighlight` does, or third-party code on the page
2. SelectionObserver handles and reports the selection change
3. Guest receives notification from SelectionObserver and calls `_onClearSelection`

Something that I realized may be confusing here is that `_onClearSelection` is called _after_ the selection is cleared, not _to clear_ the selection.

esanzgar:
Regarding your alternative idea, yes it works. I missed the function `clearSelection` on top and thought about `_onClearSelection` instead.

robertknight (approved):

PR #3905, opened 2021-11-05
Improved separation between primary (main) and secondary `Guest`s

esanzgar:
Before this PR, page notes were associated to the URL of the host frame.
This is because the `Guest` class in the host frame was given the 'main'
role.

In this PR, the logic for deciding which frame contains the main
annotatable content is delegated to the `createIntegration` function.
Using some simple heuristics, this function decides which frames
contains the primary annotatable content. For the primary annotatable
frame it returns 'main' as the frame identifier. For other annotatable
frames it returns an identifier of the form `sub_[10 alpha-numerical]`.
The heuristic could be enhanced in the future.

On a second commit, I make the type stricter. This commit could be dropped.

Closes https://github.com/hypothesis/product-backlog/issues/1235

---

**Testing:**

1. Open the [local development VS EPUB page](http://localhost:3000/document/vitalsource-epub)

2. Create a page note for the current chapter

3. Navigate to another chapter, the notes disappears because it is
   associated with the URL of the previous chapter.

In the case of the VS the main annotatable frame is associated with an
specific iframe, not the host frame. For other web pages and PDFs there
should not be any change in behaviour.

### DIFF
--- a/src/annotator/integrations/index.js
+++ b/src/annotator/integrations/index.js
@@ -1,3 +1,5 @@
+import { generateHexString } from '../../shared/random';
+
 import { HTMLIntegration } from './html';
 import { PDFIntegration, isPDF } from './pdf';
 import {
@@ -13,22 +15,32 @@ import {
 
 /**
  * Create the integration that handles document-type specific aspects of
- * guest functionality.
+ * guest functionality. Also returns a frame identifier based on the
+ * role of the frame: 'main' for the frame that holds the primary annotatable
+ * content, or 'sub_[random string]' for other annotatable frames.
  *
  * @param {Annotator} annotator
- * @return {Integration}
+ * @param {string|null} [frameIdentifier]
+ * @return {[Integration, 'main'|`sub_${string}`]}.
  */
-export function createIntegration(annotator) {
+export function createIntegration(annotator, frameIdentifier = null) {
   if (isPDF()) {
-    return new PDFIntegration(annotator);
+    return [new PDFIntegration(annotator), 'main'];
   }
+
+  const subFrameIdentifier = /** @type {const} */ (
+    `sub_${generateHexString(10)}`
+  );
 
   const vsFrameRole = vitalSourceFrameRole();
   if (vsFrameRole === 'container') {
-    return new VitalSourceContainerIntegration(annotator);
+    return [new VitalSourceContainerIntegration(annotator), subFrameIdentifier];
   } else if (vsFrameRole === 'content') {
-    return new VitalSourceContentIntegration();
+    return [new VitalSourceContentIntegration(), 'main'];
   }
 
-  return new HTMLIntegration();
+  return [
+    new HTMLIntegration(),
+    frameIdentifier === null ? 'main' : subFrameIdentifier,
+  ];
 }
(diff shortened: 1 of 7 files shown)


src/annotator/integrations/index.js
```diff
@@ -1,3 +1,5 @@
+import { generateHexString } from '../../sidebar/util/random';
```
esanzgar, line 1: This is importing from annotator but I will fix it in a different PR.

src/annotator/integrations/index.js
```diff
@@ -9,26 +11,37 @@ import {
...
+ * @return {[Integration, FrameIdentifier]}.
  */
-export function createIntegration(annotator) {
+export function createIntegration(annotator, frameIdentifier = null) {
   if (isPDF()) {
-    return new PDFIntegration(annotator);
+    return [new PDFIntegration(annotator), 'main'];
   }
 
+  const subFrameIdentifier = /** @type {const} */ (
```
esanzgar, line 32: `@type {const}` is new in TS 4.5. Very handy!

src/annotator/integrations/index.js
```diff
@@ -9,26 +11,37 @@ import {
...
  * Create the integration that handles document-type specific aspects of
- * guest functionality.
+ * guest functionality. Also returns a frame identifier based on the
+ * role of the frame: 'main' for the frame that holds the primary annotatable
+ * content, or 'sub_[random string]' for other annotatable frames.
  *
  * @param {Annotator} annotator
- * @return {Integration}
+ * @param {string|null} [frameIdentifier]
+ * @return {[Integration, FrameIdentifier]}.
```
esanzgar, line 25: Setting (internally) a different frame identifier value from the value of `subFrameIdentifier` in the configuration is a good idea to avoid conflicts.

Currently, the `subFrameIdentifier` configuration property is only used to decide if the sidebar and notebook should be shown or not in that frame. It is kind of a bad naming. It would have been much better to rename it `enableSidebar` and transform it into a boolean. This is, nonetheless, out of question now.

External users have made use of `subFrameIdentifier` configuration property to force the sidebar from showing. Here is an example of that use in the [Page containing a cross-origin guest iframe development scenario](http://localhost:3000/document/cross-origin-iframe): 

https://github.com/hypothesis/client/blob/4d145d7d52e3d3b2b7041dbd4e53cdb7af7273c9/dev-server/documents/html/doyle-embedded.mustache#L8-L11

If a user decide to set the value of the `subFrameIdentifier` to 'main', it will result in errors, possible duplication of page notes or annotations.

robertknight: I think we could get away with deprecating `subFrameIdentifier` and treating existing uses of it as a boolean - if present, run client in "guest only" mode, otherwise run in host + guest mode. We can map deprecated properties to modern equivalents as part of configuration parsing.

esanzgar: >I think we could get away with deprecating subFrameIdentifier and treating existing uses of it as a boolean - if present, run client in "guest only" mode, otherwise run in host + guest mode. We can map deprecated properties to modern equivalents as part of configuration parsing.

I think it is a good idea.

esanzgar:
Robert has found some objections to this approach:

- There could be several PDF.js readers on the page: the heuristics about PDF doesn't work correctly.

- Robert, would prefer to not instantiate or instantiate a minimal Guest for the container VS frame.

PR #3913, opened 2021-11-08
Prevent Node globals from being included in typechecking environment

robertknight:
By default TypeScript includes all `@types/<name>` packages. This includes
@types/node, though we don't use it directly, because it is a transitive
dependency (see `yarn why @types/node`).

As a result Node's globals are added to the environment TS sees, even though
they don't really exist when our code runs in a browser. These globals
include overloads for `setTimeout` and `setInterval` which return a different
type (`Timeout`), causing spurious errors when assigning the result to a number.

Fix the problem by using the `types` option [1] in tsconfig.json to explicitly
specify which `@types/<name>` packages  to include when checking code in src/
This does mean that if we ever intentionally add @types packages which declare
globals (eg. for mocha), we'll need to explicitly list them here.

In the process the ES target was updated to fix an error about a `BigInt` reference.

[1] https://www.typescriptlang.org/tsconfig#types

### DIFF
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "allowJs": true,
     "checkJs": true,
-    "lib": ["es2018", "dom", "dom.iterable"],
+    "lib": ["es2020", "dom", "dom.iterable"],
     "jsx": "react-jsx",
     "jsxImportSource": "preact",
     "module": "commonjs",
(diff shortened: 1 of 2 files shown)


robertknight:
There are alternative ways to fix the `setTimeout`/setInterval` issue, which is to use `window.setTimeout` to force use of the browser version of `setTimeout`. The approach here has the advantage of preventing accidental usage of Node globals, at the expensive of requiring more complex TS configuration.

esanzgar (approved):
We don't currently check types in the test files, that's why mocha's globals types are not needed, correct?

If ever want to type check our tests, an alternative way would be to import mocha in every file: `import 'mocha';`.

robertknight:
> We don't currently check types in the test files, that's why mocha's globals types are not needed, correct?

That's correct. If we ever did decide to do that then we'd just add mocha, chai and sinon to the `types` field here.

PR #3914, opened 2021-11-08
Add a remove method to ListenerCollection

esanzgar:
This new method allows to unregister individual event listeners.

### DIFF
--- a/src/shared/listener-collection.js
+++ b/src/shared/listener-collection.js
@@ -11,11 +11,13 @@
  */
 export class ListenerCollection {
   constructor() {
-    /** @type {Listener[]} */
-    this._listeners = [];
+    /** @type {Map<Symbol, Listener>} */
+    this._listeners = new Map();
   }
 
   /**
+   * Adds a listener and returns a listenerId
+   *
    * @param {Listener['eventTarget']} eventTarget
    * @param {Listener['eventType']} eventType
    * @param {Listener['listener']} listener
@@ -23,13 +25,28 @@ export class ListenerCollection {
    */
   add(eventTarget, eventType, listener, options) {
     eventTarget.addEventListener(eventType, listener, options);
-    this._listeners.push({ eventTarget, eventType, listener });
+    const symbol = Symbol();
+    this._listeners.set(symbol, { eventTarget, eventType, listener });
+    return symbol;
+  }
+
+  /**
+   * Removes a listener using a listenerId
+   * @param {Symbol} listenerId
+   */
+  remove(listenerId) {
+    const event = this._listeners.get(listenerId);
+    if (event) {
+      const { eventTarget, eventType, listener } = event;
+      eventTarget.removeEventListener(eventType, listener);
+      this._listeners.delete(listenerId);
+    }
   }
 
   removeAll() {
     this._listeners.forEach(({ eventTarget, eventType, listener }) => {
       eventTarget.removeEventListener(eventType, listener);
     });
-    this._listeners = [];
+    this._listeners.clear();
   }
 }
--- a/src/shared/listener-collection.js
+++ b/src/shared/listener-collection.js
@@ -16,7 +16,7 @@ export class ListenerCollection {
   }
 
   /**
-   * Adds a listener and returns a listenerId
+   * Add a listener and return an ID that can be used to remove it later
    *
    * @param {Listener['eventTarget']} eventTarget
    * @param {Listener['eventType']} eventType
@@ -31,7 +31,8 @@ export class ListenerCollection {
   }
 
   /**
-   * Removes a listener using a listenerId
+   * Remove a listener using a listenerId
+   *
    * @param {Symbol} listenerId
    */
   remove(listenerId) {
--- a/src/shared/test/listener-collection-test.js
+++ b/src/shared/test/listener-collection-test.js
@@ -22,7 +22,7 @@ describe('ListenerCollection', () => {
   });
 
   describe('#remove', () => {
-    it('unregisters some event listeners', () => {
+    it('unregisters the specified listener', () => {
       const listener1 = sinon.stub();
       const listener2 = sinon.stub();
       listeners.add(window, 'resize', listener1);
--- a/src/shared/test/listener-collection-test.js
+++ b/src/shared/test/listener-collection-test.js
@@ -11,23 +11,41 @@ describe('ListenerCollection', () => {
     listeners.removeAll();
   });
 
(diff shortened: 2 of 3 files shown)


src/shared/listener-collection.js
```diff
@@ -11,24 +11,41 @@
  */
 export class ListenerCollection {
   constructor() {
-    /** @type {Listener[]} */
+    /** @type {Array<Listener|null>} */
     this._listeners = [];
   }
 
   /**
+   * Adds a listener and returns a listenerId
```
robertknight, line 19: There should be a blank line between the description of a symbol and any JSDoc tags (`@param`, `@prop` etc.).

src/shared/test/listener-collection-test.js
```diff
@@ -19,7 +19,19 @@ describe('ListenerCollection', () => {
     assert.calledOnce(listener);
   });
 
-  it('unregisters event listeners', () => {
+  it('unregisters some event listeners', () => {
```
robertknight, line 22: Can you wrap the tests here in `describe(<method name>)` blocks to make it more obvious which functions are being tested, eg:

```js
describe('#remove', () => {
  it('removes an event listener', () => { ... });
});

describe('#removeAll', () => {
  it('unregisters all event listeners', () => { ... });
});
```

src/shared/listener-collection.js
```diff
@@ -11,24 +11,41 @@
...
    * @param {Listener['eventTarget']} eventTarget
    * @param {Listener['eventType']} eventType
    * @param {Listener['listener']} listener
    * @param {AddEventListenerOptions} [options]
    */
   add(eventTarget, eventType, listener, options) {
     eventTarget.addEventListener(eventType, listener, options);
-    this._listeners.push({ eventTarget, eventType, listener });
+    return this._listeners.push({ eventTarget, eventType, listener }) - 1;
```
robertknight, line 27: The first listener ID is returned is 0, which is a falsey value. This could cause problems if calling code tries to do `if (!listenerId) { ... }`. I think it would be safer to make IDs start from 1, like `setTimeout` IDs.

src/shared/listener-collection.js
```diff
@@ -11,24 +11,41 @@
...
+  /**
+   * Removes a listener using a listenerId
+   * @param {number} listenerId
+   */
+  remove(listenerId) {
+    const event = this._listeners[listenerId];
+    if (event) {
+      const { eventTarget, eventType, listener } = event;
       eventTarget.removeEventListener(eventType, listener);
+      this._listeners[listenerId] = null;
```
robertknight, line 39: I see steps have been taken to avoid ID reuse here, which is good. It does mean that you have a growing array of empty slots to loop over in future though. Also `removeAll` will result in ID reuse.

src/shared/listener-collection.js
```diff
@@ -11,24 +11,41 @@
...
+  }
+
+  removeAll() {
+    this._listeners.forEach(event => {
+      if (event) {
+        const { eventTarget, eventType, listener } = event;
+        eventTarget.removeEventListener(eventType, listener);
+      }
     });
     this._listeners = [];
```
robertknight, line 50: If `removeAll` is called and then new listeners are added, listener IDs will be reused.

src/shared/listener-collection.js
```diff
@@ -11,24 +11,41 @@
  */
 export class ListenerCollection {
   constructor() {
-    /** @type {Listener[]} */
+    /** @type {Array<Listener|null>} */
```
robertknight, line 14: There is currently a listener ID reuse hazard in `removeAll` (see comments for that method). Although you could fix that while continuing to use an array, I think it would be more natural to use a map of listener ID to listener, together with an ID counter that doesn't get reset.

robertknight (changes requested):
The proposed API looks good. However I spotted a couple of issues with the way listener IDs are allocated and stored internally. I also had some minor organizational feedback about the tests.

esanzgar (commented):
As you suggested, I have reimplemented the `ListenerCollection` using internally a map. The keys (`listenerId`s) are symbols. Let me know if you find any pitfall in this alternative implementation.

src/shared/listener-collection.js
```diff
@@ -11,25 +11,42 @@
...
    * @param {Listener['eventType']} eventType
    * @param {Listener['listener']} listener
    * @param {AddEventListenerOptions} [options]
    */
   add(eventTarget, eventType, listener, options) {
     eventTarget.addEventListener(eventType, listener, options);
-    this._listeners.push({ eventTarget, eventType, listener });
+    const symbol = Symbol();
+    this._listeners.set(symbol, { eventTarget, eventType, listener });
+    return symbol;
```
robertknight, line 30: Using a symbol works. I guess you could also have made `this._listeners` a set which stores the `{ eventTarget, eventType, listener }` objects and return the object reference as a key.

src/shared/listener-collection.js
```diff
@@ -11,25 +11,42 @@
...
+  }
+
+  /**
+   * Removes a listener using a listenerId
+   * @param {Symbol} listenerId
```
robertknight, line 35:
```suggestion
   * Remove a listener using a listenerId
   *
   * @param {Symbol} listenerId
```

src/shared/test/listener-collection-test.js
```diff
@@ -11,23 +11,41 @@ describe('ListenerCollection', () => {
...
   });
 
-  it('unregisters event listeners', () => {
-    const listener1 = sinon.stub();
-    const listener2 = sinon.stub();
-    listeners.add(window, 'resize', listener1);
-    listeners.add(window, 'resize', listener2);
-    listeners.removeAll();
+  describe('#remove', () => {
+    it('unregisters some event listeners', () => {
```
robertknight, line 25:
```suggestion
    it('unregisters the specified listener', () => {
```

src/shared/listener-collection.js
```diff
@@ -11,25 +11,42 @@
...
 export class ListenerCollection {
   constructor() {
-    /** @type {Listener[]} */
-    this._listeners = [];
+    /** @type {Map<Symbol, Listener>} */
+    this._listeners = new Map();
   }
 
   /**
+   * Adds a listener and returns a listenerId
```
robertknight, line 19:
```suggestion
   * Add a listener and return an ID that can be used to remove it later
```

robertknight (approved):
LGTM, once the earlier branch is merged.

Issue #3915, opened 2021-11-09
Missing annotation cards in the sidebar


esanzgar:
Sometimes annotation cards are missing from the sidebar. The number of annotations in the sidebar and the highlights in the guest page shows always correctly. The missing annotations cards are rendered when scrolling the sidebar. This bug doesn't happens all the time, but quite frequently, so a few reloading the web page where the Hypothesis client is loaded will manifest this issue.

The issue is difficult to see in very small screen.

### Steps to reproduce
1. Reload the web page where the Hypothesis client is loaded  until you see missing annotations cards.
2. Scroll the sidebar
3. Missing annotations appears

### Expected behaviour
All annotation cards should render at the same time.

### Actual behaviour
Missing annotations are rendered only when the sidebar viewport changes.

### Browser/system information
I was able to reproduce this on Chrome, Safari and Firefox.

### Additional details

https://user-images.githubusercontent.com/8555781/140902642-afde223d-1134-4ff5-bccb-ba32fbba5b21.mov

PR #3916, opened 2021-11-09
Fix scroll container initial height measurement

robertknight:
Fix an issue that caused `ThreadList` to sometimes have an incorrect (too small)
internal `scrollContainerHeight` state after the app loaded, resulting in not
enough annotation cards being rendered for the height of the sidebar.

`ThreadList` measures the height of the scroll container after the
initial render in a `useLayoutEffect` callback. It then registers handlers for
window resize events in a `useEffect` callback order to re-measure the height
when it changes.

The sequence of events that lead to the incorrect state was:

 1. The sidebar iframe starts out hidden, by a `display: none` on a container of
    the iframe in the parent frame.
 2. When the `useLayoutEffect` callback runs, the iframe is hidden and so the
    scroll container height is measured as zero.
 3. The iframe is then shown and a `resize` event is fired at the window
 4. The ThreadList's `useEffect` callback runs and registers a handler for the
    window's `resize` event

Because the initial window `resize` event is fired before the listener is
registered, the scroll container height did not get re-measured.

The fix is to combine the `useLayoutEffect` and `useEffect` callbacks from steps
(2) and (4) so that the `resize` listener is registered immediately after the
initial size is measured. Combining the effects is also beneficial from a
readability perspective as it locates all the related state and logic together
in the component.

Fixes https://github.com/hypothesis/client/issues/3915

### DIFF
--- a/src/sidebar/components/ThreadList.js
+++ b/src/sidebar/components/ThreadList.js
@@ -47,21 +47,38 @@ function roundScrollPosition(pos) {
  * @param {ThreadListProps} props
  */
 function ThreadList({ threads }) {
-  // Height of the visible area of the scroll container.
-  const [scrollContainerHeight, setScrollContainerHeight] = useState(
-    window.innerHeight
-  );
+  // Client height of the scroll container.
+  const [scrollContainerHeight, setScrollContainerHeight] = useState(0);
 
-  // Scroll offset of scroll container. This is updated after the scroll
-  // container is scrolled, with debouncing to limit update frequency.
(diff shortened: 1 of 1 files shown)


esanzgar (approved):
This solution fixes the issue and, as pointed, makes all the logic regarding the update of the container size or scrolling in one place.

For future reference, we also try to remove the `display: none` style in the `div.annotator-frame`, which also fixed the issue. However, that caused the `Toolbar` and the `BucketBar` to show in this state, which is not pleasant for the view:

Screenshot 2021-11-09 at 12 26 10

PR #3917, opened 2021-11-11
Remove two unused dependencies

robertknight:
- @actions/core was used when the @hypothesis/frontend-shared package was
  part of the client repo.
- postcss-url was used prior to the introduction of
  @hypothesis/frontend-build


lyzadanger (approved):
Yay! Like!

PR #3929, opened 2021-11-15
Simplify nomenclature

esanzgar:
For PortFinder, I followed this advice:
https://github.com/hypothesis/client/pull/3881#discussion_r743030606,
except that I made the argument of `PorFinder#discover` a string (instead of an object).

```
const portFinder = new PortFinder({ source: 'guest', hostFrame });
portFinder.discover('sidebar');
portFinder.discover('host');
```

For PortProvider, I followed this advice:
https://github.com/hypothesis/client/pull/3881#discussion_r743033013,
except that I used a getter.

```
cont portProvider = new PortProvider(...)
const bridge = new Bridge();
bridge.createChannel(portProvider.sidebarPort);
```

I have renamed the properties of `Message`:
- `source` becomes `authority`
- `channel` and `port` have been replaced by `frame1` and `frame2`

I did that to align `port1` and `frame1` and `port2` and `frame2`, while
also avoiding clashing with other names (`source`, `target`) which have
other meaning in the `window.postMessage` context.

I have removed a level of nesting in the `PortProvider#discover` that
make the method more readable.

I added some other suggestions from PR #3881.

### DIFF
--- a/src/shared/port-finder.js
+++ b/src/shared/port-finder.js
@@ -7,8 +7,7 @@ const POLLING_INTERVAL_FOR_PORT = 250;
 /**
  * @typedef {import('../types/annotator').Destroyable} Destroyable
  * @typedef {import('./port-util').Message} Message
- * @typedef {Message['channel']} Channel
- * @typedef {Message['port']} Port
+ * @typedef {import('./port-util').Frame} Frame
  */
 
 /**
@@ -16,14 +15,18 @@ const POLLING_INTERVAL_FOR_PORT = 250;
  * MessagePort-based connection to other frames. It is used together with
  * PortProvider which runs in the host frame. See PortProvider for an overview.
  *
- * Channel nomenclature is `[frame1]-[frame2]` so that:
- *   - `port1` should be owned by/transferred to `frame1`, and
- *   - `port2` should be owned by/transferred to `frame2`
- *
  * @implements Destroyable
  */
 export class PortFinder {
-  constructor() {
+  /**
+   * @param {object} options
+   *   @param {Exclude<Frame, 'host'>} options.source - the role of this frame
+   *   @param {Window} options.hostFrame - the frame where the `PortProvider` is
+   *     listening for messages.
+   */
+  constructor({ hostFrame, source }) {
+    this._hostFrame = hostFrame;
+    this._source = source;
     this._listeners = new ListenerCollection();
   }
 
@@ -34,20 +37,16 @@ export class PortFinder {
   /**
    * Request a specific port from `hostFrame`
    *
-   * @param {object} options
-   *   @param {Channel} options.channel - requested channel
-   *   @param {Window} options.hostFrame - frame where the hypothesis client is
-   *     loaded and `PortProvider` is listening for messages
-   *   @param {Port} options.port - requested port
+   * @param {Frame} target - the frame aiming to be discovered
    * @return {Promise<MessagePort>}
    */
-  async discover({ channel, hostFrame, port }) {
+  async discover(target) {
     let isValidRequest = false;
     if (
-      (channel === 'guest-host' && port === 'guest') ||
-      (channel === 'guest-sidebar' && port === 'guest') ||
-      (channel === 'host-sidebar' && port === 'sidebar') ||
-      (channel === 'notebook-sidebar' && port === 'notebook')
+      (this._source === 'guest' && target === 'host') ||
+      (this._source === 'guest' && target === 'sidebar') ||
+      (this._source === 'sidebar' && target === 'host') ||
+      (this._source === 'notebook' && target === 'sidebar')
     ) {
       isValidRequest = true;
     }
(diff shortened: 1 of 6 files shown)


src/shared/port-finder.js
```diff
@@ -34,20 +37,16 @@ export class PortFinder {
   /**
    * Request a specific port from `hostFrame`
```
robertknight, line 38:
```suggestion
   * Request a specific port from the host frame
```

`hostFrame` is no longer an argument to this method.

src/shared/port-finder.js
```diff
@@ -57,36 +56,44 @@ export class PortFinder {
...
-      // Therefore, for the `PortFinder`, we implement a polling strategy (sending a
-      // message every N milliseconds) until a response is received.
+      // Because `guest` iframes load in parallel to the `host` frame, we can
+      // not assume that the code in the `host` frame is executed before the
+      // code in a `guest` frame. Hence, we can't assume that `PortProvider` (in
+      // the `host` frame) is initialized before `PortFinder` (in the non-host
+      // frames). Therefore, for the `PortFinder`, we implement a polling
+      // strategy (sending a message every N milliseconds) until a response is
+      // received.
       const intervalId = setInterval(postRequest, POLLING_INTERVAL_FOR_PORT);
```
robertknight, line 78: If `destroy` is called while a `discover` operation is in progress, this interval doesn't get cleared and polling will continue.

src/shared/port-finder.js
```diff
@@ -57,36 +56,44 @@ export class PortFinder {
...
       }, MAX_WAIT_FOR_PORT);
 
       const listenerId = this._listeners.add(window, 'message', event => {
         const { data, ports } = /** @type {MessageEvent} */ (event);
         if (
           isMessageEqual(data, {
-            channel,
-            port,
-            source: 'hypothesis',
+            authority: 'hypothesis',
```
robertknight, line 94: I presume the change from `source` => `authority` was to avoid confusion with the concept of a source frame? Perhaps `protocol` might be a more typical choice here to indicate the format/purpose of the messages.

src/shared/port-provider.js
```diff
@@ -86,67 +80,114 @@ export class PortProvider {
...
+   * @param {MessageEventSource|null} source
+   * @return {source is Window}
    */
-  _isValidRequest(event, allowedMessage, allowedOrigin) {
-    const { data, origin, source } = event;
-    if (allowedOrigin !== '*' && origin !== allowedOrigin) {
-      return false;
-    }
-
+  _isSourceWindow(source) {
```
robertknight, line 128: This doesn't need to be a member method any more. It could be a helper.

src/shared/port-provider.js
```diff
@@ -86,67 +80,114 @@ export class PortProvider {
...
     if (
       // `source` can be of type Window | MessagePort | ServiceWorker.
       // The simple check `source instanceof Window`` doesn't work here.
       // Alternatively, `source` could be casted `/** @type{Window} */ (source)`
-      !source ||
+      source === null ||
       source instanceof MessagePort ||
       source instanceof ServiceWorker
```
robertknight, line 135: This currently covers all the kinds of `MessageEventSource` that exist. In future new ones might be added. Since we check the content of the message as well as the source before processing it, I don't think that should be a problem though.

src/shared/port-provider.js
```diff
@@ -86,67 +80,114 @@ export class PortProvider {
...
+   *   @param {MessagePort} options.port1 - the port to be sent.
+   *   @param {MessagePort} [options.port2] - if a counterpart port is provided,
+   *     send this port either, (1) to the `sidebar` frame using the `sidebar-host`
+   *     channel or (2) through the `onHostPortRequest` event listener.
    */
-  _sendPort(event, message, port, counterpartPort) {
-    const source = /** @type {Window} */ (event.source);
-
-    source.postMessage(message, event.origin, [port]);
+  _sendPort({ channel, message, origin, source, port1, port2 }) {
```
robertknight, line 178: Perhaps rename to `_sendPorts` since we will sometimes be sending two ports.

src/shared/port-util.js
```diff
@@ -3,20 +3,19 @@
...
- * @typedef {'guest-host'|'guest-sidebar'|'host-sidebar'|'notebook-sidebar'} Channel
- * @typedef {'guest'|'host'|'notebook'|'sidebar'} Port
+ * @typedef {'guest'|'host'|'notebook'|'sidebar'} Frame
  *
  * @typedef Message
- * @prop {Channel} channel
- * @prop {Port} port
- * @prop {'offer'|'request'}  type
- * @prop {SOURCE} source -
+ * @prop {AUTHORITY} authority
```
robertknight, line 15: As noted in another comment, I understand what `authority` is for, but it probably isn't the term I would choose for this. `protocol` seems the most obvious term to me. Another convention I'm aware of for specifying the protocol is to use a protocol-specific identifier with a dummy value (eg. JSON-RPC messages have `jsonrpc: "2.0"` in them).

robertknight (approved):
LGTM. I left a few minor comments, one of which was a bikeshedding comment about the `authority` field. I also see some timeouts that `PortFinder.destroy` doesn't clean up. Since that is not a new issue it might be best to handle that in a follow-up PR.

esanzgar (commented):
I have incorporated your suggestions in the last commit.

I will defer to a different PR the improvement related to:
* the `destroy` method called while `discover` operation is taking place
* the need of `authority` field in the message

PR #3930, opened 2021-11-15
Add integration test to check inter-frame communication

esanzgar:
This new integration test follows the line of the `rpc-bridge-test.js`
but adds `PortProvider` and `PortFinder` to the mix.

These three classes, `PortProvider`, `PortFinder` and `Bridge` must work
in harmony. A change in one could lead to the others no working as
planned.

### DIFF
--- a/src/shared/test/integration/inter-frame-communication-test.js
+++ b/src/shared/test/integration/inter-frame-communication-test.js
@@ -1,0 +1,241 @@
+import { delay } from '../../../test-util/wait';
+import { Bridge as BridgeOri } from '../../bridge';
+import { ListenerCollection as ListenerCollectionOri } from '../../listener-collection';
+import { PortFinder as PortFinderOri } from '../../port-finder';
+import { PortProvider as PortProviderOri } from '../../port-provider';
+import { isMessageEqual } from '../../port-util';
+
+describe('PortProvider-PortFinder-Bridge integration', () => {
+  let destroyables;
+
+  // These proxy classes are to keep track of instances that need to be destroyed.
+  class Bridge extends BridgeOri {
+    constructor() {
+      super();
+      destroyables.push(this);
+    }
+  }
+
+  class ListenerCollection extends ListenerCollectionOri {
+    constructor() {
+      super();
+      destroyables.push(this);
+    }
+    destroy() {
+      this.removeAll();
+    }
+  }
+
+  class PortFinder extends PortFinderOri {
+    constructor(...args) {
+      super(...args);
+      destroyables.push(this);
+    }
+  }
+
+  class PortProvider extends PortProviderOri {
+    constructor(...args) {
+      super(...args);
+      destroyables.push(this);
+    }
+  }
+
+  beforeEach(() => {
+    destroyables = [];
+  });
+
+  afterEach(() => {
+    destroyables.forEach(instance => instance.destroy());
+  });
+
+  it('enables the communication between sidebar-host and notebook-sidebar frames', async () => {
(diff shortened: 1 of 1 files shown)


src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,243 @@
...
+      destroyables.push(this);
+    }
+  }
+
+  class ListenerCollection extends ListenerCollectionOri {
+    constructor() {
+      super();
+      destroyables.push(this);
+    }
+    destroy() {
```
esanzgar, line 24: Makes me think if `ListenerCollection#removeAll` should be renamed `ListenerCollection#destroy` and tag the class as destroyable.

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,243 @@
...
+    const isNotebookSidebarEnable = new Promise(
+      resolve => (notebookSidebarEnable = resolve)
+    );
+
+    // register RPC method *before* creating the channel.
+    sidebarBridge1.on('ping', async () => {
+      await isNotebookSidebarEnable;
+      sidebarBridge2.call('ping');
+    });
+    sidebarBridge2.on('pong', () => sidebarBridge1.call('pong'));
```
esanzgar, line 95: One way to avoid to instantiate two `Bridge`s, could be to allow the bridge to forward messages to specific frames, in this way:
```suggestion
    const sidebarBridge = new Bridge();

    // The following promise is to prevent sending a message to the notebook
    // before the notebook-sidebar (sidebarBridge2) channel is created.
    let notebookSidebarEnable;
    const isNotebookSidebarEnable = new Promise(
      resolve => (notebookSidebarEnable = resolve)
    );

    // register RPC method *before* creating the channel.
    sidebarBridge.on('ping', async () => {
      await isNotebookSidebarEnable;
      sidebarBridge.call('ping', {to: 'notebook'});
    });
    sidebarBridge2.on('pong', () => sidebarBridge.call('pong', {to: 'host'}));
```

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,241 @@
+import { delay } from '../../../test-util/wait';
+import { Bridge as BridgeOri } from '../../bridge';
```
robertknight, line 2: I presume `Ori` is short for `Original` here? We don't have a lot of precedents for this, but appending an underscore (`Bridge_`) or using an `Orig` prefix (`OrigBridge`) are more common approaches in our code.

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,241 @@
...
+
+    // Test this fictional message path:
+    // host -- ping --> sidebar -- ping --> notebook
+    // host <- pong --- sidebar <- pong --- notebook
+
+    // host frame
+    const portProvider = new PortProvider(window.location.origin);
+    const hostBridge = new Bridge();
+
+    // register RPC method *before* creating the channel
```
robertknight, line 63:
```suggestion
    // Register RPC method *before* creating the channel
```

Same for other comments.

robertknight (changes requested):
The two tests that are added here test a lot of different scenarios. I think it will make future understanding and debugging easier to have smaller tests for each of the main channels, with an extended tests at the end if there are important scenarios that can't be covered this way.

esanzgar:
I will simplify these tests.

PR #3934, opened 2021-11-16
Rename bridge to contain the name of the target frame

esanzgar:
In a future PR (#3904), `Guest` will contain two bridges (like `FrameSync`).
Naming them by the destination frame will help identify the purpose of
it.

The agreed name for this `Bridge`s is to contain the word `RPC` to
indicate that it is a remote (inter-frame) procedure.

### DIFF
--- a/src/annotator/annotation-sync.js
+++ b/src/annotator/annotation-sync.js
@@ -28,11 +28,11 @@
 export class AnnotationSync {
   /**
    * @param {EventBus} eventBus - Event bus for communicating with the annotator code (eg. the Guest)
-   * @param {SidebarBridge} bridge - Channel for communicating with the sidebar
+   * @param {SidebarBridge} sidebarBridge - Channel for communicating with the sidebar
    */
-  constructor(eventBus, bridge) {
+  constructor(eventBus, sidebarBridge) {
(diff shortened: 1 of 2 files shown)


src/annotator/annotation-sync.js
```diff
@@ -28,11 +28,11 @@
 export class AnnotationSync {
   /**
    * @param {EventBus} eventBus - Event bus for communicating with the annotator code (eg. the Guest)
-   * @param {SidebarBridge} bridge - Channel for communicating with the sidebar
+   * @param {SidebarBridge} sidebarBridge - Channel for communicating with the sidebar
```
robertknight, line 31: We might want to rename the `Bridge` class in future to better align it with the `<frameRole>RPC` field names and `PortRPC` class that exists. Or perhaps even combine it with `PortRPC`, maybe only for certain uses? The main distinction with `PortRPC` in the current code is that a `Bridge` can be connected to zero or many ports, whereas a `PortRPC` is always connected to exactly one.

robertknight (approved):

src/annotator/annotation-sync.js
esanzgar: It makes sense.

PR #3935, opened 2021-11-16
Type and private properties methods

esanzgar:
The first commit it narrows the type of `Guest` used in the `Sidebar`. See the commit description for more details.

The second commit is very minor, it rearranges a method and a few fields. In addition, it prepends two fields with underscore to indicate that the are meant to be private.

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -67,7 +67,10 @@ export default class Sidebar {
    */
   constructor(element, eventBus, guest, config = {}) {
     this._emitter = eventBus.createEmitter();
-    const hypothesisAppsOrigin = new URL(config.sidebarAppUrl).origin;
+    this._guest = guest;
+    this._config = config;
+
+    const hypothesisAppsOrigin = new URL(this._config.sidebarAppUrl).origin;
     this._portProvider = new PortProvider(hypothesisAppsOrigin);
 
     /**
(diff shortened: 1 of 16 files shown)


src/annotator/sidebar.js
```diff
@@ -215,19 +215,6 @@ export default class Sidebar {
     });
   }
 
-  destroy() {
```
esanzgar, line 216: I prefer to have the private methods on top and the public ones a the bottom.

src/annotator/sidebar.js
```diff
@@ -18,6 +18,8 @@ import { createShadowRoot } from './util/shadow-root';
  * @typedef {import('../types/bridge-events').SidebarToHostEvent} SidebarToHostEvent
  * @typedef {import('../types/annotator').SidebarLayout} SidebarLayout
  * @typedef {import('../types/annotator').Destroyable} Destroyable
+ * @typedef {'contentContainer'|'createAnnotation'|'fitSideBySide'} DependenciesSidebar
```
robertknight, line 21: It isn't immediately obvious when looking at the code in this context what these fields relate to. I think it would be clearer with something like:

```js
/** @typedef {Pick<Guest, ...fields>} SidebarGuest */
/** @typedef {Pick<Guest, ...fields>} BucketBarGuest */

/**
  * @param {SidebarGuest & BucketBarGuest} guest
  */
constructor(...) {
}
```

That said, this kind of thing does add extra complexity to the types, so it is important to have a clear reason for doing this.

In this case we are going to need to make a larger structural change to the code soon to support host frames which are not guest frames. Do you already have a clear plan of how that will relate to these changes? If not I'd hold off on them until that problem is solved.

src/annotator/sidebar.js
robertknight: This isn't a coding convention we have as a team though. If anything our existing code leans more towards the opposite approach, which has the advantage that the public interface consumers need to know about is near the top. We can discuss this together at some point, but now is not the best time as I think it will be a distraction from functional improvements. Can you leave this method where it is for the moment.

src/annotator/sidebar.js
```diff
@@ -57,15 +59,18 @@ export default class Sidebar {
...
    *   The `Guest` instance for the current frame. It is currently assumed that
    *   it is always possible to annotate in the frame where the sidebar is
    *   displayed.
    * @param {Record<string, any>} [config]
    */
   constructor(element, eventBus, guest, config = {}) {
     this._emitter = eventBus.createEmitter();
-    const hypothesisAppsOrigin = new URL(config.sidebarAppUrl).origin;
+    this._guest = guest;
+    this._config = config;
```
robertknight, line 71: Rather than saving the whole config object as a field we could just save the data that other methods actually use, which looks like just `_showHighlights`.

robertknight (changes requested):
I had some feedback on the Guest subsets and a query about whether it makes sense to do that given upcoming changes.

Regarding moving the `destroy` method, I would prefer not to do that, at least not for the reason given. 

The change to make the `guest` field private and remove the `options` fields makes sense, although we can replace `options` with a simpler field that just has the data used by non-constructor methods.

esanzgar:
There is consensus on making fields private, but not in the ordering of private and public method.

The type added is really to itemise the dependencies of `Guest` class on the `Sidebar` and the `BucketBar` classes. I envision been able to decouple `Guest` from `Sidebar` because of a new `guest-host` channel. If that's possible, then the type will be removed.

PR #3937, opened 2021-11-17
Add stricter types for the discovery `Message`

esanzgar:
More stricter types allows to reduce errors during development and
better ergonomics as IDE are able to add accurate autocomplete. Before,
`PortProvider` allowed to listen for all short of frame combinations,
while `PortFinder#discover` allowed to all type of frames as argument.
Now, the set of these possibilities is reduced to only valid
combinations.

In `PortFinder#discover`, the addition of types, makes the check of
valid combinations of frames unnecessary.

The price to pay for these stricter types is a bit more verbose types,
and in the case of `PortFinder#discover` the use of conditional types.

Example in `PortFinder`:

https://user-images.githubusercontent.com/8555781/142216237-16c5bf25-cb6e-4578-b4f9-4be6612740a0.mov

Example in `PortProvider`:

https://user-images.githubusercontent.com/8555781/142216284-e7366305-3268-4ef3-bce8-0576c47d825c.mov

### DIFF
--- a/src/shared/port-finder.js
+++ b/src/shared/port-finder.js
@@ -7,20 +7,20 @@ const POLLING_INTERVAL_FOR_PORT = 250;
 /**
  * @typedef {import('../types/annotator').Destroyable} Destroyable
  * @typedef {import('./port-util').Message} Message
- * @typedef {import('./port-util').Frame} Frame
+ * @typedef {'guest'|'notebook'|'sidebar'} SourceFrame
  */
 
 /**
  * PortFinder is used by non-host frames in the client to establish a
  * MessagePort-based connection to other frames. It is used together with
  * PortProvider which runs in the host frame. See PortProvider for an overview.
- *
+ * @template {SourceFrame} T
  * @implements Destroyable
  */
 export class PortFinder {
   /**
    * @param {object} options
-   *   @param {Exclude<Frame, 'host'>} options.source - the role of this frame
+   *   @param {T} options.source - the role of this frame
    *   @param {Window} options.hostFrame - the frame where the `PortProvider` is
    *     listening for messages.
    */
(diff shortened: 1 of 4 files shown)


src/shared/port-finder.js
```diff
@@ -89,11 +78,14 @@ export class PortFinder {
...
         const { data, ports } = /** @type {MessageEvent} */ (event);
         if (
-          isMessageEqual(data, {
-            frame1: this._source,
-            frame2: target,
-            type: 'offer',
-          })
+          isMessageEqual(
+            data,
+            /** @type {Message} */ ({
```
esanzgar, line 83: It is a pity that TS is not able to understand that the combination of `frame1` and `frame2` is a valid combination in `Message`.

src/shared/port-finder.js
```diff
@@ -37,30 +37,19 @@ export class PortFinder {
...
-      (this._source === 'sidebar' && target === 'host') ||
-      (this._source === 'notebook' && target === 'sidebar')
-    ) {
-      isValidRequest = true;
-    }
-
-    if (!isValidRequest) {
-      throw new Error('Invalid request of channel/port');
-    }
-
```
esanzgar, line 57: This check is unnecessary because of the stricter types.

robertknight (changes requested):
I don't think we should make the changes to the `Message` type proposed here. Unfortunately the error messages that TS currently produces when trying to construct a `Message` are pretty difficult to follow, even for me as someone who knows the language pretty well. We know that it is fine to use a cast, but this might not be obvious to someone else. In contrast the equivalent runtime code for checking compatibility of source and destination was easy to follow. I also think the previous `Message` type was easier for new developers to look at and understand.

The change to parametrize `PortFinder` is OK, though I noted that generic parameter names (`T`) should only be used when the meaning is obvious.

esanzgar:
This doesn't seem to bring very obvious benefits.

PR #3939, opened 2021-11-18
Remove `PDFViewerApplication` global field after test

esanzgar:
The tests crashed if either of these commands was executed:

- `yarn test --grep anchoring`, or
- `yarn test --grep 'anchoring/test/pdf|test/integration/anchoring'`

`src/anotator/anchoring/test/pdf-test.js` cleanup the PDF viewer by setting
the `window.PDFViewerApplication` to `null`. After that, if another test
creates a `Guest` instance (like
`src/annotator/test/integration/anchoring-test.js`) it will think that the
document is a PDF and will crash. This is the code that checks if the
document is a PDF:

https://github.com/hypothesis/client/blob/492ab6d56a5685356ca6b3a48644e047bb28dca0/src/annotator/integrations/pdf.js#L53-L56

The solution is either deleting `window.PDFViewerApplication' or setting
it to `undefined`.

### DIFF
--- a/src/annotator/anchoring/test/pdf-test.js
+++ b/src/annotator/anchoring/test/pdf-test.js
@@ -75,7 +75,7 @@ describe('annotator/anchoring/pdf', () => {
   /** Clean up any resources created by the fake PDF.js viewer. */
   function cleanupViewer() {
     viewer?.dispose();
-    window.PDFViewerApplication = null;
+    delete window.PDFViewerApplication;
   }
 
   beforeEach(() => {


src/annotator/anchoring/test/pdf-test.js
```diff
@@ -85,7 +85,6 @@ describe('annotator/anchoring/pdf', () => {
   });
 
   afterEach(() => {
-    pdfAnchoring.$imports.$restore();
```
esanzgar, line 88: I was not sure about this change. I don't see any imports being mocked, so I thought it is not needed.

robertknight: There is a call to `$imports.$mock` in one specific test currently. While it would be possible to use `try ... finally` in that test, I'd consider it idiomatic to have a top-level `$imports.$restore` call if any test makes use of it.

robertknight (changes requested):
The `window.PDFViewerApplication` change looks fine. The `$imports.$restore` code that was removed is still needed though. Feel free to merge just the `window.PDFViewerApplication` part.

PR #3940, opened 2021-11-18
Use `WeakMap` to store `MessageChannel`

esanzgar:
We use `Window` objects as keys on a map to store `MessageChannel`s.
Windows can be removed. `WeakMap` has the characteristic that keys
that are claimed by the garbage collector are removed from the map.
In this way, we are memory friendly and don't retain any defunct object.

Here is an example:

https://user-images.githubusercontent.com/8555781/142408512-2c0b2218-8b6f-4595-885c-ff89e6976d96.mov


robertknight (approved):
LGTM

PR #3951, opened 2021-11-22
Revert "Adoption of `PortProvider` and `PortFinder`"

esanzgar:
Reverts hypothesis/client#3903


robertknight:
Note for posterity - we found an issue that the client stopped working on sites served over plain HTTP, rather than HTTPS, due to an `object instanceof ServiceWorker` check in the merged code. `ServiceWorker` is only available in secure contexts.

There might be an unrelated issue as well. Sentry report: https://sentry.io/organizations/hypothesis/issues/2807403156/events/5084d2ccce394417a5e4abf4e79dd259/?project=69811&query=is%3Aunresolved.

PR #3952, opened 2021-11-22
Adopt new PortProvider and PortFinder

esanzgar:
This adoption simplifies the discovery of frames.

I did a little bit of clean-up in the tests: rearranged a few variables
alphabetically, and delete unneeded lines and variables.

### DIFF
--- a/src/shared/port-util.js
+++ b/src/shared/port-util.js
@@ -60,12 +60,12 @@ export function isMessageEqual(data, message) {
  */
 export function isSourceWindow(source) {
   if (
-    // `source` can be of type Window | MessagePort | ServiceWorker.
-    // The simple check `source instanceof Window`` doesn't work here.
-    // Alternatively, `source` could be casted `/** @type{Window} */ (source)`
+    // `source` can be of type Window, MessagePort, ServiceWorker, or null.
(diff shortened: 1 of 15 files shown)


src/shared/port-util.js
```diff
@@ -60,12 +60,12 @@ export function isMessageEqual(data, message) {
...
 export function isSourceWindow(source) {
   if (
-    // `source` can be of type Window | MessagePort | ServiceWorker.
-    // The simple check `source instanceof Window`` doesn't work here.
-    // Alternatively, `source` could be casted `/** @type{Window} */ (source)`
+    // `source` can be of type Window, MessagePort, ServiceWorker, or null.
+    // `source instanceof Window` doesn't work in Chrome if `source` is a
+    // cross-origin window.
     source === null ||
-    source instanceof MessagePort ||
```
robertknight, line 67: If MessagePort doesn't exist then the entire application is going to fail, so I don't think it makes sense to check for that. Also, we've already been using MessagePort for months!

esanzgar: OK, I will remove it.

robertknight (approved):
LGTM

PR #3953, opened 2021-11-23
Bump approx-string-match to v2.0.0

robertknight:
This includes a fix for incorrect results under certain conditions when the pattern length is an exact multiple of 32. See https://github.com/robertknight/approx-string-match-js/pull/13.

This could lead to occasional failures to find a fuzzy match for long quotes, with the annotations showing up in the orphans tab when instead they should have anchored to a "close-enough" match. I don't have a manual test case for the client, but see the linked PR.

The package has also been converted to a native ES module, helping with the overall progress of ESM-ification of our dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@rollup/plugin-replace": "^3.0.0",
     "@rollup/plugin-virtual": "^2.0.3",
     "@sentry/browser": "^6.0.2",
-    "approx-string-match": "^1.1.0",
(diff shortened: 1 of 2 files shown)


robertknight:
I'm going to merge this PR to test an npm authentication change in the publishing workflow. [See thread](https://hypothes-is.slack.com/archives/C1M8NH76X/p1637663116206300).

PR #3955, opened 2021-11-23
Remove unused babel-plugin-transform-async-to-promises dependency

robertknight:
This dependency used to be used to enable the boot script to run in IE
11 and other old browsers. It is no longer used however.
See https://github.com/hypothesis/client/pull/3944#issuecomment-976289004.


robertknight:
It turns out I was mistaken about us not using this plugin. The boot script is configured to use it by `src/boot/.babelrc` rather than by configuration in rollup-boot.config.mjs.

PR #3959, opened 2021-11-24
Simplify PortProvider request handling

robertknight:
This PR simplifies some PortProvider internals to try to make the overall sequence of steps involved in handling a port request easier to follow.

- The data structure that keeps track of already handled requests has been simplified. It used to be a `Map<Channel, WeakMap<Window, MessageChannel>>`, where `Window` was the source window. The WeakMap however was effectively used as a set. Only the presence of an entry affected future behavior, not the value. It is now a `WeakMap<Window, Set<Channel>>` that maps from source window to channels requested by that source.

- The code paths for the sidebar-host channel and other channels have been unified. As a result the `_sendPorts` helper was only called in one place and it was possible to inline it to avoid indirection tax. In the process an obsolete reference to `onHostPortRequest` was removed.

- I removed an internal comment about a couple of aspects of MessagePort behavior, as it was not directly relevant for the surrounding code.

The external behavior/API is unchanged, so there are no changes to the tests.

### DIFF
--- a/src/shared/port-provider.js
+++ b/src/shared/port-provider.js
@@ -138,37 +138,6 @@ export class PortProvider {
   }
 
   /**
-   * Send a message and a port to the corresponding destinations.
-   *
-   * @param {object} options
-   *   @param {Channel} options.channel - communication channel enabled by this
-   *     port.
-   *   @param {Message} options.message - the message to be sent.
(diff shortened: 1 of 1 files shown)


esanzgar (commented):
Thanks Robert, I was thinking similar simplification.

esanzgar (approved):

PR #3960, opened 2021-11-24
Make shadow DOM support mandatory

robertknight:
All of our target browsers now support shadow DOM [1]. Make it mandatory so it
is clear that all annotator UI elements can rely on Shadow DOM and thus don't
need to worry about their internal styles affecting or being affected by the
host page.

[1] https://caniuse.com/?search=shadow%20dom

### DIFF
--- a/src/annotator/util/shadow-root.js
+++ b/src/annotator/util/shadow-root.js
@@ -22,20 +22,10 @@ function loadStyles(shadowRoot) {
  * Create the shadow root for an annotator UI component and load the annotator
  * CSS styles into it.
  *
- * In browsers that support it, shadow DOM is used to isolate annotator UI
- * components from the host page's styles.
- *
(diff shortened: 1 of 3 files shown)


src/annotator/util/shadow-root.js
```diff
@@ -22,20 +22,10 @@ function loadStyles(shadowRoot) {
...
- *   root.
+ * @return {ShadowRoot}
  */
 export function createShadowRoot(container) {
-  if (!container.attachShadow) {
-    stopEventPropagation(container);
-    return container;
-  }
-
   const shadowRoot = container.attachShadow({ mode: 'open' });
```
esanzgar, line 29: Unrelated with this PR, but I have tried `mode: 'closed'` and the applications seems to work well too.

esanzgar (approved):

PR #3962, opened 2021-11-24
Remove channel variable and replace it by the target frame

esanzgar:
Before this PR, we created a variable called `channel` that was the combination
between source frame and target frame: `${frame1}-${frame2}`. This was
used to track if a channel was already sent `this._sentChannels`.

`frame1` acts as a synonym of the source window, hence it is
redundant and strictly none necessary.

I propose in this PR to remove the `channel` variable and store only
the target frame in `this._sentChannels`.

* A `Guest` window will be able to target: `host` and `sidebar
`
* A `Notebook` window will be able to target: `host` and `sidebar`

* A `Sidebar` window will be able to target: `host`

### DIFF
--- a/src/shared/port-provider.js
+++ b/src/shared/port-provider.js
@@ -7,7 +7,6 @@ import { isMessageEqual, isSourceWindow } from './port-util';
  * @typedef {import('../types/annotator').Destroyable} Destroyable
  * @typedef {import('./port-util').Message} Message
  * @typedef {import('./port-util').Frame} Frame
- * @typedef {'guest-host'|'guest-sidebar'|'notebook-sidebar'|'sidebar-host'} Channel
  */
 
 /**
@@ -74,7 +73,7 @@ export class PortProvider {
      * Guest frames in particular may send duplicate requests (see comments in
      * PortFinder).
      *
-     * @type {WeakMap<Window, Set<Channel>>}
+     * @type {WeakMap<Window, Set<Frame>>}
      */
     this._sentChannels = new Map();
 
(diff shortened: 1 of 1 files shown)


robertknight:
> frame1 acts as a synonym of the source window, hence it is redundant and strictly none necessary.

This is true as long as a frame can only have one role. In the case of the host window however it can function as both a guest and a host. This is currently not a problem, as all `host-$OTHER` channels are established via a means other than using `PortFinder`, eg. using `hostPortFor(...)`.

src/shared/port-provider.js
```diff
@@ -74,7 +73,7 @@ export class PortProvider {
      * Guest frames in particular may send duplicate requests (see comments in
      * PortFinder).
      *
-     * @type {WeakMap<Window, Set<Channel>>}
+     * @type {WeakMap<Window, Set<Frame>>}
```
robertknight, line 76: The comment for this field needs an update. It currently reads "Map of source frame to channels". It would need to be something like "Map of source frame to destination frame role" or similar.

robertknight (changes requested):
It is the case at present that each frame only adopts a single role when connecting to a `PortProvider`, and thus we don't have a use case for creating multiple `PortFinder`s with different source roles in the same frame. This is not a restriction that is obviously implied by the API though.

I think we need to either:

1. Keep the code as it was previously OR
2. Indicate the limitation/assumption that a given window can only act as one role in appropriate places (eg. the `this._sentChannels` docs)

Personally I would tend towards (1) since the channel concept only adds minimal internal complexity. I'll leave you to make a call on this.

esanzgar:
> Personally I would tend towards (1) since the channel concept only adds minimal internal complexity. I'll leave you to make a call on this.

I agree.

PR #3963, opened 2021-11-25
Remove `PortProvider#hostPortFrom` method

esanzgar:
There are two methods to get the host port from `PortProvider`:

- `#hostPortFrom` which returns the host port exclusively from the
  `sidebar-host` channel.

- `#on` allows to register a listener that is fired upon requests
  of host frames for channels other than the `sidebar-host`.

This is an attempt to simplify the API of `PortProvider` and merge both
methods into one.

Advantages:

- simplifies the `Sidebar` constructor signature (no need to pass the
  host port)

- avoids the pitfall of creating host-sidebar RPC channel before
  registering the RPC methods

- prepares for future guest-host channel see https://github.com/hypothesis/client/blob/ad93317debdac0d2c3abec83eb9d690934632ef3/src/annotator/index.js#L64-L66 (where the `#on` method is use to create the guest-host channel).

### DIFF
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -54,14 +54,12 @@ function init() {
     const hypothesisAppsOrigin = new URL(sidebarConfig.sidebarAppUrl).origin;
     portProvider = new PortProvider(hypothesisAppsOrigin);
 
-    sidebar = new Sidebar(
-      document.body,
-      eventBus,
-      portProvider.hostPortFor('sidebar'),
-      guest,
-      sidebarConfig
-    );
+    sidebar = new Sidebar(document.body, eventBus, guest, sidebarConfig);
     notebook = new Notebook(document.body, eventBus, getConfig('notebook'));
+
+    portProvider.on('frameConnected', (source, port) =>
+      sidebar.onFrameConnected(source, port)
+    );
 
     portProvider.listen();
   }
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -56,18 +56,14 @@ export default class Sidebar {
    * @param {HTMLElement} element
    * @param {import('./util/emitter').EventBus} eventBus -
    *   Enables communication between components sharing the same eventBus
-   * @param {MessagePort} hostPort -
-   *   Host port for the host-sidebar communication channel. The sidebar app will
-   *   request its side of the channel when it starts.
    * @param {Guest} guest -
    *   The `Guest` instance for the current frame. It is currently assumed that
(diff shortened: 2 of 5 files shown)


src/shared/port-provider.js
```diff
@@ -135,24 +135,14 @@ export class PortProvider {
 
   /**
    * @param {'frameConnected'} eventName
-   * @param {(source: 'guest', port: MessagePort) => void} handler - this handler
+   * @param {(source: 'guest'|'sidebar', port: MessagePort) => void} handler - this handler
    *   fires when a request for the host frame has been granted. Currently, only
    *   the 'guest-host' channel triggers this listener.
```
robertknight, line 140: The second sentence here ("Currently, only the guest-host...") is out of date. I suggest to just remove it, as the type signature makes it obvious what the source frames may be.

src/annotator/index.js
```diff
@@ -54,15 +54,13 @@ function init() {
...
     portProvider = new PortProvider(hypothesisAppsOrigin);
 
-    sidebar = new Sidebar(
-      document.body,
-      eventBus,
-      portProvider.hostPortFor('sidebar'),
-      guest,
-      sidebarConfig
-    );
+    sidebar = new Sidebar(document.body, eventBus, guest, sidebarConfig);
```
robertknight, line 57: An advantage of the previous API is that the caller of `Sidebar` couldn't forget to supply the sidebar app connection. On the other hand, we need something like `onFrameConnected` for guest frames later.

src/annotator/sidebar.js
```diff
@@ -225,6 +218,18 @@ export default class Sidebar {
     this._emitter.destroy();
   }
 
+  /**
+   * Establish RPC channels from specific source frames
```
robertknight, line 222:
```suggestion
   * Setup communication with a frame that has connected to the host.
```

I reworded this a little to try and describe logically what it does from the caller's point of view, rather than how it is implemented.

robertknight (approved):
LGTM with a couple of requested changes to the docs.

src/annotator/index.js
esanzgar: Yes, I agree it had some advantages, but made it a little asymmetric.

src/annotator/sidebar.js
esanzgar: Thanks

src/shared/port-provider.js
esanzgar: 👍🏼

esanzgar (commented):
Thanks!

PR #3964, opened 2021-11-25
Refactor `FrameSync`

esanzgar:
1. I converted `_setupSyncToGuest` and `_setupSyncFromGuests` into
   conventional class methods.

2. Added a new method, `_setupHostEvents` for sidebar-host
   communication.

3. Invoke these three methods above, from the constructor.

### DIFF
--- a/src/sidebar/services/frame-sync.js
+++ b/src/sidebar/services/frame-sync.js
@@ -84,7 +84,11 @@ export class FrameSyncService {
      */
     this._guestRPC = new Bridge();
 
-    // Set of tags of annotations that are currently loaded into the frame
+    /**
+     * Tags of annotations that are currently loaded into guest frames.
+     *
+     * @type {Set<string>}
+     */
     this._inFrame = new Set();
 
     /** Whether highlights are visible in guest frames. */
@@ -96,7 +100,7 @@ export class FrameSyncService {
   }
 
   /**
-   * Watch for changes to the set of annotations displayed in the sidebar and
+   * Watch for changes to the set of annotations loaded in the sidebar and
    * notify connected guests about new/updated/deleted annotations.
    */
   _setupSyncToGuests() {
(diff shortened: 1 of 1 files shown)


src/sidebar/services/frame-sync.js
```diff
@@ -84,157 +85,156 @@ export class FrameSyncService {
     this._guestRPC = new Bridge();
 
     // Set of tags of annotations that are currently loaded into the frame
```
robertknight, line 87:
```suggestion
    /**
     * Tags of annotations that are currently loaded into guest frames.
     *
     * @type {Set<string>}
     */
```

- There can be multiple frames
- Specified the type of set elements, since TS can't infer it

src/sidebar/services/frame-sync.js
```diff
@@ -84,157 +85,156 @@ export class FrameSyncService {
...
+    this._setupSyncToGuests();
+    this._setupSyncFromGuests();
+    this._setupHostEvents();
+  }
 
-            if (isPublic(annot)) {
-              ++publicAnns;
-            }
+  /**
+   * Watch for changes to the set of annotations displayed in the sidebar and
```
robertknight, line 99:
```suggestion
   * Watch for changes to the set of annotations loaded in the sidebar and
```

This code cares about which annotations are loaded, rather than what is _displayed_, which depends on various filtering settings.

src/sidebar/services/frame-sync.js
```diff
@@ -260,10 +260,12 @@ export class FrameSyncService {
     };
 
     this._guestRPC.onConnect(addFrame);
+  }
 
-    this._setupSyncToGuests();
-    this._setupSyncFromGuests();
-
+  /**
+   * Listen for message coming from the host frame.
```
robertknight, line 266:
```suggestion
   * Listen for messages coming from the host frame.
```

robertknight (approved):
Good refactoring 👍. I requested a couple of minor changes.

PR #3968, opened 2021-11-26
Move random hex generator function to the `src/shared` folder

esanzgar:
Code in the `src/annotator` can and will use the `generateHexString` functionality.

I also cleanup some unusual code.


src/sidebar/util/postmessage-json-rpc.js
```diff
@@ -33,7 +28,7 @@ export function call(
   params = [],
   timeout = 2000,
   window_ = window,
-  id = generateId()
+  id = generateHexString(10)
```
robertknight, line 31: I think we can just suppress the code coverage complaint here.

robertknight (approved):
LGTM

src/sidebar/util/postmessage-json-rpc.js
esanzgar: I have removed the unnecessary test seam in a follow up commit.

src/annotator/annotation-sync.js
```diff
@@ -108,7 +110,7 @@ export class AnnotationSync {
     if (annotation.$tag) {
       return annotation;
     }
-    tag = tag || window.btoa(Math.random().toString());
+    tag = tag || generateHexString(8);
```
esanzgar, line 113: The only question I have is about this line.

This creates around 24 character long string of random alpha-numerical characters. This was different from `src/sidebar/services/annotations.js` that use `generateHexString(8)`.

Is there any value on having a different format for the `$tag` here than the one generated in the sidebar?

PR #3981, opened 2021-11-29
Call PortProvider.listen sooner

robertknight:
Call `listen` immediately after creating the PortProvider, before
creating the sidebar.

This will allow us to rule out the possibility that the sidebar sometimes fails
to connect to the PortProvider because an error occurred in the annotator code
after creating the sidebar iframe but before calling `PortProvider.listen`,
resulting in the PortProvider listener not being created.

### DIFF
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -53,6 +53,7 @@ function init() {
 
     const hypothesisAppsOrigin = new URL(sidebarConfig.sidebarAppUrl).origin;
     portProvider = new PortProvider(hypothesisAppsOrigin);
+    portProvider.listen();
 
     sidebar = new Sidebar(document.body, eventBus, guest, sidebarConfig);
     notebook = new Notebook(document.body, eventBus, getConfig('notebook'));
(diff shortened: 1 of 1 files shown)


src/annotator/index.js
```diff
@@ -53,15 +53,14 @@ function init() {
 
     const hypothesisAppsOrigin = new URL(sidebarConfig.sidebarAppUrl).origin;
     portProvider = new PortProvider(hypothesisAppsOrigin);
+    portProvider.listen();
```
esanzgar, line 56: If the portProvider starts listening at this point, there is a risk that a `Guest` able to discover the host and triggers the `frameConnected` event before the `Sidebar` is created.

robertknight: Any messages posted to the host frame won't be handled until the next turn of the event loop. With this change we are still registering the listener in the same event loop turn, just earlier.

esanzgar (commented):
`PortProvider#listen` must be called after registering the `frameConnected` callbacks.

src/annotator/index.js
esanzgar: OK, then I will approve the PR.

esanzgar (approved):
Robert has confirmed that the `frameConnected` won't be triggered before the Sidebar is created.

PR #3982, opened 2021-11-30
Remove `AnnotatorSync` class

esanzgar:
Through a progressive series of commits I was able to remove `AnnotatorSync` class and simplify the messages that are sent between the host and the sidebar frames.

Each commit explains the actual change.


src/sidebar/services/frame-sync.js
```diff
@@ -109,9 +113,15 @@ export class FrameSyncService {
     watch(
       this._store.subscribe,
       [() => this._store.allAnnotations(), () => this._store.frames()],
+      /**
+       * @param {[Annotation[], Frame[]]} arg0
+       * @param {[Annotation[]]} arg1
```
robertknight, line 118:
```suggestion
       * @param {[Annotation[], Frame[]]} current
       * @param {[Annotation[]]} prev
```

Even though `arg0` and `arg1` are not referenced here, we might as well give them more meaningful names for the reader's benefit.

src/sidebar/services/frame-sync.js
```diff
@@ -165,34 +175,36 @@ export class FrameSyncService {
...
-      if (!this._store.isLoggedIn()) {
-        this._hostRPC.call('openSidebar');
-        this._store.openSidebarPanel('loginPrompt');
-        this._guestRPC.call('deleteAnnotation', formatAnnot(annot));
-        return;
-      }
-      this._inFrame.add(event.tag);
+    this._guestRPC.on(
+      'createAnnotation',
+      (/** @type {AnnotationSafeData} */ annot) => {
```
robertknight, line 180: I think you'll need to use an `@param {Type} paramName` annotation _above_ the closure here for TS to use it. See for example uses of `@param` in various `createSelector` calls in store modules.

src/types/annotator.js
```diff
@@ -26,7 +26,7 @@
 /**
  * An object representing an annotation in the document.
  *
- * @typedef AnnotationData
+ * @typedef AnnotationSafeData
```
robertknight, line 29: I do think it would be good to clarify the difference between this type and the API `Annotation`, and to use a different name. I'm not sure about the proposed name, it's a bit unwieldy.

src/annotator/guest.js
```diff
@@ -367,6 +352,17 @@ export default class Guest {
...
+    this._sidebarRPC.on(
+      'deleteAnnotation',
+      /** @type {string} */ tag => this.detach(tag)
+    );
+
+    this._sidebarRPC.on(
+      'loadAnnotations',
+      (/** @type {AnnotationSafeData[]} */ annotations) =>
+        annotations.forEach(annotation => this.anchor(annotation))
+    );
```
robertknight, line 364: Since `anchor` is asynchronous there is a race condition here if a `deleteAnnotation` event is received for an annotation while it is being anchored. The `detach` method operates on the `this.anchors` array, but loaded annotations are not added to that until the end of the `anchor` method.

src/annotator/guest.js
```diff
@@ -367,6 +352,17 @@ export default class Guest {
       this.setHighlightsVisible(showHighlights);
     });
 
+    this._sidebarRPC.on(
+      'deleteAnnotation',
+      /** @type {string} */ tag => this.detach(tag)
```
robertknight, line 357: Using `@type` here doesn't work, and unfortunately TS doesn't complain if a JSDoc annotation is not successfully attached to a value. This works:

```js
    this._sidebarRPC.on(
      'deleteAnnotation',
      /** @param {string} tag */
      tag => this.detach(tag)
    );
```

robertknight (changes requested):
The overall direction of the changes here looks great. I like the removal of more low-value indirection in the guest <-> sidebar interactions.

I'm not keen on the `AnnotationSafeData` name. It feels a bit wordy for such a common type in the annotator and the meaning of "safe" is not explained in the docs. It is only explained in the `formatAnnotation` function in the sidebar. My suggestion would be to separate that change into a different PR.

I also found a race condition if annotations are unloaded while they are being anchored, which could cause a mismatch between what is displayed in the sidebar and what is displayed in the page.

src/annotator/guest.js
esanzgar: Yes, sorry. I have made similar mistakes it works if it is wrapped in parenthesis.
`(/** @type {string} */ tag) => this.detach(tag)`, but I have forgotten to add the parenthesis.

I

src/sidebar/services/frame-sync.js
esanzgar: Yes, you are correct. I made those changes in this PR: https://github.com/hypothesis/client/pull/3983

src/sidebar/services/frame-sync.js
esanzgar: 👍🏼

src/annotator/guest.js
esanzgar: Would an `await` here fix the race condition?

esanzgar:
```
    async annotations =>
        await Promise.all(
          annotations.map(annotation => this.anchor(annotation))
        )
```

esanzgar: This seems to be a pre-existing issue, isn't it?

robertknight: > Would an await here fix the race condition?

No. Think about what is happening here. The proposed code will just result in the `loadAnnotations` RPC method handler receiving a promise indicating when the operation is complete. `Bridge`/`PortRPC` won't do anything with that promise though. The race here is between the `detach` method, which relies on the annotation having been added to the `this.anchors` array, and `anchor` which only adds the anchor to that array asynchronously.

The one solution would be to keep track of annotations that are currently being anchored and defer de-anchoring for them in the `deleteAnnotations` callback.

> This seems to be a pre-existing issue, isn't it?

This is true. Perhaps the expedient thing would be to note the issue in a comment and address it in a separate PR.

PR #3983, opened 2021-11-30
Remove `AnnotationSync` utility

esanzgar:
AnnotationSync utility didn't have a clear purpose, except for assigning a
temporarily `$tag` to newly created annotations. For that, it used an
internal cache of annotation objects from the Guest, that were mutated
in-place. `AnnotationSync` used a very indirect method of communication
to the `Guest`, using an event emitter, even though their mutual
relationship was quite straightforward.

In this PR, we have:

* Simplified the messages that are sent between the host and the
  sidebar. We have assumed that a `$tag` value is always present if it
  comes from the `sidebar`, and a temporarily `$tag` is assigned in the
  `Guest` class for newly created annotations. Therefore, the type of
  the message sent between these two frames can be simplified to always
  be of the shape of `AnnotationData`, a safe version of `Annotation`.

* Added a few missing types.

* Changed the RPC method `deleteAnnotation` to send astring tag, similar
  to other RPC methods, like `focusAnnotations` or `scrollToAnnoation`.
  As a consequence of that we changed the `Guest#detach` method
  signature.

On a second commit I fix the tests


src/annotator/guest.js
```diff
@@ -305,55 +300,53 @@ export default class Guest {
...
-      this._focusedAnnotations.clear();
-      tags.forEach(tag => this._focusedAnnotations.add(tag));
-
-      for (let anchor of this.anchors) {
-        if (anchor.highlights) {
-          const toggle = tags.includes(anchor.annotation.$tag);
-          setHighlightsFocused(anchor.highlights, toggle);
+    this._sidebarRPC.on(
+      'focusAnnotations',
+      /** @param {string[]} tags */
```
esanzgar, line 308: No change on this callback, except adding the type.

src/annotator/guest.js
```diff
@@ -305,55 +300,53 @@ export default class Guest {
...
-      // section of the page that needs to be un-collapsed.
-      const event = new CustomEvent('scrolltorange', {
-        bubbles: true,
-        cancelable: true,
-        detail: range,
-      });
-      const defaultNotPrevented = this.element.dispatchEvent(event);
+    this._sidebarRPC.on(
+      'scrollToAnnotation',
+      /** @param {string} tag */
```
esanzgar, line 324: No change on this callback, except adding the type.

src/annotator/guest.js
```diff
@@ -499,21 +503,21 @@ export default class Guest {
     this._updateAnchors(this.anchors.concat(anchors), true /* notify */);
 
     // Let other frames (eg. the sidebar) know about the new annotation.
-    this._annotationSync.sync([annotation]);
+    this._sidebarRPC.call('syncAnchoringStatus', annotation);
```
esanzgar, line 506: Changed this RPC method to send a single (safe) annotation, instead of an array.

src/annotator/guest.js
```diff
@@ -499,21 +503,21 @@ export default class Guest {
...
 
     return anchors;
   }
 
   /**
    * Remove the anchors and associated highlights for an annotation from the document.
    *
-   * @param {AnnotationData} annotation
+   * @param {string} tag
```
esanzgar, line 514: A string is more convenient here.

src/sidebar/services/frame-sync.js
```diff
@@ -8,29 +8,33 @@ import { isReply, isPublic } from '../helpers/annotation-metadata';
...
+export function formatAnnot({ $tag, document, target, uri }) {
   return {
-    tag: ann.$tag,
-    msg: {
-      document: ann.document,
-      target: ann.target,
-      uri: ann.uri,
-    },
+    $tag,
+    document: { title: document.title, link: [{ href: target[0].source }] },
```
esanzgar, line 35: I needed to format it a little the returned object, to make it fully compatible with `AnnotationData`.

src/annotator/test/guest-test.js
```diff
@@ -427,6 +395,34 @@ describe('Guest', () => {
         );
       });
     });
+
+    describe('on "loadAnnotations" event', () => {
+      it('anchors annotations', () => {
+        const guest = createGuest();
+        const ann1 = { id: 1, $tag: 'tag1' };
+        const ann2 = { id: 2, $tag: 'tag2' };
+        sandbox.stub(guest, 'anchor');
```
robertknight, line 404: I think this follows the pattern of the previous tests. This will do for now, although stubbing an internal method is not ideal, as what you want to check really are the externally visible effects of anchoring. One way to do this is to listen for the `anchorsChanged` event that is emitted when anchoring completes.

src/annotator/test/guest-test.js
```diff
@@ -788,10 +784,10 @@ describe('Guest', () => {
       ]);
     });
 
-    it('sets `$tag` to a falsey value', async () => {
+    it('sets `$tag` to a temporarily value', async () => {
```
robertknight, line 787:
```suggestion
    it('sets `$tag` to a temporary value', async () => {
```

src/annotator/test/guest-test.js
```diff
@@ -1047,53 +1043,72 @@ describe('Guest', () => {
   });
 
   describe('#detach', () => {
+    let index = 0;
```
robertknight, line 1046: This could use a slightly more descriptive name, eg. `nextTagId`.

src/sidebar/services/frame-sync.js
```diff
@@ -8,29 +8,33 @@ import { isReply, isPublic } from '../helpers/annotation-metadata';
...
  * @typedef {import('../../types/bridge-events').SidebarToGuestEvent} SidebarToGuestEvent
  * @typedef {import('../../types/bridge-events').GuestToSidebarEvent} GuestToSidebarEvent
  * @typedef {import('../../shared/port-rpc').PortRPC} PortRPC
+ * @typedef {import('../store/modules/frames').Frame} Frame
  */
 
 /**
  * Return a minimal representation of an annotation that can be sent from the
- * sidebar app to a connected frame.
+ * sidebar app to the host frame.
```
robertknight, line 23:
```suggestion
 * sidebar app to a guest frame.
```

Annotations are displayed in guest frames, which may or may not be a host frame.

src/sidebar/services/frame-sync.js
robertknight, line 35: Does anything in the guest frame make use of the `document` property of annotations sent _to_ it as opposed to received _from_ it (when creating a new annotation). If not, then maybe we could avoid sending this information entirely.

src/sidebar/store/modules/frames.js
```diff
@@ -18,6 +18,7 @@ import { createStoreModule } from '../create-store';
  *   - Sub-frames will all have a id (frame identifier) set. The main frame's id is always `null`
  * @prop {DocumentMetadata} metadata - Metadata about the document currently loaded in this frame
  * @prop {string} uri - Current primary URI of the document being displayed
+ * @prop {boolean} [isAnnotationFetchComplete]
```
robertknight, line 21: The data model could be simplified here by making this property mandatory, in which case it will need to be set to `false` as part of the `connectFrame` action.

src/annotator/guest.js
```diff
@@ -499,21 +509,26 @@ export default class Guest {
...
    * @param {boolean} [notify] - For internal use. Whether to emit an `anchorsChanged` notification
    */
-  detach(annotation, notify = true) {
+  detach(tag, notify = true) {
+    if (!this._anchoringCompleted) {
```
robertknight, line 524: As noted above, a boolean variable won't be sufficient here.

robertknight (changes requested):
A boolean variable is not going to work for handling the race between `loadAnnotations` and `deleteAnnotation` calls. I sketched out a possible solution, but since there was an existing problem here, I think it would be pragmatic to remove the attempted workaround and file an issue to address it in a later PR.

Otherwise I just had some minor comments and once the above issue and these are addressed, we can get this merged.

src/annotator/guest.js
```diff
@@ -367,6 +361,23 @@ export default class Guest {
...
+    );
+
+    this._sidebarRPC.on(
+      'loadAnnotations',
+      /** @param {AnnotationData[]} annotations */
+      async annotations => {
+        await Promise.all(
+          annotations.map(annotation => this.anchor(annotation))
+        );
+        this._anchoringCompleted = true;
```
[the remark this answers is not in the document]
esanzgar: > annotations can be sent to guest frames in an any number of loadAnnotations calls

Ah... I see, I thought annotations were sent only once. Wrong assumption, sorry. I will delete

src/annotator/test/guest-test.js
esanzgar: 👍🏼

src/annotator/guest.js
esanzgar: Yes, it is clear, thanks.

src/sidebar/services/frame-sync.js
esanzgar: `document` is not used in `Guest#anchor`. In the last commit, I have made the `document` property optional, that forced me to add an empty `document.title` on `FrameSync`. Let me know what you think.

esanzgar:
I dropped the commit that incorrectly tried to address the issue sync issue with `Guest#detach`. This is ready for a second review.

src/annotator/test/guest-test.js
```diff
@@ -788,10 +784,10 @@ describe('Guest', () => {
       ]);
     });
 
-    it('sets `$tag` to a falsey value', async () => {
+    it('sets `$tag` to a temporary value', async () => {
```
robertknight, line 787: "Temporary" is perhaps ambiguous here. It is "temporary" in the sense that it will change when the page is reloaded and the annotation is re-fetched from the server. It is not "temporary" in the sense that it won't change as long as the annotation is present in the page. Perhaps better to just say "sets $tag property".

robertknight (approved):
LGTM.

src/annotator/test/guest-test.js
esanzgar: 👍🏼

PR #3985, opened 2021-11-30
Capture errors responding to port requests in the host frame and forward them to Sentry

robertknight:
This PR adds infrastructure that will enable us to gain visibility into errors happening in specifically wrapped blocks of Hypothesis code in the host frame. This is then used to check for errors happening during handling of port requests in `PortProvider#listen`.

This infrastructure works as follows:

1. In the annotator code, the sidebar iframe is created and a `sendErrorsTo` function is called to indicate which frame should handle captured errors from the host frame
2. In the sidebar iframe's Sentry client, `handleErrorsInFrames` is called to register a callback for errors from the host frame and send them to Sentry
3. In the annotator code, callbacks can be wrapped using a `captureErrors` function to capture any errors that the callback throws and forward them to the current handler frame. Alternatively a `sendError` function can be used from the `catch` block of a try-catch handler.

All captured errors have a `context` string to indicate where they came from, which is exposed in Sentry reports as a `context` tag.

This initial PR only supports reporting errors in browsers which can clone Errors when calling `window.postMessage`, which is currently only Chrome. In other browsers a warning is logged. We can address this in future by adding a fallback code path to pseudo-clone errors in these browsers.

This is part of the investigation into https://github.com/hypothesis/client/issues/3986.

---

Example of a Sentry report created by a dummy error thrown in `PortProvider#listen`:

Sentry forwarded error

In the context of the ongoing errors when creating the sidebar-host channel, this exception will show up in the breadcrumbs:

Sentry frame error in breadcrumbs

### DIFF
--- a/src/shared/test/frame-error-capture-test.js
+++ b/src/shared/test/frame-error-capture-test.js
@@ -1,0 +1,144 @@
+import { delay } from '../../test-util/wait';
+import {
+  captureErrors,
+  handleErrorsInFrames,
+  sendError,
+  sendErrorsTo,
+} from '../frame-error-capture';
+
+describe('shared/frame-error-capture', () => {
+  let errorEvents;
+
+  function handleMessage(event) {
+    if (event.data.type === 'hypothesis-error') {
+      errorEvents.push(event.data);
+    }
+  }
+
+  beforeEach(() => {
+    errorEvents = [];
+    window.addEventListener('message', handleMessage);
+  });
+
+  afterEach(() => {
+    window.removeEventListener('message', handleMessage);
+    sendErrorsTo(null);
+  });
+
+  describe('captureErrors', () => {
+    it('returns wrapped callback', () => {
+      const callback = captureErrors((a, b) => a * b, 'Testing captureErrors');
+      const result = callback(7, 8);
+      assert.equal(result, 56);
+    });
+
+    it('captures errors and forwards them to the current handler frame', async () => {
+      sendErrorsTo(window);
+
+      const error = new Error('Test error');
+      const callback = captureErrors(() => {
+        throw error;
+      }, 'Testing captureErrors');
+
+      assert.throws(() => {
+        callback();
+      }, 'Test error');
+
+      await delay(0);
+
+      assert.equal(errorEvents.length, 1);
+      assert.match(errorEvents[0], {
+        context: 'Testing captureErrors',
+        error: sinon.match({ message: 'Test error' }),
+        type: 'hypothesis-error',
+      });
+    });
+
+    it('does not forward errors if there is no handler frame', () => {
+      const callback = captureErrors(() => {
+        throw new Error('Test error');
+      }, 'Testing captureErrors');
+
+      assert.throws(() => {
+        callback();
+      }, 'Test error');
+
+      assert.equal(errorEvents.length, 0);
+    });
+
+    it('ignores errors forwarding error to handler frame', () => {
+      try {
+        sinon
+          .stub(window, 'postMessage')
+          .throws(new Error('postMessage error'));
+        sinon.stub(console, 'warn');
+
+        sendErrorsTo(window);
+        const callback = captureErrors(() => {
+          throw new Error('Test error');
+        }, 'Testing captureErrors');
+
+        assert.throws(() => {
+          callback();
+        }, 'Test error');
+
+        assert.calledOnce(window.postMessage);
+        assert.calledOnce(console.warn);
+      } finally {
+        console.warn.restore();
+        window.postMessage.restore();
+      }
+    });
+  });
+
+  describe('handleErrorsInFrames', () => {
+    it('invokes callback when an error is received', async () => {
+      const receivedErrors = [];
+
+      const removeHandler = handleErrorsInFrames((error, context) => {
+        receivedErrors.push({ error, context });
+      });
+
+      try {
+        sendErrorsTo(window);
+        sendError(new Error('Test error'), 'Test context');
+
+        await delay(0);
+
+        assert.equal(receivedErrors.length, 1);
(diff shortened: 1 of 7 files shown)


esanzgar (commented):
I am concern about possibly losing messages. Currently, if any errors happen before the  `Sidebar` (`src/annotator/sidebar.js`) is instantiated they will be lost.

Have you considered using `MessageChannel`?

`src/annnotator/sentry.js`
```
const {port1, port2} = new MessageChannel();

export function sendError(...) {
  port1.postMessage(....)
} 

export function captureErrors(...) {
}

logic to listen and handle the request of port2
```

`src/sidebar/util/sentry.js`
```
...
request and receive port2

port2.start()
port2.addEventListener((error, context) => Sentry.captureException(...));


export function reset(){
  port2?.stop();
}
```

src/shared/test/frame-error-capture-test.js
```diff
@@ -0,0 +1,144 @@
...
+
+      assert.equal(errorEvents.length, 1);
+      assert.match(errorEvents[0], {
+        context: 'Testing captureErrors',
+        error: sinon.match({ message: 'Test error' }),
+        type: 'hypothesis-error',
+      });
+    });
+
+    it('does not forward errors if there is no handler frame', () => {
```
esanzgar, line 57: This test belongs more to the `sendError` function.

src/shared/test/frame-error-capture-test.js
```diff
@@ -0,0 +1,144 @@
...
+      }, 'Testing captureErrors');
+
+      assert.throws(() => {
+        callback();
+      }, 'Test error');
+
+      assert.equal(errorEvents.length, 0);
+    });
+
+    it('ignores errors forwarding error to handler frame', () => {
```
esanzgar, line 69: Again this is more a test for `sendError`, which it would simplify the test.

src/shared/test/frame-error-capture-test.js
```diff
@@ -0,0 +1,144 @@
...
+        sinon.stub(console, 'warn');
+
+        sendErrorsTo(window);
+        const callback = captureErrors(() => {
+          throw new Error('Test error');
+        }, 'Testing captureErrors');
+
+        assert.throws(() => {
+          callback();
+        }, 'Test error');
```
esanzgar, line 83:
```suggestion
        sendError(new Error(), 'Text context');

```

src/shared/test/frame-error-capture-test.js
```diff
@@ -0,0 +1,144 @@
...
+        assert.equal(receivedErrors.length, 1);
+        assert.equal(receivedErrors[0].error.message, 'Test error');
+        assert.equal(receivedErrors[0].context, 'Test context');
+      } finally {
+        removeHandler();
+      }
+    });
+  });
+
+  // Integration test that combines `captureErrors` and `handleErrorsInFrames`.
```
esanzgar, line 117: I will move this test to the above `describe`, which I also consider an integration test:
`sendError` -> postMessage -> `handleErrorsInFrames`. This tests one more function: `captureErrors` -> `sendError` -> postMessage -> `handleErrorsInFrame`.

src/sidebar/util/sentry.js
```diff
@@ -12,6 +13,9 @@ import warnOnce from '../../shared/warn-once';
 let eventsSent = 0;
 const maxEventsToSendPerSession = 5;
 
+/** @type {(() => void)|null} */
+let removeFrameErrorHandler;
```
esanzgar, line 17:
```suggestion
let removeFrameErrorHandler = null;
```

esanzgar (approved):
[After a chat about the robustness of this code](https://hypothes-is.slack.com/archives/C1M8NH76X/p1638372530056200), we have agreed that this code is a simple as possible for the moment, to help with debugging the inter-frame communication issue.

> I agree that it would be better to implement a more robust delivery mechanism, once we can rule out problems with `MessageChannel`/`MessagePort` APIs.

Left a few comments about the tests.

src/shared/test/frame-error-capture-test.js
```diff
@@ -0,0 +1,144 @@
...
+    errorEvents = [];
+    window.addEventListener('message', handleMessage);
+  });
+
+  afterEach(() => {
+    window.removeEventListener('message', handleMessage);
+    sendErrorsTo(null);
+  });
+
+  describe('captureErrors', () => {
```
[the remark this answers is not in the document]
robertknight: I agree. I did this originally because `captureErrors` was the "main" API that other modules used. I've split this now so that `captureErrors` has a basic test that an error gets sent in the happy path, and there are now explicit tests for `sendErrors` that handle the failure cases.

src/shared/test/frame-error-capture-test.js
robertknight: I agree. I've added a separate block of tests for `sendError`.

src/shared/test/frame-error-capture-test.js
robertknight: Done!

src/sidebar/util/sentry.js
robertknight: Done

src/sidebar/util/test/sentry-test.js
```diff
@@ -198,6 +204,21 @@ describe('sidebar/util/sentry', () => {
       // be omitted from the report.
       assert.deepEqual(event.extra, {});
     });
+
+    it('registers a handler for errors in other frames', () => {
```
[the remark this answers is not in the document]
robertknight: We don't currently have tests for `reset` as it exists only for use in the tests. The actual application never calls it. I added a comment to clarify this.

robertknight:
Thanks for the feedback. I'm going to get this merged so we can see if it turns up any issues in production.

PR #3989, opened 2021-12-03
Add fallback for browsers that can't clone errors

robertknight:
Currently only Chrome supports performing a structured clone of Errors, though it is planned to be supported in other browsers in future [1]. Implement a fallback so that we can get information about annotator-side
errors in other browsers.

Also improve handling if a non-Error value is passed to `sendError`.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1556604. Haven't found a Safari issue for this yet.

### DIFF
--- a/src/shared/frame-error-capture.js
+++ b/src/shared/frame-error-capture.js
@@ -23,6 +23,44 @@ export function captureErrors(callback, context) {
 }
 
 /**
+ * @typedef ErrorData
+ * @prop {string} message
+ * @prop {string} [stack]
+ */
+
+/**
+ * Return a cloneable representation of an Error.
+ *
+ * This is needed in browsers that don't support structured-cloning of Error
+ * objects, or if the error is not cloneable for some reason.
+ *
+ * @param {Error|unknown} err
+ * @return {ErrorData}
+ */
+function serializeError(err) {
+  if (!(err instanceof Error)) {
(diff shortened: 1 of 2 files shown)


src/shared/frame-error-capture.js
```diff
@@ -41,12 +79,28 @@ export function sendError(error, context) {
...
-    // a fallback.
-    errorDestination.postMessage(data, '*');
+    // Try to send the error. If this fails because the browser doesn't support
+    // structured cloning of errors, use a fallback.
+    try {
+      errorDestination.postMessage(data, '*');
+    } catch (postErr) {
+      if (
+        postErr instanceof DOMException &&
+        postErr.name === 'DataCloneError'
```
esanzgar, line 96: I guess you have tested this in the browsers that don't currently support serialisation of `Error`.

esanzgar (approved):

src/shared/frame-error-capture.js
robertknight: Yes. This can be tested from the browser console like so:

```
> try { window.postMessage(new Error('test'), '*'); } catch (err) { lastErr = err };
< DataCloneError: The object can not be cloned.
> lastErr instanceof DOMException
< true
> lastErr.name
< "DataCloneError"
```

Issue #3986, opened 2021-12-02
"Unable to establish sidebar-host communication channel"


robertknight:
We are seeing a large number of [Sentry reports](https://sentry.io/organizations/hypothesis/issues/2810452318/?environment=prod&project=69811&query=is%3Aunresolved+%22sidebar-host%22&statsPeriod=14d) of errors setting up the sidebar-host channel using the new PortProvider/PortFinder infrastructure.

The error here is a timeout from the sidebar waiting for PortProvider to respond to fulfil a request for a host <-> sidebar MessageChannel.

Possible causes that we have identified:

- The user is visiting a site through a proxy which rewrites iframe URLs, making the origin of the sidebar app different from what PortProvider expects when trying to send messages to it. This can be identified in reports by the report's source URL being different than https://hypothes.is/app.html
- There is a mismatch between the version of the client used in the annotator and the version used in the sidebar. If the version used on the annotator side is from before we shipped PortProvider/PortFinder, then the sidebar-side client will fail to connect.
  - To help debug this we added a `version` field to the config metadata that the annotator passes to the sidebar. As a result we saw some reports where this may have been the problem, but others where it definitely wasn't.

esanzgar:
Regarding the mismatch of versions in annotator and sidebar, I would suggest to remove any caching from cloudflare for `https://cdn.hypothes.is/hypothesis`. This is a very minimal script (8K minimised), but very critical to load the latest version of the client in the `host` frame.

```
curl -I https://cdn.hypothes.is/hypothesis
HTTP/2 200
date: Thu, 02 Dec 2021 10:55:54 GMT
content-type: application/javascript; charset=UTF-8
content-length: 5753
x-amz-id-2: +7hmqcEcFoBNmVu2QLu63ymRSl/O94pE6Xm4wXZOHHx9HKAbLrmMXtY0JIxDmrKhCX4ifK+hGG0=
x-amz-request-id: X1VPE901AGQG5F1D
last-modified: Wed, 01 Dec 2021 13:30:34 GMT
etag: "e5ff6ba19940a34b515700aab155b90a"
cache-control: public, max-age=1800, must-revalidate
cf-cache-status: HIT
age: 616
accept-ranges: bytes
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
strict-transport-security: max-age=15552000; includeSubDomains; preload
x-content-type-options: nosniff
server: cloudflare
[...]

PR #4004, opened 2021-12-06
Report more detailed errors if port requests cannot be handled

robertknight:
This PR makes two improvements that will help us drill down into the causes of communication errors between frames that we are seeing in Sentry:

- The first commit causes failures in validating the details or sender of a message to produce an error rather than just failing silently. With the new logic, incoming messages are first checked to see if they have the correct structure and `type` value for a port request. If not the message is ignored. If the message passes this test, the same validation of the source and channel is applied as previously, but errors are now reported if the various checks fail.
- The second commit changes payload <-> message comparisons to only check known fields of the message, allowing us to rule out the possibility that unknown fields added by extensions or other sources are causing problems.

See the commit messages for more details.

Related to https://github.com/hypothesis/client/issues/3986.

### DIFF
--- a/src/shared/port-provider.js
+++ b/src/shared/port-provider.js
@@ -2,7 +2,7 @@ import { TinyEmitter } from 'tiny-emitter';
 
 import { captureErrors } from './frame-error-capture';
 import { ListenerCollection } from './listener-collection';
-import { isMessageEqual, isSourceWindow } from './port-util';
+import { isMessage, isSourceWindow } from './port-util';
 
 /**
  * @typedef {import('../types/annotator').Destroyable} Destroyable
@@ -10,6 +10,19 @@ import { isMessageEqual, isSourceWindow } from './port-util';
  * @typedef {import('./port-util').Frame} Frame
  * @typedef {'guest-host'|'guest-sidebar'|'notebook-sidebar'|'sidebar-host'} Channel
  */
+
+/**
+ * Combinations of frames that are allowed to connect to each other.
+ *
+ * The first frame in each pair is the source frame which will initiate the
+ * connection.
+ */
+const allowedChannels = [
+  'guest-sidebar',
+  'guest-host',
+  'sidebar-host',
+  'notebook-sidebar',
+];
 
 /**
  * PortProvider creates a `MessageChannel` for communication between two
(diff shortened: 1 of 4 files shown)


src/shared/port-util.js
```diff
@@ -42,14 +42,12 @@ export function isMessageEqual(data, message) {
...
-    );
-  } catch {
-    return false;
-  }
+  // We assume `JSON.stringify` cannot throw because `isMessage` verifies that
+  // all the fields we serialize here are serializable values.
+  return (
+    JSON.stringify(data, Object.keys(message).sort()) ===
+    JSON.stringify(message, Object.keys(message).sort())
+  );
```
robertknight, line 50: There is a risk here that future changes to the format of messages or the `isMessage` logic breaks this invariant. On the other hand this eliminates a code path which is currently never used.

src/shared/port-provider.js
```diff
@@ -151,29 +116,36 @@ export class PortProvider {
...
+        throw new Error(
+          `Port request for ${channel} came from a non-Window source`
+        );
+      }
+
+      const isAllowedOrigin =
+        // Guest and host frames may be hosted on any origin.
+        ['guest', 'host'].includes(frame1) ||
+        // The sidebar and notebook apps must be loaded from the expected origin.
+        origin === this._hypothesisAppsOrigin;
```
esanzgar, line 141: The association between allowed origins and channels is now less obvious than before, in my opinion.

`host` frame is never requesting port, so I will change it to:


```suggestion
      let isAllowedOrigin = true;
      // The sidebar and notebook apps must be loaded from the expected origin.
      if (['notebook', 'sidebar'].includes(frame1)){
          isAllowedOrigin = origin === this._hypothesisAppsOrigin;
      }
```

esanzgar: I think this will solve the issue of `guest` frames from opaque origin having the origin `'null'`.

https://github.com/hypothesis/client/issues/4011

src/shared/port-provider.js
```diff
@@ -151,29 +116,36 @@ export class PortProvider {
...
+      if (!isAllowedOrigin) {
+        throw new Error(
```
esanzgar, line 143: I don't agree with this approach. We are adding errors just to debug the current situation. The problem that I see with this is that the errors are not going to be removed from the code. I think the correct approach is to use `sendError` which is clearly a debugging statement, that can perfectly be removed from the code in the future and doesn't need to be tested.

src/shared/port-provider.js
```diff
@@ -192,7 +164,13 @@ export class PortProvider {
           : new MessageChannel();
 
       const message = { frame1, frame2, type: 'offer' };
-      source.postMessage(message, origin, [messageChannel.port1]);
+      try {
+        source.postMessage(message, origin, [messageChannel.port1]);
+      } catch (err) {
+        throw new Error(
```
esanzgar, line 170: I think this is a genuine problem that should crash the application. However, this is very unlikely that will err on a modern browser. Therefore, I would still suggest using `sendError` here.

src/shared/test/port-provider-test.js
```diff
@@ -1,9 +1,25 @@
 import { delay } from '../../test-util/wait';
-import { PortProvider } from '../port-provider';
+import { PortProvider, $imports } from '../port-provider';
 
 describe('PortProvider', () => {
+  let capturedErrors;
```
esanzgar, line 5: I don't think we should test debugging code.

esanzgar (changes requested):
I am not keen about throwing errors in situations that should be silently ignored for the sake of debugging.

I think debugging code should be temporal and should not be tested.

I suggest an alternative approach: use of `if` statements in combination with the `sendError` utility.

esanzgar:
> The first commit causes failures in validating the details or sender of a message to produce an error rather than just failing silently.

`'failing silently'` is not really accurate term. I would refer to `'silently ignoring'` those errors which was what the code was meant to do and I still feel that is what it suppose to do in the future.

I am fine with the second commit.

robertknight:
Thanks for the feedback. I extracted the second commit into a separate PR since we're agreed on that part. For the first part, I proposed an alternative approach, which is a much more minimal alteration to the existing code: https://github.com/hypothesis/client/pull/4015. I'm going to close this PR for now on the assumption that we'll land #4015 instead or something more in that vein.

PR #4005, opened 2021-12-06
Fix occassionally slow frame-error-capture tests

robertknight:
These tests access the `error.stack` property, which in turn triggers the `Error.prepareStackTrace` [1] handler installed by source-map-support [2] (this library is involved in translating errors thrown in tests to their original source locations). This
is expensive and has caused some CI timeouts.

The workaround is to disable the `error.stack` handler for the duration of these tests.

[1] https://v8.dev/docs/stack-trace-api#customizing-stack-traces
[2] https://github.com/evanw/node-source-map-support/blob/ac2c3e4c633c66931981ac94b44e6963addbe3f4/source-map-support.js#L584

### DIFF
--- a/src/shared/test/frame-error-capture-test.js
+++ b/src/shared/test/frame-error-capture-test.js
@@ -28,7 +28,15 @@ describe('shared/frame-error-capture', () => {
     return stub;
   }
 
+  let origPrepareStackTrace;
+
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


lyzadanger (approved):
Thanks for linking to background on this from the PR description. Good sleuthing!

PR #4008, opened 2021-12-06
Fix race condition when annotations are deleted before anchoring

esanzgar:
`deleteAnnotation` RPC event can be triggered before `loadAnnotations`.
This is a race condition in which annotations are asked to be deleted in
the `sidebar` frame before the `guest` frame finishes to anchor the
annotations.

This problem is avoided by maintaining one additional set which tracks
the annotations that should be in the page. This can be updated 1) when
an annotation is received, just before it is anchored and 2) when an
annotation is detached. After anchoring completes, the logic can check
if the annotation should still be present in the page and skip saving
the anchor if not.

Solution from here:
https://github.com/hypothesis/client/pull/4007#discussion_r763971360

### DIFF
--- a/src/annotator/test/guest-test.js
+++ b/src/annotator/test/guest-test.js
@@ -51,6 +51,14 @@ describe('Guest', () => {
     return guest;
   };
 
+  const emitSidebarEvent = (event, ...args) => {
+    for (let [evt, fn] of fakeBridge.on.args) {
+      if (event === evt) {
+        fn(...args);
+      }
+    }
+  };
+
   beforeEach(() => {
     guests = [];
     highlighter = {
@@ -195,14 +203,6 @@ describe('Guest', () => {
   });
 
   describe('events from sidebar frame', () => {
-    const emitSidebarEvent = (event, ...args) => {
-      for (let [evt, fn] of fakeBridge.on.args) {
-        if (event === evt) {
-          fn(...args);
-        }
-      }
-    };
-
     describe('on "focusAnnotations" event', () => {
       it('focuses any annotations with a matching tag', () => {
         const highlight0 = document.createElement('span');
@@ -1105,6 +1105,21 @@ describe('Guest', () => {
         anchors[0].highlights,
         true
       );
+    });
+
+    it('prevents saving of annotations that are deleted while been anchored', async () => {
+      const guest = createGuest();
+      const annotation = {
+        $tag: 'tag1',
+        target: [{ selector: [{ type: 'TextQuoteSelector', exact: 'hello' }] }],
+      };
+      fakeIntegration.anchor.resolves(range);
+
+      emitSidebarEvent('loadAnnotations', [annotation]);
+      emitSidebarEvent('deleteAnnotation', annotation.$tag);
+      await delay(0);
+
+      assert.lengthOf(guest.anchors, 0);
     });
   });
 
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -165,6 +165,12 @@ export default class Guest {
      */
     this.anchors = [];
 
+    /**
+     * Tags of annotations that are currently anchored or being anchored in
+     * the guest.
+     */
+    this._annotations = /** @type {Set<string>} */ (new Set());
+
     // Set the frame identifier if it's available.
     // The "top" guest instance will have this as null since it's in a top frame not a sub frame
     /** @type {string|null} */
@@ -527,11 +533,19 @@ export default class Guest {
     // Remove existing anchors for this annotation.
     this.detach(annotation.$tag, false /* notify */);
 
+    this._annotations.add(annotation.$tag);
+
     // Resolve selectors to ranges and insert highlights.
     if (!annotation.target) {
       annotation.target = [];
     }
     const anchors = await Promise.all(annotation.target.map(locate));
+
+    // If the annotation was removed while anchoring, don't save the anchors.
+    if (!this._annotations.has(annotation.$tag)) {
+      return [];
+    }
+
     for (let anchor of anchors) {
       highlight(anchor);
     }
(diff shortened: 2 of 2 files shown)


robertknight (commented):
I suggested a change of approach in the previous PR in this series which will impact on this one, depending on the final solution we land on there.

One thing I note about both this PR and the previous one is that we are relying on the assumption that tags will not be reused within a session, even for the same annotation (if it is loaded in the sidebar, unloaded and later re-loaded again). This is currently the case, but not called out in the `AnnotationData` typedef or the code that generates tag values. We should make that clearer.

esanzgar:
> One thing I note about both this PR and the previous one is that we are relying on the assumption that tags will not be reused within a session, even for the same annotation (if it is loaded in the sidebar, unloaded and later re-loaded again).

Yes, I was relying on the assumption that an annotation's `$tag` uniquely identifies a set of anchors. If that's not the case, or if we have plans to change this in the future, we should device a different system to pass information between the sidebar and host.

The opposite is true: it is currently not clear from any comment or typedef of that tags should/could be reused.

src/annotator/test/guest-test.js
```diff
@@ -420,15 +420,18 @@ describe('Guest', () => {
     });
 
     describe('on "deleteAnnotation" event', () => {
-      it('detaches annotation', () => {
+      it('prevents saving of annotations that are currently anchoring', async () => {
```
esanzgar, line 423:
```suggestion
      it('prevents saving of annotations deleted while been anchored', async () => {
```

src/annotator/test/guest-test.js
```diff
@@ -420,15 +420,18 @@ describe('Guest', () => {
...
         const guest = createGuest();
-        const callback = sinon.stub();
-        guest._emitter.subscribe('anchorsChanged', callback);
+        const ann = { target: [], uri: 'uri', $tag: 'tag1' };
 
-        emitSidebarEvent('deleteAnnotation', 'tag1');
+        emitSidebarEvent('loadAnnotations', [ann]);
+        emitSidebarEvent('deleteAnnotation', ann.$tag);
```
esanzgar, line 428: Deleting this line test passes, so this is not ideal.

esanzgar, line 428: If you invert both lines the test should not pass.

I think there is a caveat to this solution: if the deletion of the annotation happens _before_ the loading the annotation will be anchored. This solution works if the deletion happens _while_ the anchoring is taking place.

esanzgar:
This solution takes care of a possible deletion of annotation between these lines:

https://github.com/hypothesis/client/blob/715856835a1d965a8c7fb3578a12c10b0aaf5193/src/annotator/guest.js#L536-L545

In line 542, the `locate` method is invoke asynchronously. Therefore, it is possible that an annotation is deleted that removes the `$tag` in `this._annotation` (check in line 545) even thought it was added in line 536.

However, this solution doesn't take into consideration the async execution of the `Guest#anchor` method in this line:

https://github.com/hypothesis/client/blob/715856835a1d965a8c7fb3578a12c10b0aaf5193/src/annotator/guest.js#L421

It is not impossible for an annotation to be deleted between `Guest#anchor` executions.

src/annotator/test/guest-test.js
robertknight: Either of these works for me.

src/annotator/test/guest-test.js
robertknight: > Depending of how the browser decides to execute the waiting anchoring tasks the race condition could still persist.

The code looks safe at present. `this.anchor` is called synchronously for all annotations in the `loadAnnotations` event handler in `Guest`. This event handler is in turn called in response to our code (PortRPC) receiving a message via a MessagePort. The MessagePort in turn will deliver events in the other they were posted by the sending frame. As long as the sidebar calls `loadAnnotations` before `deleteAnnotation`, everything should work.

There is a refactoring hazard here though. If someone where to add async steps in the `loadAnnotations` event handler or `anchor` method before `this._annotations.add` were called, then it could introduce a race. This can be prevented by a test like this which calls `loadAnnotations` followed by `deleteAnnotation` as fast as possible. In the actual app, the fastest a `deleteAnnotation` call can follow a `loadAnnotations` call is on the next macrotask. Here it is being done in the same event loop turn.

robertknight: Why not add a target here?

robertknight (commented):
I responded to a couple of comments on the PR.

src/annotator/test/guest-test.js
```diff
@@ -420,15 +420,18 @@ describe('Guest', () => {
     });
 
     describe('on "deleteAnnotation" event', () => {
-      it('detaches annotation', () => {
```
robertknight, line 423: This test replaces a test for the main scenario of using `deleteAnnotation` with one that handles this edge case. I think it would make sense to have both.

robertknight (commented):
The implementation looks fine, but in guest-test.js the test for the main scenario involving `deleteAnnotation` is replaced with one for an edge case. I think it would make sense to have a test for the main use, simulating a `deleteAnnotation` call occuring after an annotation is loaded and anchored, as well as this edge case.

src/annotator/test/guest-test.js
esanzgar: > The code looks safe to me at present. this.anchor is called synchronously for all annotations in the loadAnnotations event handler in Guest. This event handler is in turn called in response to our code (PortRPC) receiving a message via a MessagePort. The MessagePort in turn will deliver events in the other they were posted by the sending frame. As long as the sidebar calls loadAnnotations before deleteAnnotation, everything should work.

Thanks for the explanation. I agree everything should work fine to avoid the race condition.

esanzgar: > There is a refactoring hazard here though. If someone where to add async steps in the loadAnnotations event handler or anchor method before this._annotations.add were called, then it could introduce a race. This can be prevented by a test like this which calls loadAnnotations followed by deleteAnnotation as fast as possible. In the actual app, the fastest a deleteAnnotation call can follow a loadAnnotations call is on the next macrotask. Here it is being done in the same event loop turn.

I agree, I will try to work on that test. In the past I found challenging to simulate race conditions. Let's see...

esanzgar:
I have fixed the tests as you suggested.

robertknight (approved):
LGTM

PR #4009, opened 2021-12-06
 Improved performance on `focusAnnotations` RPC event

esanzgar:
Instead of looping through _all_ the annotations to toggle the
highlight, we target specifically the two subsets that we are interested
on:
    
* the previously focus annotations: to turn the highlight off
* the new focused annotations: to turn the highlights on.


esanzgar:
We have decided to postpone these improvements for later.

PR #4010, opened 2021-12-06
Improved performance on `deleteAnnotation` RPC event

esanzgar:
This is a minor improvement. Instead of looping through _all_ the
annotations to check if we need to remove the highlights, we access this
subset of annotations directly, remove the highlights, delete them from
the `Guest#_annotation` map and use the reminder annotations to notify
the `Sidebar` via `anchorsChanged`.


esanzgar:
Let's postpone these changes for later.

esanzgar:
Let's postpone these changes for later.

Issue #4011, opened 2021-12-06
Guests in file:// or other opaque origins cannot connect to sidebar


robertknight:
When a guest frame is loaded from a file:// URL or some other [opaque origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin) and sends a request to the host frame for a guest <-> sidebar connection, `PortProvider#listen` will receive a `MessageEvent` whose `origin` is the _string_ "null". This string is not a legal argument to `window.postMessage`, so an error occurs when PortProvider attempts to send the port back to the guest using `source.postMessage(data, event.origin)`.

Related Slack thread: https://hypothes-is.slack.com/archives/C2BLQDKHA/p1638812510038200

One possible solution would be to special case handling of opaque origins using `source.postMessage(data, event.origin === null ? '*' : event.origin)`. Another might be to always allow any origin for certain types of source frame (eg. guests).

mattdricker:
This is the original user report: https://app.hubspot.com/contacts/6291320/ticket/690271618/

Example local HTML files attached to that ticket.

PR #4012, opened 2021-12-07
Handle port requests from guests with opaque origins

robertknight:
When a guest frame with an opaque origin, such as file:// URL or a sandboxed iframe, sends a request for a port to the host frame, the `event.origin` value will be the string "null". This is not a legal value for the `targetOrigin` argument to `postMessage`, so we need to remap it to "*" (any origin).

Related to this, I checked all of the other `window.postMessage` calls in the app and they either already use a wildcard value for `targetOrigin` or are used in a context where I think we can reasonably assume that the target can't be an opaque origin (eg. LMS frontend).

Fixes https://github.com/hypothesis/client/issues/4011.

### DIFF
--- a/src/shared/port-provider.js
+++ b/src/shared/port-provider.js
@@ -192,7 +192,16 @@ export class PortProvider {
           : new MessageChannel();
 
       const message = { frame1, frame2, type: 'offer' };
-      source.postMessage(message, origin, [messageChannel.port1]);
+
+      // If the source window has an opaque origin [1], `event.origin` will be
+      // the string "null". This is not a legal value for the `targetOrigin`
+      // parameter to `postMessage`, so remap it to "*".
+      //
+      // [1] https://html.spec.whatwg.org/multipage/origin.html#origin.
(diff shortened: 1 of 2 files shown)


esanzgar (approved):

esanzgar:
As a follow up, maybe you could convert the iframe to a sandbox type in this scenario http://localhost:3000/document/cross-origin-iframe.

robertknight:
I rebase this branch directly on master, so that it isn't dependent on the work to add additional error reporting to PortProvider.

> As a follow up, maybe you could convert the iframe to a sandbox type in this scenario http://localhost:3000/document/cross-origin-iframe so we can test more easily this case.

I think we should probably add a separate test case for this, since having a non-sandboxed iframe is probably a more common scenario for cross-origin frames.

robertknight:
I'm not sure why GitHub CI hasn't run, and unfortunately I can't see an option to re-run it manually. This might have been to do with the change of base branch. However Jenkins CI passed and I've re-run `make sure` locally. On that basis I'm going to treat this as having passed CI.

PR #4014, opened 2021-12-08
Proposed fix for clearing the selection on another iframe

esanzgar:
The proposal consist of removing the selection without re-triggering the `SelectionObserver`. 

This is done by (1) making `SelectionObserver#_cancelPendingUpdates` a public method (this would be equivalent to `lodash.debounce#cancel`) and (2) using this method to cancel the update caused by removing the selection.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -314,23 +314,24 @@ export default class Guest {
 
   async _connectHostEvents() {
     this._hostRPC.on(
-      'deselectTextExcept',
-      /** @type {string} */ frameIdentifier => {
+      'clearSelectionExceptIn',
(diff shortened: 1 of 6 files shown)


robertknight (commented):
On the upside this is very simple. There is a risk that a change in browser behavior might break it, if delivery of `selectionchange` was made to happen later. The [spec](https://w3c.github.io/selection-api/#selectionchange-event) does say that event delivery must happen on the next tick (rather than immediately), but I'm not certain if that rules out it being deferred to later. I think we could tolerate that risk in this context though.

esanzgar:
Close in favour of https://github.com/hypothesis/client/pull/4021

PR #4015, opened 2021-12-08
Capture message validation errors when handling port provider requests

robertknight:
Following discussion yesterday, this is a more minimalist alternative to https://github.com/hypothesis/client/pull/4004 which keeps the structure of PortProvider#listen and tests the same.

To help us debug issues with sidebar-host connection setup [1], report an
error to the sidebar if PortProvider receives a message which looks like
a port request (data has the structure `{ frame1: string, frame2:
string, type: 'request' }`) but comes from a non-Window source, invalid
origin or is not a valid combination of frames.

There are some minimal tests added because we do need logic to avoid sending the same report many times.

[1] https://sentry.io/organizations/hypothesis/issues/2810452318/

---

To test this:

1. Modify `sendError` in `frame-error-capture.js` to log the errors it is passed
2. Remove or modify entries in the `this._allowedMessages` array in PortProvider to force a message validation error.
3. Reload the client's dev server and you should see an error

### DIFF
--- a/src/shared/port-provider.js
+++ b/src/shared/port-provider.js
@@ -1,8 +1,8 @@
 import { TinyEmitter } from 'tiny-emitter';
 
-import { captureErrors } from './frame-error-capture';
+import { captureErrors, sendError } from './frame-error-capture';
 import { ListenerCollection } from './listener-collection';
-import { isMessageEqual, isSourceWindow } from './port-util';
+import { isMessage, isMessageEqual, isSourceWindow } from './port-util';
 
 /**
  * @typedef {import('../types/annotator').Destroyable} Destroyable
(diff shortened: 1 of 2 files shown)


esanzgar (commented):
I think this approach is better.

I don't fully understand the need to store the errors in `sentErrors` instead of just send them directly.

robertknight:
> I don't fully understand the need to store the errors in sentErrors instead of just send them directly.

The reason is that if a validation error occurs when responding to a real PortFinder message, then PortFinder will keep sending the same message until it gets a response or a timeout occurs. This will lead to us reporting the same error to the sidebar many times (or doing whatever other logging `sendError` may do).

esanzgar (approved):

PR #4018, opened 2021-12-08
Make `Guest#selectedRanges` part of the public API

esanzgar:
I defined a type and initialise the value of `Guest#selectedRanges` in
the constructor.

As far as I know, this is used only in tests.

Follow up of https://github.com/hypothesis/client/pull/3904/files#r764767461

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -136,6 +136,8 @@ export default class Guest {
     this._highlightsVisible = false;
     this._isAdderVisible = false;
     this._informHostOnNextSelectionClear = true;
+    /** @type {Range[]} - Ranges of the current text selection. */
+    this.selectedRanges = [];
 
     this._adder = new Adder(this.element, {
       onAnnotate: async () => {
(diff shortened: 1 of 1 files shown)


src/annotator/guest.js
```diff
@@ -131,6 +131,7 @@ export default class Guest {
     this._hostFrame = hostFrame;
     this._highlightsVisible = false;
     this._isAdderVisible = false;
+    this.selectedRanges = /** @type {Range[]} */ []; // ranges of the current text selection
```
robertknight, line 134:
```suggestion
    /** Ranges of the current text selection. */
    this.selectedRanges = /** @type {Range[]} */ ([]);
```

- TS requires the `@type`-d expression to be wrapped in parens
- In order for the symbol documentation to be attached to the symbol (eg. for use in VS Code tooltips and similar in other IDEs), you need to use JSDoc comments.

robertknight (changes requested):
Be careful to include parens in `@type` expressions. It would be good if a lint check could catch this kind of issue, but we don't currently have one set up.

robertknight (approved):

PR #4020, opened 2021-12-08
Actually use a WeakMap for window <-> channel map

robertknight:
🤦


esanzgar (approved):

esanzgar:
It is difficult to catch this mistake because the interface is identical.

robertknight:
> It is difficult to catch this mistake because the interface is identical.

The WeakMap interface is a subset of the Map interface - no iteration support and more constrained key types. I guess this assignment has to be legal because its the same situation as a derived => base class type cast, which you'd want to allow.

PR #4021, opened 2021-12-09
Fix clearing the selection on another iframe

esanzgar:
After the proposal in
https://github.com/hypothesis/client/pull/4014/commits/c32b7013df49fae87adca077e571f42ef1162e7d,
we agree this alternative approach for clearing the selection on another iframe is more straight forward.

### To test this PR:

1. Go to http://localhost:3000/document/cross-origin-iframe
2. Make a text selection in one frame. The adder should appear.
3. Make a text selection in the other frame. The adder should appear.
   The adder in the other frame should disappear _and_ the selection
   should be removed.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -90,6 +90,10 @@ function resolveAnchor(anchor) {
   }
 }
 
+function removeTextSelection() {
+  document.getSelection()?.removeAllRanges();
+}
+
 /**
  * `Guest` is the central class of the annotator that handles anchoring (locating)
  * annotations in the document when they are fetched by the sidebar, rendering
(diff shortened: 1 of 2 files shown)


src/annotator/guest.js
```diff
@@ -318,10 +323,8 @@ export default class Guest {
           return;
         }
 
-        this._onClearSelection({ informHostFrame: false });
-        // TODO: clear the selection calling `document.getSelection()?.removeAllRanges();`
-        // without triggering the selection observer (that in turns calls
-        // `_onClearSelection({informHostFrame: true})` and changes `Sidebar#._guestWithSelection`).
+        removeTextSelection();
+        this._informHostOnNextSelectionClear = false;
```
robertknight, line 327: This ordering works because `selectionchanged` is dispatched asynchronously. Logically I think it would be nicer to set the flag before removing the selection.

robertknight (approved):
LGTM.

src/annotator/guest.js
esanzgar: 👍🏼

PR #4022, opened 2021-12-09
Check annotator-sidebar version consistency and sidebar origin on startup

robertknight:
This PR adds checks for three potential problems with the environment when the sidebar app starts. If the checks fail client startup is allowed to continue, but we log a warning in the browser console and disable Sentry error reporting to reduce noise.

The problems that are checked for currently are:

- Sidebar and annotator versions do not match, which _may_ cause problems with communication between the sidebar and other frames
- The sidebar is loaded in a sandboxed iframe, indicated by `window.origin` being "null"
- The sidebar is served from the wrong origin

The code changes are:

- Add an `origin` parameter to the `#config` fragment in the sidebar/notebook app launch URLs which indicates the expected origin of the sidebar app.
- Add a `checkEnvironment` function in `src/sidebar/config/check-env.js` which performs the checks and returns true they passed or logs warnings and returns false otherwise.
- In `src/sidebar/index.js`, call `checkEnvironment` and disable Sentry reporting if it fails.

## Background

When deploying the new inter-frame communication code we saw errors [1] setting up communication between the sidebar and host frames caused by:

 1. Mismatches between the version of the client used in the annotator and the sidebar
 2. The sidebar being served from an unexpected origin (eg. due to use of a web proxy) or an opaque origin (eg. due to the host being loaded in a sandboxed iframe). This is generally outside of our control.

We should be able to reduce occurrences of (1) by changing caching settings on the client's boot script. See https://github.com/hypothesis/client/pull/4050. However there are also situations out of our control where it can happen, eg. if a web page embedding Hypothesis is saved as an archive using certain tools, or a proxy overrides the boot script's caching headers.

Situation (2) is outside of our control and we are not going to relax origin checks in the login process as they are essential for security.

[1] See https://github.com/hypothesis/client/issues/3986#issuecomment-989685057 for more details.

## Testing

1. In a normal startup (ie. not in one of the environments where the client is expected to fail), no new warnings should be logged.
2. Change the `checkEnvironment` call near the top of `sidebar/index.js` to either of the following variants. The client should log a warning on startup:

```js
// Simulate sidebar being loaded from another origin
const envOk = checkEnvironment({ origin: 'https://hypothesis.someproxy.edu', location });

// Simulate sidebar being loaded in sandboxed iframe
const envOk = checkEnvironment({ origin: 'null', location });
```
3. Change the `appConfig.version` initialization in `src/annotator/config/app.js` to `appConfig.version = '1.0.0-other-version'` and reload the page. You should see a warning about the version mismatch.


robertknight:
We could handle the sidebar being reloaded within a client session in a similar way. The host could pass in a unique random token in the client's config and the sidebar would remember used tokens for some period of time.

A sketch of this:

```diff
diff --git a/src/annotator/config/app.js b/src/annotator/config/app.js
index c7452758c..f9a70528d 100644
--- a/src/annotator/config/app.js
+++ b/src/annotator/config/app.js
@@ -1,3 +1,5 @@
+import { generateHexString } from '../../shared/random';
+
 /**
  * Create the JSON-serializable subset of annotator configuration that should
  * be passed to the sidebar or notebook applications.
@@ -32,6 +34,7 @@ export function createAppConfig(config) {
   // Pass the version of the client, so we can check if it is the same as the
   // one used in the sidebar/notebook.
   appConfig.version = '__VERSION__';
+  appConfig.launchId = generateHexString(6);
 
   // Pass the URL of the page that embedded the client.
   const hostURL = new URL(window.location.href);
diff --git a/src/sidebar/config/host-config.js b/src/sidebar/config/host-config.js
index 7e6fc8772..b6969ab84 100644
--- a/src/sidebar/config/host-config.js
+++ b/src/sidebar/config/host-config.js
@@ -18,7 +18,9 @@ import {
 export function checkVersionAndOrigin(window) {
   let ok = true;
 
-  const { version, origin } = parseConfigFragment(window.location.href);
+  const { version, origin, launchId } = parseConfigFragment(
+    window.location.href
+  );
   if (version !== '__VERSION__') {
     console.warn(
       `The Hypothesis sidebar is using a different version (__VERSION__) than the host page (${version}). It may not work.`
@@ -33,6 +35,14 @@ export function checkVersionAndOrigin(window) {
     ok = false;
   }
 
+  if (window.sessionStorage.getItem(launchId) !== null) {
+    console.warn(
+      'The Hypothesis sidebar was reloaded in the same client session. It may not work.'
+    );
+    ok = false;
+  }
+  window.sessionStorage.setItem(launchId, '');
+
   return ok;
 }
 ```

**Comment by robertknight** [comment]
I found another caching issue when testing the idea in the previous comment. In Safari when using the browser's back button to revisit a page that embeds the client, the sidebar app may see the config fragment from the initial load instead of the most recent load. I couldn't reproduce this in Chrome and Firefox so far.

**Review by esanzgar — APPROVED** [review]
Well done! Thanks.

File: src/sidebar/config/check-env.js [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+    console.warn(
+      `The Hypothesis sidebar is using a different version (__VERSION__) than the host page (${version}). It may not work.`
+    );
+    return false;
+  }
+
+  // The client can't work if loaded in the wrong origin or an opaque origin
+  // because various cross-frame messaging (eg. login window, host <-> sidebar
+  // RPC) set an origin filter on `postMessage` calls.
+  if (window.origin === 'null') {
```
esanzgar, line 32: I know that this is referring to the `sidebar` frame been sandboxed, however, if only the `guest` iframe is sandboxed, it should still work OK (if the `sidebar` is not sandboxed). Like the this scenario: https://github.com/hypothesis/client/pull/4024

robertknight: This is true, I'll update this comment to clarify that it refers to the sidebar specifically.

Issue #3681, opened 2021-08-16
Sentry issues from different releases are not combined


robertknight:
It seems that Sentry is not automatically combining the same reports when they come from two different client releases, even when the exceptions are the same.

See for example https://sentry.io/organizations/hypothesis/issues/2569105765/ and https://sentry.io/organizations/hypothesis/issues/2574958144/.

I suspect this is caused by the client version number appearing in the source file URLs associated with stack frames, meaning that a frame from the same function will be considered to have occurred in different files across two different releases.

The source file URLs come from resolving the relative source file URLs in the source map files against the URL where the source map is served from. For example [annotator.bundle.js](https://cdn.hypothes.is/hypothesis/1.853.0/build/scripts/annotator.bundle.js) has a source map reference at the bottom to [annotator.bundle.js.map](https://cdn.hypothes.is/hypothesis/1.853.0/build/scripts/annotator.bundle.js.map). The sourcemap file lists module paths relative to the root of the client repository, but those files don't exist on https://cdn.hypothes.is. The relative URLs in the sourcemap files are consistent across releases, but the resolved absolute URLs are not.

robertknight:
In the [Issue Grouping settings](https://sentry.io/settings/hypothesis/projects/client/issue-grouping/) page for the client I noticed a big red **Upgrade Grouping Strategy** button at the bottom. I'm a sucker for big red buttons, so I clicked it to see if that helps at all.

I did click through a warning that the old and new strategies are not compatible and will result in new issues being created.

robertknight:
Docs for issue grouping: https://docs.sentry.io/product/data-management-settings/event-grouping/

robertknight:
>  I'm a sucker for big red buttons, so I clicked it to see if that helps at all.

It didn't help.

The bottom of each Sentry issue shows the data that was used in grouping. The client version number is appearing in the data that is used:

<img width="739" alt="Sentry event grouping" src="https://user-images.githubusercontent.com/2458/145562725-3e5630b4-78ab-4a06-8969-4e966ae5aabc.png">

I think what we may need to do is use the `RewriteFrames` integration to strip everything up to and including `hypothesis/<version>/` from the script URL before it is submitted to Sentry, such that the filename is consistent across releases.

If we do this, we'll also need to upload sourcemaps using sentry-cli as part of the release process because Sentry will no longer be able to retrieve them by following the sourcemap URL in the bundle.

This will look like:

```sh
[...]

PR #4023, opened 2021-12-10
Upload sourcemaps as part of Jenkins build

robertknight:
In fixing https://github.com/hypothesis/client/issues/3681, we will be stripping parts of the script filename in stacktraces sent to Sentry. As a result it will no longer be possible for Sentry to fetch the script and associated sourcemap from https://cdn.hypothes.is/. Instead we need to upload sourcemaps directly to Sentry as part of the build process.

See https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -108,6 +108,7 @@ node {
             nodeEnv.inside("-e HOME=${workspace}") {
                 withCredentials([
                     string(credentialsId: 'npm-token', variable: 'NPM_TOKEN'),
+                    string(credentialsId: 'sentry-token', variable: 'SENTRY_AUTH_TOKEN'),
                     usernamePassword(credentialsId: 'github-jenkins-user',
                                      passwordVariable: 'GITHUB_TOKEN_NOT_USED',
                                      usernameVariable: 'GITHUB_USERNAME'),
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -124,6 +125,9 @@ node {
                     export SIDEBAR_APP_URL=https://qa.hypothes.is/app.html
                     export NOTEBOOK_APP_URL=https://qa.hypothes.is/notebook
                     yarn version --no-git-tag-version --new-version ${qaVersion}
+
+                    yarn run sentry-cli releases --org hypothesis --project client new ${qaVersion}
+                    yarn run sentry-cli releases --org hypothesis --project client files ${qaVersion} upload-sourcemaps build/scripts/*.map
```
robertknight, line 130: See https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/cli/

robertknight:
Related non-blocking playbook issue: https://github.com/hypothesis/playbook/issues/791

lyzadanger (approved):
OK, this makes sense. When we publish, we create a new release/version in Sentry (which is an idempotent action, according to your notes) and upload all `*.map` files out of the `build/scripts/` directory and associate them with that release/version. 

👍🏻

PR #4024, opened 2021-12-10
Create an dev test scenario for a sandboxed guest frame

esanzgar:
This will allow us to test the effects of sandboxing the guest iframes.
In the past, it affected the inter-frame communication (#4012) but can
also affect other process. For example, it has revealed a possible
collision with the `createIntegration` function:

```
Sandbox access violation: Blocked a frame at "http://localhost:3002" from accessing a frame at "http://localhost:3000".  The frame requesting access is sandboxed and lacks the "allow-same-origin" flag.
```

The above error is from Safari.

### DIFF
--- a/dev-server/documents/html/cross-origin-iframe.mustache
+++ b/dev-server/documents/html/cross-origin-iframe.mustache
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <title>Cross-origin guest iframe</title>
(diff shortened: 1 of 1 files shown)


robertknight (changes requested):
We should create a separate test page for sandbox-related scenarios. Sandboxing an iframe effectively assigns it to a new, temporary origin. To see this, evaluate `window.origin` in the iframe. It returns `"null"`. This effectively overrides the cross-origin aspect of this test.

esanzgar:
I agree, I will create a new scenario.

PR #4043, opened 2021-12-15
Remove event emitter from `Guest`

esanzgar:
The main `Guest` was assumed to be always in the same frame as the
`host` frame. In that scenario (which is currently the most common),
`Guest` was able to communicate with the `host` via `TinyEmitter`.

This PR removes that assumption an enforce all communication between the
`Guest` and the `host` to occur via an inter-frame communication
channel. This paves the work for the main `Guest` to be in a different
frame than the `host`, which is the case in the Ebook readers and
VitalSource.

The bucket bar should continue to work exactly the same as in the past.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -30,7 +30,6 @@ import { normalizeURI } from './util/url';
  * @typedef {import('../types/bridge-events').GuestToHostEvent} GuestToHostEvent
  * @typedef {import('../types/bridge-events').GuestToSidebarEvent} GuestToSidebarEvent
  * @typedef {import('../types/bridge-events').SidebarToGuestEvent} SidebarToGuestEvent
- * @typedef {import('./util/emitter').EventBus} EventBus
  */
 
 /**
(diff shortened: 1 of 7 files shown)


src/annotator/guest.js
```diff
@@ -593,8 +589,8 @@ export default class Guest {
    */
   _updateAnchors(anchors, notify) {
     this.anchors = anchors;
-    if (notify) {
-      this._emitter.publish('anchorsChanged', this.anchors);
```
esanzgar, line 597: `anchors` was sent as a parameter but it was not used by the `sidebar`

src/annotator/guest.js
```diff
@@ -593,8 +589,8 @@ export default class Guest {
...
+    if (notify && this._frameIdentifier === null) {
```
esanzgar, line 592: Special treatment of the `Guest` that is in the `host` frame. However, this assumption could be lifted in the future.

robertknight (approved):
The code change here looks good. I haven't tested this manually so I'll leave it to you to do that.

src/annotator/guest.js
```diff
@@ -122,16 +122,13 @@ export default class Guest {
    * @param {HTMLElement} element -
    *   The root element in which the `Guest` instance should be able to anchor
    *   or create annotations. In an ordinary web page this typically `document.body`.
-   * @param {EventBus} eventBus -
```
robertknight, line 125: There is a reference to `EventBus` in the typedefs in this file which needs to be removed.

esanzgar: Thanks, I remove it the reference.

PR #4044, opened 2021-12-15
Reduce `Sidebar` dependency on `Guest`

esanzgar:
Replace the 'sidebarLayoutChanged' event emitter for an RPC call using
the `guest-host` communication channel. This change is a step further
into be able to instantiate `Sidebar` independently from the `Guest`.
This may be desirable in scenarios that has several iframes, like Ebooks
and VitalSource.

In the second commit I make a few general improvements to the `Sidebar` and `Guest` tests.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -342,6 +342,16 @@ export default class Guest {
       }
     );
 
+    this._hostRPC.on(
+      'sidebarLayoutChanged',
+      /** @param {SidebarLayout} sidebarLayout */
+      sidebarLayout => {
+        if (this._frameIdentifier === null) {
+          this.fitSideBySide(sidebarLayout);
+        }
+      }
+    );
+
     // Discover and connect to the host frame. All RPC events must be
     // registered before creating the channel.
     const hostPort = await this._portFinder.discover('host');
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -134,8 +134,6 @@ export default class Sidebar {
       sendErrorsTo(this.iframe.contentWindow);
     }
 
-    this.guest = guest;
-
     this._listeners = new ListenerCollection();
 
     // Set up the toolbar on the left edge of the sidebar.
@@ -430,9 +428,7 @@ export default class Sidebar {
       this.onLayoutChange(layoutState);
     }
 
-    this.guest.fitSideBySide(layoutState);
-
-    this._emitter.publish('sidebarLayoutChanged', layoutState);
+    this._guestRPC.call('sidebarLayoutChanged', layoutState);
   }
 
   /**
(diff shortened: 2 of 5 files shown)


src/annotator/sidebar.js
```diff
@@ -426,9 +424,7 @@ export default class Sidebar {
       this.onLayoutChange(layoutState);
     }
 
-    this.guest.fitSideBySide(layoutState);
-
-    this._emitter.publish('sidebarLayoutChanged', layoutState);
```
esanzgar, line 431: Currently, it was doing nothing.

src/annotator/guest.js
```diff
@@ -341,6 +341,16 @@ export default class Guest {
       }
     );
 
+    this._hostRPC.on(
+      'sidebarLayoutChanged',
+      /** @param {SidebarLayout} sidebarLayout */
+      sidebarLayout => {
+        if (this._frameIdentifier === null) {
```
esanzgar, line 348: Reacting to the sidebar layout changes makes sense only to the `Guest` that is in the `host` frame. Currently, we know that's the case if `frameIdentifier` is `null`. In the future however, that can change, which will require adaptation.

src/annotator/test/guest-test.js
```diff
@@ -754,18 +804,18 @@ describe('Guest', () => {
   });
 
   describe('#createAnnotation', () => {
-    it('creates an annotation if host calls "createAnnotationIn" RPC method', () => {
-      const guest = createGuest();
-      sinon.stub(guest, 'createAnnotation');
```
esanzgar, line 759: Remove this reliance on mocking a public method. Instead a check that the 'createAnnotation' RPC event is sent to the sidebar.

src/annotator/test/sidebar-test.js
```diff
@@ -599,29 +618,29 @@ describe('Sidebar', () => {
       const { port1 } = new MessageChannel();
       sidebar.onFrameConnected('dummy', port1);
 
-      assert.notCalled(fakeBridge.createChannel);
+      assert.notCalled(sidebarBridge().createChannel);
```
robertknight, line 621: You'll also need an `assert.notCalled(guestBridge().createChannel)` check here.

robertknight (approved):
LGTM. I think it may be useful to create a shared fake bridge class which can be used by multiple tests in future. I noted one missing assert in a test.

src/annotator/test/sidebar-test.js
```diff
@@ -18,7 +18,8 @@ describe('Sidebar', () => {
   let containers;
   let sidebars;
 
-  let fakeBridge;
+  let FakeBridge;
```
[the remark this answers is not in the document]
esanzgar: Like in `frame-sync-test.js`?

src/annotator/test/sidebar-test.js
esanzgar: That's a detailed catch! Thanks, fixed.

PR #4048, opened 2021-12-16
Move scroll and resize listeners to their own service

esanzgar:
This is a step towards making the bucket bar works for a `guest` frame
that is not in the `host` frame. When that's the case the observation of
scrolling and resize can't happen in the `host` frame.

A new service has been created, `BucketClient`. The service is
instantiated only in the `guest` frame that contains the main
annotatable content. The service observes changes in the elements that
could impact the relative position of the anchors relative to the
viewport. If these changes are detected, as well as addition or deletion
of anchors, it sends the `anchorsChanged` RPC event to the `host` frame
so that the positions can be recalculated. As a result of the new
service the `BucketBar` has been simplified.

### DIFF
--- a/src/annotator/bucket-bar-client.js
+++ b/src/annotator/bucket-bar-client.js
@@ -1,0 +1,58 @@
+import { ListenerCollection } from '../shared/listener-collection';
+
+/**
+ * @typedef {import('../shared/bridge').Bridge<GuestToHostEvent, HostToGuestEvent>} HostBridge
+ * @typedef {import('../types/annotator').Destroyable} Destroyable
+ * @typedef {import('../types/bridge-events').HostToGuestEvent} HostToGuestEvent
+ * @typedef {import('../types/bridge-events').GuestToHostEvent} GuestToHostEvent
+ */
+
+/**
+ * Communicate to the `host` frame when:
+ * 1. The set of anchors has been changed (due to annotations being added or removed)
+ * 2. The position of anchors relative to the viewport of the guest has changed
+ *
+ * @implements Destroyable
+ */
+export class BucketBarClient {
+  /**
+   * @param {object} options
+   *   @param {Element} options.contentContainer - The scrollable container element for the
+   *     document content. All of the highlights that the bucket bar's buckets point
+   *     at should be contained within this element.
+   *   @param {HostBridge} options.hostRPC
+   */
+  constructor({ contentContainer, hostRPC }) {
+    this._hostRPC = hostRPC;
+    this._updatePending = false;
+    this._listeners = new ListenerCollection();
+
+    this._listeners.add(window, 'resize', () => this.update());
+    this._listeners.add(window, 'scroll', () => this.update());
+    this._listeners.add(contentContainer, 'scroll', () => this.update());
+  }
+
+  destroy() {
+    this._listeners.removeAll();
+  }
+
+  /**
+   * Notifies the BucketBar in the host frame when:
+   * 1. The set of anchors has been changed (due to annotations being added or removed)
+   * 2. The position of anchors relative to the viewport of the guest has changed
+   *
+   * Updates are debounced to reduce the overhead of gathering and sending anchor
+   * position data across frames.
+   */
+  update() {
+    if (this._updatePending) {
+      return;
+    }
+
+    this._updatePending = true;
+    requestAnimationFrame(() => {
+      this._hostRPC.call('anchorsChanged');
+      this._updatePending = false;
+    });
+  }
+}
--- a/src/annotator/bucket-bar-client.js
+++ b/src/annotator/bucket-bar-client.js
@@ -8,7 +8,8 @@ import { ListenerCollection } from '../shared/listener-collection';
  */
 
 /**
- * Communicate to the `host` frame when:
+ * Communicate to the host frame when:
+ *
  * 1. The set of anchors has been changed (due to annotations being added or removed)
  * 2. The position of anchors relative to the viewport of the guest has changed
  *
--- a/src/annotator/bucket-bar.js
+++ b/src/annotator/bucket-bar.js
@@ -1,16 +1,9 @@
 import { render } from 'preact';
 
-import { ListenerCollection } from '../shared/listener-collection';
-
 import Buckets from './components/Buckets';
 import { anchorBuckets } from './util/buckets';
 
 /**
- * @typedef BucketBarOptions
- * @prop {Element} [contentContainer] - The scrollable container element for the
- *   document content. All of the highlights that the bucket bar's buckets point
- *   at should be contained within this element.
- *
  * @typedef {import('../types/annotator').Destroyable} Destroyable
  */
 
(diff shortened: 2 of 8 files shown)


src/annotator/guest.js
```diff
@@ -739,15 +750,6 @@ export default class Guest {
...
 
-  /**
-   * Return the scrollable element that contains the main document content.
-   *
-   * @return {HTMLElement}
-   */
-  contentContainer() {
-    return this._integration.contentContainer();
-  }
-
```
esanzgar, line 750: This public method is not longer needed.

src/annotator/guest.js
```diff
@@ -577,7 +587,8 @@ export default class Guest {
    * Remove the anchors and associated highlights for an annotation from the document.
    *
    * @param {string} tag
-   * @param {boolean} [notify] - For internal use. Whether to emit an `anchorsChanged` notification
+   * @param {boolean} [notify] - For internal use. Whether to inform the `host`
```
robertknight, line 590:
```suggestion
   * @param {boolean} [notify] - For internal use. Whether to inform the host
```

"host frame" is a concept rather than an identifier. If you are introducing a concept in documentation, you can use italics for that. In this case though I think we can assume that the concept of a host vs guest is explained elsewhere.

src/annotator/test/bucket-bar-test.js
```diff
@@ -1,34 +1,27 @@
-import BucketBar from '../bucket-bar';
-import { $imports } from '../bucket-bar';
+import BucketBar, { $imports } from '../bucket-bar';
```
robertknight, line 1: At some point we should convert BucketBar to a named export.

src/annotator/test/bucket-service-test.js
```diff
@@ -0,0 +1,95 @@
+import { delay } from '../../test-util/wait';
+import { BucketService } from '../bucket-service';
+
+describe('BucketBar', () => {
```
robertknight, line 4: This doesn't match the name of the class.

robertknight (changes requested):
The overall structure of having the `Guest` use a helper class which acts as a client for the bucket bar, which in turn gathers the relevant information and forwards it to the host frame, makes sense. Moving the debouncing to the guest frame also makes sense.

I mostly had feedback on the name and documentation of the `BucketService` class, as well as handling `requestAnimationFrame` in tests. There was also a behavioral query about `requestAnimationFrame` which would be useful to clarify.

src/annotator/test/bucket-service-test.js
```diff
@@ -0,0 +1,95 @@
...
+    });
+    bucketServices.push(bucketService);
+    return bucketService;
+  };
+
+  it('should update buckets when the window is resized', async () => {
+    createBucketService();
+
+    window.dispatchEvent(new Event('resize'));
+    await delay(20);
```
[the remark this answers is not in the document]
esanzgar: I agree. I stubbed `requestAnimationFrame` by a `setTimeout` that is triggered on the next event loop.

src/annotator/bucket-service.js
```diff
@@ -0,0 +1,64 @@
...
+  _update() {
+    this._hostRPC.call('anchorsChanged');
+  }
+
+  destroy() {
+    this._listeners.removeAll();
+  }
+
+  /**
+   * Throttles the `_update` method to be invoked at most once per 1s/60 or
```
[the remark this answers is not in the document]
esanzgar: > Regarding hidden frames and backgrounded tabs, does the callback not run at all or is it merely deferred until the frame becomes visible?

It is deferred until it is visible.

esanzgar (commented):
I have addressed your comments. Please, let me know what you think.

src/annotator/bucket-bar.js
```diff
@@ -1,66 +1,33 @@
...
    * @param {Pick<import('./guest').default, 'anchors'|'scrollToAnchor'|'selectAnnotations'>} guest
-   * @param {BucketBarOptions} [options]
    */
-  constructor(container, guest, { contentContainer = document.body } = {}) {
-    this._contentContainer = contentContainer;
-
-    this._bucketsContainer = document.createElement('div');
-    container.appendChild(this._bucketsContainer);
+  constructor(container, guest) {
+    this._bucketBar = document.createElement('bucket-bar');
```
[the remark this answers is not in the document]
robertknight: > Do you see any problems changing the div to bucket-bar?

Technically not really, but we don't do this anywhere else, so the use of a custom element implies there is some functional reason which doesn't actually exist. I would suggest to revert this change. If we need to reference the bucket bar in tests, use a `data-testid`.

src/annotator/bucket-service.js
```diff
@@ -0,0 +1,64 @@
...
+    this._updatePending = false;
+    this._listeners = new ListenerCollection();
+
+    this._listeners.add(window, 'resize', () => this.update());
+    this._listeners.add(window, 'scroll', () => this.update());
+    this._listeners.add(contentContainer, 'scroll', () => this.update());
+  }
+
+  _update() {
+    this._hostRPC.call('anchorsChanged');
```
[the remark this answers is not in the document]
robertknight: > The anchorsChanged is also sent when an annotation is anchored or detached.

Right, my thinking was that adding or removing an annotation obviously implies the set of positions/locations where anchors exist has also changed.

src/annotator/test/bucket-bar-test.js
robertknight: > But yes, we can be more strict, and everything that is not in a components folder could be converted to a named export. That would include Sidebar, Guest and others.

I think we should do this. Applying the strict rule that only Preact components, from a file of the same name, use default exports, is in line with our [JS style guide](https://github.com/hypothesis/frontend-toolkit/blob/master/docs/js-guide.md#module-naming) and is an easier convention for future devs to follow.

src/annotator/bucket-bar-client.js
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * @typedef {import('../shared/bridge').Bridge<GuestToHostEvent, HostToGuestEvent>} HostBridge
+ * @typedef {import('../types/annotator').Destroyable} Destroyable
+ * @typedef {import('../types/bridge-events').HostToGuestEvent} HostToGuestEvent
+ * @typedef {import('../types/bridge-events').GuestToHostEvent} GuestToHostEvent
+ */
+
+/**
+ * Communicate to the `host` frame when:
+ * 1. The set of anchors has been changed (due to annotations being added or removed)
```
robertknight, line 12:
```suggestion
 * Communicate to the host frame when:
 *
 * 1. The set of anchors has been changed (due to annotations being added or removed)
```

src/annotator/bucket-bar-client.js
```diff
@@ -0,0 +1,58 @@
...
+ * 2. The position of anchors relative to the viewport of the guest has changed
+ *
+ * @implements Destroyable
```
robertknight, line 15: I notice we mix `@implements Name` and `@implements {Name}` in our code. It seems that TS accepts either, but we should standardize. I'm inclined to include the parens as this seems to be the preferred style from the [TS handbook](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#implements).

src/annotator/bucket-bar-client.js
```diff
@@ -0,0 +1,58 @@
...
+   * Updates are debounced to reduce the overhead of gathering and sending anchor
+   * position data across frames.
+   */
+  update() {
+    if (this._updatePending) {
+      return;
+    }
+
+    this._updatePending = true;
+    requestAnimationFrame(() => {
```
robertknight, line 53: The `destroy` method should really cancel any pending update, though it likely won't ever matter in the context of the app.

src/annotator/test/guest-test.js
```diff
@@ -1273,16 +1284,24 @@ describe('Guest', () => {
...
+
+    createGuest();
+
+    assert.calledWith(FakeBucketBarClient, {
+      contentContainer,
+      hostRPC: hostBridge(),
     });
   });
 
+  it('does not configure the BucketBarClient if guest is the main annotatable frame', () => {
```
robertknight, line 1299:
```suggestion
  it('does not configure the BucketBarClient if guest is not the main annotatable frame', () => {
```

robertknight (approved):
LGTM with some very minor comments and corrections.

esanzgar (commented):
I acted on your suggestions, thanks.

Issue #4046, opened 2021-12-15
`allowUrls` Sentry configuration is not set correctly


robertknight:
We use the [allowUrls](https://docs.sentry.io/platforms/javascript/configuration/options/#allow-urls) configuration option in Sentry to try and filter out errors in third party code. This is not currently set correctly as the `currentScriptOrigin` helper is returning `null`. This helper relies on [document.currentScript](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) which, per the MDN docs, does not work in scripts loaded using `<script type="module">`. We need to use `import.meta.url` instead.

PR #4049, opened 2021-12-16
Fix script origin detection in Sentry initialization

robertknight:
The Sentry initialization in the sidebar used `document.currentScript` which does not work for module scripts. As a result the [`allowUrls`](https://docs.sentry.io/platforms/javascript/configuration/options/#allow-urls) Sentry configuration was not set as intended. Replace this with [`import.meta.url`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta).

Fixing the `allowUrls` Sentry configuration will hopefully prevent some reports from external code (eg. browser extensions) from being sent to Sentry. I'll verify after deployment.

Fixes https://github.com/hypothesis/client/issues/4046.


src/tsconfig.json
```diff
@@ -5,7 +5,8 @@
     "lib": ["es2020", "dom", "dom.iterable"],
     "jsx": "react-jsx",
     "jsxImportSource": "preact",
-    "module": "commonjs",
+    "module": "es2020",
+    "moduleResolution": "node",
```
robertknight, line 9: The `module` configuration update is needed to use `import.meta`. The `moduleResolution` change is needed to allow the use of Node-style import paths (eg. without a file extension, and lookup in `node_modules` if the path doesn't start with `./`) together with `es2020`.

src/sidebar/util/test/sentry-test.js
```diff
@@ -1,8 +1,6 @@
 import * as sentry from '../sentry';
 
 describe('sidebar/util/sentry', () => {
-  let fakeDocumentReferrer;
```
robertknight, line 4: This is a fake for `document.referrer` which used to be used by `sentry.js` but no longer is.

src/sidebar/util/test/sentry-test.js
```diff
@@ -38,22 +36,15 @@ describe('sidebar/util/sentry', () => {
...
 
+    beforeEach(() => {
       // Reset rate limiting counters.
       sentry.reset();
     });
 
     afterEach(() => {
-      fakeDocumentCurrentScript.restore();
-      fakeDocumentReferrer.restore();
+      import.meta.url = originalURL;
```
robertknight, line 47: It turns out that the native `import.meta.url` property is, conveniently, writable. Hence we can modify it during tests.

esanzgar: Interesting...

esanzgar (approved):
Very good to know about the `import.meta`.

src/sidebar/util/test/sentry-test.js
```diff
@@ -100,8 +94,8 @@ describe('sidebar/util/sentry', () => {
       );
     });
 
-    it('disables the URL allowlist if `document.currentScript` is inaccessible', () => {
-      fakeDocumentCurrentScript.get(() => null);
+    it('disables the URL allowlist if the script URL is unavailable', () => {
+      import.meta.url = null;
```
esanzgar, line 98: Is it `null` for sandboxed iframes?

robertknight: The URL refers to the script rather than the document, so it won't be null in a sandboxed iframe. Re-reading the [relevant spec](https://html.spec.whatwg.org/multipage/webappapis.html#hostgetimportmetaproperties) it doesn't look like the URL _can_ be falsey. It is possible for `URL(import.meta.url).origin` to be the string "null" if the script is an inline script and the document URL is `about:blank`. This script is always loaded externally though, so that shouldn't happen.

robertknight: I suppose some third-party JS on the page could manipulate this URL, as we found is possible to do in the tests.

PR #4050, opened 2021-12-16
Reduce boot script cache expiry time

robertknight:
Reduce the potential for mismatches between the client versions loaded in the annotator and sidebar when rolling out a new release by shortening the cache expiry time on the boot script.

Set a very short expiry time on the browser cache TTL and a slightly longer time on the edge cache.

### DIFF
--- a/scripts/deploy-to-s3.js
+++ b/scripts/deploy-to-s3.js
@@ -132,7 +132,7 @@ async function uploadPackageToS3(bucket, options) {
     // nb. When deploying to cdn.hypothes.is, the max-age seen by the browser is
     // the higher of the value here and CloudFlare's "Browser Cache TTL"
     // setting.
-    aliasCacheControl = 'public, max-age=1800, must-revalidate';
+    aliasCacheControl = 'public, s-maxage=300, max-age=60, must-revalidate';
   } else {
     aliasCacheControl = 'no-cache';
   }


scripts/deploy-to-s3.js
```diff
@@ -132,7 +132,7 @@ async function uploadPackageToS3(bucket, options) {
     // nb. When deploying to cdn.hypothes.is, the max-age seen by the browser is
     // the higher of the value here and CloudFlare's "Browser Cache TTL"
     // setting.
-    aliasCacheControl = 'public, max-age=1800, must-revalidate';
+    aliasCacheControl = 'public, s-maxage=300, max-age=60, must-revalidate';
```
robertknight, line 135: `max-age` is for all caches, including the browser. `s-maxage` is specifically for shared caches, meaning Cloudflare.

esanzgar (approved):

esanzgar:
We have verified that this works (click on `Path` in the following [`Live` graph](https://dash.cloudflare.com/304c52f571c1c0a02368cf9618254292/hypothes.is/analytics/traffic?path~ends=%2Fbuild%2Fscripts%2Fsidebar.bundle.js&time-window=30)):

image

The release of the new version happened at 12:35pm. One need to use the `Live` graph to observe changes with frequencies of minutes, otherwise the sampling is longer the one minute and it can still appear that there is around 30 min. caching.

Contrast the above graph with this other and you will notice that versions of the Hypothesis client are cache for much shorter time now:

https://github.com/hypothesis/client/issues/3986#issuecomment-984608181

P.S.: it would have been nice to compare `live` graph to `live` graph so the sampling is the same.

PR #4051, opened 2021-12-17
Remove unused 'sidebarOpened' event emitter

esanzgar:
There is both a 'sidebarOpened' event emitter and RPC event by the same name. The RPC
event is used, while the event emitter is not.

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -499,7 +499,6 @@ export default class Sidebar {
 
   open() {
     this._sidebarRPC.call('sidebarOpened');
-    this._emitter.publish('sidebarOpened');
 
     if (this.iframeContainer) {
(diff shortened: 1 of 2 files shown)


esanzgar:
I can confirm that the `sidebarOpened` is not used.

One possible reason this was left was because, it was included in one test that intended to test the `sidebarLayoutChanged` emitter and that this was included as an explanation for the number of calls to the emitter.

https://github.com/hypothesis/client/blob/d4b659606fa6d0201ee8e1314a5917355aa846ff/src/annotator/test/sidebar-test.js#L781-L787

lyzadanger (approved):
If you feel confident that removing this is safe, go for it! I'd love a tour of how the RPC events are working these days, and understand what is leftover for the `emitter` here...I don't have my head around it at all.

PR #4052, opened 2021-12-17
Increase timeout for port discovery

robertknight:
On a slow device if the host page is heavily loaded with JavaScript, the 10-second timeout can be exceeded. Increase this threshold to 20s to see what effect that has on the volume of reports we receive.

Part of https://github.com/hypothesis/client/issues/3986.

### DIFF
--- a/src/shared/port-finder.js
+++ b/src/shared/port-finder.js
@@ -1,8 +1,11 @@
 import { ListenerCollection } from './listener-collection';
 import { isMessageEqual } from './port-util';
 
-const MAX_WAIT_FOR_PORT = 1000 * 10;
-const POLLING_INTERVAL_FOR_PORT = 250;
+/** Timeout for waiting for the host frame to respond to a port request. */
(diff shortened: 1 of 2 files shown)


src/shared/port-finder.js
```diff
@@ -1,8 +1,11 @@
 import { ListenerCollection } from './listener-collection';
 import { isMessageEqual } from './port-util';
 
-const MAX_WAIT_FOR_PORT = 1000 * 10;
-const POLLING_INTERVAL_FOR_PORT = 250;
+/** Timeout for waiting for the host frame to respond to a port request. */
+export const MAX_WAIT_FOR_PORT = 1000 * 20;
```
esanzgar, line 5: I will go for even longer, 30 seconds.

On my machine loading on an iframe https://html.spec.whatwg.org/ can take around 10 seconds, if you add a slow network even minutes.

esanzgar (approved):

PR #4055, opened 2021-12-17
Improve uniformity in tests

esanzgar:
I included two utility functions, similar to `sidebar-test.js` and
`guest-test.js`, to send RPC calls. This provides more uniformity
testing inter-frame communication in these three classes.


lyzadanger (approved):
@robertknight is assigned as reviewer, but this does look straightforward and fine to me.

PR #4056, opened 2021-12-17
Create utility function to avoid repetition

esanzgar:
A minor streamline improvement.


esanzgar:
This is a minor improvement not worth to move forward at this time.

PR #4069, opened 2021-12-22
Focus on annotations rely on inter-frame communication

esanzgar:
`BucketBar` was setting the focus when hovering on the left-pointed
buckets. That relied on direct access to the anchors, which it is not
always possible. Instead, now the `BucketBar` communicates with a new
RPC event, `focusAnnotations`, in the `guest-host` channel (analogous to
the `focusAnnotations` in the `sidebar-guest` channel.

### New
I have made a small improvement: when hovering a left-pointed bucket,
not only focus the anchor's highlights but also the corresponding
annotation cards in the sidebar. The same can be done for top and
bottom-pointed buckets.

https://user-images.githubusercontent.com/8555781/147094899-e40a022c-b045-4d0f-8ba2-9456389a936d.mov

### DIFF
--- a/src/annotator/test/guest-test.js
+++ b/src/annotator/test/guest-test.js
@@ -203,6 +203,20 @@ describe('Guest', () => {
       });
     });
 
+    describe('on "focusAnnotations" event', () => {
+      it('calls "Guest#._focusAnnotations"', () => {
+        const guest = createGuest();
+        sandbox.stub(guest, '_focusAnnotations').callThrough();
+        const tags = ['t1', 't2'];
+        sidebarRPC().call.resetHistory();
+
+        emitHostEvent('focusAnnotations', tags);
+
+        assert.calledWith(guest._focusAnnotations, tags);
+        assert.calledWith(sidebarRPC().call, 'focusAnnotations', tags);
+      });
+    });
+
     describe('on "selectAnnotations" event', () => {
       it('calls "Guest#selectAnnotations"', () => {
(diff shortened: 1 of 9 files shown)


src/annotator/test/guest-test.js
```diff
@@ -204,6 +204,20 @@ describe('Guest', () => {
       });
     });
 
+    describe('on "focusAnnotations" event', () => {
+      it('calls "Guest#._focusAnnotations"', () => {
+        const guest = createGuest();
+        sandbox.stub(guest, '_focusAnnotations').callThrough();
```
robertknight, line 210: It would be preferable if this test didn't refer to internal methods, though you also don't to repeat the tests for the "focusAnnotations" event from the host frame. One way to handle this would be to do something similar to `selectAnnotations` where there is a public method for focusing annotations, plus tests that check that this public method is invoked in response to various events. One benefit of this is that it might enable the code that handles events from the host/sidebar frames to be separated into a helper class in future to make the overall Guest class easier to follow.

robertknight (approved):
LGTM. I did have some design feedback regarding testing of the focusing behavior in the guest.

src/annotator/test/guest-test.js
esanzgar: My intent is not to test the `_focusAnnotations` method but make sure that the `focusAnnotations` RPC event triggers the method. I could test the final outcome.

PR #4070, opened 2021-12-22
Add `scrollToClosestOffScreenAnchor` RPC event

esanzgar:
In order to decouple `BucketBar` from `Guest` we need to send RPC events
through the `guest-host` inter-frame communication channel.  The
messages must trigger identical behaviour on the `Guest`.

In this PR, we replace the direct use of `Guest#scrollToAnchor` by a new
RPC event, `scrollToClosestOffScreenAnchor`.

### DIFF
--- a/src/annotator/components/Buckets.js
+++ b/src/annotator/components/Buckets.js
@@ -1,10 +1,6 @@
 import classnames from 'classnames';
 
-import { findClosestOffscreenAnchor } from '../util/buckets';
-
 /**
- * @typedef {import('../../types/annotator').AnnotationData} AnnotationData
- * @typedef {import('../../types/annotator').Anchor} Anchor
  * @typedef {import('../util/buckets').Bucket} Bucket
  */
 
@@ -13,9 +9,9 @@ import { findClosestOffscreenAnchor } from '../util/buckets';
  * or selects associated annotations.
  *
  * @param {object} props
- *  @param {Bucket} props.bucket
- *  @param {(tags: string[]) => void} props.onFocusAnnotations
- *  @param {(tags: string[], toggle: boolean) => void} props.onSelectAnnotations
+ *   @param {Bucket} props.bucket
+ *   @param {(tags: string[]) => void} props.onFocusAnnotations
+ *   @param {(tags: string[], toggle: boolean) => void} props.onSelectAnnotations
  */
 function BucketButton({ bucket, onFocusAnnotations, onSelectAnnotations }) {
   const buttonTitle = `Select nearby annotations (${bucket.anchors.length})`;
@@ -55,18 +51,19 @@ function BucketButton({ bucket, onFocusAnnotations, onSelectAnnotations }) {
  *
  * @param {object} props
  *   @param {Bucket} props.bucket
- *   @param {'up'|'down'} props.direction
- *   @param {(a: Anchor) => void} props.scrollToAnchor - Callback invoked to
- *     scroll the document to a given anchor
+ *   @param {'down'|'up'} props.direction
+ *   @param {(tags: string[], direction: 'down'|'up') => void} props.onScrollToClosestOffScreenAnchor
  */
-function NavigationBucketButton({ bucket, direction, scrollToAnchor }) {
+function NavigationBucketButton({
+  bucket,
+  direction,
+  onScrollToClosestOffScreenAnchor,
+}) {
   const buttonTitle = `Go ${direction} to next annotations (${bucket.anchors.length})`;
 
   function scrollToClosest() {
-    const closest = findClosestOffscreenAnchor(bucket.anchors, direction);
-    if (closest) {
-      scrollToAnchor(closest);
-    }
+    const tags = bucket.anchors.map(anchor => anchor.annotation.$tag);
+    onScrollToClosestOffScreenAnchor(tags, direction);
   }
 
   return (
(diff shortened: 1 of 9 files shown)


src/types/bridge-events.d.ts
```diff
@@ -88,7 +88,12 @@ export type HostToGuestEvent =
   /**
    * The host informs guests that the sidebar layout has been changed.
    */
-  | 'sidebarLayoutChanged';
+  | 'sidebarLayoutChanged'
+
+  /**
+   * The host informs guests to scroll to the closest annotation from the set.
+   */
+  | 'scrollToAnnotations';
```
esanzgar, line 96: This name is very similar to `sidebar-host` event, `scrollToAnnotation`. On acts on a single annotation tag, but this new event acts on a list of annotation tags.

I am open to alternative names: `scrollToClosestAnnotation` for example.

src/annotator/test/guest-test.js
```diff
@@ -120,6 +121,14 @@ describe('Guest', () => {
...
     FakeBucketService = sinon.stub().returns(fakeBucketService);
 
+    fakeFindClosestOffscreenAnchor = sinon.stub();
+
+    fakeHypothesisInjector = {
+      destroy: sinon.stub(),
+      injectClient: sinon.stub().resolves(),
+    };
+    FakeHypothesisInjector = sinon.stub().returns(fakeHypothesisInjector);
+
```
esanzgar, line 131: Just ordering alphabetically.

src/annotator/guest.js
```diff
@@ -739,15 +765,6 @@ export default class Guest {
...
 
-  /**
-   * Scroll the document content so that `anchor` is visible.
-   *
-   * @param {Anchor} anchor
-   */
-  scrollToAnchor(anchor) {
-    return this._integration.scrollToAnchor(anchor);
-  }
-
```
esanzgar, line 750: This public method is not longer needed.

src/annotator/guest.js
```diff
@@ -18,6 +18,7 @@ import { HypothesisInjector } from './hypothesis-injector';
 import { createIntegration } from './integrations';
 import * as rangeUtil from './range-util';
 import { SelectionObserver } from './selection-observer';
+import { findClosestOffscreenAnchor } from './util/buckets';
```
esanzgar, line 21: I am happy to move this piece of code to another place if it makes more sense.

robertknight (commented):
There is a 1:N association between annotations and anchors, where each anchor corresponds to a different entry in an annotation's `target` array. This is the case at least in the data model, although the h backend only supports one target per annotation and a few places in the client code treat the first entry in the array specially. Focusing and selecting actions apply to an annotation and all of its anchors, whereas scrolling targets a specific anchor. We need to make an intentional call about how we're going to handle this:

1. Allow the bucket bar to distinguish which anchor, rather than just annotation, should be scrolled to? (eg. by giving each anchor an ID or by using a combination of annotation tag and target index to identify an anchor)
1. Restrict the bucket bar to only show the first anchor for each annotation?
1. Something else?

src/annotator/guest.js
```diff
@@ -682,6 +692,22 @@ export default class Guest {
     this._sidebarRPC.call('focusAnnotations', tags);
   }
 
+  /**
+   * Scroll to the closest off screen anchor.
```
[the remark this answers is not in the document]
esanzgar: 👍🏼

src/annotator/components/Buckets.js
```diff
@@ -55,18 +51,20 @@ function BucketButton({ bucket, onFocusAnnotations, onSelectAnnotations }) {
  *
  * @param {object} props
  *   @param {Bucket} props.bucket
- *   @param {'up'|'down'} props.direction
- *   @param {(a: Anchor) => void} props.scrollToAnchor - Callback invoked to
- *     scroll the document to a given anchor
+ *   @param {'down'|'up'} props.direction
+ *   @param {(tags: string[], direction: 'down'|'up') => void} props.onScrollToClosestOffScreenAnchor
+ *     - Callback invoked to scroll the document to a given anchor
```
robertknight, line 56: The comment refers to _anchor_ singular, but there are multiple tags. IMO the purpose of this prop is clear from the name, so I'd just remove the description.

src/annotator/components/Buckets.js
```diff
@@ -90,17 +88,17 @@ function NavigationBucketButton({ bucket, direction, scrollToAnchor }) {
  *   @param {Bucket} props.below
  *   @param {Bucket[]} props.buckets
  *   @param {(tags: string[]) => void} props.onFocusAnnotations
+ *   @param {(tags: string[], direction: 'down'|'up') => void} props.onScrollToClosestOffScreenAnchor
+ *     - Callback invoked to scroll the document to a given anchor
```
robertknight, line 92: Same comment as for the other `onScrollToClosestOffScreenAnchor` prop.

src/annotator/guest.js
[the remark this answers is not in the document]
robertknight: Ah, I see.

src/annotator/test/guest-test.js
```diff
@@ -220,6 +232,29 @@ describe('Guest', () => {
       });
     });
 
+    describe('on "scrollToClosestOffScreenAnchor" event', () => {
+      it('calls "Guest#._scrollToClosestOffScreenAnchor"', () => {
```
robertknight, line 236: The test description references an internal method, although the test itself only checks interaction with external classes, which is the right thing to do. I would suggest to reword this description as "scrolls to the nearest anchor in the given set" or similar.

src/types/port-rpc-events.d.ts
```diff
@@ -91,7 +91,12 @@ export type HostToGuestEvent =
   /**
    * The host informs guests that the sidebar layout has been changed.
    */
-  | 'sidebarLayoutChanged';
+  | 'sidebarLayoutChanged'
+
+  /**
+   * The host informs guests to scroll to the closest off-screen anchor from the set of annotations.
```
robertknight, line 97:
```suggestion
   * The host informs guests to scroll to the closest off-screen anchor associated with a set of annotations.
```

robertknight (approved):
This looks great. Just some small comments regarding a test description and some prop descriptions.

src/annotator/components/Buckets.js
esanzgar: 👍🏼

src/annotator/test/guest-test.js
esanzgar: 👍🏼

src/types/port-rpc-events.d.ts
esanzgar: 👍🏼

esanzgar (commented):
I applied all the suggestions.

PR #4071, opened 2021-12-23
Remove anchors from the `BucketSet` data structure

esanzgar:
This PR makes the `Buckets` component to accept a simplified data
structure that doesn't contain `anchor`s. The `anchor` objects are no
longer needed and they can't be serialised (because they contain DOM
elements).

On the second commit, I changed the term `anchor` to `annotation` when referring to certain events.

### DIFF
--- a/src/annotator/util/buckets.js
+++ b/src/annotator/util/buckets.js
@@ -14,7 +14,7 @@ import { getBoundingClientRect } from '../highlighter';
 /**
  * @typedef BucketSet
  * @prop {Bucket} above - A single bucket containing all the annotation
- *   tags which anchors are offscreen upwards
+ *   tags whose anchors are offscreen upwards
  * @prop {Bucket} below - A single bucket containing all the annotation
  *   tags which anchors are offscreen downwards
  * @prop {Bucket[]} buckets - On-screen buckets
--- a/src/annotator/util/buckets.js
+++ b/src/annotator/util/buckets.js
@@ -6,27 +6,27 @@ import { getBoundingClientRect } from '../highlighter';
 
 /**
  * @typedef Bucket
- * @prop {Anchor[]} anchors - The anchors in this bucket
+ * @prop {Set<string>} tags - The annotation tags in this bucket
  * @prop {number} position - The vertical pixel offset where this bucket should
- *                           appear in the bucket bar.
+ *   appear in the bucket bar
  */
 
 /**
  * @typedef BucketSet
- * @prop {Bucket} above - A single bucket containing all of the anchors that
- *                        are offscreen upwards
- * @prop {Bucket} below - A single bucket containing all of the anchors that are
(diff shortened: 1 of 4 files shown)


src/annotator/util/buckets.js
```diff
@@ -166,7 +169,7 @@ export function anchorBuckets(anchors) {
     const anchorHeight = anchorPosition.bottom - anchorPosition.top;
     const bucketPosition = anchorPosition.top + anchorHeight / 2;
     const bucket = /** @type WorkingBucket */ ({
-      anchors: [anchorPosition.anchor],
+      tags: new Set([anchorPosition.anchor.annotation.$tag]),
```
esanzgar, line 172: For uniformity, I decided to make it also a `Set`

src/annotator/util/test/buckets-test.js
```diff
@@ -116,44 +125,32 @@ describe('annotator/util/buckets', () => {
   describe('anchorBuckets', () => {
     it('puts anchors that are above the screen into the `above` bucket', () => {
       const bucketSet = anchorBuckets(fakeAnchors);
-      assert.deepEqual(bucketSet.above.anchors, [
-        fakeAnchors[0],
-        fakeAnchors[1],
-      ]);
+      assert.deepEqual([...bucketSet.above.tags], [getTag(0), getTag(1)]);
```
esanzgar, line 128: If the conversion of a `Set` to an array changes the order of the elements this test will fail.

src/annotator/util/test/buckets-test.js
```diff
@@ -58,6 +62,11 @@ describe('annotator/util/buckets', () => {
     stubbedInnerHeight.restore();
   });
 
+  /** @param {number} index */
+  const getTag = index => {
```
esanzgar, line 66: Maybe `getTagFromFakeBucket` would be clearer.

robertknight (approved):
LGTM.

I have some thoughts related to annotation tags and supporting annotations with multiple targets. I think we can afford not to support this for now, but I'll mention it in case it comes up in future.

The `Annotation` data structure allows for an annotation to have multiple targets, and thus multiple anchors. This means that using an annotation tag to identify an anchor is technically ambiguous. However all annotations that the client creates or that are stored in h will currently only have one target and thus one anchor. The assumption of one anchor per annotation has probably crept into the code in other ways by now.

I think what we could probably do here is to use annotation tags to identify anchors in the bucket bar for now, and if and when we have a use case for multiple anchors per annotation, replace the tags with anchor-specific IDs.

esanzgar:
> The Annotation data structure allows for an annotation to have multiple targets, and thus multiple anchors. This means that using an annotation tag to identify an anchor is technically ambiguous.

I totally get this and I agree about replacing the tags with anchor-specific IDs.

PR #4072, opened 2021-12-24
Compute buckets on two steps (guest/host frames)

esanzgar:
~~Depends on #4071~~ 

The functionality to compute the buckets has been split in two steps:

1. Guest frame computes and sends a list of the anchors's top/bottom
   positions with the `anchorsChanged` RPC event.

2. In the host frame, `BucketBar` uses the list anchor's positions to
   compute the buckets.

There is one issue with the current approach: if an iframe doesn't cover
the entire host frame's viewport buckets can appear offset.

This is a video that issue:

https://user-images.githubusercontent.com/8555781/147335655-0223571e-57ca-4a28-8278-5b5878729814.mov

### DIFF
--- a/src/annotator/bucket-bar-client.js
+++ b/src/annotator/bucket-bar-client.js
@@ -65,8 +65,6 @@ export class BucketBarClient {
 
     this._updatePending = true;
     requestAnimationFrame(() => {
-      // Computing anchor positions may be expensive when there are many
-      // annotations or a forced layout reflow
       const positions = computeAnchorPositions(this._anchors);
       this._hostRPC.call('anchorsChanged', positions);
       this._updatePending = false;
--- a/src/annotator/bucket-bar-client.js
+++ b/src/annotator/bucket-bar-client.js
@@ -65,8 +65,8 @@ export class BucketBarClient {
 
     this._updatePending = true;
     requestAnimationFrame(() => {
-      // In document with many annotations computing the anchor positions can
-      // block the JS event loop for up to 200 ms.
+      // Computing anchor positions may be expensive when there are many
+      // annotations or a forced layout reflow
       const positions = computeAnchorPositions(this._anchors);
       this._hostRPC.call('anchorsChanged', positions);
       this._updatePending = false;
--- a/src/annotator/bucket-bar-client.js
+++ b/src/annotator/bucket-bar-client.js
@@ -1,7 +1,11 @@
 import { ListenerCollection } from '../shared/listener-collection';
 
+import { computeAnchorPositions } from './util/buckets';
+
 /**
  * @typedef {import('../shared/messaging').PortRPC<HostToGuestEvent, GuestToHostEvent>} HostRPC
+ * @typedef {import('../types/annotator').Anchor} Anchor
+ * @typedef {import('../types/annotator').AnchorPosition} AnchorPosition
  * @typedef {import('../types/annotator').Destroyable} Destroyable
  * @typedef {import('../types/port-rpc-events').HostToGuestEvent} HostToGuestEvent
  * @typedef {import('../types/port-rpc-events').GuestToHostEvent} GuestToHostEvent
@@ -26,6 +30,8 @@ export class BucketBarClient {
   constructor({ contentContainer, hostRPC }) {
     this._hostRPC = hostRPC;
     this._updatePending = false;
+    /** @type {Anchor[]} */
+    this._anchors = [];
     this._listeners = new ListenerCollection();
 
     this._listeners.add(window, 'resize', () => this.update());
@@ -44,15 +50,25 @@ export class BucketBarClient {
    *
    * Updates are debounced to reduce the overhead of gathering and sending anchor
    * position data across frames.
+   *
+   * @param {Anchor[]} [anchors] - pass this option when anchors are added or
+   *   deleted
    */
-  update() {
+  update(anchors) {
+    if (anchors) {
+      this._anchors = anchors;
+    }
+
     if (this._updatePending) {
       return;
     }
 
     this._updatePending = true;
     requestAnimationFrame(() => {
-      this._hostRPC.call('anchorsChanged');
+      // In document with many annotations computing the anchor positions can
+      // block the JS event loop for up to 200 ms.
+      const positions = computeAnchorPositions(this._anchors);
+      this._hostRPC.call('anchorsChanged', positions);
       this._updatePending = false;
     });
   }
--- a/src/annotator/bucket-bar.js
+++ b/src/annotator/bucket-bar.js
@@ -1,9 +1,10 @@
 import { render } from 'preact';
 
 import Buckets from './components/Buckets';
-import { anchorBuckets } from './util/buckets';
+import { computeBuckets } from './util/buckets';
 
 /**
+ * @typedef {import('../types/annotator').AnchorPosition} AnchorPosition
  * @typedef {import('../types/annotator').Destroyable} Destroyable
  */
 
@@ -16,7 +17,6 @@ import { anchorBuckets } from './util/buckets';
 export default class BucketBar {
   /**
    * @param {HTMLElement} container
-   * @param {Pick<import('./guest').default, 'anchors'>} guest
    * @param {object} options
    *   @param {(tags: string[]) => void} options.onFocusAnnotations
    *   @param {(tags: string[], direction: 'down'|'up') => void} options.onScrollToClosestOffScreenAnchor
@@ -24,7 +24,6 @@ export default class BucketBar {
    */
   constructor(
     container,
-    guest,
     {
       onFocusAnnotations,
       onScrollToClosestOffScreenAnchor,
@@ -34,13 +33,12 @@ export default class BucketBar {
     this._bucketsContainer = document.createElement('div');
     container.appendChild(this._bucketsContainer);
 
-    this._guest = guest;
     this._onFocusAnnotations = onFocusAnnotations;
     this._onScrollToClosestOffScreenAnchor = onScrollToClosestOffScreenAnchor;
     this._onSelectAnnotations = onSelectAnnotations;
 
-    // Immediately render the buckets for the current anchors.
-    this.update();
+    // Immediately render the bucket bar
+    this.update([]);
   }
 
   destroy() {
@@ -48,8 +46,11 @@ export default class BucketBar {
     this._bucketsContainer.remove();
   }
 
-  update() {
-    const buckets = anchorBuckets(this._guest.anchors);
+  /**
+   * @param {AnchorPosition[]} positions
+   */
+  update(positions) {
+    const buckets = computeBuckets(positions);
     render(
       <Buckets
         above={buckets.above}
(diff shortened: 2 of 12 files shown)


src/annotator/util/test/buckets-test.js
```diff
@@ -1,44 +1,65 @@
...
+      { annotation: { $tag: 't2' }, highlights: [201, 50] },
+      // top: 301, bottom: 351 — on screen
+      { annotation: { $tag: 't3' }, highlights: [301, 50] },
+      // top: 401, bottom: 451 — below screen
+      { annotation: { $tag: 't4' }, highlights: [401, 50] },
+      // top: 501, bottom: 551 - below screen
+      { annotation: { $tag: 't5' }, highlights: [501, 50] },
+    ];
+
+    fakeAnchorPositions = [
```
robertknight, line 30: To check my understanding: `fakeAnchorPositions` is the expected _output_ when calling `computeAnchorPositions(fakeAnchors)`, right? If so I think it would help to name this `expectedAnchorPositions` or similar to make it clearer that this is an expected output (although also an input to `computeBuckets`), or add a comment here to indicate the connection.

src/types/annotator.js
```diff
@@ -53,6 +53,16 @@
  * @prop {Target} target
  */
 
+/**
+ * Top and bottom positions of the bounding box created by the union of the
+ * highlight elements associated to an anchor.
+ *
+ * @typedef AnchorPosition
+ * @prop {string} tag - annotation tag
+ * @prop {number} top
```
robertknight, line 62: What coordinate system and units are `top` and `bottom` in?

src/annotator/util/test/buckets-test.js
```diff
@@ -1,44 +1,65 @@
...
-  offsetIncrement = 100,
-  boxHeight = 50
-) {
-  let highlightIndex = offsetStart;
-  return () => {
-    // In a normal `Anchor` object, `highlights` would be an array of
-    // DOM elements. Here, `highlights[0]` is the vertical offset (top) of the
-    // fake anchor's highlight box and `highlights[1]` is the height of the
-    // box. This is in used in conjunction with the mock for
-    // `getBoundingClientRect`, below
```
robertknight, line 16: This is a useful comment that got removed in the refactoring of the tests.

src/annotator/util/test/buckets-test.js
```diff
@@ -1,44 +1,65 @@
...
+      { annotation: { $tag: 't1' }, highlights: [101, 50] },
+      // top: 201, bottom: 251 — on screen
+      { annotation: { $tag: 't2' }, highlights: [201, 50] },
+      // top: 301, bottom: 351 — on screen
+      { annotation: { $tag: 't3' }, highlights: [301, 50] },
+      // top: 401, bottom: 451 — below screen
+      { annotation: { $tag: 't4' }, highlights: [401, 50] },
+      // top: 501, bottom: 551 - below screen
+      { annotation: { $tag: 't5' }, highlights: [501, 50] },
+    ];
```
robertknight, line 28: All of these test cases currently has only a single "highlight". That's not too much of an issue since the whole collection is passed to `getBoundingClientRect` which returns the combined bounding box.

src/annotator/util/test/buckets-test.js
```diff
@@ -1,44 +1,65 @@
...
 describe('annotator/util/buckets', () => {
-  let fakeGetBoundingClientRect;
-
   let fakeAnchors;
+  let fakeAnchorPositions;
+  let fakeGetBoundingClientRect;
   let stubbedInnerHeight;
 
   beforeEach(() => {
-    const fakeAnchor = fakeAnchorFactory();
```
robertknight, line 33: In addition to the functional changes in this PR, it looks like there has been some general refactoring to the approach the tests use. Reviewing this code was more difficult because it wasn't clear to me initially how the changes related to the functional changes in the code. Would be very helpful here is to call out the non-functional changes in the PR description, a commit message or comments on the relevant parts of the PR.

robertknight (commented):
From a functional perspective, everything works as I expected. I had a query about the comment relating to performance, and also a question / ask about the rationale for the changes in the bucket functions tests' which seem to be a refactoring rather than a necessity of the functional changes. I think I can guess what the reasoning might be, but it wasn't mentioned explicitly.

robertknight:
I [discussed](https://hypothes-is.slack.com/archives/C1M8NH76X/p1643627791167819?thread_ts=1643386945.652579&cid=C1M8NH76X) the performance issue with Eduardo in Slack. To summarize: this is not a new issue in this branch, but rather the comment was calling attention to the fact that `computeAnchorPositions` may be expensive as it can force a layout update in order to compute the bounding boxes of highlights. I suggested to revise the comment on this basis.

src/annotator/bucket-bar-client.js
```diff
@@ -44,15 +50,25 @@ export class BucketBarClient {
...
+
     if (this._updatePending) {
       return;
     }
 
     this._updatePending = true;
     requestAnimationFrame(() => {
-      this._hostRPC.call('anchorsChanged');
+      // In document with many annotations computing the anchor positions can
+      // block the JS event loop for up to 200 ms.
```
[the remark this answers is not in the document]
esanzgar: I amended the comment.

src/annotator/bucket-bar.js
```diff
@@ -34,22 +33,24 @@ export default class BucketBar {
...
 
-    this._guest = guest;
     this._onFocusAnnotations = onFocusAnnotations;
     this._onScrollToClosestOffScreenAnchor = onScrollToClosestOffScreenAnchor;
     this._onSelectAnnotations = onSelectAnnotations;
 
-    // Immediately render the buckets for the current anchors.
-    this.update();
+    // Immediately render the bucket bar
+    this.update([]);
```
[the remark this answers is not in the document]
esanzgar: Yes, I think so.

src/annotator/sidebar.js
```diff
@@ -301,9 +298,13 @@ export default class Sidebar {
...
     // The listener will do nothing if the sidebar doesn't have a bucket bar
     // (clean theme), but it is still actively listening.
-    guestRPC.on('anchorsChanged', () => {
-      this.bucketBar?.update();
-    });
+    guestRPC.on(
+      'anchorsChanged',
+      /** @param {AnchorPosition[]} positions  */
+      positions => {
+        this.bucketBar?.update(positions);
```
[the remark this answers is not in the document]
esanzgar: Good idea, I will add a comment.

src/annotator/test/sidebar-test.js
```diff
@@ -481,9 +478,10 @@ describe('Sidebar', () => {
         const sidebar = createSidebar();
         connectGuest(sidebar);
 
-        emitGuestEvent('anchorsChanged');
+        emitGuestEvent('anchorsChanged', [1, 2]);
```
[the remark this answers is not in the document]
esanzgar: Yes, I agree. Let me change that.

src/types/annotator.js
esanzgar: I added the unit (pixel) as a comment.

src/annotator/util/test/buckets-test.js
esanzgar: Yes, sorry. I should have suggest this change in a follow up.

The reasoning for the change was that the `fakeAnchor` utility created an structure that was simple to recreate. I feel looking at  `{ annotation: { $tag: 't0' }, highlights: [1, 50] }, ...` is simpler to know the `fakeAnchors` structure.

src/annotator/util/test/buckets-test.js
esanzgar: I assumed that testing multiple (overlapping) highlights was done in `highlighter-test.js`

esanzgar (commented):
I included your suggestions in a second commit. Please, let me know if that addresses all your comments.

src/annotator/bucket-bar-client.js
```diff
@@ -65,8 +65,8 @@ export class BucketBarClient {
 
     this._updatePending = true;
     requestAnimationFrame(() => {
-      // In document with many annotations computing the anchor positions can
-      // block the JS event loop for up to 200 ms.
+      // Computing anchor positions may be expensive when there are many
+      // annotations or a forced layout reflow
```
robertknight, line 69: Computing anchor positions can be expensive if it triggers a layout reflow. If no reflows are required then this would be fast even if there were a lot of annotations. Depending on how the browser does reflows, the cost might be spread out over the `getBoundingClientRect` calls or might be paid all at once in the first `getBoundingClientRect` call.

src/annotator/test/sidebar-test.js
```diff
@@ -478,10 +478,11 @@ describe('Sidebar', () => {
         const sidebar = createSidebar();
         connectGuest(sidebar);
 
-        emitGuestEvent('anchorsChanged', [1, 2]);
+        const anchorPositions = [{ tag: 't0', top: 1, bottom: 2 }];
+        emitGuestEvent('anchorsChanged', anchorPositions);
```
robertknight, line 482: 👍

src/annotator/util/test/buckets-test.js
robertknight: `highlighter-test.js` does indeed test that `getBoundingClientRect` returns the union if a collection is passed in. There would be value in having an integration test somewhere that includes the case of an annotation with multiple highlights. That's out of scope for this PR.

src/types/annotator.js
```diff
@@ -59,8 +59,8 @@
  *
  * @typedef AnchorPosition
  * @prop {string} tag - annotation tag
- * @prop {number} top
- * @prop {number} bottom
+ * @prop {number} top - in pixel
```
robertknight, line 62: This answers the question about _units_ but what about coordinate system? Are these relative to the top of the document? The top of the viewport? Something else?

src/annotator/bucket-bar-client.js
esanzgar: My observation in Chrome is that this could be expensive even when no reflows are involved.

In the following video, I enable the browser extension on a previously loaded document: (1) Anchors are added and (2) the console outputs the time that takes to compute the anchors positions for the new anchors.

https://user-images.githubusercontent.com/8555781/151801287-875721b8-f12e-4785-b601-3a4d0f1602b8.mp4

I can see that in one case it takes 53 ms.

src/types/annotator.js
esanzgar: Good point. Zero is the top of the viewport. Elements that are above the viewport have negative values.
Would that be enough?

robertknight: Yes. Saying "viewport relative" or similar would be fine.

src/annotator/bucket-bar-client.js
esanzgar: Based on our conversation in [Slack](https://hypothes-is.slack.com/archives/C1M8NH76X/p1643637320348609?thread_ts=1643386945.652579&cid=C1M8NH76X) I decided to remove the comment. I realised I don't fully understand this well enough what the browsing is doing.

robertknight (approved):
LGTM

PR #4090, opened 2022-01-03
Add integration test to check inter-frame communication

esanzgar:
This new integration test checks that `PortFinder`, `PortProvider` and
`PortRPC` work in harmony. A change in one could lead to the others no
working as planned.

### DIFF
--- a/src/shared/test/integration/inter-frame-communication-test.js
+++ b/src/shared/test/integration/inter-frame-communication-test.js
@@ -1,0 +1,178 @@
+import { delay } from '../../../test-util/wait';
+import { PortRPC as PortRPC_ } from '../../port-rpc';
+import { ListenerCollection as ListenerCollection_ } from '../../listener-collection';
+import { PortFinder as PortFinder_ } from '../../port-finder';
+import { PortProvider as PortProvider_ } from '../../port-provider';
+import { isMessageEqual } from '../../port-util';
+
+describe('PortProvider-PortFinder-PortRPC integration', () => {
+  let destroyables;
+
+  // These proxy classes are to keep track of instances that need to be destroyed.
+  class PortRPC extends PortRPC_ {
+    constructor() {
+      super();
+      destroyables.push(this);
+    }
+  }
+
+  class ListenerCollection extends ListenerCollection_ {
+    constructor() {
+      super();
+      destroyables.push(this);
+    }
+    destroy() {
+      this.removeAll();
+    }
+  }
+
+  class PortFinder extends PortFinder_ {
+    constructor(...args) {
+      super(...args);
+      destroyables.push(this);
+    }
+  }
+
+  class PortProvider extends PortProvider_ {
+    constructor(...args) {
+      super(...args);
+      destroyables.push(this);
+    }
+  }
+
+  beforeEach(() => {
+    destroyables = [];
+  });
+
+  afterEach(() => {
+    destroyables.forEach(instance => instance.destroy());
+  });
+
+  it('enables the communication between guest-host', async () => {
+    let done;
+    const promise = new Promise(resolve => (done = resolve));
+
+    // guest frame
+    const portFinder = new PortFinder({
+      hostFrame: window,
+      source: 'guest',
+    });
+    const hostRPC = new PortRPC();
+
+    portFinder.discover('host').then(port => {
+      hostRPC.connect(port);
+      hostRPC.call('ping', response => {
+        assert.equal(response, 'pong');
+        done();
+      });
+    });
+
+    await delay(10); // add some realism
+
+    // host frame
+    const portProvider = new PortProvider(window.location.origin);
+    portProvider.listen();
+    const guestRPC = new PortRPC();
+
+    // Register RPC method *before* connection
+    guestRPC.on('ping', cb => cb('pong'));
+
+    portProvider.on('frameConnected', (source, port) => {
+      if (source === 'guest') {
+        guestRPC.connect(port);
+      }
+    });
+
+    return promise;
+  });
+
+  it('enables the communication between guest-sidebar', async () => {
+    let done;
+    const promise = new Promise(resolve => (done = resolve));
+
+    // guest frame;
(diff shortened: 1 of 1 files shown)


esanzgar:
I have created three simpler tests the following three communication channels:
* guest-host
* guest-sidebar (a little bit more complex)
* sidebar-host

esanzgar:
We exchanged comments about the `subclassing` vs using helper functions [here](https://github.com/hypothesis/client/pull/4092/files#r778077255)

Please, let me know if you want me to using helper functions.

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,178 @@
...
+    });
+    const hostRPC = new PortRPC();
+
+    portFinder.discover('host').then(port => {
+      hostRPC.connect(port);
+      hostRPC.call('ping', response => {
+        assert.equal(response, 'pong');
+        done();
+      });
+    });
```
robertknight, line 68: The `done` resolver could potentially be replaced by doing the `portFinder.discover` call in the Promise callback:

```js
const done = new Promise(resolve => {
  portFinder.discover('host').then(port => {
    hostRPC.connect(port);
    hostRPC.call('ping', response => {
      assert.equal(response, 'pong');
      resolve();
    });
  });
});

...

return done;
```

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,178 @@
...
+
+    await delay(10); // add some realism
+
+    // host frame
+    const portProvider = new PortProvider(window.location.origin);
+    portProvider.listen();
+    const guestRPC = new PortRPC();
+
+    // Register RPC method *before* connection
```
robertknight, line 77: I suggest to remove these comments, and leave it to the PortRPC docs to explain that correct usage of the class requires calling `on` before `connect`.

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,178 @@
...
+
+  afterEach(() => {
+    destroyables.forEach(instance => instance.destroy());
+  });
+
+  it('enables the communication between guest-host', async () => {
+    let done;
+    const promise = new Promise(resolve => (done = resolve));
+
+    // guest frame
```
robertknight, line 55:
```suggestion
    // Guest frame
```

.. and similar for other comments.

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,178 @@
...
+
+    return promise;
+  });
+
+  it('enables the communication between guest-sidebar', async () => {
+    let done;
+    const promise = new Promise(resolve => (done = resolve));
+
+    // guest frame;
+    const portFinder1 = new PortFinder({
```
robertknight, line 94:
```suggestion
    const guestPortFinder = new PortFinder({
```

Alternatively, if you wrap each actor in a separate function, the variable names won't conflict.

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,178 @@
...
+    // host frame
+    const portProvider = new PortProvider(window.location.origin);
+    portProvider.listen();
+
+    return promise;
+  });
+
+  it('enables the communication between sidebar-host', async () => {
+    let done;
+    const promise = new Promise(resolve => (done = resolve));
```
robertknight, line 145: If you are going to use this pattern, I'd suggest to let `done` be the promise as it is a name of a state (similar to `loaded` or `completed`), and call the resolver `resolveDone`. Alternatively, skip the resolver variable and just use the promise callback:

```js
const done = new Promise(resolve => {
   ...
   hostRPC.call('ping', response => { ...; resolve(); });
});
```

robertknight (approved):
I think the subclassing is OK for now. There are ways it could be abstracted, but the current implementation is easy enough to follow.

I had a suggestion regarding the grouping of the code that represents each of the different frames in the test, namely to split them into separate functions, to make it clear that they are meant to represent activities happening in different frames which run concurrently and don't share any state (except via a message channel).

Generally LGTM though.

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,178 @@
...
+
+    portFinder.discover('host').then(port => {
+      hostRPC.connect(port);
+      hostRPC.call('ping', response => {
+        assert.equal(response, 'pong');
+        done();
+      });
+    });
+
+    await delay(10); // add some realism
```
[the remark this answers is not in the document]
esanzgar: `Guest` can be ready before the `Host` frame. If that's the case, `Guest` will polls the `Host`. I was trying to simulate that scenario.

src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -0,0 +1,178 @@
...
+
+  beforeEach(() => {
+    destroyables = [];
+  });
+
+  afterEach(() => {
+    destroyables.forEach(instance => instance.destroy());
+  });
+
+  it('enables the communication between guest-host', async () => {
```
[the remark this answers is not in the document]
esanzgar: Very good idea! Thanks!

src/shared/test/integration/inter-frame-communication-test.js
esanzgar: Great idea! Thanks!

src/shared/test/integration/inter-frame-communication-test.js
esanzgar: 👍🏼

src/shared/test/integration/inter-frame-communication-test.js
esanzgar: I use separate functions.

esanzgar (commented):
I incorporated your feedback on the second commit.

PR #4092, opened 2022-01-03
Make `PortProvider#listen` a private  method

esanzgar:
I can't see any negative consequence of starting to listen to
postMessage discovery message immediately after `PortProvider` is
instantiated. (In fact, this is what it was currently happening). These
messages won't be lost, and the `PortRPC`s won't create the
communication channels before the RPC methods are registered.

On the other hand, it simplifies the use of `PortProvider` and
eliminates the potential problem of `PortProvider#listen` been called
more than once.

### DIFF
--- a/src/shared/port-provider.js
+++ b/src/shared/port-provider.js
@@ -60,6 +60,8 @@ import { isMessage, isMessageEqual, isSourceWindow } from './port-util';
  */
 export class PortProvider {
   /**
+   * Begin listening to port requests from other frames.
+   *
    * @param {string} hypothesisAppsOrigin - the origin of the hypothesis apps
    *   is use to send the notebook and sidebar ports to only the frames that
    *   match the origin.
@@ -112,41 +114,11 @@ export class PortProvider {
         type: 'request',
       },
     ];
+
+    this._listen();
   }
 
-  /**
-   * Check that data and origin matches the expected values.
-   *
-   * @param {object} options
-   *   @param {Message} options.allowedMessage - the `data` must match this
-   *     `Message`.
-   *   @param {string} options.allowedOrigin - the `origin` must match this
-   *     value. If `allowedOrigin` is '*', the origin is ignored.
-   *   @param {any} options.data - the data to be compared with `allowedMessage`.
-   *   @param {string} options.origin - the origin to be compared with
-   *     `allowedOrigin`.
-   */
-  _messageMatches({ allowedMessage, allowedOrigin, data, origin }) {
-    if (allowedOrigin !== '*' && origin !== allowedOrigin) {
-      return false;
-    }
-
-    return isMessageEqual(data, allowedMessage);
-  }
-
-  /**
-   * @param {'frameConnected'} eventName
-   * @param {(source: 'guest'|'sidebar', port: MessagePort) => void} handler - this handler
-   *   fires when a request for the host frame has been granted.
-   */
-  on(eventName, handler) {
-    this._emitter.on(eventName, handler);
-  }
-
-  /**
-   * Initiate the listener of port requests by other frames.
-   */
-  listen() {
+  _listen() {
     const errorContext = 'Handling port request';
     const sentErrors = /** @type {Set<string>} */ (new Set());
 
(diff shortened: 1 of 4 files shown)


src/shared/test/integration/inter-frame-communication-test.js
```diff
@@ -132,8 +130,7 @@ describe('PortProvider-PortFinder-Bridge integration', () => {
     });
 
     // host frame
-    const portProvider = new PortProvider(window.location.origin);
-    portProvider.listen();
+    new PortProvider(window.location.origin);
```
robertknight, line 133: When I initially read this I thought there was a leak because the `PortProvider` instance was not cleaned up. Looking above I see that `PortProvider` is not in fact the original `PortProvider` class but a subclass. I think it would be clearer what was going on if instead of subclassing `PortProvider` you instead used a helper function to populate the `destroyables` list.

src/shared/port-provider.js
```diff
@@ -112,41 +112,14 @@ export class PortProvider {
...
-   * @param {(source: 'guest'|'sidebar', port: MessagePort) => void} handler - this handler
-   *   fires when a request for the host frame has been granted.
-   */
-  on(eventName, handler) {
-    this._emitter.on(eventName, handler);
+    this._listen();
   }
 
   /**
    * Initiate the listener of port requests by other frames.
```
robertknight, line 120: The documentation for `listen` needs to be moved to / combined with the existing constructor documentation.

robertknight (commented):
If we don't currently need the flexibility that comes with separating initialization/configuration and activation of this class then the change seems reasonable and eliminates a potential misuse of the API. If we do later find that we need this flexibility we might have to reintroduce the separation again. If you have a class constructor that does something "active" then it is important to call that out in the documentation though.

src/shared/test/integration/inter-frame-communication-test.js
esanzgar: Yes, I thought about adding a helper function, however, I thought it was going to be a little bit repetitive (`createBridge`, `createPortProvider`, `createPortFinder`). If implemented more generically, like `addToDestroyables(new Bridge())`, there is the possibility that one can forget the wrapping utility and we leak resources. 

I thought the subclassing was a middle ground, avoid leaking resources and not verbose, but I understand it is less intuitive.

I also thought about overriding the `new` operator, but that's not possible.

Any way, I am happy to swap subclassing for the helper utility if you want. https://github.com/hypothesis/client/pull/4090 would be a better PR to discuss that.

src/shared/port-provider.js
esanzgar: Agree.

src/shared/port-provider.js
```diff
@@ -246,6 +221,35 @@ export class PortProvider {
...
+   * Check that data and origin matches the expected values.
+   *
+   * @param {object} options
+   *   @param {Message} options.allowedMessage - the `data` must match this
+   *     `Message`.
+   *   @param {string} options.allowedOrigin - the `origin` must match this
+   *     value. If `allowedOrigin` is '*', the origin is ignored.
+   *   @param {any} options.data - the data to be compared with `allowedMessage`.
+   *   @param {string} options.origin - the origin to be compared with
+   *     `allowedOrigin`.
```
robertknight, line 234: Given that this is an internal method that is very short and simple, I think it would be fine to remove the descriptions here and just specify the types.

src/shared/port-provider.js
```diff
@@ -246,6 +221,35 @@ export class PortProvider {
...
+      return false;
+    }
+
+    return isMessageEqual(data, allowedMessage);
+  }
+
+  /**
+   * @param {'frameConnected'} eventName
+   * @param {(source: 'guest'|'sidebar', port: MessagePort) => void} handler - this handler
+   *   fires when a request for the host frame has been granted.
```
robertknight, line 247:
```suggestion
   *   fires when a frame connects to the host frame
```

The concept of requests and offers are really internal to PortProvider and PortFinder. Consumers only really need to know that a non-host frame connected to the host.

src/shared/port-provider.js
```diff
@@ -56,7 +56,9 @@ import { isMessage, isMessageEqual, isSourceWindow } from './port-util';
  *                                                           (eg. via MessageChannel connection
  *                                                           between host and other frame)
  *
- * @implements Destroyable
+ * Immediately after the class is instantiated it starts listening for port requests.
```
robertknight, line 59: It probably would make sense to move this to the constructor docs and phrase it as eg. "Begin listening to port requests from other frames."

src/shared/port-provider.js
```diff
@@ -112,41 +114,14 @@ export class PortProvider {
...
-  on(eventName, handler) {
-    this._emitter.on(eventName, handler);
+    this._listen();
   }
 
   /**
    * Initiate the listener of port requests by other frames.
    */
-  listen() {
+  _listen() {
```
robertknight, line 124: Given that `_listen` is now an internal method with a pretty obvious purpose, assuming someone has read the general class documentation, I think it would be fine to remove the JSDoc for it.

robertknight (approved):
LGTM, with some comments regarding docs. In general there are pros and cons to having server-like classes listen as soon as constructed or require a separate step. Given the current usage of `PortProvider` it is simpler this way. We can always revisit if it becomes problematic.

src/shared/port-provider.js
esanzgar: 👍🏼

src/shared/port-provider.js
esanzgar: 👍🏼

src/shared/port-provider.js
esanzgar: 👍🏼

src/shared/port-provider.js
esanzgar: 👍🏼

Issue #3182, opened 2021-03-22
Synchronise annotations in the Notebook


esanzgar:
Currently, there is no mechanism to synchronise annotations in the notebook. We circumvent this by [reloading the notebook's iframe each time it is opened](https://github.com/hypothesis/client/pull/3172). However, reloading the iframe is a costly and slow event.

It would be good to implement a mechanism that enable new/deleted/updated annotations in the sidebar to be synchronised on the notebook without the need to reload the iframe. Probably, through a websocket communication.

esanzgar:
We have implemented a synchronisation mechanism in the notebook.

The notebook still reloads the iframe on every `openNotebook` event, which takes quite a lot of time.

PR #4096, opened 2022-01-06
WIP: proof of concept for notebook-sidebar communication

esanzgar:
This PR enables a new communication channel between the notebook and the
sidebar. This in turn prevents the notebook from having to re-initialise
the notebook app (re-rendering parent iframe) each time the notebook is
opened, hence speeding up the app after the first rendering.

This is very crude WIP PR, just to demonstrate the feasibility of the
channel and to get preliminary feedback.

You can compare the speed difference between the before and after videos:

### Before
https://user-images.githubusercontent.com/8555781/148407477-da4436f3-64e4-4326-9df2-d375ce5be819.mov

### After
https://user-images.githubusercontent.com/8555781/148407493-3edc3c97-5f87-4b26-9e70-e8c853cd06f8.mov

Related to: #3313
Closes https://github.com/hypothesis/client/issues/3182

### DIFF
--- a/src/sidebar/components/GroupList/GroupListItem.js
+++ b/src/sidebar/components/GroupList/GroupListItem.js
@@ -16,6 +16,7 @@ import MenuItem from '../MenuItem';
  * @prop {boolean} [isExpanded] - Whether the submenu for this group is expanded
  * @prop {(expand: boolean) => any} onExpand -
  *   Callback invoked to expand or collapse the current group
+ * @prop {import('../../services/frame-sync').FrameSyncService} frameSync
  * @prop {import('../../services/groups').GroupsService} groups
  * @prop {import('../../services/toast-messenger').ToastMessengerService} toastMessenger
  */
@@ -30,6 +31,7 @@ import MenuItem from '../MenuItem';
  */
 function GroupListItem({
   isExpanded,
+  frameSync,
   group,
   groups: groupsService,
   onExpand,
@@ -48,6 +50,7 @@ function GroupListItem({
     store.clearDirectLinkedGroupFetchFailed();
     store.clearDirectLinkedIds();
     groupsService.focus(group.id);
+    frameSync.notifyNotebook('groupChanged', group.id);
   };
 
   const leaveGroup = async () => {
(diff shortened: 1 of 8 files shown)


esanzgar:
A good place to start reviewing this PR is `src/sidebar/components/GroupList/GroupListItem.js`. Each time a new group is selected it sends a message to the notebook.

src/sidebar/services/frame-sync.js
```diff
@@ -87,6 +89,16 @@ export class FrameSyncService {
...
     this._guestRPC = new Bridge();
 
+    /**
+     * Channel for sidebar-notebook communication.
+     *
+     * @type {Bridge<SidebarToNotebookEvent,NotebookToSidebarEvent>}
+     */
+    this._notebookRPC = new Bridge();
+    this._notebookRPC.onConnect(() =>
+      this._notebookRPC.call('groupChanged', store.focusedGroup()?.id)
```
robertknight, line 99: Related to my investigation about potentially removing `Bridge#onConnect` in future, this call can be moved to just after the `this._notebookRPC.createChannel` call.

src/sidebar/components/NotebookView.js
```diff
@@ -64,6 +78,28 @@ function NotebookView({ loadAnnotationsService, streamer }) {
 
   const hasFetchedProfile = store.hasFetchedProfile();
 
+  // Establish communication channel with the sidebar
+  useEffect(() => {
```
robertknight, line 82: It might make sense to move some of this non-UI logic to a service in future.

src/sidebar/components/GroupList/GroupListItem.js
```diff
@@ -48,6 +50,7 @@ function GroupListItem({
     store.clearDirectLinkedGroupFetchFailed();
     store.clearDirectLinkedIds();
     groupsService.focus(group.id);
+    frameSync.notifyNotebook('groupChanged', group.id);
```
robertknight, line 53: The focused group can change as a result of other events as well - for example when the logged-in user changes. One way to handle this is to have a service watch `store.focusedGroup()` and make this call in response.

src/sidebar/components/NotebookView.js
```diff
@@ -35,8 +47,10 @@ function NotebookView({ loadAnnotationsService, streamer }) {
 
   const rootThread = useRootThread();
 
+  // TODO: should we compute this more directly
```
robertknight, line 50: Can you expand on what you are thinking here?

PR #4104, opened 2022-01-07
Re-organize inter-frame communication code under shared/messaging

robertknight:
Make it easier for future developers to understand the inter-frame communication
code by grouping it in a directory with a short README and an index module that
defines the API for use by other parts of the client.

~~Looking at this again, I've decided that I don't really like the `shared/ipc` name, as we don't use the term anywhere else. How about `shared/ports`? Any other suggestions?~~ (_Changed to `shared/messaging` after [Slack discussion](https://hypothes-is.slack.com/archives/C1M8NH76X/p1643201099083000)_)


robertknight:
This is ready for review, but I could use some input on naming of the `shared/ipc` directory. See https://hypothes-is.slack.com/archives/C1M8NH76X/p1643201099083000.

esanzgar (approved):
I really like this file structure!

I wonder if the integration test `inter-frame-communication-test.js` shouldn't be renamed and moved to `src/shared/messaging/test/integration/messaging.js`?

PR #4111, opened 2022-01-10
Combine Bridge and PortRPC classes

robertknight:
Reduce the number of layers involved in inter-frame communication by combining
the Bridge and PortRPC classes. PortRPC still only supports sending requests to
a single port, but now has an API that is more like the Bridge class, where
construction of the class, registering of RPC method handlers and connecting to
a MessagePort are separate steps. This separation is convenient because it
allows the RPC client to be set up before the MessagePort retrieval process,
which is async in non-host frames, has completed.

Existing uses of Bridge have been replaced either with a single PortRPC
instance, for 1:1 connections between frames, or an array of PortRPC
instances for 1:N connections between the sidebar/host and guest frames.
Migrating to an array for 1:N connections paves the way to sending certain
requests only to specific guests, whereas they are currently sent to _all_
guests.

Bridge's `call` method had logic for gathering results from multiple guests,
timing out and disconnecting clients in the event of an error.  This logic was
broken in various ways and has not been re-implemented in PortRPC:

 - The timeout value was hardcoded at 1000ms for all calls. This may be too
   short depending on what the call does or what code is executing in the
   destination frame.
 - Disconnecting clients in the event of _any_ error, whether with the
   connection or an error on the client end in handling the call, doesn't make sense.
 - The `call` method returned a single Promise, so an error from any client
   would replace successful results from other clients. Also the client would
   have to wait until the results from all callers was available before sending
   a response.
 - The `call` method always passed a callback function to `PortRPC.call`,
   resulting in PortRPC always sending a response message back over the message
   channel, even when the code calling `bridge.call` did not use the response.
   Currently _no_ calls to `bridge.call` used the response, so this was just
   wasted traffic.

The changes in detail are:

 1. Change the PortRPC API to be more like Bridge, where an instance is
    constructed, RPC handlers are added via `on` and the client is then
    connected to a port via `connect`.

 2. Replace use of Bridge with either a single PortRPC or array of PortRPCs in
    the `Guest`, `Sidebar` and `FrameSyncService` classes.

 3. Rename `bridge-events.d.ts` => `port-rpc-events.d.ts` to match (2)

 4. Remove the `Bridge` class, its tests and the PortRPC-Bridge integration
    tests, as they are now unused.

Two existing issues were identified in this work but are not yet resolved:

 1. `FrameSyncService` does not tear down the PortRPC client for a guest when
     the guest frame disconnects. The changes in https://github.com/hypothesis/client/pull/4113 will make this easier.
 2. If `PortRPC.call` is called before a port has been connected using `PortRPC.connect`, nothing happens.  The follow-up PR https://github.com/hypothesis/client/pull/4112 provides a solution by buffering calls that are made before `connect`. The follow-up PR https://github.com/hypothesis/client/pull/4113 makes use of this buffering behavior.

---

Review notes: I suggest looking at the changes to `PortRPC` first, then the `Guest`, `Sidebar` and `FrameSyncService` classes and finally the tests.


src/annotator/sidebar.js
```diff
@@ -80,16 +80,18 @@ export default class Sidebar {
     /**
      * Channel for host-guest communication.
      *
-     * @type {Bridge<HostToGuestEvent,GuestToHostEvent>}
+     * TODO - Make this a Map<frame ID, PortRPC<...>>?
+     *
+     * @type {PortRPC<GuestToHostEvent, HostToGuestEvent>[]}
      */
-    this._guestRPC = new Bridge();
+    this._guestRPC = [];
```
robertknight, line 87: In future it will probably make sense to make this a `Map<frame ID, PortRPC>`. It's problematic to do that now however because the frame ID is determined by the guest itself and we only learn about it after the guest has connected. Hence for now it is just an array.

src/annotator/sidebar.js
```diff
@@ -140,7 +142,9 @@ export default class Sidebar {
     const toolbarContainer = document.createElement('div');
     this.toolbar = new ToolbarController(toolbarContainer, {
       createAnnotation: () =>
-        this._guestRPC.call('createAnnotationIn', this._guestWithSelection),
+        this._guestRPC.forEach(rpc =>
+          rpc.call('createAnnotationIn', this._guestWithSelection)
+        ),
```
robertknight, line 147: A follow-up step here will be to call `createAnnotationIn` only on the specific guest with the selection.

src/shared/port-rpc.js
```diff
@@ -51,39 +51,63 @@ const PROTOCOL = 'frame-rpc';
...
- * It uses the Channel Messaging API [1] for inter-frame communication.
+ * To communicate between two frames with this class, construct a PortRPC
+ * instance in each and register method handlers with {@link on}. Create a
+ * {@link MessageChannel} and send one of its two ports to each frame. Then call
+ * {@link connect} to make the PortRPC instance in each frame use the corresponding
+ * port.
  *
  * [1] https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API
  *
- * Code adapted from https://github.com/substack/frame-rpc.
```
robertknight, line 60: The copyright note for frame-rpc is still at the top of this file, although the code has evolved so much that it has only small connections to this package.

src/shared/port-rpc.js
```diff
@@ -93,20 +117,23 @@ export class PortRPC {
...
+   * which is invoked with the response in the form of (error, result) arguments.
    *
-   * @param {string} method
+   * @param {CallMethod} method
    * @param {any[]} args
    */
   call(method, ...args) {
-    if (this._destroyed) {
+    // TODO - What should happen here if the port is not connected? Buffer
+    // method calls until the port is connected?
```
robertknight, line 134: This TODO is addressed by https://github.com/hypothesis/client/pull/4112.

src/sidebar/components/test/HypothesisApp-test.js
```diff
@@ -289,9 +289,8 @@ describe('HypothesisApp', () => {
 
     it('sends "loginRequested" event to host page if using a third-party service', async () => {
       // If the client is using a third-party annotation service then clicking
-      // on a login button should send the "loginRequested" event over the bridge
-      // (so that the partner site we're embedded in can do its own login
-      // thing).
+      // on a login button should notify the host frame (so that the partner
+      // site we're embedded in can do its own login thing).
```
robertknight, line 293: I changed this description to be less specific about implementation details.

src/sidebar/components/test/TopBar-test.js
```diff
@@ -111,7 +111,7 @@ describe('TopBar', () => {
       });
 
       context('help service handler configured in services', () => {
-        it('fires a bridge event if help clicked and service is configured', () => {
+        it('notifies host frame if help clicked and service is configured', () => {
```
robertknight, line 114: Description changed to be less specific about implementation details.

src/sidebar/services/frame-sync.js
```diff
@@ -76,16 +75,18 @@ export class FrameSyncService {
...
     /**
      * Channel for sidebar-guest(s) communication.
      *
-     * @type {Bridge<SidebarToGuestEvent,GuestToSidebarEvent>}
+     * TODO - Make this a Map<frame ID, PortRPC<...>>?
+     *
+     * @type {PortRPC<GuestToSidebarEvent, SidebarToGuestEvent>[]}
      */
-    this._guestRPC = new Bridge();
+    this._guestRPC = [];
```
robertknight, line 89: Similar to the corresponding field in the `Sidebar` class, it will probably make sense to make this a `Map<frame ID, PortRPC>` in future. It is an array for now because we don't have a frame ID at the point where we want to add the PortRPC to this map.

src/sidebar/services/frame-sync.js
```diff
@@ -292,13 +302,14 @@ export class FrameSyncService {
       if (frame) {
         this._store.destroyFrame(frame);
       }
+      // TODO - Remove the guest connection associated with this frame identifier.
```
robertknight, line 305: I left this as a TODO to avoid making this PR larger and because this is a pre-existing issue.

src/sidebar/services/frame-sync.js
```diff
@@ -347,7 +357,7 @@ export class FrameSyncService {
    */
   focusAnnotations(tags) {
     this._store.focusAnnotations(tags);
-    this._guestRPC.call('focusAnnotations', tags);
+    this._guestRPC.forEach(rpc => rpc.call('focusAnnotations', tags));
```
robertknight, line 360: In future it might be possible to be smarter about only sending these events to the right guest. However it isn't that important because guests ignore these messages if they are not displaying a particular annotation.

src/sidebar/services/test/frame-sync-test.js
```diff
@@ -216,42 +210,48 @@ describe('FrameSyncService', () => {
       frameSync.connect();
     });
 
-    it('sends a "loadAnnotations" message to the frame', () => {
+    it('sends a "loadAnnotations" message to the frame', async () => {
+      await connectGuest();
```
robertknight, line 214: Several tests for messages sent to guests did not actually simulate a guest connecting before checking that the message was sent. This didn't matter before because events were sent on the bridge whether or not any guests were connected. Now there will only be `PortRPC` connections if a guest is actually connected, so we have to simulate a guest connecting first.

src/annotator/annotation-counts.js
```diff
@@ -7,11 +7,11 @@ const ANNOTATION_COUNT_ATTR = 'data-hypothesis-annotation-count';
...
  * @param {Element} rootEl - The DOM element which contains the elements that
  *   display annotation count.
- * @param {import('../shared/bridge').Bridge} bridge - Channel for host-sidebar communication
+ * @param {import('../shared/port-rpc').PortRPC} rpc - Channel for host-sidebar communication
  */
 
-export function annotationCounts(rootEl, bridge) {
-  bridge.on('publicAnnotationCountChanged', updateAnnotationCountElems);
+export function annotationCounts(rootEl, rpc) {
+  rpc.on('publicAnnotationCountChanged', updateAnnotationCountElems);
```
esanzgar, line 14: Unrelated with this PR, but I think how should consolidate how we deal with all the `data-hypothesis-[...]` attributes within the same file. So unifying `annotation-count.js` with `sidebar-trigger.js` (by the way, `sidebar-trigger.js` doesn't remove the event handlers correctly)

src/annotator/features.js
```diff
@@ -8,10 +8,10 @@ const _set = features => {
 
 export const features = {
   /**
-   * @param {import('../shared/bridge').Bridge} bridge - Channel for host-sidebar communication
+   * @param {import('../shared/port-rpc').PortRPC} rpc - Channel for host-sidebar communication
    */
-  init: function (bridge) {
-    bridge.on('featureFlagsUpdated', _set);
+  init: function (rpc) {
```
esanzgar, line 13: Also unrelated, we should refactor this and convert it into a class.

esanzgar (approved):
I really like this simplification! Thank you very much!

I have left a few minor comments and suggestions.

src/shared/port-rpc.js
robertknight: I agree.

src/shared/port-rpc.js
```diff
@@ -93,20 +117,23 @@ export class PortRPC {
...
    *
    * If the final argument in `args` is a function, it is treated as a callback
-   * which is invoked with the response.
+   * which is invoked with the response in the form of (error, result) arguments.
    *
-   * @param {string} method
+   * @param {CallMethod} method
    * @param {any[]} args
    */
   call(method, ...args) {
```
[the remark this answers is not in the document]
robertknight: I realized when working on this that there is a disadvantage to using a Promise-based API here. The callback argument is both a way to return the result, but also a signal that the caller is interested in the result. With a Promise-based API, the `call` method has no way to tell if the caller wants the result, and so result messages may be sent via the port even if not used.

src/annotator/sidebar.js
```diff
@@ -251,40 +251,54 @@ export default class Sidebar {
...
         break;
     }
   }
 
-  _setupGuestEvents() {
-    this._guestRPC.on(
+  /**
+   * @param {MessagePort} port
+   */
+  _connectGuest(port) {
```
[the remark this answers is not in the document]
robertknight: > Personally, I lean towards _connectGuest, it is simpler.

I prefer this  too.

src/annotator/annotation-counts.js
robertknight: I agree. It would make sense to combine the code for all these elements into one module.

src/annotator/features.js
robertknight: Yes, indeed.

src/sidebar/services/frame-sync.js
```diff
@@ -98,7 +97,6 @@ export class FrameSyncService {
     this._highlightsVisible = false;
 
     this._setupSyncToGuests();
-    this._setupSyncFromGuests();
     this._setupHostEvents();
```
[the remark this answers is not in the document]
robertknight: Unlike `_connectGuest` this method doesn't connect the host RPC channel, it just registers method handlers, so I think the existing name makes sense.

PR #4112, opened 2022-01-10
Buffer inter-frame RPC calls made before port is connected

robertknight:
**Depends on https://github.com/hypothesis/client/pull/4111**

Fix a race condition where inter-frame RPC calls would silently fail if made before the MessagePort discovery process had completed. Fix the issue by queuing calls in PortRPC and then dispatching them once a port is connected. This is similar to how MessagePort buffers messages until they are consumed by calling `start`.

An alternate approach would be to disallow RPC calls before the port is connected and force callers to deal with the possibility. Given that all RPC calls in the application bar one are one-way notifications, I suspect callers would deal with this by implementing a similar queue at a higher level.

----

**Testing:**

1. Apply the following diff to simulate port discovery taking a long time:

```diff
diff --git a/src/shared/port-finder.js b/src/shared/port-finder.js
index 9cbba2a5c..8358ffff1 100644
--- a/src/shared/port-finder.js
+++ b/src/shared/port-finder.js
@@ -58,6 +58,8 @@ export class PortFinder {
       throw new Error('Invalid request of channel/port');
     }
 
+    await new Promise(resolve => setTimeout(resolve, 5000));
+
     return new Promise((resolve, reject) => {
       const postRequest = () => {
         this._hostFrame.postMessage(
```

2. Open the client's dev server, select some text and click "Annotate" before the sidebar has appeared

On master, these steps will not result in a new draft annotation being created. On this branch a new draft annotation should appear once the sidebar has opened.

### DIFF
--- a/src/shared/port-rpc.js
+++ b/src/shared/port-rpc.js
@@ -80,6 +80,13 @@ export class PortRPC {
     this._callbacks = new Map();
 
     this._listeners = new ListenerCollection();
+
+    /**
+     * Method and arguments of pending RPC calls made before a port was connected.
+     *
+     * @type {Array<[CallMethod, any[]]>}
+     */
+    this._pendingCalls = [];
   }
 
   /**
@@ -108,6 +115,11 @@ export class PortRPC {
       this._handle(/** @type {MessageEvent} */ (event))
     );
     port.start();
+
+    for (let [method, args] of this._pendingCalls) {
+      this.call(method, ...args);
+    }
+    this._pendingCalls = [];
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/shared/port-rpc.js
```diff
@@ -123,15 +135,19 @@ export class PortRPC {
...
    * which is invoked with the response in the form of (error, result) arguments.
    *
    * @param {CallMethod} method
    * @param {any[]} args
    */
   call(method, ...args) {
-    // TODO - What should happen here if the port is not connected? Buffer
-    // method calls until the port is connected?
+    if (!this._port) {
+      this._pendingCalls.push([method, args]);
```
esanzgar, line 149: Personally, I would check if the port is `null` only once, and leave just `this._destroyed` in the next `if` statement.
```suggestion
      this._pendingCalls.push([method, args]);
      return;
```

esanzgar (approved):
It looks good!

src/shared/port-rpc.js
robertknight: Drat, I missed this before merging. I'll have to come back to this.

Issue #4094, opened 2022-01-06
Optimize document URI discovery for main guest


robertknight:
The sidebar needs to register the initial/main guest frame in the store and query its document metadata/URIs before it can fetch any annotations. There is currently a lengthy sequence of steps that happens during startup of the Hypothesis client that leads up to this. There is room to optimize this which would improve the perceived startup time.

The order of events is typically:

1. The host creates the sidebar frame
2. The guest requests a connection to the sidebar
3. The host creates ports for the sidebar-guest connection and sends them to the corresponding frames
4. The sidebar requests its port for the host-sidebar channel from the host
5. The host responds with the corresponding port
6. The sidebar registers a `message` handler on this port to receive new guest ports
7. The sidebar receives the guest port sent in step (3) and calls `Bridge.createChannel` to connect to the guest. The Bridge class internally invokes a `connect` RPC method on the new guest-sidebar channel.
8. When the `connect` RPC response is received in the sidebar `Bridge.onConnect` handlers are invoked
9. One of the `onConnect` handlers makes an `getDocumentInfo` RPC call to the guest
10. The guest responds to the `getDocumentInfo` call
11. The sidebar receives the `getDocumentInfo` response and registers the new frame in the store

Steps 7-11 have some obvious room for improvement:

 - The `connect` RPC call in steps 7 & 8 is no longer necessary. This used to be required so that the sidebar could know when the guest is ready for `getDocumentInfo` to be invoked. Because the `Bridge` class now uses the channel messaging API as its transport, RPC requests can be invoked without needing to wait for the receiving end to be ready (MessagePort will buffer any received messages until the receiver is ready to consume them). See https://github.com/hypothesis/client/pull/4103.
 - We know that the sidebar will always query the guest for document metadata as soon as it connects. Therefore it would be more efficient for the guest to preemptively _push_ document metadata to the sidebar (eg. via a `documentChanged` RPC call from guest -> sidebar) in step 2. This RPC call would then be received by the sidebar immediately after step 7. The ability of the guest to push metadata to the sidebar will be a prerequisite for supporting client-side navigations in the guest frame (ie. where the document in the guest frame changes after it has connected to the sidebar). See https://github.com/hypothesis/client/pull/4113.

[...]

PR #4113, opened 2022-01-11
Push document metadata from guest to sidebar

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/4112**~~

Change the way document metadata and URIs for guest frames gets from the guest to the sidebar, so that the guest _pushes_ the information via a `documentInfoChanged` call instead of the sidebar requesting it via `getDocumentInfo`. This achieves two things:

 1. It enables the sidebar to get the initial document URI faster, as
    the guest can fetch the information and send it to the sidebar over
    the message channel while the sidebar is still loading. This in turn
    allows the initial annotation search request to be made sooner and
    for annotations to appear quicker.

    This fixes https://github.com/hypothesis/client/issues/4094.

    In practice this doesn't improve the best case metadata fetch time
    significantly (only 5-10ms or so) but does seem to reduce the variance in
    fetch times.

 2. It will in future enable the guest to inform the sidebar of
    URI/metadata changes after the initial load. This can happen if a
    client-side navigation happens in an HTML document for example or if
    the currently loaded PDF is changed in PDF.js.

A known weak point of both the previous and the new implementation is that there isn't good error handling of the guest fails to read the document metadata. The sidebar will just remain in a loading state. The main case where this can happen currently is if a PDF fails to load. In that case though it is obvious that something is wrong.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -184,6 +184,12 @@ export default class Guest {
     });
 
     /**
+     * Integration that handles document-type specific functionality in the
+     * guest.
+     */
+    this._integration = createIntegration(this);
+
+    /**
      * Channel for host-guest communication.
      *
      * @type {PortRPC<HostToGuestEvent, GuestToHostEvent>}
@@ -202,12 +208,6 @@ export default class Guest {
     // Set up automatic and integration-triggered injection of client into
     // iframes in this frame.
     this._hypothesisInjector = new HypothesisInjector(this.element, config);
-
-    /**
-     * Integration that handles document-type specific functionality in the
-     * guest.
-     */
-    this._integration = createIntegration(this);
 
     this._bucketBarClient =
       this._frameIdentifier === null
@@ -415,13 +415,6 @@ export default class Guest {
       }
     );
 
-    // Handler for when sidebar requests metadata for the current document
-    this._sidebarRPC.on('getDocumentInfo', cb => {
-      this.getDocumentInfo()
-        .then(info => cb(null, info))
-        .catch(reason => cb(reason));
-    });
-
     // Handler for controls on the sidebar
     this._sidebarRPC.on('setHighlightsVisible', showHighlights => {
       this.setHighlightsVisible(showHighlights);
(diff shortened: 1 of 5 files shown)


src/annotator/guest.js
```diff
@@ -183,6 +183,12 @@ export default class Guest {
       source: 'guest',
     });
 
+    /**
+     * Integration that handles document-type specific functionality in the
+     * guest.
+     */
+    this._integration = createIntegration(this);
```
robertknight, line 190: Creation of the integration was moved to happen earlier because `this._connectSidebarEvents` now relies on it via calling `this.getDocumentInfo()`.

src/annotator/guest.js
```diff
@@ -439,10 +432,16 @@ export default class Guest {
...
+    // Connect to sidebar and send document info/URIs to it.
+    //
+    // RPC calls are deferred until a connection is made, so these steps can
+    // complete in either order.
+    this._portFinder.discover('sidebar').then(port => {
+      this._sidebarRPC.connect(port);
+    });
+    this.getDocumentInfo().then(metadata =>
+      this._sidebarRPC.call('documentInfoChanged', metadata)
+    );
```
robertknight, line 444: If getting document info fails, the sidebar will remain in a loading state indefinitely. This was the case before as well. To improve this we'd need to add an appropriate error UI in the sidebar and also have a way for the guest to notify the sidebar about the error.

src/sidebar/services/test/frame-sync-test.js
```diff
@@ -35,17 +35,6 @@ const fixtures = {
...
-  // Response to the `getDocumentInfo` channel message for a frame displaying
-  // a PDF
-  pdfDocumentInfo: {
-    uri: 'http://example.org/paper.pdf',
-    metadata: {
-      documentFingerprint: '1234',
-      link: [{ href: 'http://example.org/paper.pdf' }, { href: 'urn:1234' }],
-    },
-    frameIdentifier: null,
-  },
```
robertknight, line 47: This fixture was unused.

src/annotator/guest.js
esanzgar, line 444: Why not continue sending the `error` and `data` (following the Node.js convention)?

```suggestion
    this.getDocumentInfo().then(metadata =>
      this._sidebarRPC.call('documentInfoChanged', null, metadata)
    ).catch(error => 
      this._sidebarRPC.call('documentInfoChanged', 'Failed to get the document information', null)
    );
```

Alternatively, `metadata` could be a `string` (for the text of the error) or the current data structure. If the text is meaningless, then a `null` could indicate the error.

In the future, once `Exceptions` are able to be transmitted among frames, it could be an `Exception` or the current data structure.

esanzgar (approved):
It looks good and the change in the name of the RPC event paves the way to better support single page applications.

I have provided a suggestion about maintaining the possibility of sending error information.

src/annotator/guest.js
robertknight: Yes, I think so.

src/annotator/guest.js
robertknight: > Alternative, we could continue to send an error and data (following the Node.js convention).

You mean make the arguments to `documentInfoChanged` have the form `(err, data)`? That could work. Another option would be to use a variant/union.

robertknight:
I have updated a couple of internal method names. I'm going to defer changing error handling to a subsequent PR - it wasn't properly implemented previously so we're not losing anything here.

PR #4126, opened 2022-01-20
Fix error when VitalSource integration injects client into content frame

robertknight:
Construction of the integration was recently moved to happen earlier in the
Guest.  This caused a regression in the VitalSource integration, which may
synchronously call `Guest#injectClient` when constructed. `injectClient` in turn
relies on `_hypothesisInjector` being initialized, which was not the case.

The error could be seen in the dev server's VS test case at
http://localhost:3000/document/vitalsource-epub.

Fix the issue by moving `_hypothesisInjector` initialization to happen earlier
in the Guest constructor, before `createIntegration` is called.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -176,6 +176,10 @@ export default class Guest {
       source: 'guest',
     });
 
+    // Set up automatic and integration-triggered injection of client into
+    // iframes in this frame.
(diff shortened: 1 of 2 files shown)


esanzgar (approved):
These hidden interdependences are tricky and dangerous. It is good to have a test that, from now on, check the correct order of the class instantiations.

However, this situation could indicate that the integration may be the process that drives by the injection.

PR #4127, opened 2022-01-20
Format `@implements {...}` JSDoc tag

esanzgar:
The `@implements` tag works with or without braces. We decided to
include braces following the documentation at JSDoc:

https://jsdoc.app/tags-implements.html

See also: https://github.com/hypothesis/client/pull/4048/files#r786462923


robertknight (approved):
👍

Issue #3992, opened 2021-12-03
Inconsistent re-anchoring in multi-guest scenarios


esanzgar:
Re-anchoring of annotations is inconsistent when there are several `guest` frames. This is because the `sidebar` send to all the `guest` frames the annotations. Every frame tries to anchor the best it can, and depending of the timing of the reporting the annotation can be tagged as `orphan`.

### Steps to reproduce
1. Go to http://localhost:3000/document/cross-origin-iframe
2. Make an annotation on a very common word
3. Reload the page
4. The annotation will be anchor in both frames

### Expected behaviour
It should only be anchored to the frame where the annotation was made.

### Actual behaviour
It appears on both frames

### Browser/system information
All browsers

https://user-images.githubusercontent.com/8555781/144642866-72108810-cfdb-44e5-a90d-7ea8e7e8b072.mov

### Steps to reproduce
1. Go to http://localhost:3000/document/cross-origin-iframe
2. Make an annotation on very long phrase or paragraph
3. Reload the page
4. The annotation will correctly anchor in one frames, but it _could_ be reported as orphan.

### Expected behaviour
It should only be anchored to the frame where the annotation was made, and it should be numbered in the `Annotations` tab.

### Actual behaviour
It sometimes appears on `Orphan` tab.

### Browser/system information
All browsers

https://user-images.githubusercontent.com/8555781/144643209-861728fe-6b5c-4b02-95fb-04cc4a8d8927.mov

robertknight:
This is related to one of the sub-tasks in https://github.com/hypothesis/client/issues/3798, so I edited that issue's description to reference this one. The description in that issue mentions two different ways of fixing the problem in the VitalSource context:

1. Don't load the guest functionality in the host frame, so there is only one guest.
2. Send annotations only to the correct frame when there are multiple guests

This issue discusses the second solution, which we need to support multiple guests in the general context. In an ebook reader, the first would be sufficient.

PR #4129, opened 2022-01-20
Revise how guest notifies host and sidebar when it is unloaded

robertknight:
Change how the guest notifies other frames, specifically the sidebar and host,
when it is unloaded. The host and sidebar now receive a `frameDestroyed` message
from the corresponding guest port, which allows them to easily close the right
port and remove it from the list of active guest ports.

Due to a Safari bug (see code comments) we can't send the `frameDestroyed`
messages from the guest frame while it is being unloaded. However it is possible
to first transfer the port to the host frame and then have the host frame send
the message on the same port. I also tried sending the message from the guest
frame, and then transferring the port to the host frame, but that didn't work.
This workaround has the advantage that it is transparent to the receiver of the
`frameDestroyed` message.

This change is also a step towards possibly not relying on user-provided guest
frame identifiers in the sidebar, which only become available once the
`documentInfoChanged` call has been received. Instead the sidebar could
use its own internal IDs for guest frames, avoiding the possibility for
conflicts.

Changes in detail:

 - Add `disconnect` method to PortRPC. This disconnects the MessagePort (if any) from the PortRPC and returns it. This allows the port to be transferred to another frame.

 - When guest is unloaded, transfer ports to the host frame in a
   `hypothesisGuestUnloaded` message, and make the host frame dispatch
   `frameDestroyed` calls on these ports.

 - Handle `frameDestroyed` in sidebar by closing port, removing it from
   the active guest list and removing the associated frame from the
   store

 - Handle `frameDestroyed` in host frame by closing port and removing it
   from the active guest list

---

**Testing:**

In Safari, Chrome and Firefox:

1. Load the example at http://localhost:3000/document/vitalsource-epub
2. Click the "Next chapter" and "Prev chapter" buttons to navigate backwards and forwards
3. Click on Help => About this version in the sidebar, and check that the "URLs" field shows two entries, one for the host page and one for the current guest chapter.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -225,7 +225,7 @@ export default class Guest {
      */
     this._focusedAnnotations = new Set();
 
-    this._listeners.add(window, 'unload', () => this._notifyGuestUnload());
+    this._listeners.add(window, 'unload', () => this._sendUnloadNotification());
   }
 
   // Add DOM event listeners for clicks, taps etc. on the document and
@@ -452,8 +452,12 @@ export default class Guest {
   }
 
   destroy() {
+    this._sendUnloadNotification();
+
     this._portFinder.destroy();
-    this._notifyGuestUnload();
+    this._hostRPC.destroy();
+    this._sidebarRPC.destroy();
+
     this._hypothesisInjector.destroy();
     this._listeners.removeAll();
 
@@ -464,21 +468,35 @@ export default class Guest {
     removeAllHighlights(this.element);
 
     this._integration.destroy();
-    this._sidebarRPC.destroy();
   }
 
   /**
-   * Notify the host frame that the guest is being unloaded.
+   * Notify other frames that the guest is being unloaded.
    *
-   * The host frame in turn notifies the sidebar app that the guest has gone away.
+   * In order to work around a bug in Safari 15 and below [1], this is done by
+   * first sending ports to the host frame and then, from the host frame,
+   * sending messages on the ports.
+   *
+   * [1] https://bugs.webkit.org/show_bug.cgi?id=231167.
    */
-  _notifyGuestUnload() {
-    const message = {
-      type: 'hypothesisGuestUnloaded',
-      frameIdentifier: this._frameIdentifier,
-    };
+  _sendUnloadNotification() {
+    const ports = [];
 
-    this._hostFrame.postMessage(message, '*');
+    const sidebarPort = this._sidebarRPC.disconnect();
+    if (sidebarPort) {
+      ports.push(sidebarPort);
+    }
+
+    const hostPort = this._hostRPC.disconnect();
+    if (hostPort) {
+      ports.push(hostPort);
+    }
+
+    this._hostFrame.postMessage(
+      { type: 'hypothesisGuestUnloaded' },
+      '*',
+      ports
+    );
   }
 
   /**
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -482,6 +482,16 @@ export default class Guest {
   _sendUnloadNotification() {
     const ports = [];
 
+    // There is currently a race condition where the guest can fail to notify
+    // the host and sidebar when it unloads if it has not yet received its ports
+    // for guest-host and guest-sidebar commmunication.
+    //
+    // For the moment this can be tolerated as guest frames are not displayed in
+    // the sidebar until the guest has received the port and sent a
+    // `documentInfoChanged` notification on it. Nevertheless the host / sidebar
+    // will be left with a disconnected PortRPC for the guest, so we should aim
+    // to fix this in future.
+
     const sidebarPort = this._sidebarRPC.disconnect();
     if (sidebarPort) {
       ports.push(sidebarPort);
--- a/src/annotator/test/guest-test.js
+++ b/src/annotator/test/guest-test.js
@@ -109,6 +109,7 @@ describe('Guest', () => {
         call: sinon.stub(),
         connect: sinon.stub(),
         destroy: sinon.stub(),
+        disconnect: sinon.stub(),
         on: sinon.stub(),
       };
       fakePortRPCs.push(rpc);
@@ -1244,7 +1245,12 @@ describe('Guest', () => {
     });
 
     it('notifies host frame that guest has been unloaded', () => {
+      const sidebarPort = {};
+      const hostPort = {};
+
       const guest = createGuest({ subFrameIdentifier: 'frame-id' });
+      hostRPC().disconnect.returns(hostPort);
+      sidebarRPC().disconnect.returns(sidebarPort);
 
       guest.destroy();
 
@@ -1252,15 +1258,19 @@ describe('Guest', () => {
         hostFrame.postMessage,
         {
           type: 'hypothesisGuestUnloaded',
-          frameIdentifier: 'frame-id',
         },
-        '*'
+        '*',
+        sinon.match.array.contains([sidebarPort, hostPort])
       );
     });
   });
 
   it('notifies host frame when guest frame is unloaded', () => {
+    const sidebarPort = {};
+    const hostPort = {};
     createGuest({ subFrameIdentifier: 'frame-id' });
+    hostRPC().disconnect.returns(hostPort);
+    sidebarRPC().disconnect.returns(sidebarPort);
 
     window.dispatchEvent(new Event('unload'));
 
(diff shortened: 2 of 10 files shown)


robertknight:
Note for my future self: I found some odd behavior with `unload` events when testing VitalSource this afternoon. I want to look into that next week before this lands. See https://github.com/hypothesis/client/pull/4131#issuecomment-1018586367.

esanzgar (commented):
This revised unloading notification mechanism achieves the same results as the previous mechanism: informing the host and sidebar frames that the guest has been unloaded. Both, previous and revised approaches, are workarounds to the bug found in Safari. 

Previously, the guest frame sent a `window.postMessage` (not using the `MessageChannel` API) to the host frame, who, in turn, relayed that information (using the `MessageChannel` API) to the sidebar. A frame identifier needed to be passed to identify the particular guest frame that was disconnected.

The new approach continues to use a hybrid approach: A message is sent outside the `MessageChannel` API (`window.postMessage`) between the guest and host, but then (difference) the host frame receives the ports from the guest frames (the opposite to what happens with `PortFinder#discover`), and on behalf of the guest frame, sends the `frameDestroyed` RPC events using those ports. In order to do that: the ports in the guest frame needs to be disconnected and reconnected again in the host frame. It relies on the `frameEvent` event registered by the guest, never de-registered, but called through the host.

@robertknight, outlined two advantages of this new approach:
1. no need to use frame identifiers to distinguish which frame is disconnected
2. the `frameDestroyed` RPC event flows `guest -> host` and `guest -> sidebar`, without relay

Personally, I don't see a big problem using frame identifiers to distinguish frames.

Regarding, point #2, while it is true that the message is not relayed, the guest and host frames have to do quite a number of unique, a bit counterintuitive, steps to achieve it the same as previously. 

The introduction of the public `PortRPC#disconnect`, in addition to `PortRPC#destroy`, could cause some confusion for developers. Since, `PortRPC#destroy` closes the port, it is unclear what happens if  `PortRPC#disconnect` and `PortRPC#destroy` were called in the same event loop.

Overall, I am concern about the complexity introduced for the small gains. I believe I may failing to use other future benefits.

robertknight:
> Personally, I don't see a big problem using frame identifiers to distinguish frames.

There are several problems with frame identifiers as they currently exist:

1. They are not a generic feature of `PortProvider`/`PortFinder` but rather a custom thing used by guests. This means that it is not possible for eg. the sidebar to know the frame ID that corresponds to a new guest immediately after it has received the port. This means that if a guest is unloaded before it has responded to `documentInfoChanged`, the sidebar cannot close the relevant port and do other cleanup.
2. Frame identifiers are currently allocated by whoever loads the client into the guest frame - either `HypothesisInjector` or end-user code. When end-users are allocating the values, there are no guarantees that they are unique, making them unsuitable for use in a `Map<frame ID, guest port>` map in the sidebar. Such a map will be useful in order to send messages to only specific guests (part of the solution to https://github.com/hypothesis/client/issues/3992).
3. The frame identifier value `null` is treated specially in some places as the "main" frame. This special value complicates the type of the frame ID, but also the concept doesn't really work in some scenarios. The "main" frame is implicitly assumed to have the same URL as the host frame in various places in the sidebar. In ebooks those places should really use the host frame URL instead.

> Since, PortRPC#destroy closes the port, it is unclear what happens if PortRPC#disconnect and PortRPC#destroy were called in the same event loop.

`disconnect` detaches the PortRPC from the port, so after it is called, no subsequent `PortRPC` operations will affect the port in any way. Whether the `disconnect` and `destroy` calls happen in the same event loop turn doesn't matter.

src/annotator/guest.js
```diff
@@ -465,21 +469,35 @@ export default class Guest {
...
+  _sendUnloadNotification() {
+    const ports = [];
+
+    const sidebarPort = this._sidebarRPC.disconnect();
+    if (sidebarPort) {
+      ports.push(sidebarPort);
+    }
 
-    this._hostFrame.postMessage(message, '*');
+    const hostPort = this._hostRPC.disconnect();
```
esanzgar, line 491: I have been thinking about what would happen if the guest frame is unloaded while there is an in-flight host discovery process. In other words, what would happened if  `this._hostRPC.disconnect()` is called before `this._hostRPC.connect(port)`. 

In that scenario, `this._hostRPC.disconnect()` will return `null` and it will fail to notify the host or sidebar frames of the unloading.

src/annotator/test/sidebar-test.js
```diff
@@ -486,6 +497,23 @@ describe('Sidebar', () => {
...
+        guestRPC().call.resetHistory();
+
+        emitGuestEvent('frameDestroyed');
+
+        assert.called(guestRPC().destroy);
+
+        // Trigger a notification to all connected guests. This should no longer
+        // be sent to the guest that has just been disconnected.
+        sidebar.open();
+        assert.notCalled(guestRPC().call);
```
esanzgar, line 514: 👍🏼 good way to test that the array of guestRPC doesn't contain any longer that particular guestRPC object.

src/sidebar/services/frame-sync.js
```diff
@@ -181,19 +181,33 @@ export class FrameSyncService {
...
+      frameIdentifier = info.frameIdentifier;
       this._store.connectFrame({
         id: info.frameIdentifier,
         metadata: info.metadata,
         uri: info.uri,
       });
     });
 
+    guestRPC.on('frameDestroyed', () => {
+      const frame = this._store.frames().find(f => f.id === frameIdentifier);
```
esanzgar, line 203: It feels that `this._store.frames` could be a `Map` instead of an array.

src/sidebar/services/frame-sync.js
```diff
@@ -181,19 +181,33 @@ export class FrameSyncService {
...
     guestRPC.on('documentInfoChanged', info => {
+      frameIdentifier = info.frameIdentifier;
       this._store.connectFrame({
         id: info.frameIdentifier,
         metadata: info.metadata,
         uri: info.uri,
       });
     });
 
+    guestRPC.on('frameDestroyed', () => {
```
esanzgar, line 202: Could `frameDestroyed` RPC event be received before the `documentInfoChanged`? If so, it will fail to destroy the frame.

esanzgar (approved):
I left a comment about a possible race condition in which an unload event is triggered in the guest frame before the discovery process has completed (still in-flight).

I could see that happening if there is a JS event navigation in the guest frame.

src/annotator/test/guest-test.js
```diff
@@ -1244,33 +1245,42 @@ describe('Guest', () => {
...
 
       assert.calledWith(
         hostFrame.postMessage,
         {
           type: 'hypothesisGuestUnloaded',
-          frameIdentifier: 'frame-id',
         },
-        '*'
+        '*',
+        sinon.match.array.contains([sidebarPort, hostPort])
```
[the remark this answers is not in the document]
robertknight: The proposed change works, although the reason for using `sinon.match.array.contains` here was to ignore the order of the ports. Same for the call below.

src/annotator/guest.js
robertknight: > In that scenario, this._hostRPC.disconnect() will return null and it will fail to notify the host or sidebar frames of the unloading.

This is true. Sadly the Channel Messaging API lacks a built-in solution for this. In the case of the guest-sidebar connection, we can use the lack of `documentInfoChanged` message as an indicator. I think it would make sense to have a generic solution in the `Port*` classes.

There might also be scenarios where the frame is unloaded without having the ability to send a notification on the port. I'm not sure yet if that can happen in any browsers or whether browsers will guarantee to unload entire tabs (eg. to save memory) as an atomic unit.

src/sidebar/services/frame-sync.js
robertknight: The reason that `frames` returns an array currently is because it returns data from a Redux store which only contains plain objects and arrays that are easily made immutable. It might make sense to add support for a lightweight immutable maps/sets in the store in future.

src/sidebar/services/frame-sync.js
robertknight: After `frameDestroyed` is received, no more RPC messages will be received from the guest. If we receive `frameDestroyed` first, then we will not receive `documentInfoChanged` and the frame will never be added to the store.

esanzgar: It is true, good point.

robertknight:
I added a comment in `Guest#_sendUnloadNotification` about the known race condition that can prevent the host/sidebar from receiving the `frameDestroyed` notification. We have discussed / are exploring solutions for this, but I believe we can tolerate it for the moment, which will allow https://github.com/hypothesis/client/pull/4131 to be shipped and then we can rectify the race later.

PR #4130, opened 2022-01-20
Improve consistency on bucket hovering and focusing

esanzgar:
On #4069, we introduced a small improvement when hovering a left-pointed
bucket: focus the corresponding annotation card, in addition to the
anchor's highlight.

In this PR, we introduce the same improvement to the up and down-pointed
buckets.

In addition, I have realised we handled `onBlur` but not `onFocus`
events. I have added more complete support for keyboard navigation.

I substituted `onMouseMove` for `onMouseEnter` because it is triggered
less frequently.

### Before

https://user-images.githubusercontent.com/8555781/150384308-5a032316-878c-4980-bd54-d8bde22d5dc1.mov

### After

https://user-images.githubusercontent.com/8555781/150383191-28cfff05-fa12-4052-a81e-72a4fd42cf15.mov

### DIFF
--- a/src/annotator/components/Buckets.js
+++ b/src/annotator/components/Buckets.js
@@ -20,6 +20,7 @@ function BucketButton({ bucket, onFocusAnnotations, onSelectAnnotations }) {
     onSelectAnnotations([...bucket.tags], event.metaKey || event.ctrlKey);
   }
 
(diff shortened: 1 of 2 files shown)


robertknight (approved):
This is a nice improvement. I appreciate the attention to detail!

PR #4131, opened 2022-01-21
Only send annotations to matching frame

robertknight:
**Depends on https://github.com/hypothesis/client/pull/4129**

When the sidebar is connected to multiple guest frames it will send all
incoming annotations to all frames. The result is typically that the
annotation will anchor in one frame and orphan in the others. Depending
on what order this happens in, the annotation will non-deterministically
show up as an Annotation or Orphan in the sidebar.

In order to determine which frames an annotation should be sent to in
all cases, we'd either need the backend to return information about which
search URIs an annotation matches or make a separate search request for
each frame and record the associated frame with the results. This
will require some significant refactoring of the annotation search
service.

As an interim step, make `FrameSyncService` send annotations only to a
single frame based on matching URL, with a fallback to sending to the
main frame if there is no exact match. This will work as expected for
most pages, and is at least deterministic when it does fail. When we
have a solution for being able to match annotations to frames more
generally, we can adapt this code to use it.

This is a partial solution to https://github.com/hypothesis/client/issues/3992.

**TODO:**

- [x] FrameSyncService needs some additional tests to cover the new behaviors.
- [x] Test this with VitalSource using the browser extension

**Testing:**

In one of the dev server test pages that contains an annotation-enabled iframe, try annotating the host frame (if possible) and the iframe, and then reload the page several times.

On master, annotations may sometimes appear in the "Orphans" tab. On this branch, all the annotations should consistently anchor in the expected frame.

### DIFF
--- a/src/sidebar/services/frame-sync.js
+++ b/src/sidebar/services/frame-sync.js
@@ -225,12 +225,15 @@ export class FrameSyncService {
     /** @type {PortRPC<GuestToSidebarEvent, SidebarToGuestEvent>} */
     const guestRPC = new PortRPC();
 
-    // Generate a temporary ID for this guest until we learn its "real" ID.
+    // Add guest RPC to map with a temporary ID until we learn the real ID.
+    //
+    // We need to add the guest to the map immediately so that any notifications
+    // sent from this service to all guests, before we learn the real frame ID,
+    // are sent to this new guest.
     ++this._nextGuestId;
     let frameIdentifier = /** @type {string|null} */ (
       `temp-${this._nextGuestId}`
     );
-
     this._guestRPC.set(frameIdentifier, guestRPC);
 
     // Update document metadata for this guest. We currently assume that the
--- a/src/sidebar/services/frame-sync.js
+++ b/src/sidebar/services/frame-sync.js
@@ -35,22 +35,18 @@ export function formatAnnot({ $tag, target, uri }) {
 }
 
 /**
- * Return the frame in `frames` which best matches `ann`.
+ * Return the frame which best matches an annotation.
+ *
+ * If there is a frame whose URL exactly matches the annotation, it will be
+ * returned. Otherwise the main frame will be returned. Currently this may fail
+ * to return the expected frame if an annotation was fetched from h that
+ * does not match one of the frame URIs. This can happen due to URI expansion /
+ * document equivalence in h.
  *
  * @param {Frame[]} frames
  * @param {Annotation} ann
  */
 function frameForAnnotation(frames, ann) {
-  // Choose the frame whose URL exactly matches this annotation. If there is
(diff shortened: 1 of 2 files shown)


robertknight:
I'm seeing an issue when testing this branch with VitalSource using the browser extension. The guest frame appears to be receiving an `unload` just after the guest has loaded and connected to the sidebar. This results in the guest being removed from the sidebar immediately after it is added.

robertknight:
> I'm seeing an issue when testing this branch with VitalSource using the browser extension. The guest frame appears to be receiving an unload just after the guest has loaded and connected to the sidebar

It looks like what is happening here is that chapter content loads in several phases:

1. The iframe for the previous chapter is removed
2. An iframe is created for the new chapter and navigated to an initial URL. For example https://jigsaw.vitalsource.com/books/9781319367077/epub/OEBPS/xhtml/lun_9781319056261_ch01_04.xhtml?brand=vitalsource&create=true
3. The iframe created in step 2 is then navigated to a new URL via a form submission. The form action URL has the same URL as step 2 but with `favre=brett` query params. For example https://jigsaw.vitalsource.com/books/9781319367077/epub/OEBPS/xhtml/lun_9781319056261_ch01_04.xhtml?brand=vitalsource&create=true&favre=brett.

If the client gets injected into the new frame after step 2, then it gets unloaded in step 3 and doesn't get re-injected, because the VitalSource integration only monitors for creation of iframes, not navigation of existing iframes.

src/sidebar/services/frame-sync.js
```diff
@@ -181,17 +226,28 @@ export class FrameSyncService {
...
+    // Add guest RPC to map with a temporary ID until we learn the real ID.
+    //
+    // We need to add the guest to the map immediately so that any notifications
+    // sent from this service to all guests, before we learn the real frame ID,
+    // are sent to this new guest.
+    ++this._nextGuestId;
+    let frameIdentifier = /** @type {string|null} */ (
+      `temp-${this._nextGuestId}`
+    );
+    this._guestRPC.set(frameIdentifier, guestRPC);
```
robertknight, line 239: In this PR I have not changed how frame IDs are allocated. If we change this in future so that frame IDs are either allocated entirely within the sidebar or in such a way that we can know them as soon as the guest connects, then we can simplify this code.

robertknight:
Regarding the VitalSource comments above, this was addressed by #4145.

esanzgar (approved):
It looks good. I left a couple of comments.

src/sidebar/services/frame-sync.js
robertknight: > In the future, we could easily add a capability to PortFinder#discover to send a frame identifier.

Yes. That's definitely one possibility.

PR #4144, opened 2022-01-24
Fix port discovery after a guest iframe is navigated

robertknight:
PortProvider previously used a `WeakMap<Window, Set<Channel>>` to keep
track of which ports had been requested from a particular source frame,
where the map keys came from `MessageEvent.source`. A problem with this
approach is that for messages from an iframe, `MessageEvent.source` refers to a
`WindowProxy` which retains the same identity across page navigations.
As a result if a guest iframe was navigated to a new page which also
embedded the client as a guest (or into which the client was
subsequently injected as a guest), the guest in the new document would
not be able to retrieve any ports.

Implement a fix whereby each call to `PortFinder.discover` generates a
random request ID which is sent with the initial port request plus any
re-sent requests. PortProvider then keeps track of the request IDs it
has seen and ignores messages with a previously seen request ID.

This change would also fix port discovery if the guest in a frame was
unloaded and later re-loaded.

### DIFF
--- a/src/shared/messaging/port-finder.js
+++ b/src/shared/messaging/port-finder.js
@@ -1,4 +1,5 @@
 import { ListenerCollection } from '../listener-collection';
+import { generateHexString } from '../random';
 import { isMessageEqual } from './port-util';
 
 /** Timeout for waiting for the host frame to respond to a port request. */
@@ -58,6 +59,8 @@ export class PortFinder {
       throw new Error('Invalid request of channel/port');
     }
 
+    const requestId = generateHexString(6);
+
     return new Promise((resolve, reject) => {
       const postRequest = () => {
         this._hostFrame.postMessage(
@@ -65,6 +68,7 @@ export class PortFinder {
             frame1: this._source,
             frame2: target,
             type: 'request',
+            requestId,
           },
           '*'
         );
@@ -96,6 +100,7 @@ export class PortFinder {
             frame1: this._source,
             frame2: target,
             type: 'offer',
+            requestId,
           })
         ) {
           clearInterval(intervalId);
--- a/src/shared/messaging/port-provider.js
+++ b/src/shared/messaging/port-provider.js
@@ -139,7 +139,7 @@ export class PortProvider {
     const handleRequest = event => {
       const { data, origin, source } = /** @type {MessageEvent} */ (event);
 
-      if (!isMessage(data) || data.type !== 'request' || !data.requestId) {
+      if (!isMessage(data) || data.type !== 'request') {
         // If this does not look like a message intended for us, ignore it.
         return;
       }
--- a/src/shared/messaging/port-provider.js
+++ b/src/shared/messaging/port-provider.js
@@ -71,15 +71,15 @@ export class PortProvider {
     this._emitter = new TinyEmitter();
 
     /**
-     * Map of source frame to channels that have been created for this frame.
+     * IDs of port requests that have been handled.
      *
      * This is used to avoid responding to the same request multiple times.
      * Guest frames in particular may send duplicate requests (see comments in
      * PortFinder).
      *
-     * @type {WeakMap<Window, Set<Channel>>}
+     * @type {Set<string>}
      */
-    this._sentChannels = new WeakMap();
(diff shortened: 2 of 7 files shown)


src/shared/port-provider.js
```diff
@@ -71,23 +71,23 @@ export class PortProvider {
...
      *
      * This is used to avoid responding to the same request multiple times.
      * Guest frames in particular may send duplicate requests (see comments in
      * PortFinder).
      *
-     * @type {WeakMap<Window, Set<Channel>>}
+     * @type {Set<string>}
      */
-    this._sentChannels = new WeakMap();
+    this._handledRequests = new Set();
```
robertknight, line 82: There is no expiry of used request IDs here. We just rely on the total number being small enough that it doesn't use significant memory to remember them forever.

src/shared/port-provider.js
```diff
@@ -71,23 +71,23 @@ export class PortProvider {
...
 
     // Create the `sidebar-host` channel immediately, while other channels are
     // created on demand
     this._sidebarHostChannel = new MessageChannel();
 
     this._listeners = new ListenerCollection();
 
-    /** @type {Array<Message & {allowedOrigin: string}>} */
+    /** @type {Array<Partial<Message> & {allowedOrigin: string}>} */
```
robertknight, line 90: `Partial` is needed because `_allowedMessages` doesn't include the `requestId` field, since that changes with each request.

src/shared/port-util.js
```diff
@@ -35,7 +37,7 @@ export function isMessage(data) {
  * Return true if the data payload from a MessageEvent matches `message`.
  *
  * @param {any} data
- * @param {Message} message
+ * @param {Partial<Message>} message
```
robertknight, line 40: This allows `isMessageEqual` to match `data` against a subset of fields of a message. In particular, to ignore the `requestId`.

src/shared/messaging/port-provider.js
```diff
@@ -139,12 +139,12 @@ export class PortProvider {
     const handleRequest = event => {
       const { data, origin, source } = /** @type {MessageEvent} */ (event);
 
-      if (!isMessage(data) || data.type !== 'request') {
+      if (!isMessage(data) || data.type !== 'request' || !data.requestId) {
```
esanzgar, line 142: Why don't you update `isMessage` to check for the presence of the `requestId` property?

src/shared/messaging/port-finder.js
```diff
@@ -58,13 +59,16 @@ export class PortFinder {
       throw new Error('Invalid request of channel/port');
     }
 
+    const requestId = generateHexString(6);
```
esanzgar, line 62: Something to think for the future: it may be valuable to return the requestId value together with the port in the promise of `PortFinder#discover`.

`const [sidebarPort, requestId] = await PortFinder.discover('sidebar')`

The `requestId` could be used to identify the provenance of certain RPC events, in cases of one to many connections: `1 sidebar : N guests` or `1 host: N guests`

src/shared/messaging/port-util.js
```diff
@@ -35,7 +37,7 @@ export function isMessage(data) {
  * Return true if the data payload from a MessageEvent matches `message`.
  *
  * @param {any} data
- * @param {Message} message
+ * @param {Partial<Message>} message
```
esanzgar, line 40: Isn't it too lenient to mark all the `Message` properties as optional? As suggested earlier, why not to test for `requestId` too?

src/shared/messaging/test/port-util-test.js
```diff
@@ -54,6 +57,7 @@ describe('port-util', () => {
           frame1,
           frame2,
           type,
+          requestId,
```
esanzgar, line 60: I would remove these `requestId` properties. They are not checked and a message is valid without them.

esanzgar (approved):
I left a few comments.

src/shared/messaging/port-provider.js
robertknight: That would make sense. I think I didn't initially because `requestId` was an optional property only present on `request` messages initially.

PR #4146, opened 2022-01-24
Remove obsolete reference to `__hypothesis` global

robertknight:
The `hasHypothesis` function always returned false because it relied on a global
variable that no longer exists. The fact that it was broken didn't cause
noticeable problems because the client's boot script also has a check for
whether Hypothesis is already injected.

Replace the broken check with a working one that corresponds to the nearby
`injectHypothesis` function.

### DIFF
--- a/src/annotator/hypothesis-injector.js
+++ b/src/annotator/hypothesis-injector.js
@@ -85,13 +85,13 @@ export class HypothesisInjector {
 }
 
 /**
- * Check if the Hypothesis client has already been injected into an iframe
+ * Check if the client was added to a frame by {@link injectHypothesis}.
  *
  * @param {HTMLIFrameElement} iframe
  */
 function hasHypothesis(iframe) {
-  const iframeWindow = /** @type {Window} */ (iframe.contentWindow);
-  return '__hypothesis' in iframeWindow;
+  const iframeDocument = /** @type {Document} */ (iframe.contentDocument);
+  return iframeDocument.querySelector('script.js-hypothesis-config') !== null;
 }
 
 /**
--- a/src/annotator/test/integration/hypothesis-injector-test.js
+++ b/src/annotator/test/integration/hypothesis-injector-test.js
@@ -155,7 +155,11 @@ describe('HypothesisInjector integration test', () => {
 
   it('excludes injection from already injected iframes', async () => {
     const iframe = createAnnotatableIFrame();
-    iframe.contentWindow.eval('window.__hypothesis = {}');
+
+    // Add a client config in the same way as the injector.
+    const configScript = document.createElement('script');
+    configScript.className = 'js-hypothesis-config';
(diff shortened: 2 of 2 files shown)


src/annotator/hypothesis-injector.js
```diff
@@ -85,13 +85,13 @@ export class HypothesisInjector {
...
 /**
- * Check if the Hypothesis client has already been injected into an iframe
+ * Check if the client was added to a frame by {@link injectHypothesis}.
  *
  * @param {HTMLIFrameElement} iframe
  */
 function hasHypothesis(iframe) {
-  const iframeWindow = /** @type {Window} */ (iframe.contentWindow);
-  return '__hypothesis' in iframeWindow;
+  const iframeDocument = /** @type {Document} */ (iframe.contentDocument);
```
esanzgar, line 93: This will raise an error if accessing a cross-origin frame.

src/annotator/hypothesis-injector.js
```diff
@@ -85,13 +85,13 @@ export class HypothesisInjector {
...
+  return iframeDocument.querySelector('script.js-hypothesis-config') !== null;
```
esanzgar, line 94: Strictly, Hypothesis client could be injected without the need of a configuration (default values), isn't?

Wouldn't be better to check for the boot script?

esanzgar (commented):
I had a query for you.

src/annotator/hypothesis-injector.js
esanzgar: The problem with the boot script is that it can have several names:

* `.../build/boots.js`
* `https://cdn.hypothes.is/hypothesis`

esanzgar: An alternative suggestion: check for `link[type="application/annotator+html"][rel="sidebar"]`. This element is created just before the boot script executes. We use this link to destroy the application (see 'src/annotator/index`), so maybe it is also good as a check if the application has started. It also is specific enought.

src/annotator/hypothesis-injector.js
robertknight: In the context `hasHypothesis` is used, the `injectClient` method, it is specified that a prerequisite is for the frame to be same-origin. An option would be to simply mention this requirement in the docs.

src/annotator/hypothesis-injector.js
robertknight: > This element is created just before the boot script executes. 

This element is created _by_ the boot script. I think this would be a good suggestion except that it leads to a race condition - `hasHypothesis` will return true immediately after the boot script is injected. This is why the code currently looks for the configuration that is injected prior to the boot script. It is true that the client can be injected without any configuration, but in that case it will function as a host frame rather than a guest only frame.

esanzgar (approved):
👍🏼

PR #4168, opened 2022-01-31
Remove a reference to `Bridge` and strengthen type

esanzgar:
1. Remove a reference to the defunct class `Bridge`.

2. Make the type of `OnMethod` and `CallMethod` the union of all allowed
   RPC event names, instead of a generic string.


esanzgar:
@robertknight, has suggested to keep `PortRPC` generic and not bound to only events in the client.

PR #4175, opened 2022-02-03
Add built-in "connect" and "close" events to PortRPC

robertknight:
Add two built-in events to PortRPC which are dispatched when PortRPC connects to
a port and when it is destroyed or the containing frame is unloaded. These
events can be used in the counterpart PortRPC to confirm that the sender has
successfully received and connected to the port, and to get notified when the
port goes away.

Sending the `close` event in the context of a window unloading in Safari
<= 15 requires a workaround that involves registering a handler in the
parent frame. This handler is currently installed in the host frame.

The new "close" event is used to replace the "frameDestroyed" message
that was used by guests to notify the sidebar when it went away. This solves
several problems:

 - It centralizes the workaround for https://bugs.webkit.org/show_bug.cgi?id=231167 in `post-rpc.js`, instead of having it spread between several modules. A benefit of this is that the workaround can be tested end-to-end more easily. See the PortRPC tests.
 - It provides a generic way to for a PortRPC user to know when the channel has closed, rather than needing to implement this for each scenario where it is needed
 - If MessagePort gains a built-in close event in future, which I hope it will, since this really requires browser support to handle all scenarios (eg. process crashes), it will be easy to adapt PortRPC to use it
 - It provides a way to, in future, handle the case where a guest is unloaded before it has received and connected to the port, by having the host/sidebar frames expect the "connect" call within a timeout. This is not yet implemented.

**Changes in detail:**

 - Add built-in `close` and `connect` events to `PortRPC`
 - Remove `frameDestroyed` event handler in `FrameSyncService` and replace with `close` handler
 - Install workaround for `close` event delivery in host frame, in the annotator entry point
 - Add `close` event handler in `Sidebar` class, which cleans up the host side of the guest-host connection when the guest goes away
 - Remove the `PortRPC#disconnect` method. It is no longer used and it complicates the internals and usage of PortRPC as it adds additional state transitions (initialized => connected => disconnected => connected etc.) that have to be handled.

The `connect` event is not yet used in this PR. That will be used in a follow-up to clean up guest connections if a guest navigates away before it receives a port.

**Testing:**

Try the following in Safari, Chrome and Firefox:

1. Go to http://localhost:3000/document/vitalsource-epub
2. Add an annotation at the top of each book chapter
3. Navigate back and forth between the chapters, and verify that the correct annotations are shown in the sidebar after each navigation

### DIFF
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -56,7 +56,8 @@ function init() {
 
   if (hostFrame === window) {
     // Ensure port "close" notifications from eg. guest frames are delivered properly.
-    installPortCloseWorkaroundForSafari();
+    const removeWorkaround = installPortCloseWorkaroundForSafari();
+    destroyables.push({ destroy: removeWorkaround });
 
     const sidebarConfig = getConfig('sidebar');
 
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -9,7 +9,10 @@ import { registerIcons } from '@hypothesis/frontend-shared';
 import { annotatorIcons } from './icons';
 registerIcons(annotatorIcons);
 
-import { PortProvider } from '../shared/messaging';
+import {
+  PortProvider,
+  installPortCloseWorkaroundForSafari,
+} from '../shared/messaging';
 import { getConfig } from './config/index';
 import { Guest } from './guest';
 import { HypothesisInjector } from './hypothesis-injector';
@@ -52,6 +55,9 @@ function init() {
   const destroyables = [];
 
   if (hostFrame === window) {
+    // Ensure port "close" notifications from eg. guest frames are delivered properly.
+    installPortCloseWorkaroundForSafari();
+
     const sidebarConfig = getConfig('sidebar');
 
     const hypothesisAppsOrigin = new URL(sidebarConfig.sidebarAppUrl).origin;
--- a/src/shared/messaging/test/port-rpc-test.js
+++ b/src/shared/messaging/test/port-rpc-test.js
@@ -175,23 +175,54 @@ describe('PortRPC', () => {
     assert.calledWith(testMethod, 'second', 'call');
   });
 
-  describe('#disconnect', () => {
-    it('disconnects and returns port', async () => {
-      const rpc = new PortRPC();
-      const testMethod = sinon.stub();
-      rpc.on('test', testMethod);
+  it('should send "connect" event after connection', async () => {
+    const { port1, port2 } = new MessageChannel();
+    const sender = new PortRPC();
+    const receiver = new PortRPC();
+    const connectHandler = sinon.stub();
 
-      assert.isNull(rpc.disconnect());
+    receiver.on('connect', connectHandler);
+    receiver.connect(port2);
+    sender.connect(port1);
 
-      const { port1 } = new MessageChannel();
-      rpc.connect(port1);
+    await waitForMessageDelivery();
 
-      assert.equal(rpc.disconnect(), port1);
-      assert.isNull(rpc.disconnect());
+    assert.calledWith(connectHandler);
+  });
 
-      rpc.call('test');
-      await waitForMessageDelivery();
-      assert.notCalled(testMethod);
-    });
+  it('should send "close" event after port is destroyed', async () => {
+    const { port1, port2 } = new MessageChannel();
+    const sender = new PortRPC();
+    const receiver = new PortRPC();
+    const closeHandler = sinon.stub();
+
+    receiver.on('close', closeHandler);
+    receiver.connect(port2);
+    sender.connect(port1);
+    await waitForMessageDelivery();
+
+    closeHandler.resetHistory();
+    sender.destroy();
+    await waitForMessageDelivery();
+
+    assert.calledWith(closeHandler);
+  });
+
+  it('should send "close" event when window is unloaded', async () => {
+    const { port1, port2 } = new MessageChannel();
+    const sender = new PortRPC();
+    const receiver = new PortRPC();
+    const closeHandler = sinon.stub();
+
+    receiver.on('close', closeHandler);
+    receiver.connect(port2);
+    sender.connect(port1);
+    await waitForMessageDelivery();
+
+    closeHandler.resetHistory();
+    window.dispatchEvent(new Event('unload'));
+    await waitForMessageDelivery();
+
+    assert.calledWith(closeHandler);
   });
 });
--- a/src/shared/messaging/test/port-rpc-test.js
+++ b/src/shared/messaging/test/port-rpc-test.js
@@ -1,4 +1,4 @@
-import { PortRPC } from '../port-rpc';
+import { PortRPC, installPortCloseWorkaroundForSafari } from '../port-rpc';
 
 describe('PortRPC', () => {
   let port1;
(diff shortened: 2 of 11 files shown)


src/shared/messaging/test/port-rpc-test.js
```diff
@@ -175,23 +175,158 @@ describe('PortRPC', () => {
...
+    const sender = new PortRPC();
+    const receiver = new PortRPC();
+    const closeHandler = sinon.stub();
+
+    receiver.on('close', closeHandler);
+    receiver.connect(port2);
+    sender.connect(port1);
+    await waitForMessageDelivery();
+
+    closeHandler.resetHistory();
```
esanzgar, line 204: I believe this line should not be here. `closeHandler` handler should be invoked only once.
```suggestion
```

src/shared/messaging/test/port-rpc-test.js
```diff
@@ -175,23 +175,158 @@ describe('PortRPC', () => {
...
+    sender.destroy();
+    await waitForMessageDelivery();
+
+    assert.calledWith(closeHandler);
```
esanzgar, line 208:
```suggestion
    assert.calledOnce(closeHandler);
    assert.calledWith(closeHandler);
```

src/shared/messaging/test/port-rpc-test.js
```diff
@@ -175,23 +175,158 @@ describe('PortRPC', () => {
...
 
-      const { port1 } = new MessageChannel();
-      rpc.connect(port1);
+    closeHandler.resetHistory();
+    window.dispatchEvent(new Event('unload'));
+    await waitForMessageDelivery();
 
-      assert.equal(rpc.disconnect(), port1);
-      assert.isNull(rpc.disconnect());
+    assert.calledWith(closeHandler);
```
esanzgar, line 226:
```suggestion
    assert.calledOnce(closeHandler);
    assert.calledWith(closeHandler);
```

src/shared/messaging/test/port-rpc-test.js
```diff
@@ -175,23 +175,158 @@ describe('PortRPC', () => {
...
 
-      assert.isNull(rpc.disconnect());
+    receiver.on('close', closeHandler);
+    receiver.connect(port2);
+    sender.connect(port1);
+    await waitForMessageDelivery();
 
-      const { port1 } = new MessageChannel();
-      rpc.connect(port1);
+    closeHandler.resetHistory();
```
esanzgar, line 222:
```suggestion
```

src/shared/messaging/test/port-rpc-test.js
```diff
@@ -175,23 +175,158 @@ describe('PortRPC', () => {
...
+
+    await waitForMessageDelivery();
+
+    assert.calledOnce(closeHandler);
+  });
+
+  /** Transfer a MessagePort to another frame and return the transferred port. */
+  async function transferPort(port, targetWindow) {
+    const transferredPort = new Promise(resolve => {
+      targetWindow.addEventListener('message', e => {
```
esanzgar, line 252: That's interesting... setting an event listener in another window from the current window. And then drive the port from the current window. 🤯

src/shared/messaging/test/port-rpc-test.js
```diff
@@ -175,23 +175,158 @@ describe('PortRPC', () => {
...
+
+  [
+    // Chrome 100
+    'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4867.0 Safari/537.36',
+
+    // Firefox 96
+    'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:96.0) Gecko/20100101 Firefox/96.0',
+  ].forEach(userAgent => {
+    it('does not use workaround in unaffected browsers', () => {
+      sinon.stub(window, 'addEventListener');
```
esanzgar, line 325: Shouldn't a sinon `sandbox` would be better here? So that `sandbox.restore` is called on `afterEach`?

src/shared/messaging/test/port-rpc-test.js
```diff
@@ -175,23 +175,158 @@ describe('PortRPC', () => {
...
+      const receiver = new PortRPC();
+      const closeHandler = sinon.stub();
+
+      receiver.on('close', closeHandler);
+      receiver.connect(port2);
+      sender.connect(transferredPort);
       await waitForMessageDelivery();
-      assert.notCalled(testMethod);
+
+      closeHandler.resetHistory();
```
esanzgar, line 300: Doesn't this send the message that `closeHandler` could be triggered several times?

src/shared/messaging/test/port-rpc-test.js
```diff
@@ -175,23 +175,158 @@ describe('PortRPC', () => {
...
+
+      // Emulate the Safari bug by disabling `postMessage` on the sending port
+      // in the original frame. When the port is transferred to the "parent"
+      // frame, it will reconstituted as a new MessagePort instance.
+      transferredPort.postMessage = () => {};
```
esanzgar, line 305: Smart!

esanzgar (approved):
Great job, Robert!

I left a couple of suggestion. For symmetry with other abstractions in the annotator, I think `installPortCloseWorkaroundForSafari` could/should be converted into a class that implements the `Destroyable` interface.

src/shared/messaging/test/port-rpc-test.js
robertknight: The purpose of this was to make sure that `close` is definitely dispatched in response to `sender.destroy()` and not any earlier. I changed this to an assert to be clearer about that. Same for other `closeHandler.resetHistory` calls.

src/shared/messaging/test/port-rpc-test.js
robertknight: I changed this to an assert. See https://github.com/hypothesis/client/pull/4175#discussion_r801355241.

src/shared/messaging/test/port-rpc-test.js
robertknight: I tried a few different approaches before I landed on the one currently used in this PR. Hopefully this will be a useful reference for some other test in future.

src/shared/messaging/test/port-rpc-test.js
robertknight: I agree, except that the `sinon` object is itself a sandbox in modern Sinon versions so these mocks can be reset with `sinon.restore()` in a suite-level `afterEach`.

src/shared/messaging/test/port-rpc-test.js
robertknight: I changed the various `resetHistory` calls to asserts.

src/annotator/index.js
```diff
@@ -52,6 +55,9 @@ function init() {
   const destroyables = [];
 
   if (hostFrame === window) {
+    // Ensure port "close" notifications from eg. guest frames are delivered properly.
+    installPortCloseWorkaroundForSafari();
```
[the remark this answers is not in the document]
robertknight: Good point. I used a very simple adapter to convert the returned unsubscriber to a `Destroyable` and add it to `destroyables`:

```diff
diff --git a/src/annotator/index.js b/src/annotator/index.js
index 8da6fffcf..1389decc6 100644
--- a/src/annotator/index.js
+++ b/src/annotator/index.js
@@ -56,7 +56,8 @@ function init() {

   if (hostFrame === window) {
     // Ensure port "close" notifications from eg. guest frames are delivered properly.
-    installPortCloseWorkaroundForSafari();
+    const removeWorkaround = installPortCloseWorkaroundForSafari();
+    destroyables.push({ destroy: removeWorkaround });

     const sidebarConfig = getConfig('sidebar');
```

I deferred changing `installPortCloseWorkaroundForSafari` to a class for the moment as its class-ness would only be used for the `destroy` method, and we do have a pattern in the code of functions/methods which return unsubscribe functions. A variation on this idea would be to have the function return some object which implements `Destroyable`.

PR #4176, opened 2022-02-03
Do not create a `Guest` instance in the VS container

esanzgar:
We want to avoid creating an unnecessary `Guest` instance in the
VitalSource container. We follow these steps:

* The `VitalSourceContainerIntegration` is stripped down and converted
  to a class, `VitalSourceInjector`, that does not implement the
  Integration interface but is only responsible for injecting the client
  into content frames.

* The host frame checks the VS frame role and either constructs a Guest
  or sets up the `VitalSourceInjector` as appropriate.

* The `HypothesisInjector`-related logic in `Guest` is extracted out of
  that class and set up in the annotator entry point instead.

Closes https://github.com/hypothesis/product-backlog/issues/1235

### DIFF
--- a/src/annotator/hypothesis-injector.js
+++ b/src/annotator/hypothesis-injector.js
@@ -26,7 +26,7 @@ export class HypothesisInjector {
     this._config = config;
     this._frameObserver = new FrameObserver(
       element,
-      frame => this.injectClient(frame), // Frame added callback
+      frame => injectClient(frame, config), // Frame added callback
       () => {} // Frame removed callback
     );
   }
@@ -36,51 +36,6 @@ export class HypothesisInjector {
    */
   destroy() {
     this._frameObserver.disconnect();
-  }
-
-  /**
-   * Inject Hypothesis client into a frame.
-   *
-   * IMPORTANT: This method requires that the iframe is "accessible"
-   * (frame.contentDocument|contentWindow is not null).
-   *
-   * This waits for the frame to finish loading before injecting the client.
-   * See {@link onDocumentReady}.
-   *
-   * @param {HTMLIFrameElement} frame
-   */
-  async injectClient(frame) {
-    if (hasHypothesis(frame)) {
-      return;
-    }
-
-    await onDocumentReady(frame);
-
-    // Propagate the client resource locations from the current frame.
-    //
-    // These settings are set only in the browser extension and not by the
-    // embedded client (served by h).
-    //
-    // We could potentially do this by allowing these settings to be part of
-    // the "annotator" config (see `annotator/config/index.js`) which gets passed
-    // to the constructor.
-    const { assetRoot, notebookAppUrl, sidebarAppUrl } =
-      parseJsonConfig(document);
-
-    // Generate a random string to use as a frame ID. The format is not important.
-    const subFrameIdentifier = generateHexString(10);
-    const injectedConfig = {
-      ...this._config,
-
-      assetRoot,
-      notebookAppUrl,
-      sidebarAppUrl,
-
-      subFrameIdentifier,
-    };
-
-    const { clientUrl } = this._config;
-    injectHypothesis(frame, clientUrl, injectedConfig);
   }
 }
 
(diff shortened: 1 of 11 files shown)


src/annotator/hypothesis-injector.js
```diff
@@ -95,24 +50,57 @@ function hasHypothesis(iframe) {
...
 
 /**
- * Inject the client's boot script into the iframe. The iframe must be from the
- * same origin as the current window.
+ * Inject Hypothesis client into a frame.
  *
- * @param {HTMLIFrameElement} iframe
- * @param {string} scriptSrc
- * @param {Record<string, any>} config
+ * IMPORTANT: This method requires that the iframe is "accessible"
```
robertknight, line 55:
```suggestion
 * IMPORTANT: This method requires that the iframe is same-origin
```

The specific test used to determine whether one document can reach into another is whether they are [same-origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-bcc-content-document).

src/annotator/index.js
```diff
@@ -41,8 +46,11 @@ function init() {
 
   const hostFrame = annotatorConfig.subFrameIdentifier ? window.parent : window;
 
+  /** @type {Sidebar|undefined} */
   let sidebar;
```
robertknight, line 50: I expected that TypeScript would infer the type of these variables automatically based on subsequent assignments. It appears however that it doesn't when the variable is accessed in a callback.

In this code:

```js
sidebar = new Sidebar(document.body, eventBus, sidebarConfig);
notebook = new Notebook(document.body, eventBus, getConfig('notebook'));

portProvider.on('frameConnected', (source, port) =>
  /** @type {Sidebar} */ (sidebar).onFrameConnected(source, port)
);
```

The type of `sidebar` will be inferred as `Sidebar` in the top-level function after the assignment, but it is inferred as `any` within the callback.

We could actually avoid the need for `let` here. The only use of `sidebar` outside the `if (hostFrame === window) { ... }` block is to call `destroy` later. The same applies to the other conditionally-defined variables. We could collect all of these in a `Destroyable[]` and just loop over the array to cleanup. This would allow the variables to be scoped to the block where they are assigned and assigned to a `const` binding.

src/annotator/index.js
```diff
@@ -55,18 +63,34 @@ function init() {
...
+  let guest;
+  /** @type {VitalSourceInjector|undefined} */
+  let vitalSourceInjector;
+  /** @type {HypothesisInjector|undefined} */
+  let hypothesisInjector;
+  const vsFrameRole = vitalSourceFrameRole();
+  if (vsFrameRole === 'container') {
+    vitalSourceInjector = new VitalSourceInjector(annotatorConfig);
+  } else {
+    // Set up automatic and integration-triggered injection of client into
```
robertknight, line 80: The "integration-triggered injection" part no longer applies.

src/annotator/integrations/index.js
```diff
@@ -15,7 +14,7 @@ import {
  * Create the integration that handles document-type specific aspects of
  * guest functionality.
  *
- * @param {Annotator} annotator
+ * @param {Pick<Annotator, 'anchor'|'anchors'>} annotator
```
robertknight, line 17: The definition of the `Annotator` type is:

```js
/**
 * Subset of the `Guest` class that is exposed to integrations.
 *
 * @typedef Annotator
 * @prop {Anchor[]} anchors
 * @prop {(ann: AnnotationData) => Promise<Anchor[]>} anchor
 */
```

It is already narrowly scoped for the purpose of use by integrations and only has these two fields, so there is no need to refine it further.

src/annotator/integrations/vitalsource.js
```diff
@@ -38,18 +39,14 @@ export function vitalSourceFrameRole(window_ = window) {
 }
 
 /**
- * Integration for the container frame in VitalSource's Bookshelf ebook reader.
- *
- * This frame cannot be annotated directly. This integration serves only to
- * load the client into the frame that contains the book content.
- *
- * @implements {Integration}
+ * Observe the book content iframe and load the client into this frame.
```
robertknight, line 42: This description could be clearer about where this code runs. eg. "VitalSourceInjector runs in the book container frame and loads the client into book content frames."

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -64,37 +68,36 @@ describe('HypothesisInjector integration test', () => {
...
-      assert.match(config.subFrameIdentifier, /[0-9]+/);
+      assert.match(config.subFrameIdentifier, /[a-f0-9]+/);
       assert.notOk(config.assetRoot);
       assert.notOk(config.notebookAppUrl);
       assert.notOk(config.sidebarAppUrl);
     });
 
     it('copies client asset locations from host frame', async () => {
       hostJSONConfig = {
+        clientUrl: 'chrome-extension:///o',
```
robertknight, line 91: It looks like this line is incomplete. Can we make it a realistic-looking URL that matches the pattern of the others.

robertknight (changes requested):
The overall structure and functionality of this PR is good. I added feedback on some details.

esanzgar (commented):
I addressed your suggestion in the follow up commit.

src/annotator/test/integration/hypothesis-injector-test.js
```diff
@@ -64,37 +68,36 @@ describe('HypothesisInjector integration test', () => {
...
-      assert.match(config.subFrameIdentifier, /[0-9]+/);
+      assert.match(config.subFrameIdentifier, /[a-f0-9]+/);
       assert.notOk(config.assetRoot);
       assert.notOk(config.notebookAppUrl);
       assert.notOk(config.sidebarAppUrl);
     });
 
     it('copies client asset locations from host frame', async () => {
       hostJSONConfig = {
+        clientUrl: 'chrome-extension://abc/client/build/boots.js',
```
robertknight, line 91:
```suggestion
        clientUrl: 'chrome-extension://abc/client/build/boot.js',
```

robertknight (approved):
Changes LGTM.

PR #4177, opened 2022-02-04
Show buckets in the bucket-bar for VitalSource

esanzgar:
### Background information

Currently, the bucket-bar should display anchor positions only from a
_single_ guest frame. This is because there is no merging mechanism for
anchor positions from multiple guest frames.

### Before this PR

* The host frame listened for `anchorsChanged` events from
_all_ the guest frames.

* Only _one_ guest frame sent this event. We referred to this guest
  frame as having the 'main' annotatable content.  The 'main'
  annotatable guest frame was identified by not having a
  `subFrameIdentifier` (a configuration option added to a guest frame
  when the Hypothesis client was injected). Hence, the 'main'
  annotatable guest frame was always the frame where the Hypothesis
  loaded initially (in contrast with the injection mechanism).

### In this PR

We have reversed the logic of how the `anchorChanged` RPC
events are send and received:

* _Every_ guest frame sends `anchorsChanged` RPC events to the host
  frame.

* The host frame chooses to listen to only _one_ guest frame for the
  `anchorsChanged` RPC events: the guest frame that connects first with
  the host.

For non-VitalSource scenarios, the first guest frame that connects to the
host frame is the one where the Hypothesis client was initially loaded
(doesn't contain `subFrameIdentifier` option).

For the VitalSource, because of #4176, there is no guest frame in the
host frame. This allows the guest frame in the book content, where the
Hypothesis client was injected, to send anchor positions, and positions
to be reflected in the host's bucket-bar.

Closes https://github.com/hypothesis/product-backlog/issues/1234

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -301,15 +301,20 @@ export class Sidebar {
     );
 
     // The listener will do nothing if the sidebar doesn't have a bucket bar
-    // (clean theme), but it is still actively listening.
-    guestRPC.on(
-      'anchorsChanged',
-      /** @param {AnchorPosition[]} positions  */
-      positions => {
-        // Currently, only one guest frame sends anchor positions to the bucket bar
-        this.bucketBar?.update(positions);
-      }
-    );
+    // (clean theme)
+    const bucketBar = this.bucketBar;
+    // Currently, we ignore `anchorsChanged` for all the guests except the first connected guest.
+    if (bucketBar) {
+      guestRPC.on(
+        'anchorsChanged',
+        /** @param {AnchorPosition[]} positions  */
+        positions => {
+          if (this._guestRPC.indexOf(guestRPC) === 0) {
+            bucketBar.update(positions);
+          }
+        }
+      );
+    }
 
     guestRPC.on('frameDestroyed', () => {
       guestRPC.destroy();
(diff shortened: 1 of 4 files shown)


src/annotator/sidebar.js
```diff
@@ -301,15 +301,16 @@ export class Sidebar {
...
-      /** @param {AnchorPosition[]} positions  */
-      positions => {
-        // Currently, only one guest frame sends anchor positions to the bucket bar
-        this.bucketBar?.update(positions);
-      }
-    );
+    // (clean theme)
+    const bucketBar = this.bucketBar;
+    // Currently, we ignore `anchorsChanged` for all the guests except the first connected guest.
+    if (this._guestRPC.length === 0 && bucketBar) {
```
robertknight, line 307: There is a potential issue here if a new guest is added and then one is removed. For example if in an ebook the new chapter iframe was created and the old one was removed and the new guest was able to register itself before the old one unloaded. What we could do to avoid this is to always register the listener, but check if the guest is the first entry in the `this._guestRPC` array when the callback is executed.

robertknight (commented):
Functionally this works correctly in our VS test pages and in the real VS viewer, however it relies on the ordering in which guests are added and removed when changing chapters. I suggested changing the logic slightly to make it more robust.

src/annotator/sidebar.js
esanzgar: 👍🏼

robertknight (approved):

Issue #4191, opened 2022-02-07
Send annotations to correct frame when annotation URL does not match frame URL


robertknight:
https://github.com/hypothesis/client/pull/4131 implemented initial support for sending annotations to a specific frame based on a match between the annotation URL and the frame URL. This means that if an annotation is fetched for a non-main guest frame based on some other identifier (eg. a DOI, a fingerprint) or the backend returns an annotation with URL A when viewing frame with URL B because URLs A and B are deemed equivalent by the h backend, then the annotation will not be sent to the frame.

The client currently requests annotations for all frames with a single sequence of search requests, and the backend does not provide information about why a particular URL matched the query (eg. which of the search URLs matched). In order to fix this we'll either need to make separate calls to the `/api/search` endpoint for each frame, or get h to provide information about why a particular annotation was returned in the response.

Related issues:
- https://github.com/hypothesis/client/issues/3992
- https://github.com/hypothesis/client/issues/4153#issuecomment-1031383538

robertknight:
I think it is likely we will need to solve this as part of https://github.com/hypothesis/product-backlog/issues/1253.

PR #4193, opened 2022-02-08
WIP - Fetch annotations for each frame separately

robertknight:
Explore a client-side solution to sending annotations to correct frame
if annotation URL does not exactly match frame URL (see [1]). In the
client-side solution, a separate search request is made for each frame
and the returned annotations are then associated with that frame.

Some issues to resolve:

 - How can we test this locally with the dev server?
 - What if searches for different frames return the same annotation?
    - Allow annotations to be associated with multiple frames by making the `$frameId` property an array?
    - Preferentially associate the annotation with the main frame?
 - How should the WebSocket be adapted?
    - A separate WebSocket connection per frame?
 - What do about frame IDs for non-sidebar applications (notebook, stream, single annotation page) where annotations are not associated with frames?
    - Treat the annotation these views like the sidebar when there is only one main frame?

[1] https://github.com/hypothesis/client/issues/4191

### DIFF
--- a/src/sidebar/components/NotebookView.js
+++ b/src/sidebar/components/NotebookView.js
@@ -81,6 +81,7 @@ function NotebookView({ loadAnnotationsService, streamer }) {
     store.setSortKey('Newest');
     if (groupId) {
       loadAnnotationsService.load({
+        frameId: null,
         groupId,
         // Load annotations in reverse-chronological order because that is how
         // threads are sorted in the notebook view. By aligning the fetch
--- a/src/sidebar/components/SidebarView.js
+++ b/src/sidebar/components/SidebarView.js
@@ -51,7 +51,7 @@ function SidebarView({
     ? tabForAnnotation(linkedAnnotation)
     : 'annotation';
 
-  const searchUris = store.searchUris();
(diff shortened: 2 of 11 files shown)


robertknight:
> How should the WebSocket be adapted? A separate WebSocket instance per frame?

This is not ideal, for a few reasons:

- Some WebSocket messages pertain to annotations, but others pertain to objects which are not associated with frames, such as the user's groups
- WebSocket connections have a per-connection cost on the backend
- On the client side there will be added complexity for managing multiple WS connections

On the client side the main problem with WS notifications currently is that we don't know which frame(s) a new, but not updated or deleted, annotation should be assigned to. If the backend could provide information about which URIs in the WS configuration the annotation matches, that would solve the problem.

robertknight:
Related [Slack discussion](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1644478075890059).

robertknight:
> How should the WebSocket be adapted? A separate WebSocket instance per frame?

A possible incremental step is for the WebSocket to initially only receive updates for the main frame, or whichever one connected first and is still connected.

PR #4197, opened 2022-02-08
Move logic to show the sidebar UI to `_setupSidebarEvents` method

esanzgar:
[Based on this comment](https://github.com/hypothesis/client/pull/3903/files#r751261292) I have moved the logic to show the sidebar UI with other sidebar RPC event
listeners.

### DIFF
--- a/src/annotator/sidebar.js
+++ b/src/annotator/sidebar.js
@@ -93,7 +93,7 @@ export class Sidebar {
      */
     this.iframe = createSidebarIframe(config);
 
-    this.options = config;
+    this._config = config;
 
     /** @type {BucketBar|null} */
     this.bucketBar = null;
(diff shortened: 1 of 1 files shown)


src/annotator/sidebar.js
```diff
@@ -315,6 +293,27 @@ export class Sidebar {
...
 
+    this._sidebarRPC.on('connect', () => {
+      // Show the UI
+      if (this.iframeContainer) {
+        this.iframeContainer.style.display = '';
+      }
+
+      // Set initial highlight visibility. We do this only once the sidebar app
+      // is ready because `setHighlightsVisible` needs to reflect this state to
+      // the sidebar app.
```
robertknight, line 304: Given that we can now queue messages to be sent to the sidebar app before it has actually loaded, I wonder if this comment is still valid.

robertknight (approved):
The code change looks fine. Please give your branches specific-enough names that I can tell what they might relate to when looking at a list of branch names though.

src/annotator/sidebar.js
esanzgar: I agree it is not relevant anymore. I personally think that the state for toggling the highlights visibility should be the host frame. Currently, there is an indirection that I don't think it is necessary: host -> sidebar -> guests. It could be simplified as host -> guests.

PR #4200, opened 2022-02-09
Ignore meta tag if `content` attribute is empty

esanzgar:
If the meta tag contains a empty `content` attribute it is ignored.
Example: `<meta name="citation_doi" content>`

Closes https://github.com/hypothesis/product-backlog/issues/1290


robertknight (approved):
LGTM. Is there a similar issue with `<link>` tags?

esanzgar:
> Is there a similar issue with <link> tags?

I will review if `<link>` suffers the same issue and create a follow up PR if so.

PR #4202, opened 2022-02-10
Send annotation to first frame when there is no URI match or main frame

robertknight:
This is a refinement of https://github.com/hypothesis/client/pull/4131.

Handle the case when an annotation is returned from the backend whose URL does
not match the frame URL and where there is no "main" guest frame. This can
happen in VitalSource, where no guest is created in the host frame.

The general solution to this is to change the client and h so that it can always
determine which query URL an annotation from h matched. See [1]. This is an
interim step which will work in VitalSource and similar scenarios.

The second commit in this PR reworks how `store.frames()` is implemented in the `FrameSyncService` tests to make tests easier to write and encourage the results to be consistent with the real application.

You won't see any immediate effects in manual testing because we don't yet rely on document equivalence in the VS integration.

[1] https://github.com/hypothesis/client/issues/4191

### DIFF
--- a/src/sidebar/services/frame-sync.js
+++ b/src/sidebar/services/frame-sync.js
@@ -37,21 +37,26 @@ export function formatAnnot({ $tag, target, uri }) {
 /**
  * Return the frame which best matches an annotation.
  *
- * If there is a frame whose URL exactly matches the annotation, it will be
- * returned. Otherwise the main frame will be returned. Currently this may fail
- * to return the expected frame if an annotation was fetched from h that
- * does not match one of the frame URIs. This can happen due to URI expansion /
- * document equivalence in h.
- *
  * @param {Frame[]} frames
  * @param {Annotation} ann
  */
 function frameForAnnotation(frames, ann) {
-  const frame = frames.find(f => f.uri === ann.uri);
-  if (frame) {
-    return frame;
+  // Choose frame with an exact URL match if possible. In the unlikely situation
+  // where multiple frames have the same URL, we'll use whichever connected first.
+  const uriMatch = frames.find(f => f.uri === ann.uri);
+  if (uriMatch) {
+    return uriMatch;
   }
-  return frames.find(f => f.id === null);
+
+  // If there is no exact URL match, choose the main/host frame for consistent results.
+  const mainFrame = frames.find(f => f.id === null);
+  if (mainFrame) {
+    return mainFrame;
+  }
+
+  // If there is no main frame (eg. in VitalSource), fall back to whichever
+  // frame connected first.
+  return frames[0];
 }
 
 /**
--- a/src/sidebar/services/test/frame-sync-test.js
+++ b/src/sidebar/services/test/frame-sync-test.js
@@ -38,13 +38,6 @@ const fixtures = {
     // This should match the guest frame ID from `framesListEntry`.
     frameIdentifier: 'abc',
   },
-
-  // The entry in the list of frames currently connected
-  framesListEntry: {
-    id: 'abc',
-    uri: 'http://example.com',
-    isAnnotationFetchComplete: true,
-  },
 };
 
 describe('FrameSyncService', () => {
@@ -96,16 +89,34 @@ describe('FrameSyncService', () => {
     };
 
     fakeStore = fakeReduxStore(
-      { annotations: [] },
+      { annotations: [], frames: [] },
       {
         allAnnotations() {
           return this.getState().annotations;
         },
-        connectFrame: sinon.stub(),
-        destroyFrame: sinon.stub(),
+
+        connectFrame(frame) {
+          const frames = [
+            ...this.getState().frames,
+            {
+              ...frame,
+              isAnnotationFetchComplete: true,
+            },
+          ];
+          this.setState({ frames });
+        },
+
+        destroyFrame(frame) {
+          const frames = this.getState().frames;
+          this.setState({ frames: frames.filter(f => f.id !== frame.id) });
+        },
+
+        frames() {
+          return this.getState().frames;
+        },
+
         findIDsForTags: sinon.stub(),
         focusAnnotations: sinon.stub(),
-        frames: sinon.stub().returns([fixtures.framesListEntry]),
         isLoggedIn: sinon.stub().returns(false),
         openSidebarPanel: sinon.stub(),
         selectAnnotations: sinon.stub(),
@@ -244,12 +255,6 @@ describe('FrameSyncService', () => {
       const mainGuestRPC = fakePortRPCs[1];
       const iframeGuestRPC = fakePortRPCs[2];
 
-      const frames = [];
-      fakeStore.connectFrame.callsFake(frame => {
-        frames.push({ id: frame.id, uri: frame.uri });
-      });
-      fakeStore.frames.returns(frames);
-
       mainGuestRPC.emit('documentInfoChanged', {
         frameIdentifier: null,
         uri: mainFrameAnn.uri,
@@ -282,12 +287,6 @@ describe('FrameSyncService', () => {
         id: 'abc',
         uri: 'urn:book-id:1234',
       };
-
-      const frames = [];
-      fakeStore.connectFrame.callsFake(frame => {
-        frames.push({ id: frame.id, uri: frame.uri });
-      });
-      fakeStore.frames.returns(frames);
 
       // Connect a single guest which is not the main/host frame. This simulates
       // what happens in VitalSource for example.
@@ -349,10 +348,15 @@ describe('FrameSyncService', () => {
       frameSync.connect();
     });
 
-    it('sends a "publicAnnotationCountChanged" message to the frame when there are public annotations', () => {
+    it('sends a "publicAnnotationCountChanged" message to the frame when there are public annotations', async () => {
+      await connectGuest();
+      emitGuestEvent('documentInfoChanged', fixtures.htmlDocumentInfo);
+      fakeStore.frames()[0].isAnnotationFetchComplete = true;
+
       fakeStore.setState({
         annotations: [annotationFixtures.publicAnnotation()],
       });
+
       assert.calledWithMatch(
         hostRPC().call,
         'publicAnnotationCountChanged',
@@ -360,7 +364,11 @@ describe('FrameSyncService', () => {
       );
     });
 
-    it('sends a "publicAnnotationCountChanged" message to the frame when there are only private annotations', () => {
+    it('sends a "publicAnnotationCountChanged" message to the frame when there are only private annotations', async () => {
+      await connectGuest();
+      emitGuestEvent('documentInfoChanged', fixtures.htmlDocumentInfo);
+      fakeStore.frames()[0].isAnnotationFetchComplete = true;
+
       const annot = annotationFixtures.defaultAnnotation();
       delete annot.permissions;
 
@@ -375,8 +383,14 @@ describe('FrameSyncService', () => {
       );
     });
 
-    it('does not send a "publicAnnotationCountChanged" message to the frame if annotation fetch is not complete', () => {
-      fakeStore.frames.returns([{ uri: 'http://example.com' }]);
+    it('does not send a "publicAnnotationCountChanged" message to the frame if annotation fetch is not complete', async () => {
+      await connectGuest();
+      emitGuestEvent('documentInfoChanged', fixtures.htmlDocumentInfo);
+      fakeStore.frames()[0].isAnnotationFetchComplete = false;
+
+      // Discard 'publicAnnotationCountChanged' message sent after frame connected.
+      hostRPC().call.resetHistory();
+
       fakeStore.setState({
         annotations: [annotationFixtures.publicAnnotation()],
       });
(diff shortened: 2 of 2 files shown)


src/sidebar/services/frame-sync.js
```diff
@@ -37,21 +37,26 @@ export function formatAnnot({ $tag, target, uri }) {
 /**
  * Return the frame which best matches an annotation.
  *
- * If there is a frame whose URL exactly matches the annotation, it will be
- * returned. Otherwise the main frame will be returned. Currently this may fail
- * to return the expected frame if an annotation was fetched from h that
- * does not match one of the frame URIs. This can happen due to URI expansion /
- * document equivalence in h.
```
robertknight, line 44: I moved the comments to be above the relevant implementation lines, so they are less likely to get out of date. Also this is an internal function so it isn't really important for this to be "public" documentation.

src/sidebar/services/test/frame-sync-test.js
```diff
@@ -96,16 +89,34 @@ describe('FrameSyncService', () => {
...
 
     fakeStore = fakeReduxStore(
-      { annotations: [] },
+      { annotations: [], frames: [] },
       {
         allAnnotations() {
           return this.getState().annotations;
         },
-        connectFrame: sinon.stub(),
-        destroyFrame: sinon.stub(),
```
robertknight, line 105: Instead of `frames`, `connectFrame` and `destroyFrame` all being independent mocks, they now functions that operate on the same Redux state. This ensures that the result of `frames()` is consistent with any prior `connectFrame` and `destroyFrame` calls.

src/sidebar/services/test/frame-sync-test.js
```diff
@@ -315,18 +348,27 @@ describe('FrameSyncService', () => {
       frameSync.connect();
     });
 
-    it('sends a "publicAnnotationCountChanged" message to the frame when there are public annotations', () => {
+    it('sends a "publicAnnotationCountChanged" message to the frame when there are public annotations', async () => {
+      await connectGuest();
+      emitGuestEvent('documentInfoChanged', fixtures.htmlDocumentInfo);
+      fakeStore.frames()[0].isAnnotationFetchComplete = true;
```
robertknight, line 354: This step should have been present previously, but the test could get away without it due to the way that `store.frames()` was faked in the tests.

src/sidebar/services/test/frame-sync-test.js
robertknight: > A philosophical comment: I am a little bit divided about the mocking dependencies that are so intimately coupled, as for example the redux store and FrameSyncService.
> ...
> I know that integration tests, as we understand them in our group, are mean to test these interdependencies, but I sometimes wonder if we are mocking too much.

I think this is a very reasonable question to ask. I'm not a complete purist when it comes to unit vs integration testing. In this case I agree there are some risks to this more complex mocking. The downside of mocking complex dependencies are typically issues like:

- Tests running slowly
- Tests requiring a lot of effort to set up dependencies of dependencies
- Problems caused by side effects of dependencies
- Debugging test failures becoming more difficult because of the increased surface area they cover
- Negative impact on unit API design because there is less incentive to ensure that interfaces between modules are simple (and thus easy to mock)

An entirely pure Redux store (ours is "mostly pure") actually avoids many of these problems. If we removed the remaining side effects from the store it would be even easier.

Out of interest I looked into what it would be like if this test used the real store, but with various store methods stubbed. It turned out not to be too difficult.

```diff
diff --git a/src/sidebar/services/test/frame-sync-test.js b/src/sidebar/services/test/frame-sync-test.js
index 79882c75f..e7b53fcfd 100644
--- a/src/sidebar/services/test/frame-sync-test.js
+++ b/src/sidebar/services/test/frame-sync-test.js
@@ -6,6 +6,7 @@ import { fakeReduxStore } from '../../test/fake-redux-store';
 import { delay } from '../../../test-util/wait';
 
 import { FrameSyncService, $imports, formatAnnot } from '../frame-sync';
+import { createSidebarStore } from '../../store';
 
 class FakeWindow extends EventTarget {
   constructor() {
@@ -88,44 +89,16 @@ describe('FrameSyncService', () => {
       discover: sinon.stub().resolves(sidebarPort),
     };
 
-    fakeStore = fakeReduxStore(
-      { annotations: [], frames: [] },
-      {
-        allAnnotations() {
-          return this.getState().annotations;
-        },
-
-        connectFrame(frame) {
-          const frames = [
-            ...this.getState().frames,
-            {
-              ...frame,
-              isAnnotationFetchComplete: true,
-            },
-          ];
-          this.setState({ frames });
-        },
-
-        destroyFrame(frame) {
-          const frames = this.getState().frames;
-          this.setState({ frames: frames.filter(f => f.id !== frame.id) });
-        },
-
-        frames() {
-          return this.getState().frames;
-        },
-
-        findIDsForTags: sinon.stub(),
-        focusAnnotations: sinon.stub(),
-        isLoggedIn: sinon.stub().returns(false),
-        openSidebarPanel: sinon.stub(),
-        selectAnnotations: sinon.stub(),
-        selectTab: sinon.stub(),
-        setSidebarOpened: sinon.stub(),
-        toggleSelectedAnnotations: sinon.stub(),
-        updateAnchorStatus: sinon.stub(),
-      }
-    );
+    fakeStore = createSidebarStore({});
+    sinon.stub(fakeStore, 'findIDsForTags');
+    sinon.stub(fakeStore, 'focusAnnotations');
+    sinon.stub(fakeStore, 'isLoggedIn');
+    sinon.stub(fakeStore, 'openSidebarPanel');
+    sinon.stub(fakeStore, 'selectAnnotations');
+    sinon.stub(fakeStore, 'selectTab');
+    sinon.stub(fakeStore, 'setSidebarOpened');
+    sinon.stub(fakeStore, 'toggleSelectedAnnotations');
+    sinon.stub(fakeStore, 'updateAnchorStatus');
 
     fakeWindow = new FakeWindow();
     fakeWindow.parent = new FakeWindow();
@@ -214,9 +187,7 @@ describe('FrameSyncService', () => {
       await connectGuest();
       emitGuestEvent('documentInfoChanged', frameInfo);
 
-      fakeStore.setState({
-        annotations: [fixtures.ann],
-      });
+      fakeStore.addAnnotations([fixtures.ann]);
 
       assert.calledWithMatch(
         guestRPC().call,
@@ -228,12 +199,14 @@ describe('FrameSyncService', () => {
     it('sends annotations only to matching frame', async () => {
       // Annotation whose URI exactly matches the main frame.
       const mainFrameAnn = {
+        $tag: 't1',
         id: 'abc',
         uri: 'https://example.com',
       };
 
       // Annotation whose URI exactly matches the iframe.
       const iframeAnn = {
+        $tag: 't2',
         id: 'def',
         uri: 'https://example.com/iframe',
       };
@@ -243,6 +216,7 @@ describe('FrameSyncService', () => {
       // The search API can return such annotations if the annotation's URI
       // is deemed equivalent to one of the searched-for URIs.
       const unknownFrameAnn = {
+        $tag: 't3',
         id: 'ghi',
         uri: 'https://example.org',
       };
@@ -265,9 +239,7 @@ describe('FrameSyncService', () => {
         uri: iframeAnn.uri,
       });
 
-      fakeStore.setState({
-        annotations: [mainFrameAnn, iframeAnn, unknownFrameAnn],
-      });
+      fakeStore.addAnnotations([mainFrameAnn, iframeAnn, unknownFrameAnn]);
 
       // Check that expected annotations were sent to each frame.
       assert.calledWithMatch(
@@ -284,6 +256,7 @@ describe('FrameSyncService', () => {
 
     it('sends annotation to first frame if there is no frame with matching URL or main frame', async () => {
       const annotation = {
+        $tag: 't1',
         id: 'abc',
         uri: 'urn:book-id:1234',
       };
@@ -299,9 +272,7 @@ describe('FrameSyncService', () => {
         uri: 'https://publisher.com/books/1234/chapter1.html',
       });
 
-      fakeStore.setState({
-        annotations: [annotation],
-      });
+      fakeStore.addAnnotations([annotation]);
 
       assert.calledWithMatch(
         guestRPC().call,
@@ -316,14 +287,10 @@ describe('FrameSyncService', () => {
       emitGuestEvent('documentInfoChanged', frameInfo);
 
       const ann2 = Object.assign({}, fixtures.ann, { $tag: 't2', id: 'a2' });
-      fakeStore.setState({
-        annotations: [fixtures.ann],
-      });
+      fakeStore.addAnnotations([fixtures.ann]);
       guestRPC().call.reset();
 
-      fakeStore.setState({
-        annotations: [fixtures.ann, ann2],
-      });
+      fakeStore.addAnnotations([ann2]);
 
       assert.calledWithMatch(
         guestRPC().call,
@@ -335,9 +302,7 @@ describe('FrameSyncService', () => {
     it('does not send a "loadAnnotations" message for replies', async () => {
       await connectGuest();
 
-      fakeStore.setState({
-        annotations: [annotationFixtures.newReply()],
-      });
+      fakeStore.addAnnotations([annotationFixtures.newReply()]);
 
       assert.isFalse(guestRPC().call.calledWith('loadAnnotations'));
     });
@@ -351,11 +316,9 @@ describe('FrameSyncService', () => {
     it('sends a "publicAnnotationCountChanged" message to the frame when there are public annotations', async () => {
       await connectGuest();
       emitGuestEvent('documentInfoChanged', fixtures.htmlDocumentInfo);
-      fakeStore.frames()[0].isAnnotationFetchComplete = true;
+      fakeStore.updateFrameAnnotationFetchStatus(fixtures.htmlDocumentInfo.uri, true);
 
-      fakeStore.setState({
-        annotations: [annotationFixtures.publicAnnotation()],
-      });
+      fakeStore.addAnnotations([annotationFixtures.publicAnnotation()]);
 
       assert.calledWithMatch(
         hostRPC().call,
@@ -367,14 +330,12 @@ describe('FrameSyncService', () => {
     it('sends a "publicAnnotationCountChanged" message to the frame when there are only private annotations', async () => {
       await connectGuest();
       emitGuestEvent('documentInfoChanged', fixtures.htmlDocumentInfo);
-      fakeStore.frames()[0].isAnnotationFetchComplete = true;
+      fakeStore.updateFrameAnnotationFetchStatus(fixtures.htmlDocumentInfo.uri, true);
 
       const annot = annotationFixtures.defaultAnnotation();
       delete annot.permissions;
 
-      fakeStore.setState({
-        annotations: [annot],
-      });
+      fakeStore.addAnnotations([annot]);
 
       assert.calledWithMatch(
         hostRPC().call,
@@ -386,21 +347,17 @@ describe('FrameSyncService', () => {
     it('does not send a "publicAnnotationCountChanged" message to the frame if annotation fetch is not complete', async () => {
       await connectGuest();
       emitGuestEvent('documentInfoChanged', fixtures.htmlDocumentInfo);
-      fakeStore.frames()[0].isAnnotationFetchComplete = false;
 
       // Discard 'publicAnnotationCountChanged' message sent after frame connected.
       hostRPC().call.resetHistory();
 
-      fakeStore.setState({
-        annotations: [annotationFixtures.publicAnnotation()],
-      });
+      fakeStore.addAnnotations([annotationFixtures.publicAnnotation()]);
+
       assert.isFalse(hostRPC().call.calledWith('publicAnnotationCountChanged'));
     });
 
     it('does not send a "publicAnnotationCountChanged" message if there are no connected frames', () => {
-      fakeStore.setState({
-        annotations: [annotationFixtures.publicAnnotation()],
-      });
+      fakeStore.addAnnotations([annotationFixtures.publicAnnotation()]);
       assert.isFalse(hostRPC().call.calledWith('publicAnnotationCountChanged'));
     });
   });
@@ -412,13 +369,8 @@ describe('FrameSyncService', () => {
       frameSync.connect();
       await connectGuest();
 
-      fakeStore.setState({
-        annotations: [ann],
-      });
-
-      fakeStore.setState({
-        annotations: [],
-      });
+      fakeStore.addAnnotations([ann]);
+      fakeStore.removeAnnotations([ann]);
 
       assert.calledWithMatch(guestRPC().call, 'deleteAnnotation', ann.$tag);
     });
@@ -567,10 +519,6 @@ describe('FrameSyncService', () => {
           id: frameInfo.frameIdentifier,
           metadata: frameInfo.metadata,
           uri: frameInfo.uri,
-
-          // This would be false in the real application initially, but in these
-          // tests we pretend that the fetch completed immediately.
-          isAnnotationFetchComplete: true,
         },
       ]);
     });
```

src/sidebar/services/test/frame-sync-test.js
robertknight: > isAnnotationFetchComplete is already set it to true by the mocking logic. 

Indeed. This was left over from an earlier revision by mistake.

src/sidebar/services/frame-sync.js
```diff
@@ -37,21 +37,26 @@ export function formatAnnot({ $tag, target, uri }) {
...
+  // If there is no exact URL match, choose the main/host frame for consistent results.
+  const mainFrame = frames.find(f => f.id === null);
+  if (mainFrame) {
+    return mainFrame;
   }
-  return frames.find(f => f.id === null);
+
+  // If there is no main frame (eg. in VitalSource), fall back to whichever
+  // frame connected first.
+  return frames[0];
```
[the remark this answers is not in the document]
robertknight: > Could this ever return undefined? Maybe if the guest frame is destroyed while fetching annotations?

Yes. In this case the annotation will simply not be sent to any frames.

PR #4206, opened 2022-02-11
Create a helper function to add HTML element to document's head

esanzgar:
The addition of the helper function I believe could encourage the
creation of more atomic tests and improve the developer experience.

It also improves the formatting and re-formatting.


robertknight:
Having a helper to insert the `<meta>` and `<link>` elements might be useful, although the indirection makes it less obvious that no escaping or sanitization of the content happens. I'm going to close this issue for the moment since Eduardo is away this week and moving on from Hypothesis next week, and there are more important things that Lyza and myself need to work on at present.

PR #4221, opened 2022-02-14
Attempt to fix flakey PortRPC tests

robertknight:
Tests sometimes failed because a PortRPC method handler was not called
after the fixed message delivery delay that the tests allowed.

I have not been able to reproduce locally, but attempt to resolve the
issue by waiting for a specific message to be delivered to the
MessagePort instead of waiting a fixed amount of time.

For some test cases which assert that a message was *not* delivered, a
fixed delay is still used. This is acceptable because the test will
still pass in the rare cases where the fixed delay is too short.

### DIFF
--- a/src/shared/messaging/test/port-rpc-test.js
+++ b/src/shared/messaging/test/port-rpc-test.js
@@ -8,7 +8,36 @@ describe('PortRPC', () => {
   let plusOne;
 
   /**
-   * Wait for messages enqueued via `postMessage` to be delivered.
+   * Wait for a specific PortRPC method call to be delivered to a port.
+   *
+   * @param {MessagePort} port
+   * @param {string} method
+   * @param {number} [count]
+   */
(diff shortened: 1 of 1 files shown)


esanzgar (approved):
Super, thanks!

> This is acceptable because the test will
still pass in the rare cases where the fixed delay is too short.

That's true, but the test is not as stricter if the delay is too short. 

I wonder if having two pairs of PortRPCs: one pair that finish the task, while the other pair that doesn't (because they haven't been connected or have been disconnected).

robertknight:
> I wonder if having two pairs of PortRPCs: one pair that finish the task, while the other pair that doesn't (because they haven't been connected or have been disconnected).

Given a test which can be written as "wait for event B that will definitely occur and only after event A, which may occur" that seems preferable to "wait for event A to not happen". We'd have to be very certain about the ordering guarantees on events A and B though.

PR #4222, opened 2022-02-14
Remove adder by clicking once on a different frame

esanzgar:
Before this PR, when there was two frames, A and B, if the adder was in
frame A, it was necessary to __double click__ on frame B to remove the
adder and the selection from frame A.

After this PR, __a single click__  on frame B removes the adder and text
selection from frame A. This behaviour is similar to what happens on a
single frame scenario, where a single click removes the adder and the
selection.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -15,7 +15,7 @@ import {
 } from './highlighter';
 import { createIntegration } from './integrations';
 import * as rangeUtil from './range-util';
-import { SelectionObserver } from './selection-observer';
+import { SelectionObserver, selectedRange } from './selection-observer';
 import { findClosestOffscreenAnchor } from './util/buckets';
 import { normalizeURI } from './util/url';
 
@@ -303,8 +303,10 @@ export class Guest {
 
   async _connectHost() {
     this._hostRPC.on('clearSelection', () => {
-      this._informHostOnNextSelectionClear = false;
-      removeTextSelection();
+      if (selectedRange(document)) {
+        this._informHostOnNextSelectionClear = false;
+        removeTextSelection();
+      }
     });
 
     this._hostRPC.on('createAnnotation', () => this.createAnnotation());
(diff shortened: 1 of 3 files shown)


src/annotator/test/guest-test.js
```diff
@@ -672,26 +676,51 @@ describe('Guest', () => {
...
+
+        // Guest has no text selected
+        fakeSelectedRange.returns(null);
+
+        hostRPC().call.resetHistory();
+        emitHostEvent('clearSelection');
+
+        assert.notCalled(removeAllRanges);
+
+        // On next selection clear it should be inform the host.
```
robertknight, line 719:
```suggestion
        // On next selection clear it should inform the host.
```

src/annotator/test/guest-test.js
```diff
@@ -672,26 +676,51 @@ describe('Guest', () => {
...
+        sandbox.stub(document, 'getSelection').returns({ removeAllRanges });
+        const guest = createGuest();
+        guest.selectedRanges = [1];
 
-      simulateSelectionWithText();
-      hostRPC().call.resetHistory();
-      emitHostEvent('clearSelection');
+        // Guest has text selected
+        simulateSelectionWithText();
+        fakeSelectedRange.returns({});
```
robertknight, line 689: I noticed a typo in the existing "should be inform the host" comment below. Should say "should inform the host". GitHub only changed code or code near changed code to be commented on during reviews.

robertknight (approved):
Assuming my understanding of the change is correct (see my comment), this looks sensible to me.

src/annotator/guest.js
```diff
@@ -303,8 +303,10 @@ export class Guest {
 
   async _connectHost() {
     this._hostRPC.on('clearSelection', () => {
-      this._informHostOnNextSelectionClear = false;
-      removeTextSelection();
+      if (selectedRange(document)) {
+        this._informHostOnNextSelectionClear = false;
+        removeTextSelection();
```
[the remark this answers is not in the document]
esanzgar: Correct.

PR #4223, opened 2022-02-14
Remove outdated test for `clearSelectionExceptIn` RPC event.

esanzgar:
In #4195, we removed `clearSelectionExceptIn` RPC event. Instead, the
host frame sends `clearSelection` RPC events to specific guest frames.


robertknight (approved):
LGTM

PR #4224, opened 2022-02-15
Only load PDF.js style overrides in PDF.js

robertknight:
Only load the `pdfjs-overrides.css` bundle when the client is loaded in
PDF.js.

This fixes an issue where a global styling change that was recently
added specifically for PDF.js [1] ended up affecting other web pages
[2].

We will eventually need to find an alternative solution to the problem that prompted https://github.com/hypothesis/client/pull/4208, that works for all web pages. However it still makes sense not to load style overrides intended for PDF.js in other contexts.

[1] https://github.com/hypothesis/client/pull/4208
[2] https://github.com/hypothesis/product-backlog/issues/1270#issuecomment-1039647361

---

**Testing:**

1. Load the client in a normal web page, verify that the pdfjs-overrides.css stylesheet is not added to the `<head>` element
2. Load the client in PDF.js, verify that the pdfjs-overrides.css stylesheet is added to the `<head>` element

### DIFF
--- a/src/boot/boot.js
+++ b/src/boot/boot.js
@@ -29,6 +29,10 @@ const commonPolyfills = [
  */
 
 /**
+ * @typedef {Window & { PDFViewerApplication?: object }} MaybePDFWindow
+ */
+
+/**
  * Mark an element as having been added by the boot script.
  *
  * This marker is later used to know which elements to remove when unloading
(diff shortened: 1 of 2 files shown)


src/boot/boot.js
```diff
@@ -197,17 +201,23 @@ export function bootHypothesisClient(doc, config) {
...
     config,
     [
-      // Vendor code and polyfills
       ...polyfills,
+      ...viewerStyles,
 
       'scripts/annotator.bundle.js',
-
       'styles/highlights.css',
-      'styles/pdfjs-overrides.css',
```
robertknight, line 210: We are changing the order of the style bundles here. That shouldn't cause any problems, and I think the new order arguably makes more sense.

esanzgar (approved):
I have tested both in the local development server and in https://musicbrainz.org/ and the solution works well.

PR #4226, opened 2022-02-16
Invoke `gulp watch` indirectly through `yarn`

esanzgar:
In the Makefile, a number of JS command line executables (for example,
`prettier`, `eslint`, `tsc`) are currently invoked indirectly using
`yarn`, instead of using the whole path `node_modules/.bin/[...]`. I
think it makes a more cleaner and consistent option to invoke `gulp
watch` through `yarn`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ help:
 
 .PHONY: dev
 dev: node_modules/.uptodate
-	node_modules/.bin/gulp watch
+	yarn gulp watch
 
 .PHONY: test
 test: node_modules/.uptodate


robertknight:
I wonder how much overhead this indirection has when running commands.

robertknight:
To answer my own question, running `yarn gulp <task>` vs `gulp <task>` consistently adds 200-300ms of latency on my system. If anything we'd be better off running the various tools directly from the Makefile in all cases. The downside is that having the commands in the `Makefile` as opposed to `scripts` in `package.json` is less consistent with the general JS ecosystem.

This is not a hugely important issue to tackle right now so I'm going to close this and revisit another time.

PR #4227, opened 2022-02-16
Use same `Guest._frameIdentifier` as assigned by `PortProvider`

esanzgar:
`Guest` class relies on a frame identifier set by the injector to
uniquely identify itself when communicating with the sidebar frame. This
identifier can have two types: `null`, when the hypothesis client was
initially loaded, and `string`, when the client is injected.

In this PR, the `Guest` waits until `PortFinder` to return a `requestId`
set by `PortProvider`. This `requestId` can be used to uniquely identify
guest frame during its lifetime.

This change allows to remove a mechanism in `FrameSync` that replaces a
temporal ID by a permanent one that comes with the metadata information
in the `documentInfoChanged` RPC event.

### DIFF
--- a/src/annotator/guest.js
+++ b/src/annotator/guest.js
@@ -165,10 +165,18 @@ export class Guest {
      */
     this._annotations = /** @type {Set<string>} */ (new Set());
 
+    /** @type {(value: string|null) => void} */
+    this._resolveFrameIdentifier = () => {};
+
     // Set the frame identifier if it's available.
     // The "top" guest instance will have this as null since it's in a top frame not a sub frame
-    /** @type {string|null} */
-    this._frameIdentifier = config.subFrameIdentifier || null;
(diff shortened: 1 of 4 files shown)


robertknight (commented):
The changes in this PR lead to the frame identifier for the main frame in the store being different than the corresponding frame identifier in the `FrameSyncService._guestRPC` map. This is going to be a problem for future changes like https://github.com/hypothesis/client/pull/4193 which try to record information about the associated frames with annotations. I think it would make sense to consider eliminating the special `null` value for the main frame identifier before trying to make this change.

robertknight:
While I agree it would be nice to avoid the use of a temporary frame identifier in `FrameSyncService`, there is a bigger issue with the special `null` frame ID that we need to address first. I'm closing the issue as Eduardo is moving on from Hypothesis as of the end of this week and won't have time to address this larger issue before he leaves.

PR #4245, opened 2022-02-23
Implement side-by-side mode for VitalSource EPUB books

robertknight:
This PR is a subset of https://github.com/hypothesis/client/pull/4151, re-organized into more digestible commits. See that one for testing notes. The scroll anchoring logic has been removed and will be re-added in a follow-up PR.

### DIFF
--- a/src/annotator/integrations/html.js
+++ b/src/annotator/integrations/html.js
@@ -102,7 +102,9 @@ export class HTMLIntegration {
     // of content. If we can't find the "main content" then we just assume that
     // everything on the page is potentially content that the user might want
     // to annotate and so try to keep it all visible.
-    const padding = 10;
+
+    // nb. 12px padding is a multiple of the 4px grid unit in our design system.
+    const padding = 12;
     const rightMargin = sidebarWidth + padding;
 
     // nb. Adjusting the body size this way relies on the page not setting a
@@ -115,7 +117,7 @@ export class HTMLIntegration {
       // sidebar overlap stuff in the document to the right of the main content.
       const freeSpace = Math.max(
         0,
-        window.innerWidth - sidebarWidth - contentArea.right
+        window.innerWidth - rightMargin - contentArea.right
       );
       if (freeSpace > 0) {
         const adjustedMargin = Math.max(0, rightMargin - freeSpace);
@@ -124,7 +126,7 @@ export class HTMLIntegration {
 
       // If the main content appears to be right up against the edge of the
       // window, add padding for readability.
-      if (contentArea.left < 10) {
+      if (contentArea.left < padding) {
         document.body.style.marginLeft = `${padding}px`;
       }
     } else {
--- a/src/annotator/integrations/html.js
+++ b/src/annotator/integrations/html.js
@@ -3,10 +3,12 @@ import scrollIntoView from 'scroll-into-view';
 import { anchor, describe } from '../anchoring/html';
 
 import { HTMLMetadata } from './html-metadata';
+import { guessMainContentArea } from './html-side-by-side';
 
 /**
  * @typedef {import('../../types/annotator').Anchor} Anchor
  * @typedef {import('../../types/annotator').Integration} Integration
+ * @typedef {import('../../types/annotator').SidebarLayout} SidebarLayout
  */
 
 /**
@@ -24,6 +26,14 @@ export class HTMLIntegration {
     this.describe = describe;
 
     this._htmlMeta = new HTMLMetadata();
+
+    /**
+     * Whether to attempt to resize the document contents when {@link fitSideBySide}
+     * is called.
+     *
+     * Currently disabled by default.
+     */
+    this.sideBySideEnabled = false;
   }
 
   canAnnotate() {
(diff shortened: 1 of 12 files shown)


src/sidebar/store/test/use-store-test.js
```diff
@@ -82,7 +82,10 @@ describe('sidebar/store/use-store', () => {
       assert.deepEqual(proxy.getThing('bar'), { id: 'bar' });
 
       // Test proxied action dispatch.
-      proxy.addThing('baz');
+      act(() => {
+        proxy.addThing('baz');
+      });
```
robertknight, line 87: This change is unrelated to the rest of the PR, it aims to fix a flakey test I encountered a couple of times while working on this branch. See https://github.com/hypothesis/client/pull/4245/commits/cba6232848412384525c5506f6265ead72bd8772.

src/annotator/integrations/test/html-test.js
```diff
@@ -58,9 +64,94 @@ describe('HTMLIntegration', () => {
...
+      it('does nothing if the content area cannot be determined', () => {
+        const integration = createIntegration();
+        fakeGuessMainContentArea.returns(null);
+
+        integration.fitSideBySide({ expanded: true, width: sidebarWidth });
+
+        assert.deepEqual(getMargins(), [null, null]);
+      });
+
+      it('resets margins on body element when side-by-side mode is deactivated', () => {
```
lyzadanger, line 145: This might be more accurately described as "removes" than "resets".

src/annotator/integrations/test/html-side-by-side-test.js
```diff
@@ -0,0 +1,92 @@
+import { guessMainContentArea } from '../html-side-by-side';
+
+describe('annotator/integrations/html-side-by-side', () => {
+  let contentElements;
+
+  function createContent(paragraphs) {
+    const paraElements = paragraphs.map(({ content, left, width }) => {
+      const el = document.createElement('p');
+      el.textContent = content;
+      el.style.position = 'absolute';
```
lyzadanger, line 10: Presumably `absolute` per the initial containing block of the `html` element, as no ancestors have an explicit `position` property? Would the behavior of tests change if the containing `div` had a positioning context?

src/annotator/util/frame.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  if (!frame.frameElement) {
+    // This is a cross-origin iframe. In this case we can't tell if it fills
+    // the parent frame or not.
+    return false;
+  }
+
+  const frameBox = frame.frameElement.getBoundingClientRect();
+  return frameBox.width / frame.parent.innerWidth >= 0.8;
```
lyzadanger, line 24: I had a query about the origin of the value `0.8`. Does it have meaning? Do we need a test for the boundary case?

src/annotator/test/guest-test.js
```diff
@@ -189,18 +195,21 @@ describe('Guest', () => {
 
   describe('events from host frame', () => {
     describe('on "sidebarLayoutChanged" event', () => {
-      it('calls fitSideBySide if `Guest` is the main annotatable frame', () => {
```
lyzadanger, line 192: Are these two removed tests no longer meaningful? Is "main annotatable frame" implicitly the case if the guest frame is filling the host frame?

src/annotator/integrations/html.js
```diff
@@ -38,9 +48,97 @@ export class HTMLIntegration {
...
+    document.body.style.marginRight = `${rightMargin}px`;
+
+    const contentArea = guessMainContentArea(document.body);
+    if (contentArea) {
+      // Check if we can give the main content more space by letting the
+      // sidebar overlap stuff in the document to the right of the main content.
+      const freeSpace = Math.max(
+        0,
+        window.innerWidth - sidebarWidth - contentArea.right
+      );
```
lyzadanger, line 119: I believe there may be a small computational bug here in that this is not taking the added `padding` into account. `freeSpace` ends up being `10` (i.e. `padding` value) when I test it on local HTML documents (e.g. enable side-by-side for HTML and then load up _Anna Karenina_: http://localhost:3000/document/tolstoy). Thus the margin gets re-computed and we lose the right-padding and the text runs up against the toolbar:

Screen Shot 2022-02-24 at 8 30 49 AM

If I update this computation to take `padding` into account

```suggestion
      const freeSpace = Math.max(
        0,
        window.innerWidth - sidebarWidth - padding - contentArea.right
      );
```
The result seems to be what I believe is intended?

Screen Shot 2022-02-24 at 8 37 19 AM

lyzadanger (changes requested):
Hi! I've completed a code review here and done just a bit of initial testing (will do more, but want to get this code review submitted). Looking mostly good here, but I think there may be a small right-margin computation bug (maybe?) and I had a few other comments on the code itself.

src/annotator/integrations/html.js
robertknight: You are correct. The `sidebarWidth` value here should be replaced by `rightMargin`, which is the sidebar width plus the padding.

src/annotator/test/guest-test.js
robertknight: > Is "main annotatable frame" implicitly the case if the guest frame is filling the host frame?

In all current situations, yes. It is possible in future there may be a situation where the main content is in an iframe that doesn't fill the host frame. We'll probably have to change how side-by-side mode works in that scenario.

src/annotator/integrations/html.js
```diff
@@ -38,9 +48,97 @@ export class HTMLIntegration {
...
+    // resulting page layout and determine whether there is significant "free space"
+    // (ie. anything that is not the main content of the document, such as ads or
+    // links to related stories) to the right of the main content. If there is,
+    // we make the document wider again to allow more space for the main content.
+    //
+    // These heuristics assume a typical "article" page with one central block
+    // of content. If we can't find the "main content" then we just assume that
+    // everything on the page is potentially content that the user might want
+    // to annotate and so try to keep it all visible.
+    const padding = 10;
```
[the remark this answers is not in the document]
robertknight: Done

src/annotator/integrations/html.js
```diff
@@ -38,9 +48,97 @@ export class HTMLIntegration {
...
+        window.innerWidth - sidebarWidth - contentArea.right
+      );
+      if (freeSpace > 0) {
+        const adjustedMargin = Math.max(0, rightMargin - freeSpace);
+        document.body.style.marginRight = `${adjustedMargin}px`;
+      }
+
+      // If the main content appears to be right up against the edge of the
+      // window, add padding for readability.
+      if (contentArea.left < 10) {
```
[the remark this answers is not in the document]
robertknight: Done

src/annotator/integrations/test/html-side-by-side-test.js
robertknight: Positioning is used as a convenient way to set the X coordinate of the paragraphs. If the containing div was positioned, then the final coordinate would be the left value here plus the X coordinate of the containing `div`.

src/annotator/util/frame.js
robertknight: This is an arbitrary threshold. I'll give it a descriptive name. The only boundary case I can immediately see is if `frame.parent.innerWidth` was 0. I don't think that is possible. If it does somehow happen, the function will return `true`. Nothing will crash though.

robertknight:
I have corrected the calculation to take account of padding and adjusted it to 12px per suggestion. I've also tweaked a few other comments based on the above feedback. PTAL.

src/annotator/integrations/test/html-test.js
robertknight: True. I've updated the description.

src/annotator/integrations/test/html-test.js
```diff
@@ -58,9 +64,94 @@ describe('HTMLIntegration', () => {
...
+      const rightMargin = bodyStyle.marginRight
+        ? parseInt(bodyStyle.marginRight)
+        : null;
+      return [leftMargin, rightMargin];
+    }
+
+    const sidebarWidth = 200;
+
+    // Return a rect for content that occupies the full width of the viewport,
+    // minus space for the opened sidebar, as `fitSideBySide` only calls this
```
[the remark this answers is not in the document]
robertknight: I have revised this comment to explain what the function's result is used for.

src/annotator/integrations/test/html-test.js
```diff
@@ -58,9 +64,94 @@ describe('HTMLIntegration', () => {
...
+      const leftMargin = bodyStyle.marginLeft
+        ? parseInt(bodyStyle.marginLeft)
+        : null;
+      const rightMargin = bodyStyle.marginRight
+        ? parseInt(bodyStyle.marginRight)
+        : null;
+      return [leftMargin, rightMargin];
+    }
+
+    const sidebarWidth = 200;
```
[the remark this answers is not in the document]
robertknight: It is an arbitrary value of the same order of magnitude as the real sidebar width (which is actually ~460px on my system).

lyzadanger (approved):
Based on [discussion in Slack](https://hypothes-is.slack.com/archives/C1M8NH76X/p1645778349057179) (private to Hypothesis team members), we feel that this is at a point that it's good to release, with some identified follow-up tasks.

PR #4272, opened 2022-03-01
Add missing parameter and variable types in src/annotator/

robertknight:
Add missing types found while trying to typecheck src/annotator with `noImplicitAny` enabled.

There are more to come. This PR contains simple cases that required no or only very minimal changes to surrounding code.

### DIFF
--- a/src/annotator/components/Buckets.js
+++ b/src/annotator/components/Buckets.js
@@ -11,7 +11,7 @@ import classnames from 'classnames';
  * the sidebar.
  *
  * @param {object} props
- *   @prop {Children} props.children
+ *   @param {Children} props.children
  */
 function BucketList({ children }) {
   return (
@@ -38,8 +38,8 @@ function BucketList({ children }) {
  * Render a vertically-positioned bucket-list item.
  *
  * @param {object} props
- *  @prop {Children} props.children
- *  @prop {number} props.topPosition - The vertical top position, in pixels,
+ *  @param {Children} props.children
+ *  @param {number} props.topPosition - The vertical top position, in pixels,
  *   for this bucket item relative to the top of the containing BucketList
  */
 function BucketItem({ children, topPosition }) {
--- a/src/annotator/features.js
+++ b/src/annotator/features.js
@@ -1,7 +1,9 @@
 import { warnOnce } from '../shared/warn-once';
 
+/** @type {Record<string, boolean>} */
 let _features = {};
 
+/** @param {Record<string, boolean>} features */
 const _set = features => {
   _features = features || {};
 };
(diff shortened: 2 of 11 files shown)


src/annotator/components/Buckets.js
```diff
@@ -11,7 +11,7 @@ import classnames from 'classnames';
  * the sidebar.
  *
  * @param {object} props
- *   @prop {Children} props.children
+ *   @param {Children} props.children
```
robertknight, line 14: `@prop` is only for use with `@typedef`.

src/annotator/features.js
```diff
@@ -1,7 +1,9 @@
 import { warnOnce } from '../shared/warn-once';
 
+/** @type {Record<string, boolean>} */
```
robertknight, line 3: This whole file needs a larger refactor, but that's out of scope here.

src/annotator/highlighter.js
```diff
@@ -246,11 +246,11 @@ export function highlightRange(range, cssClass = 'hypothesis-highlight') {
   const whitespace = /^\s*$/;
   textNodeSpans = textNodeSpans.filter(span =>
     // Check for at least one text node with non-space content.
-    span.some(node => !whitespace.test(node.nodeValue))
+    span.some(node => !whitespace.test(node.data))
```
robertknight, line 249: For `Text` nodes, `nodeValue` and `data` are the same. The difference is that `nodeValue` is nullable because it is part of the `Node` interface, whereas `data` is specific to `CharacterData` nodes (text, comments, cdata) that definitely have text.

src/annotator/highlighter.js
```diff
@@ -259,7 +259,8 @@ export function highlightRange(range, cssClass = 'hypothesis-highlight') {
     const highlightEl = document.createElement('hypothesis-highlight');
     highlightEl.className = cssClass;
 
-    nodes[0].parentNode.replaceChild(highlightEl, nodes[0]);
+    const parent = /** @type {Node} */ (nodes[0].parentNode);
+    parent.replaceChild(highlightEl, nodes[0]);
```
robertknight, line 263: `nodes` is now typed, so TS complains that `parentNode` may be null. We know that in this context it isn't.

src/annotator/range-util.js
```diff
@@ -66,16 +66,17 @@ export function forEachNodeInRange(range, callback) {
...
   const whitespaceOnly = /^\s*$/;
-  const textNodes = [];
+  const textNodes = /** @type {Text[]} */ ([]);
   forEachNodeInRange(range, node => {
     if (
       node.nodeType === Node.TEXT_NODE &&
       !(/** @type {string} */ (node.textContent).match(whitespaceOnly))
     ) {
-      textNodes.push(node);
+      textNodes.push(/** @type {Text} */ (node));
```
robertknight, line 75: TS doesn't understand that the `node.nodeType...` check above implies that `Node` is a `Text`.

lyzadanger (approved):
LGTM.

PR #4281, opened 2022-03-04
Make src/annotator/ and src/boot/ typecheck with `noImplicitAny`

robertknight:
Continue the progression towards making the client codebase typecheck with `noImplicitAny` by adding the types needed to make src/annotator and src/boot typecheck with these options.

In the process I found an issue in the recently added code in `frame-observer.js` where some variables were unintentionally untyped, as the type was not automatically inferred as I had expected.

 - Add missing types in src/annotator and src/boot.
 - Simplify the return types of some config parsing functions by
   coercing invalid values to `null` or some other default. This reduces
   complexity with the types elsewhere and also potentially avoids some
   confusing errors if invalid values are specified here.
 - Add type definitions for hammerjs, scroll-into-view and
   lodash.debounce. The definitions for hammerjs and scroll-into-view
   come from DefinitelyTyped. For lodash a minimal custom definition has
   been written because the @types/lodash.debounce package includes
   types for the whole of lodash.

   In order to use these types the `allowSyntheticDefaultImports` flag
   had to be enabled in the TS configs.

   A consequence of adding types for scroll-into-view is that some
   missing null checks in sidebar components were found.

### DIFF
--- a/src/annotator/anchoring/html.js
+++ b/src/annotator/anchoring/html.js
@@ -1,7 +1,10 @@
 import { RangeAnchor, TextPositionAnchor, TextQuoteAnchor } from './types';
 
 /**
+ * @typedef {import('../../types/api').RangeSelector} RangeSelector
  * @typedef {import('../../types/api').Selector} Selector
+ * @typedef {import('../../types/api').TextPositionSelector} TextPositionSelector
+ * @typedef {import('../../types/api').TextQuoteSelector} TextQuoteSelector
  */
 
 /**
@@ -26,9 +29,9 @@ async function querySelector(anchor, options = {}) {
  *   @param {number} [options.hint]
  */
 export function anchor(root, selectors, options = {}) {
-  let position = null;
-  let quote = null;
-  let range = null;
+  let position = /** @type {TextPositionSelector|null} */ (null);
+  let quote = /** @type {TextQuoteSelector|null} */ (null);
+  let range = /** @type {RangeSelector|null} */ (null);
 
   // Collect all the selectors
   for (let selector of selectors) {
@@ -64,22 +67,25 @@ export function anchor(root, selectors, options = {}) {
   let promise = Promise.reject('unable to anchor');
 
   if (range) {
+    const range_ = range;
     promise = promise.catch(() => {
-      let anchor = RangeAnchor.fromSelector(root, range);
+      let anchor = RangeAnchor.fromSelector(root, range_);
       return querySelector(anchor, options).then(maybeAssertQuote);
     });
   }
 
   if (position) {
+    const position_ = position;
     promise = promise.catch(() => {
-      let anchor = TextPositionAnchor.fromSelector(root, position);
+      let anchor = TextPositionAnchor.fromSelector(root, position_);
       return querySelector(anchor, options).then(maybeAssertQuote);
     });
   }
 
   if (quote) {
+    const quote_ = quote;
     promise = promise.catch(() => {
-      let anchor = TextQuoteAnchor.fromSelector(root, quote);
+      let anchor = TextQuoteAnchor.fromSelector(root, quote_);
       return querySelector(anchor, options);
     });
   }
--- a/src/annotator/anchoring/html.js
+++ b/src/annotator/anchoring/html.js
@@ -67,6 +67,7 @@ export function anchor(root, selectors, options = {}) {
   let promise = Promise.reject('unable to anchor');
 
   if (range) {
+    // Const binding assures TS that it won't be re-assigned when callback runs.
     const range_ = range;
     promise = promise.catch(() => {
       let anchor = RangeAnchor.fromSelector(root, range_);
--- a/src/annotator/annotation-counts.js
+++ b/src/annotator/annotation-counts.js
@@ -7,15 +7,16 @@ const ANNOTATION_COUNT_ATTR = 'data-hypothesis-annotation-count';
  *
  * @param {Element} rootEl - The DOM element which contains the elements that
  *   display annotation count.
- * @param {import('../shared/messaging').PortRPC} rpc - Channel for host-sidebar communication
+ * @param {import('../shared/messaging').PortRPC<'publicAnnotationCountChanged', string>} rpc - Channel for host-sidebar communication
  */
 export function annotationCounts(rootEl, rpc) {
   rpc.on('publicAnnotationCountChanged', updateAnnotationCountElems);
 
+  /** @param {number} newCount */
   function updateAnnotationCountElems(newCount) {
     const elems = rootEl.querySelectorAll('[' + ANNOTATION_COUNT_ATTR + ']');
     Array.from(elems).forEach(elem => {
-      elem.textContent = newCount;
+      elem.textContent = newCount.toString();
     });
   }
 }
--- a/src/annotator/config/settings.js
+++ b/src/annotator/config/settings.js
@@ -13,9 +13,13 @@ import { urlFromLinkTag } from './url-from-link-tag';
  * @prop {string} clientUrl
  * @prop {string} sidebarAppUrl
  * @prop {string} notebookAppUrl
- 
- * @prop {(name: string, options?: object) => (string|null)} hostPageSetting
+ * @prop {(name: string) => unknown} hostPageSetting
  */
+
+/** @param {unknown} value */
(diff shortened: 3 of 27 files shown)


src/annotator/anchoring/html.js
```diff
@@ -64,22 +67,25 @@ export function anchor(root, selectors, options = {}) {
   let promise = Promise.reject('unable to anchor');
 
   if (range) {
+    const range_ = range;
```
robertknight, line 70: This `const` alias is needed as otherwise TS complains the variable might have changed inside the `catch` callback, since it is declared with `let` and could potentially be re-assigned before the callback runs.

This whole function could be reworked to avoid the unusual control flow here, but that's out of scope for this PR.

src/annotator/config/settings.js
```diff
@@ -56,7 +61,7 @@ export function settingsFrom(window_) {
       return null;
     }
 
-    return jsonConfigs.annotations || annotationsFromURL();
+    return checkIfString(jsonConfigs.annotations) || annotationsFromURL();
```
robertknight, line 64: Adding this check avoids downstream code from needing to deal with the case where `annotations` is not a string or `null`. Same for other functions in this file.

src/annotator/config/settings.js
```diff
@@ -78,23 +83,24 @@ export function settingsFrom(window_) {
       return null;
     }
 
-    return jsonConfigs.group || groupFromURL();
+    return checkIfString(jsonConfigs.group) || groupFromURL();
   }
 
-  // TODO: Move this to a coerce method
   function showHighlights() {
```
robertknight, line 89: Previously this function could have returned arbitrary values, if the value in the configuration was not one of the allowed cases. Now it always returns either `always`, `never` or `whenSidebarOpen`.

src/annotator/frame-observer.js
```diff
@@ -162,7 +162,9 @@ export function onNextDocumentReady(frame) {
  * @return {() => void} Callback that unsubscribes from future changes
  */
 export function onDocumentReady(frame, callback, { pollInterval = 10 } = {}) {
+  /** @type {number|undefined} */
   let pollTimer;
+  /** @type {() => void} */
   let pollForDocumentChange;
```
robertknight, line 168: I thought TS would infer these, but it didn't.

src/annotator/frame-observer.js
```diff
@@ -171,7 +173,7 @@ export function onDocumentReady(frame, callback, { pollInterval = 10 } = {}) {
 
   const cancelPoll = () => {
     clearTimeout(pollTimer);
-    pollTimer = null;
+    pollTimer = undefined;
```
robertknight, line 176: `clearTimeout` expects `number|undefined`, so it was more convenient to use `undefined` as the null value for `pollTimer`.

src/annotator/index.js
```diff
@@ -61,7 +61,9 @@ function init() {
 
     const sidebarConfig = getConfig('sidebar');
 
-    const hypothesisAppsOrigin = new URL(sidebarConfig.sidebarAppUrl).origin;
+    const hypothesisAppsOrigin = new URL(
+      /** @type {string} */ (sidebarConfig.sidebarAppUrl)
+    ).origin;
```
robertknight, line 66: `sidebarConfig` is now a `Record<string, unknown>` so I'm adding a cast here to say "trust me, this is a string".

src/annotator/sidebar.js
```diff
@@ -365,7 +367,8 @@ export class Sidebar {
   _setupGestures() {
     const toggleButton = this.toolbar.sidebarToggleButton;
     if (toggleButton) {
-      this._hammerManager = new Hammer.Manager(toggleButton).on(
+      this._hammerManager = new Hammer.Manager(toggleButton);
```
robertknight, line 370: The previous code worked because `on` returns `this` (ie. it chains). However that information is missing from the types in the `@types/hammerjs` package.

src/annotator/sidebar.js
```diff
@@ -482,6 +485,7 @@ export class Sidebar {
     }
   }
 
+  /** @param {HammerInput} event */
```
robertknight, line 488: This type comes from `@types/hammerjs`.

src/sidebar/helpers/groups.js
```diff
@@ -4,7 +4,6 @@
  * @typedef {import('../../types/api').GroupIdentifier} GroupIdentifier
  */
 
-// @ts-expect-error - Ignore error about default-importing a CommonJS module.
 import escapeStringRegexp from 'escape-string-regexp';
```
robertknight, line 7: This error no longer occurs due to the addition of `allowSyntheticDefaultImports` to the TS config.

robertknight:
Marked as a draft because I noticed an issue on this branch where the sidebar doesn't automatically open on http://localhost:3000.

src/annotator/config/settings.js
```diff
@@ -13,10 +13,14 @@ import { urlFromLinkTag } from './url-from-link-tag';
  * @prop {string} clientUrl
  * @prop {string} sidebarAppUrl
  * @prop {string} notebookAppUrl
- 
- * @prop {(name: string, options?: object) => (string|null)} hostPageSetting
+ * @prop {(name: string) => unknown} hostPageSetting
```
robertknight, line 16: The declared return type was wrong. The function returns whatever value the client configuration has for a particular key, and we expect other value types (eg. `openSidebar` may be a boolean).

lyzadanger (approved):
whoof! That was _dense_, man :). Took a while to get my head around the whole set. Some questions and comments sprinkled.

src/annotator/anchoring/html.js
```diff
@@ -1,7 +1,10 @@
 import { RangeAnchor, TextPositionAnchor, TextQuoteAnchor } from './types';
 
 /**
+ * @typedef {import('../../types/api').RangeSelector} RangeSelector
```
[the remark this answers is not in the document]
robertknight: Indeed. I will add some comments.

src/annotator/annotation-counts.js
```diff
@@ -7,15 +7,16 @@ const ANNOTATION_COUNT_ATTR = 'data-hypothesis-annotation-count';
  *
  * @param {Element} rootEl - The DOM element which contains the elements that
  *   display annotation count.
- * @param {import('../shared/messaging').PortRPC} rpc - Channel for host-sidebar communication
+ * @param {import('../shared/messaging').PortRPC<'publicAnnotationCountChanged', string>} rpc - Channel for host-sidebar communication
```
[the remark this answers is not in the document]
robertknight: The `PortRPC` class accepts two type parameters. The first is the names of methods that be responded to via its `on` method. The second is the names of methods that can be called via the `call` method.

src/annotator/config/settings.js
```diff
@@ -78,23 +83,24 @@ export function settingsFrom(window_) {
...
+      case 'always':
+      case 'never':
+      case 'whenSidebarOpen':
+        return value;
+      case true:
+        return 'always';
+      case false:
+        return 'never';
+      default:
+        return 'always';
```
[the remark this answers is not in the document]
robertknight: I also look forward to a more capable successor to `switch` hopefully coming to JS - https://github.com/tc39/proposal-pattern-matching

src/boot/parse-json-config.js
```diff
@@ -1,5 +1,10 @@
-// `Object.assign()`-like helper. Used because this script needs to work
-// in IE 10/11 without polyfills.
+/**
+ * `Object.assign()`-like helper. Used because this script needs to work
+ * in IE 10/11 without polyfills.
```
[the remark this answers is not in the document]
robertknight: At the time we dropped IE support, the boot script was designed to still run in IE but log a console error instead of crashing. We've dropped IE support for a couple of years now, so we could probably look at dropping support for it in the boot script.

src/types/lodash.debounce.d.ts
```diff
@@ -0,0 +1,23 @@
+// Self-contained types for lodash.debounce that include only the functionality
+// we use and avoids adding types for the whole of lodash to our dependencies.
```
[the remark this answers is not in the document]
robertknight: I wrote this myself based on what we actually use.

PR #4303, opened 2022-03-10
Increase timeout in test

robertknight:
`waitFor` has a short default timeout of 10ms and this test sometimes
failed in CI. Try increasing it to reduce flakiness. We might want to
consider changing the `waitFor` default in future, although we also want
to encourage test authors to avoid writing tests that frequently wait
for long (> 10ms) periods of time.


lyzadanger (approved):
LGTM.

I agree with exercising caution before bumping up the default timeout across all tests.

PR #4308, opened 2022-03-11
Add `contentPartner` configuration option to show content provider banner

robertknight:
Add support for a new `contentPartner` configuration option that configures the client to show a banner with more information about where the content is from. The initial use case is for JSTOR. Showing a banner is in part a contractual obligation, but may also provide useful associated links (eg. to the article on a publisher's website).

The configuration option currently accepts the string value `'jstor'`. In future it could potentially accept an object with additional data used by the banner. This option is not expected to be used by anyone other than us, so we don't plan to maintain backwards compatibility if we make changes.

A summary of the changes:

- Add `contentPartner` option in `src/annotator/config/index.js`
- Plumb the `contentPartner` option from the `Guest` through to the `PDFIntegration`. We might want to make this configuration plumbing more general in future so that we don't need to handle each option in `Guest` and `createIntegration` specifically, but I avoided trying to generalize yet.
- Make `PDFIntegration`'s banner rendering more general so it can display content above the PDF viewer beyond the existing "no selectable text" warning
- Add support for customizing the client configuration for specific PDF test cases by adding `{basename}.config.json` file alongside the PDF. This is JSON file should contain any custom client configuration under a `clientConfig` key.
- Add a test case for the JSTOR banner at http://localhost:3000/pdf/jstor

There are some details I'm not sure about yet:

- Whether we'll want to pass additional data for use in the content partner banner. Currently the config just indicates which banner to show
- Whether we'll need to use the banner with other integrations in future
- Whether we'll have other configuration besides `contentPartner` that needs to be passed to the integration

For the moment this PR implements code that is fairly specific to the current use case. We can generalize it as it evolves.


src/annotator/integrations/pdf.js
```diff
@@ -100,12 +104,24 @@ export class PDFIntegration {
     this._reanchoringMaxWait = options.reanchoringMaxWait ?? 3000;
 
     /**
-     * A banner shown at the top of the PDF viewer warning the user if the PDF
-     * is not suitable for use with Hypothesis.
+     * Banners shown at the top of the PDF viewer.
```
lyzadanger, line 107: What's the intended behavior if both banners are applicable?

src/annotator/components/ContentPartnerBanner.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param {object} props
+ *   @param {ContentPartner} props.provider
+ *   @param {() => void} props.onClose
+ */
+export default function ContentPartnerBanner({ provider, onClose }) {
+  return (
+    <div className="bg-white p-2">
+      {provider === 'jstor' && 'Content provided by JSTOR'}
+      <button onClick={onClose}>Close</button>
```
lyzadanger, line 16: The no-selectable-text warning banner is not currently close-able. Should it be? Should this one be?

src/annotator/integrations/pdf.js
robertknight: You should see both banners. I don't know if JSTOR will serve us any PDFs that have no text at all, not even a front matter page, but there is content on JSTOR that is non-textual (eg. art, old hand-written letters that can't/haven't been OCR-ed).

src/annotator/integrations/pdf.js
```diff
@@ -241,13 +261,26 @@ export class PDFIntegration {
...
+        {this._bannerState.contentPartner && (
+          <ContentPartnerBanner
+            provider={this._bannerState.contentPartner}
+            onClose={() => this._updateBannerState({ contentPartner: null })}
+          />
+        )}
+        {this._bannerState.noTextWarning && <WarningBanner />}
+      </>,
+      /** @type {ShadowRoot} */ (this._banner.shadowRoot)
+    );
```
lyzadanger, line 281: If we're potentially rendering two banners, we may need to render an outer container for them. I'll think on that.

src/annotator/config/index.js
```diff
@@ -106,6 +106,11 @@ const configDefinitions = {
     defaultValue: null,
     getValue: settings => settings.clientUrl,
   },
+  contentPartner: {
+    allowInBrowserExt: false,
+    defaultValue: null,
+    getValue: getHostPageSetting,
+  },
```
lyzadanger, line 113: Nothing new from these changes, but typing for config across the annotator and sidebar is...challenging :). Is typing in `/types/config` _only_ for sidebar-related configuration? (Yes, I do see this typed in `types/annotator`).

lyzadanger:
@robertknight  I pushed a commit with some styling and the JSTOR logo as an SVG.

Looks like this for now:

image

It may catch your eye that the vertical alignment of the text and the JSTOR logo _seems_ slightly off. That's an optical trick of the JSTOR SVG including the JSTOR text and even some white space underneath it. That is, it is vertically centered, but the composition of the logo makes it seem like it is not. We can do some manual tweaking of margins to make it look more aligned, but I want to wait until we're sure we're moving forward with this general look.

robertknight:
There is a change to Via that can be used to test this:

1. Check out https://github.com/hypothesis/via/pull/707 and this client branch
2. Open http://localhost:9083/pdf?url=jstor://resrep06908. The PDF won't load due to a missing `jstor.ip` query param in this URL (ping me on Slack for that) but it will show the banner.

robertknight:
The current iteration of the banner looks like this:

JSTOR banner

robertknight:
This will need a rebase before merge to streamline the commit history. Otherwise I think it is ready for a review pass.

src/annotator/integrations/pdf.js
```diff
@@ -241,13 +262,26 @@ export class PDFIntegration {
...
+    if (!this._banner) {
+      this._banner = document.createElement('hypothesis-banner');
+      document.body.prepend(this._banner);
+      createShadowRoot(this._banner);
+    }
 
-    const warningBannerContent = createShadowRoot(this._warningBanner);
-    render(<WarningBanner />, warningBannerContent);
+    render(
+      <Banners>
```
lyzadanger, line 272: I pushed _just one more commit_ with a dead simple `Banners` component. Explanation:

* Tailwind only includes utility CSS classes in output that are used in any of the modules/files in its defined `content` globs
* I'd like to avoid expanding the globs to include non-component source (for cleanliness and performance) except where necessary—this opinion is certainly open for discussion!
* These classes "worked" here because `flex flex-col` are commonly used and have been used in other components in the application (and thus were included in output styles)

The `Banners` component is dead simple and is exercised by this module's tests (i.e. I don't think that display-only component needs "real" tests).

src/annotator/integrations/pdf.js
```diff
@@ -210,29 +227,33 @@ export class PDFIntegration {
...
     }
   }
 
   /**
-   * Set whether the warning about a PDF's suitability for use with Hypothesis
-   * is shown.
+   * Update banners shown above the PDF viewer.
    *
-   * @param {boolean} showWarning
+   * @param {Partial<typeof PDFIntegration.prototype._bannerState>} state
```
lyzadanger, line 240: This typing leaves me scratching my head a little bit...

lyzadanger (approved):
Just to complicate things one last bit, I pushed another commit with a `Banners` component; see my comments inline.

I think this is looking good to merge for our current intentions...thanks!

src/annotator/integrations/pdf.js
robertknight: > Tailwind only includes utility CSS classes in output that are used in any of the modules/files in its defined content globs

Ah - I forgot about that. Thanks for the fixup.

> The Banners component is dead simple and is exercised by this module's tests (i.e. I don't think that display-only component needs "real" tests).

I agree. Such a trivial component just needs _some_ test to render it.

src/annotator/integrations/pdf.js
robertknight: This is defining the type of an internal function's parameters in terms of a field on the class. I wouldn't do this for a public method, but this is private.

robertknight:
I have rebased this branch and merged several commits to clean up the Git history.

PR #4322, opened 2022-03-15
Make JSTOR logo open link in a new tab

robertknight:
In the context of our LMS app, this banner is displayed inside a nested
iframe, so it is important to open the link in a new tab.


robertknight:
Change agreed in Slack here: https://hypothes-is.slack.com/archives/C1M8NH76X/p1647358623847689?thread_ts=1647358559.840649&cid=C1M8NH76X.

PR #4327, opened 2022-03-17
Add missing types to various callbacks / props in src/sidebar/components

robertknight:
The only notable code change is in `PaginationNavigation` where I removed unnecessary checks for `event.target` being null in the `onClick` event for some `LabeledButton`s.


lyzadanger (approved):
👍🏻

PR #4349, opened 2022-03-24
Remove `DefaultIfAny` helper

robertknight:
Remove a cryptic piece of code that is no longer needed given the ability to
specify default values for template arguments [1]

[1] https://github.com/microsoft/TypeScript/pull/45483

### DIFF
--- a/src/sidebar/store/create-store.js
+++ b/src/sidebar/store/create-store.js
@@ -58,16 +58,6 @@ import { createReducer, bindSelectors } from './util';
  */
 
 /**
- * Replace a type `T` with `Fallback` if `T` is `any`.
- *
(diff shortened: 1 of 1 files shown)


lyzadanger (approved):
Please do.

src/sidebar/store/create-store.js
```diff
@@ -196,7 +186,7 @@ export function createStore(modules, initArgs = [], middleware = []) {
  * @template State
  * @template Actions
  * @template {SelectorMap<State>} Selectors
- * @template RootSelectors
+ * @template [RootSelectors={}]
```
robertknight, line 189: The `rootSelectors` property of store modules is optional. In store modules without it, this treats it as if it were an empty object.

Issue #4364, opened 2022-03-28
Keyboard shortcuts for creating annotations don't work in XHTML documents


robertknight:
While testing the XHTML pages mentioned in https://github.com/hypothesis/client/issues/4350#issuecomment-1079101754 I found an issue where the keyboard shortcuts for creating annotations don't work in XHTML pages.

**Steps to reproduce:**

1. Load client on https://www.nayuki.io/page/practical-guide-to-xhtml or https://www.w3.org/People/mimasa/test/xhtml2/spec-examples/
2. Select some text
3. Press "a" or "h" to create an annotation or higlight

**Expected:** These shortcuts create a new annotation/highlight
**Actual:** In XHTML documents, these shortcuts do not work

I experience the issue in Safari 15.4 and Chrome 101. I did not see this problem in Firefox 98.

robertknight:
Added VitalSource label as ebooks are the most common source of XHTML documents that are users will encounter.

robertknight:
This is weird. In Chrome and Safari, the body element doesn't seem to receive `keydown` events in XHTML documents. In Firefox it does.

To test, evaluate the following in the console on https://www.nayuki.io/page/practical-guide-to-xhtml then click on the document and press some keys. In Firefox - log messages. In Chrome/Safari, no logs.

```js
document.body.addEventListener('keydown', e => console.log('body received key event', e))
```

robertknight:
The default keyboard focusability of the body element seems to differ between HTML and XHTML documents.

1. Add a `keydown` event listener to the document that logs its event
2. Click in any non-interactive part of the document
3. Press keys and look at the `path` property of the event objects

In Safari and Chrome the `keydown` event is fired at the body element in an HTML document, but the root `<html>` element in an XHTML document. If the body is explicitly made focusable by setting `document.body.tabIndex = 1` then the body becomes focusable.

robertknight:
In Chrome I think this results from the logic which chooses an event target for keyboard events when there is no focused element:

I believe that in XHTML documents, the `IsA<HTMLDocument>` check returns false here:

<img width="636" alt="Default event target" src="https://user-images.githubusercontent.com/2458/160594191-870f7360-b2c1-477f-9bed-fd44891e82f1.png">

PR #4366, opened 2022-03-29
Install shortcut listener on document element

robertknight:
Listen for key events for keyboard shortcuts on the document element rather than
body element, to make them work in XHTML documents.

When no interactive element (eg. an input field) is focused in an HTML document,
keyboard events are sent to the body element. In an XHTML document however
keyboard events are sent to the document element instead in Safari and Chrome
(in Firefox they are still sent to the body). The key event listeners used for
the adder's shortcuts were installed on the body element, so they didn't work in
XHTML documents in Safari and Chrome.

My guess is that the reason for this behavioral quirk is that HTML documents are
guaranteed to have a body element - an empty one will be generated if none
is present in the markup, whereas XHTML documents are not.

Fixes https://github.com/hypothesis/client/issues/4364

---

**Testing:**

Visit http://localhost:3000/document/xhtml-test and select some text, then press "a" or "h". On master, this will not create a new annotation in Safari / Chrome but will in Firefox. On this branch the keyboard shortcuts should work in all browsers.

### DIFF
--- a/src/shared/shortcut.js
+++ b/src/shared/shortcut.js
@@ -80,7 +80,13 @@ export function matchShortcut(event, shortcut) {
 export function installShortcut(
   shortcut,
   onPress,
-  { rootElement = document.body } = {}
+  {
+    // We use `documentElement` as the root element rather than `document.body`
+    // which is used as a root element in some other places because the body
+    // element is not keyboard-focusable in XHTML documents in Safari/Chrome.
+    // See https://github.com/hypothesis/client/issues/4364.
+    rootElement = document.documentElement,
+  } = {}
 ) {
   /** @param {KeyboardEvent} event */
   const onKeydown = event => {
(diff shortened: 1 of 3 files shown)


src/shared/shortcut.js
```diff
@@ -108,11 +114,7 @@ export function installShortcut(
  * @param {(e: KeyboardEvent) => any} onPress - A function to call when the shortcut matches
  * @param {ShortcutOptions} [options]
  */
-export function useShortcut(
-  shortcut,
-  onPress,
-  { rootElement = document.body } = {}
-) {
+export function useShortcut(shortcut, onPress, { rootElement } = {}) {
```
robertknight, line 117: The default argument was removed here so that we only specify a default for `rootElement` in one place.

lyzadanger (approved):
Nice sleuthing! Looks great and works.

PR #4372, opened 2022-03-31
Add missing types to annotation and route store modules

robertknight:
- Add state and action types to reducers
 - Replace Array.forEach calls with for..of loops to enable better type inference.
 - Replace plain objects used as sets with ES sets

### DIFF
--- a/src/sidebar/store/modules/annotations.js
+++ b/src/sidebar/store/modules/annotations.js
@@ -3,7 +3,17 @@
  * sidebar.
  */
 
+import { createSelector } from 'reselect';
+
+import * as metadata from '../../helpers/annotation-metadata';
+import { isSaved } from '../../helpers/annotation-metadata';
+import { countIf, toTrueMap, trueKeys } from '../../util/collections';
+import { createStoreModule, makeAction } from '../create-store';
+
+import { routeModule } from './route';
+
 /**
+ * @typedef {'anchored'|'orphan'|'timeout'} AnchorStatus
  * @typedef {import('../../../types/api').Annotation} Annotation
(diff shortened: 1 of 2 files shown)


src/sidebar/store/modules/annotations.js
```diff
@@ -192,19 +201,28 @@ const reducers = {
...
+   * @param {{ highlighted: Record<string, boolean> }} action
+   */
   HIGHLIGHT_ANNOTATIONS(state, action) {
     return { highlighted: action.highlighted };
   },
 
-  /** @param {State} state */
+  /**
+   * @param {State} state
+   * @param {{ annotationsToRemove: AnnotationStub[], remainingAnnotations: Annotation[] }} action
```
robertknight, line 214: The `remainingAnnotations` field looks like a code smell to me, but in this PR I'm just documenting/typing the existing code.

src/sidebar/store/modules/annotations.js
```diff
@@ -125,17 +125,20 @@ const initialState = {
 /** @typedef {typeof initialState} State */
 
 const reducers = {
-  /** @param {State} state */
+  /**
+   * @param {State} state
+   * @param {{ annotations: Annotation[], currentAnnotationCount: number }} action
```
robertknight, line 130: `currentAnnotationCount` looks like a code smell to me, but I'm just documenting/typing the existing structure in this PR.

lyzadanger (approved):
I'm having a little bit of trouble following the flow of the `addAnnotations` action creator in `annotation`. There are several moving parts!

There's a long-standing ambiguity across `annotation` regarding tags and ids, but that's nothing new.

PR #4374, opened 2022-03-31
Simplify login popup creation flow

robertknight:
The process of creating and navigating the login popup used to involve two
steps, first creating a blank window and then navigating it to the final
authorization URL. This was needed because, in Firefox, the popup window had to
be created in the same turn of the event loop as the user's click on the "Log
in" button (otherwise the popup blocker would trigger) but generating the
authorization URL involved an async "fetch" of API links.

The major browsers have now all settled on a more flexible model for allowing
popups in response to user gestures, where the popup must be opened within a
certain time window of the gesture. In practice the timeout seems to be ~1000ms
in Safari and longer than that in other browsers.

In this context we expect the async delay between the user clicking the "Log in"
button and us creating the popup to be ~0ms, since the API links should already
have been fetched at this point and so we're just fetching locally cached values.

Based on this assumption, the flow for creating the login popup window has been
simplified to create the popup window at the final URL immediately, removing the
need to open a blank window as a first step.

Simplifying the code here will make it easier to change how the popup window and
sidebar communicate, eg. to resolve an issue with the new
Cross-Origin-Opener-Policy header [1].

[1] https://github.com/hypothesis/product-backlog/issues/1333

**Testing:**

Try logging into the client in Safari, Firefox and Chrome. It should behave the same as before.

**Related reading:**

- [Original Firefox issue I filed, closed 4 years ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1425419)
- [Related standardization discussion](https://github.com/whatwg/html/issues/1903)
- [Relevant HTML spec section, updated 4 years ago after the previous discussion](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation)

### DIFF
--- a/src/sidebar/services/auth.js
+++ b/src/sidebar/services/auth.js
@@ -270,9 +270,15 @@ export class AuthService extends TinyEmitter {
      * then exchange for access and refresh tokens.
      */
     async function login() {
-      const authWindow = OAuthClient.openAuthPopupWindow($window);
+      // Any async steps before the call to `client.authorize` must complete
+      // in less than ~1 second, otherwise the browser's popup blocker may block
+      // the popup.
+      //
+      // `oauthClient` is async in case in needs to fetch links from the API.
+      // This should already have happened by the time this function is called
+      // however, so it will just be returning a cached value.
       const client = await oauthClient();
-      const code = await client.authorize($window, authWindow);
+      const code = await client.authorize($window);
 
       // Save the auth code. It will be exchanged for an access token when the
       // next API request is made.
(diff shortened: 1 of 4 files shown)


robertknight:
According to my sleuthing, the exact timeouts for showing a popup after a user action appear to be:

- WebKit: 2 seconds (search for `defaultTransientActivationDuration` in source)
- Chrome: 5 seconds (search for `UserActivationState`/`kActivationLifespan` in source)
- Firefox: 5 seconds (see `dom.user_activation.transient.timeout` preference in about:config)

robertknight:
An existing issue I noticed when testing this is that clicking the "Log in" link if the popup has already been opened does not show either the existing popup or a new one in Safari. Instead the user just sees an error that the login popup was closed. In Chrome the user also sees this error but a new popup is immediately created. In both cases if the user switches to an open popup and logs in, the login will be propagated back to the sidebar.

lyzadanger (approved):
Cool. Tested in Safari, Firefox, Chrome, looking good.

PR #4393, opened 2022-04-06
Reuse `translateOffsets` utility in PDF quote anchoring

robertknight:
When anchoring quote selectors in PDFs, the quote is anchored in a version of the PDF page text that has whitespace stripped out [1]. The offsets of the best match is then translated back to the whitespace-inclusive version of the text.

This commit refactors this translation to re-use the `translateOffsets` utility added in https://github.com/hypothesis/client/pull/4355. It also re-introduces a check that was removed in that PR, which checks for non-whitespace differences in the matched quote in the PDF text extracted by `getPageTextContent`, and the corresponding content in the text layer. Such differences indicate an unexpected difference between the text layer and the text extracted via PDF.js APIs for a page, and it is useful to flag this issue.

[1] This makes whitespace changes "free" from the perspective of the fuzzy anchoring cost function (see `matchQuote`)

### DIFF
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -1,5 +1,6 @@
 /* global PDFViewerApplication */
 
+import { warnOnce } from '../../shared/warn-once';
 import { translateOffsets } from '../util/normalize';
 import { matchQuote } from './match-quote';
 import { createPlaceholder } from './placeholder';
@@ -323,6 +324,16 @@ async function anchorByPosition(pageIndex, start, end) {
       end,
       isNotSpace
     );
+
+    const textLayerQuote = stripSpaces(
+      textLayerStr.slice(textLayerStart, textLayerEnd)
+    );
+    const pageTextQuote = stripSpaces(pageText.slice(start, end));
+    if (textLayerQuote !== pageTextQuote) {
+      warnOnce(
+        'Text layer text does not match page text. Highlights will be mis-aligned.'
+      );
+    }
 
     const startPos = new TextPosition(root, textLayerStart);
     const endPos = new TextPosition(root, textLayerEnd);
--- a/src/annotator/anchoring/pdf.js
+++ b/src/annotator/anchoring/pdf.js
@@ -431,6 +431,8 @@ async function anchorQuote(quoteSelector, positionHint) {
       if (page < expectedPageIndex) {
         strippedHint = strippedText.length; // Prefer matches closer to end of page.
       } else if (page === expectedPageIndex) {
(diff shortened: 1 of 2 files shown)


src/annotator/anchoring/pdf.js
```diff
@@ -401,33 +410,36 @@ async function anchorQuote(quoteSelector, positionHint) {
...
+        strippedHint = strippedText.length; // Prefer matches closer to end of page.
       } else if (page === expectedPageIndex) {
-        hint = expectedOffsetInPage;
+        [strippedHint] = translateOffsets(
+          text,
+          strippedText,
+          expectedOffsetInPage,
+          expectedOffsetInPage,
+          isNotSpace
+        );
```
robertknight, line 440: In words, this code is translating the expected position of the quote, from an annotation's position selector, from the whitespace-inclusive version of a PDF page's text to the whitespace-stripped version of the text.

src/annotator/anchoring/pdf.js
```diff
@@ -442,19 +454,18 @@ async function anchorQuote(quoteSelector, positionHint) {
     }
 
     if (!bestMatch || match.score > bestMatch.match.score) {
+      const [start, end] = translateOffsets(
+        strippedText,
+        text,
+        match.start,
+        match.end,
+        isNotSpace
+      );
```
robertknight, line 463: This is performing the opposite translation to the above `translateOffsets` usage: mapping the offset of a quote match from the whitespace-stripped version of a page's text to the whitespace-inclusive version of the page's text.

src/annotator/anchoring/test/pdf-test.js
```diff
@@ -531,6 +531,11 @@ describe('annotator/anchoring/pdf', () => {
     });
 
     it('anchors with mismatch if text layer differs from PDF.js text API output', async () => {
+      const warnOnce = sinon.stub();
+      pdfAnchoring.$imports.$mock({
+        '../../shared/warn-once': { warnOnce },
+      });
```
robertknight, line 537: nb. There is no `$imports.$restore` call added in this PR to match this `$mock` call because one already exists in a higher-level `afterEach`.

src/annotator/anchoring/pdf.js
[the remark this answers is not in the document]
robertknight: > When passing the same value for start and end to translateOffsets, is it to be expected that it will return [startOffset, endOffset] where startOffset === endOffset in all cases?

Yes. In this case the number of non-ignored characters between the start and end offsets in the input is zero, and `translateOffsets` will always return a `[start, end]` tuple with `start` === `end` in that scenario.

PR #4398, opened 2022-04-07
Add missing types to session store module

robertknight:
This found apparent missing checks for a null `profile.userid` (ie. current user
is logged out) in several places. These places should not be reachable
when the user is logged out due to checks elsewhere, but add explicit
checks for the user being logged in.


lyzadanger (approved):
LGTM.

PR #4404, opened 2022-04-08
Set permissions for GitHub actions

naveensrinivasan:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,5 +1,8 @@
 name: Continuous integration
 on: pull_request
+permissions:
+  contents: read
+
 jobs:
   ci:
     runs-on: ubuntu-latest
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -4,8 +4,14 @@ on:
     - cron: '0 10 * * *'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   stale:
+    permissions:
+      issues: write  # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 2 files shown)


robertknight:
Thanks for the information.

Note for Hypothesis developers: As an alternative or addition to setting `contents: read` for CI workflows, there is an organization-wide default setting we could set at https://github.com/organizations/hypothesis/settings/actions. This chooses between the "Default access (restrictive)" and "Default access (permissive)" tables listed at https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token.

naveensrinivasan:
> Thanks for the information.
> 
> Note for Hypothesis developers: As an alternative or addition to setting `contents: read` for CI workflows, there is an organization-wide default setting we could set at https://github.com/organizations/hypothesis/settings/actions. This chooses between the "Default access (restrictive)" and "Default access (permissive)" tables listed at https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token.

These permissions are required to modify things

https://github.com/hypothesis/client/blob/717d8dfc48b8248e3c1655309c998e3e4ab3fe37/.github/workflows/stale.yml#L13-L14

PR #4449, opened 2022-04-21
Replace incorrect use of `any` with `void` in callback return types

robertknight:
When declaring a prop or parameter callback whose result is unused, the
return type should be `void` rather than `any`. In this context the
return type specifies what the callee _requires_, not what the caller
_provides_.


lyzadanger (approved):
👍🏻 Yep.

Issue #4332, opened 2022-03-18
Ship HTML side-by-side mode behind a feature flag


robertknight:
We have implemented support for side-by-side mode for HTML documents, but it is currently only enabled for EPUB documents. I'd like to ship this for all HTML documents. This is a significant change to the user experience, and we may still find that there are issues that need to iron out before it is good enough to ship by default, or we may need an option to turn it on/off explicitly, so I think it would make sense to ship it behind a feature flag.

Aside from enabling us to test a valuable improvement to the client, putting it behind a feature flag will also involve us building the infrastructure to support feature-flagging other changes in the annotator code. 

- [ ] Modernize the existing feature flag code in the client (in src/annotator/features.js) (https://github.com/hypothesis/client/pull/4495)
- [ ] Support feature flags in guests (currently feature flags are only accessible to the host frame) (https://github.com/hypothesis/client/pull/4495)
- [ ] Implement an `html_side_by_side` feature flag that toggles whether side-by-side mode is enabled in HTMLIntegration

PR #4496, opened 2022-05-03
Enable side-by-side mode for web pages behind a feature flag

robertknight:
Enable support for side-by-side mode for all web pages, behind an `html_side_by_side` feature flag.

- Pass feature flags from the `Guest` through to the `HTMLIntegration`
- In the VitalSource integration, instead pass a dummy `FeatureFlags` which has side-by-side mode always enabled

Marked as a draft because I want to review the plumbing of feature flags through to the HTMLIntegration and the approach used to forcibly enable it for the VitalSource integration.

**Testing:**

1. Enable the `html_side_by_side` feature flag in h (see https://github.com/hypothesis/h/pull/7428)
2. Visit a test HTML document (eg. http://localhost:3000/document/doyle) and open the sidebar. The page should be resized.
3. Repeat previous step with feature flag turned off. The page should not be resized.

A variation on Step 3 is to turn on the feature flag only for certain users. When you log in / log out the feature should be enabled/disabled.

Fixes https://github.com/hypothesis/client/issues/4332.

### DIFF
--- a/src/annotator/features.js
+++ b/src/annotator/features.js
@@ -3,14 +3,20 @@ import { TinyEmitter } from 'tiny-emitter';
 import { warnOnce } from '../shared/warn-once';
 
 /**
+ * @typedef {import('../types/annotator').FeatureFlags} IFeatureFlags
+ */
+
+/**
  * List of feature flags that annotator code tests for.
  *
  * @type {string[]}
  */
-const annotatorFlags = [];
+const annotatorFlags = ['html_side_by_side'];
 
 /**
  * An observable container of feature flags.
+ *
+ * @implements {IFeatureFlags}
  */
 export class FeatureFlags extends TinyEmitter {
   /**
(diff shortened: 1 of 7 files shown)


src/annotator/features.js
```diff
@@ -2,15 +2,21 @@ import { TinyEmitter } from 'tiny-emitter';
...
  *
  * @type {string[]}
  */
-const annotatorFlags = [];
+const annotatorFlags = ['html_side_by_side'];
 
 /**
  * An observable container of feature flags.
+ *
+ * @implements {IFeatureFlags}
```
robertknight, line 19: I don't love the fact that the interface and the main implementation have the same name. The main distinction between the interface and this class is that the interface only provides methods to read flags and subscribe to updates, whereas the class has methods to update them.

robertknight:
Thanks for the feedback Lyza. I have added some code comments in response.

PR #4497, opened 2022-05-04
Replace various uses of `any` with `unknown`

robertknight:
This includes the results of RPC calls to the LMS app, which must now be explicitly cast to the expected result type. Otherwise I didn't have to change any code which used the values.


lyzadanger (approved):
LGTM

PR #4513, opened 2022-05-10
Add missing types for PortRPC method handlers

robertknight:
Previously the arguments of the callback passed to `PortRPC.on` were
inferred as `any`. Make the callback type generic so that the caller is
forced to specify what types the arguments have.

Internally within PortRPC, use `unknown` rather than `any` for RPC
method arguments where possible.

Note that this change does not ensure that an RPC method is called (via `call`)
using arguments of the same types that the handler expects, but at least it is
now easier to check that the `call` and `on` uses match up and that arguments
are typed within the handler.

### DIFF
--- a/src/shared/messaging/port-rpc.js
+++ b/src/shared/messaging/port-rpc.js
@@ -33,14 +33,14 @@ const PROTOCOL = 'frame-rpc';
  * See https://github.com/substack/frame-rpc#protocol
  *
  * @typedef RequestMessage
- * @prop {any[]} arguments
+ * @prop {unknown[]} arguments
  * @prop {string} method
  * @prop {PROTOCOL} protocol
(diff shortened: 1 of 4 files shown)


src/shared/messaging/port-rpc.js
```diff
@@ -275,7 +290,7 @@ export class PortRPC {
 
     const seq = this._sequence++;
     const finalArg = args[args.length - 1];
-    if (typeof finalArg === 'function') {
+    if (isCallback(finalArg)) {
```
robertknight, line 293: This change alters the inferred type of `finalArg` slightly. Without it, TS complains about the `set(...)` call below.

lyzadanger (approved):
LGTM

PR #4626, opened 2022-07-04
Document issue with `useStore` + `useMemo` and fix potential missing re-render

robertknight:
This PR fixes a potential missing re-render in `useStore` if the store changes in between the initial render of a component, and `useEffect` callbacks firing, by swapping out `useEffect` for `useLayoutEffect`. We've never actually encountered a problem due to this issue IIRC, but we might as well prevent it from happening.

I also documented a hazard with using the returned wrapper together with hooks that accept dependencies. I think the hazard could be eliminated with some API / behavior changes [1], but this requires some further experimentation.

[1] Roughly what I had in mind was that the `Proxy` returned by `useStore` would get re-created whenever relevant data changes in the store. This would mean that any `useMemo`s etc. would get invalidated as necessary. However we have a bunch of effects which use the store to dispatch actions, and those shouldn't be re-run unnecessarily.

### DIFF
--- a/src/sidebar/store/use-store.js
+++ b/src/sidebar/store/use-store.js
@@ -38,6 +38,24 @@ class CacheEntry {
  *
  * The returned wrapper has the same API as the store itself.
  *
+ * The returned wrapper does not change its identity if the store updates. This
+ * means you need to be careful when using it with hooks that have dependencies,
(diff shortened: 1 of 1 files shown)


lyzadanger (approved):
This was good thinking.

PR #4630, opened 2022-07-08
Revalidate canonical URL after a navigation

robertknight:
This is a proof of concept for a solution to https://github.com/hypothesis/product-backlog/issues/1288.

After a client-side navigation or when the document's URI is fetched for the first time, the `<link rel=canonical>` element in the page is compared against the document URL. If, based on some heuristics, we determine that it is potentially out of sync due to the issue described in https://github.com/hypothesis/product-backlog/issues/1288, then the page's HTML is refetched from the origin server and the canonical link is extracted from the response. The re-fetched canonical link is then used to update the one in the page. If re-fetching the page or parsing the response fails, revalidation aborts and the logic continues as before.

In some brief testing on YouTube and Reddit, this resolves the issue with client-side navigations do not update the document URL displayed in the sidebar. There are however a myriad of ways in which this could potentially fail, so I need to think through them and decide if the risk is acceptable. Also refetching the entire page's HTML is potentially quite slow, which could lead to a confusing user experience. Potential race conditions also need to be considered.

### DIFF
--- a/src/annotator/integrations/canonical-url-revalidator.js
+++ b/src/annotator/integrations/canonical-url-revalidator.js
@@ -1,0 +1,114 @@
+/** @param {URL} url */
+function extractPathAndQuery(url) {
+  const path = url.pathname;
+  if (!url.search) {
+    return path;
+  }
+  // TODO - Sort query params
+  return `${path}?${url.search}`;
+}
+
+/**
+ * CanonicalURLRevalidator revalidates `<link rel=canonical>` if it is
+ * potentially out of sync with the current document URL.
+ *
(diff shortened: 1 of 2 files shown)


robertknight:
We might want to de-risk initial deployment of this via some combination of:

1. Limiting its usage to certain domains (eg. youtube.com, reddit.com) known to be affected and for which we have tested that it works
2. Putting canonical URL revalidation behind a feature flag

Additionally it would be good to have some kind of reporting to indicate how often revalidation occurs, how often it succeeds and how long it takes.

robertknight:
This is not a high priority at present. I'll come back to it when I have more time.

PR #4657, opened 2022-07-18
Remove obsolete ESLint rule overrides

robertknight:
Remove ESLint rule overrides that are no longer required once
https://github.com/hypothesis/frontend-toolkit/pull/42 is shipped.

While doing this I also added a clarification about why one of the remaining overrides _is_ still required.


lyzadanger (commented):
Consider this approved one the upstream dependency is merged and CI passes...

PR #4669, opened 2022-07-26
Delete ISSUE_TEMPLATE.md

seanh:
If we want a generic issue template like this I think we should add it to https://github.com/hypothesis/.github where it'll apply to all Hypothesis repos. But I'm not sure that we really want this issue template so perhaps we should just delete it?


robertknight:
There are some frontend specifics that mean we'd probably want a different template (if any) for this project compared to a Python project of a JS library. I have to say that since we don't a large volume of bug reports via GitHub, the template is probably not that important in practice.

lyzadanger (approved):
I agree that where feasible, it makes sense to have a centralized `ISSUE_TEMPLATE`. It may be useful in some situations to have a specific `ISSUE_TEMPLATE` for a given project but not if it's not maintained or not cogent to the project at hand.

